PR #24, opened 2019-10-02
chore(release): Prepare 0.1.0-alpha.1 release

LucioFranco:


tonic/src/macros.rs
```diff
@@ -1,6 +1,12 @@
 /// Include generated proto server and client items.
 ///
 /// You must specify the gRPC package name.
+///
+/// ```rust,ignore
+/// mod pb {
+///     tonic::include_proto("hellworld");
```
JohnDoneth: Should be "helloworld"

LucioFranco: good catch...

PR #28, opened 2019-10-02
feat(transport): Expose http/2 settings

JohnDoneth:
Wraps and exposes h2 options from hyper.


LucioFranco:
@JohnDoneth should we include the client as well?

Issue #47, opened 2019-10-05
Expose granular control of TLS configuration for OpenSSL and Rustls


jen20:
### Crates

- `tonic`

### Motivation

Currently a small subset of of the configuration options available for OpenSSL or Rustls are exposed by the builder API, via the `openssl_tls` and `rustls_tls` functions:

- Certificate and corresponding private key for servers (via [`openssl_tls`][server_openssl_tls] for servers and [`rustls_tls`][server_rustls_tls] for clients)
- A root certificate and domain name for client endpoints (via [`openssl_tls`][client_openssl_tls] for servers and [`rustls_tls`][client_rustls_tls] for clients)

Clearly many more configuration points exist in each of the supported TLS libraries which are _not_ exposed. If these are needed, users lose the benefit of the 'batteries included' nature of the `transport` implementations which are part of Tonic.

### Proposal

Currently the TLS library in use can be switched out in Tonic by enabling the appropriate Cargo feature, and changing the method name called from the builder - the arguments are compatible for each.

Wrapping every available option in a cross-TLS library compatible way in this way could lead to a sprawling API, and there may still be divergence between libraries. Instead, it would be better to allow the native configuration structure from each library to be passed into the builders:

- [`rustls::ServerConfig`][rustls_server_config] and [`rustls::ClientConfig`][rustls_client_config] for Rustls
- [`openssl::ssl::SslAcceptorBuilder`][openssl_acceptor_builder] and [`openssl::ssl::SslConnectorbuilder`][openssl_connector_builder] for OpenSSL. 

This does add considerable complexity though, and there are some cases where the existing mechanism is sufficient - so it is desirable to keep the simpler interface for those that prefer that.

It's probably better for these different mechanisms for configuring TLS to be mutually exclusive with one another - that is, it should not be possible to pass in both a `ServerConfig` and an `Identity`. Consequently, I propose making the existing methods on the builder take an enumeration instead, with separate variants for each configuration mechanism.

### Alternatives

An alternative to the enumeration is to have builder methods for each configuration mechanism and ensure via some other mechanism that they are not used together. This seems more complex to me, though there is currently not much in the way of precedent for the style suggested above in the builder API so this might be preferable.

[server_openssl_tls]: https://docs.rs/tonic/0.1.0-alpha.1/tonic/transport/server/struct.Server.html#method.openssl_tls
[...]

PR #48, opened 2019-10-05
Jen20/expose tls configuration

jen20:
This is the start of an implementation of the solution for exposing more granular control over TLS as described in #47.

It is currently incomplete, but does not lose any functionality over the previous API and should help inform whether this is the path we want to go down or not.


jen20:
After playing with this some more, and the suggestions in #47 I've reworked this to use a new builder for TLS options. This allows us to expose several different levels of abstraction for configuring TLS while not creating sprawl in the `Server` and `Endpoint` builders.

.gitignore
```diff
@@ -1,3 +1,5 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.idea/
```
LucioFranco: minor but I'd prefer to not introduce this to the git ignore since they are not related to the project.

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+            .field("provider", &self.provider)
+            .finish()
+    }
+}
+
+#[cfg(feature = "tls")]
+impl ClientTlsConfig {
+    /// Creates a new `ClientTlsConfig` backed by the specified provider. Enable the `openssl` or
+    /// `rustls` features of the `tonic` crate to use OpenSSL or Rustls respectively.
+    pub fn new(provider: TlsProvider) -> Self {
```
LucioFranco: what do you think about providing constructors like so? This also allows us to make the enum private thus allowing us to add new variants without breaking code!

```rust
#[cfg(feature = "openssl")]
pub fn with_openssl() -> Self {...}

#[cfg(feature = "rustls")]
pub fn with_rustls() -> Self {...}
```

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+            cert: None,
+            #[cfg(feature = "openssl")]
+            openssl_raw: None,
+            #[cfg(feature = "rustls")]
+            rustls_raw: None,
+        }
+    }
+
+    /// Sets the domain name against which to verify the server's TLS certificate. If set to `None`
+    /// (the default), the address specified
```
LucioFranco: seems like the sentence got cut off?

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+    pub fn domain_name(mut self, domain_name: impl Into<Option<String>>) -> Self {
```
LucioFranco: Since this is a builder fn and is optional probably makes sense to just take a String or `impl AsRef<str>`?

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+        self.domain = domain_name.into();
+        self
+    }
+
+    /// Sets the CA Certificate against which to verify the server's TLS certificate.
+    pub fn ca_certificate(mut self, ca_certificate: Certificate) -> Self {
```
LucioFranco: What do you think about this taking `&mut self` and returning `&mut Self` like the other builders?

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+    /// This overrides all other TLS options set via other means.
+    #[cfg(feature = "rustls")]
+    pub fn rustls_client_config(mut self, config: tokio_rustls::rustls::ClientConfig) -> Self {
+        self.rustls_raw = Some(config);
+        self
+    }
+
+    fn tls_connector(self, uri: Uri) -> Option<TlsConnector> {
+        match self.provider {
+            TlsProvider::None => None,
```
LucioFranco:
```suggestion
            TlsProvider::None => unreachable!("Created a tls provider with none enabled."),
```

tonic/src/transport/endpoint.rs
```diff
@@ -252,3 +203,96 @@ impl fmt::Debug for Endpoint {
...
+            #[cfg(feature = "openssl")]
+            TlsProvider::OpenSsl => {
+                if let Some(connector) = self.openssl_raw {
+                    return Some(TlsConnector::new_with_openssl_raw(connector).unwrap());
```
LucioFranco: do we want to bubble this result up? or is this never going to panic?

tonic/src/transport/service/tls.rs
```diff
@@ -56,41 +56,67 @@ enum Connector {
...
-
-        config.cert_store_mut().add_cert(ca)?;
+        if cert.is_some() {
+            let cert = cert.unwrap();
+            let ca = X509::from_pem(&cert.pem[..])?;
+            config.cert_store_mut().add_cert(ca)?;
+        }
 
-        let config = config.build();
+        //        let config = config.build();
```
LucioFranco:
```suggestion
```

tonic/src/transport/service/tls.rs
```diff
@@ -56,41 +56,67 @@ enum Connector {
...
     ) -> Result<Self, crate::Error> {
         let mut config = SslConnector::builder(SslMethod::tls())?;
-
         config.set_alpn_protos(ALPN_H2_WIRE)?;
 
-        let ca = X509::from_pem(&cert.pem[..])?;
-
-        config.cert_store_mut().add_cert(ca)?;
+        if cert.is_some() {
+            let cert = cert.unwrap();
```
LucioFranco: you probably wanna use:

```rust
if let Some(cert) = cert {...}
```

tonic/src/transport/service/tls.rs
```diff
@@ -56,41 +56,67 @@ enum Connector {
...
         })
     }
 
+    #[cfg(feature = "rustls")]
+    pub(crate) fn new_with_rustls_raw(
+        config: tokio_rustls::rustls::ClientConfig,
+    ) -> Result<Self, crate::Error> {
+        Ok(Self {
+            inner: Connector::Rustls(Arc::new(config)),
+            domain: Arc::new("".into()),
```
LucioFranco: do you think an option might be better here than an empty string?

tonic/src/transport/endpoint.rs
jen20: This definitely seems better and leaves fewer types exposed.

tonic/src/transport/endpoint.rs
jen20: Yes, although actually it just occurred to me that there is no value in having this be an `Option` in any case now - if it doesn't need to be set, just omit the method call.

.gitignore
jen20: Fixed.

tonic/src/transport/endpoint.rs
jen20: I've implemented this and made `TlsProvider` private to the crate - it's a lot more ergonomic and allows us to remove the `None` variant too.

tonic/src/transport/endpoint.rs
jen20: I've made this take an `Into<String>` instead, and fixed the doc comment.

tonic/src/transport/endpoint.rs
jen20: I've removed the `None` variant since it's no longer possible to construct it with the `with_openssl` and `with_rustls` functions being the entry point to the TLS builders.

tonic/src/transport/service/tls.rs
jen20: Fixed.

tonic/src/transport/endpoint.rs
jen20: I've made this bubble up to the top level of the builder, but we should probably have a wider think about how errors are handled during construction?

tonic/src/transport/service/tls.rs
jen20: Done.

tonic/src/transport/service/tls.rs
jen20: Actually it turns out this wouldn't have worked as implemented anyway: the domain name is required even with a raw config. That has simplified this area.

tonic/src/transport/endpoint.rs
jen20: I've had a look at this, but we have the issue that we want to take the `SslConnector` and `ClientConfig` values from the builder? Happy to look at alternatives though.

jen20:
Thanks for the feedback @LucioFranco - most of the issues you raised are now addressed, with a couple of additional questions inline.

tonic/src/transport/endpoint.rs
LucioFranco: I believe the configs should be cheaply clonable right? So that means we can take a &self and get a full version.

tonic/src/transport/endpoint.rs
LucioFranco: Yeah, I agree, I have some other thoughts about the builders I'd like to change. For now I think the best thing is to at least bubble it up and let the user decide if they want to panic or not.

jen20: Right now I've adjusted it such that the various builders return a `Result<Tls{Connector,Acceptor}, crate::Error>`, and the unwrap happens in the `tls_config()` method on the builder - in the same way as it was before. Not sure I see a good way to propagate errors in those builders without some kind of change to the builder pattern as you allude to there.

LucioFranco: You're right! I totally unwrapped before, I think this is fine then. Sorry about that. 😅

jen20:
Thanks! Next up I’m going to look at mutual TLS, I’ll likely have a write up of the options some time tomorrow morning (EU time).

PR #51, opened 2019-10-05
fix(transport): Avoid exit after bad TLS handshake

domodwyer:
Prevents the server exiting after a bad TLS handshake / error during
accept(). Instead the connection is dropped and the server continues to
serve new clients.

Previously an error would bubble up from the TLS library (tested with
rustls) and would cause:

```rust
	Server::builder()
	.rustls_tls(Identity::from_pem(&cert, &key))
	.clone()
	.serve(config.bind_addr(), AnExampleServer::new(endpoint))
	.await?;
```

to exit with:
```
	[src/main.rs:85] &e = Error(
		Server,
		Error(
			Accept,
			Custom {
				kind: InvalidData,
				error: CorruptMessage,
			},
		),
	)
```

This can be reproduced by dialling to a tonic server's socket and sending junk
until it crashes.

Unfortunately I'm a little unsure how to test this!

### DIFF
--- a/tonic/src/transport/server.rs
+++ b/tonic/src/transport/server.rs
@@ -28,6 +28,8 @@ use tower::{
     ServiceBuilder,
 };
 use tower_make::MakeService;
+#[cfg(feature = "tls")]
+use tracing::error;
 
 type BoxService = tower::util::BoxService<Request<Body>, Response<BoxBody>, crate::Error>;
 type Interceptor = Arc<dyn Layer<BoxService, Service = BoxService> + Send + Sync + 'static>;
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server.rs
```diff
@@ -207,7 +207,10 @@ impl Server {
                 #[cfg(feature = "tls")]
                 {
                     if let Some(tls) = &self.tls {
-                        let io = tls.connect(stream.into_inner()).await?;
+                        let io = match tls.connect(stream.into_inner()).await {
+                            Ok(io) => io,
+                            Err(_) => continue,
```
LucioFranco: What do you think about providing an error message here?

```suggestion
                            Err(error) => error!(message = "Unable to accept incoming connection.", %error);
```

domodwyer: I'd be happy to - this has just introduced me to the `tracing` crate, thanks!

domodwyer:
Thank you for building out tonic!

gRPC services are a fantastic way to incrementally introduce rust to an organisation, having a reliable crate to do so is critical - thanks for the time you've put into it.

LucioFranco:
@domodwyer I agree! Let me know if there is anything we can help with :)

Issue #44, opened 2019-10-04
Add benchmarks


LucioFranco:
It would be handy to add benchmarks mostly around encoding and decoding.

blittable:
@LucioFranco 

A few notes while chipping away at this...  something might be of interest and inspire additional tests.  IMHO, most of the request/response building is unimpressive (in a good way).  

Planning to look at concurrency/loads next and then, hopefully we can shape a pull request that provides solid regressions.


1) Request/Response construction comparison with 1 field and field size varying by byte size (log scaled):

![image](https://user-images.githubusercontent.com/53058687/66470699-4f7b3c00-eab4-11e9-8822-289cc8e58c71.png)

2) The next one compares the above with a different .proto - containing two structs with a total of 54 fields for 444 bytes,  varying one (String) by size as the above.  The bigger .proto has a nested sub-message.  We should be able to calculate the overhead of the additional field is the construction from the difference.
![image](https://user-images.githubusercontent.com/53058687/66474353-56597d00-eabb-11e9-929e-36952fd89654.png)


3) The raw logs.  The throughput measurement... it's not concurrent requests based, simply bytes processed.  It spike a bit in middle as the # of bytes increases.  That can be duplicated.

```
Request_Response/request/100                                                                             
                        time:   [29.072 ns 29.336 ns 29.644 ns]
                        thrpt:  [3.1417 GiB/s 3.1747 GiB/s 3.2035 GiB/s]
                 change:
                        time:   [-14.790% -13.916% -12.931%] (p = 0.00 < 0.05)
                        thrpt:  [+14.851% +16.166% +17.358%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  4 (4.00%) high mild
  3 (3.00%) high severe
Request_Response/response/100                                                                             
                        time:   [30.985 ns 31.229 ns 31.500 ns]
                        thrpt:  [2.9566 GiB/s 2.9822 GiB/s 3.0057 GiB/s]
                 change:
                        time:   [-7.4938% -6.9291% -6.3729%] (p = 0.00 < 0.05)
                        thrpt:  [+6.8067% +7.4450% +8.1009%]
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) high mild
  5 (5.00%) high severe
Request_Response/request/1000                                                                             
                        time:   [42.263 ns 42.528 ns 42.845 ns]
[...]

PR #53, opened 2019-10-06
chore: Add initial benchmarks

blittable:
Thanks for your work.

This is a sketch of a Criterion project/impl potentially addressing issue #44 

If the overall approach looks ok - I can dig into specific benches (data types, size, compression, etc.) and lift some interesting pb tests before a full pull-request.  

Obviously, it doesn't make sense to just test the serialization of Prost, so if you have specifics on the encoding/decoding you mentioned that would be helpful.

Longer-term (this week), I'd like to do some tests over-the-wire - so this is kind of a stepping-stone.

A few details:

The thought is to extend the lib.rs benches (possibly as a test-suite) to incorporate additional benches.

 I pulled one interesting pb file from the original protobuf project benchmark directory.  So, it seems like a good candidate for experimentation.

The dependencies are a bit funky b/c of build => bench and tonic integration.  Likewise the project/task might best be tucked inside of another.  I can probably get it cleaner.  This is just for review/discussion.

It's `cargo build, cargo run pb_build, cargo bench`  and result-free at this point.


LucioFranco:
@blittable hi! This looks like a really good start! I'll write down a couple thoughts:

- So I think direct serialization benchmarks are totally good to have in general. I think what I'm mostly thinking would be useful to bench are specifically the decoding of http bodies done in `codec/decode.rs` and encoding the prost message plus the rest of the body into a http body.
- I think it would also be really useful to have full end to end benchmarks which setup a server and see how many requests it can make. Then do the same for the server see how many requests it can handle. https://github.com/hyperium/hyper/tree/master/benches might contain some good starting points.

Overall, I think your approach is good as well! Let me know what you think about the above. 😄

blittable:
@LucioFranco   Great.  Will proceed.  

I like hyper's approach, but I'd like to see if the new throughput feature in Criterion might be useful: [throughput](https://github.com/blittable/tonic/blob/602369f190a7bb784348f3a9b751e5e7da610222/tonic-benches/benches/serialization.rs#L18)

1) Your unit tests in the encode/decode go deeper than the public api.  Is there a line for the benchmarks?  If we are building requests/responses, what's missing?

2) The project is currently sans-ci-cd.   I can open an issue and assist - let me know.  That *might* be a good place to build over-the-wire perf tests.

LucioFranco:
> if the new throughput feature in Criterion might be useful

Yes, this would be awesome!

> Your unit tests in the encode/decode go deeper than the public api. Is there a line for the benchmarks? If we are building requests/responses, what's missing?

Yeah, now that I think about it you may be right. Mostly, I want some form of benchmark that can catch regressions with the codecs.

> The project is currently sans-ci-cd. I can open an issue and assist - let me know. That might be a good place to build over-the-wire perf tests.

Not quite sure what you mean here? We do have CI setup but I don't think we really need to run the benchmarks on CI. They should mostly be for PRs introducing changes.

RetWolf:
> We do have CI setup but I don't think we really need to run the benchmarks on CI. They should mostly be for PRs introducing changes.
I'm not sure I entirely agree on this point. Tonic says it is 'focused on high performance', and I believe that having performance tests ran on CI will allow the maintainers to have more insight into the effect changes are having on the performance of the library.

Most importantly, while tests around this area can't cover all possibilities, they can help catch bugs that don't necessarily break logic, but degrade performance. Altogether I think performance tests in a project like this would just be a good thing to keep code quality high and give users confidence that they will always get great performance out of the library.

A big downside I can think of regarding this is the obvious maintenance overhead introduced. Maintainers and contributors would have to write and keep up-to-date these tests around the performance of the library otherwise we risk them just becoming an obsolete metric everyone kind of just ignores.

@LucioFranco I'd be interested to hear if you have some more thoughts about why we shouldn't have benchmarks in CI and what you think of the benefits I've mentioned here.

blittable:
For production, some  will need something in *their* ci/cd process.   For adoption, reproducible benchmarks, especially integrating cpu metrics, would be huge.

LucioFranco:
@RetWolf Hi! So mostly im not confident that github actions will give us non noisy VMs so running them on CI may not be the best. That said, I do want to have benchmarks and I do want to run them often. Grpc itself has a set of benchmarks that I have started to implement that would allow us to compare aganist other languages.

Cargo.toml
```diff
@@ -4,4 +4,5 @@ members = [
     "tonic-build",
     "tonic-examples",
     "tonic-interop",
+    "tonic-benches",
```
LucioFranco: I want to see if its possible to include these within the main tonic crate? We can avoid the build.rs if we just hand write what the codegen does. What do you think?

RetWolf:
@LucioFranco Thanks, that actually makes a lot of sense and was something I hadn't previously weighed.

Cargo.toml
blittable: @LucioFranco It simplifies everything.  

I've updated the PR here for review, pushing the criterion tests inside tonic with no proto build.    There are some comments in a readme.  There is a new dev-dep in the cargo.toml

If it makes sense, I'll open a clean PR b/c I obviously still don't understand everything about git. ;)

blittable:
@LucioFranco @RetWolf 

My 2 cents on e-2-e testing for perf - it might be a bit of overkill for Tonic to go down that path for ci/cd.  It's great - but Tower and Tokio might be more natural fits for that effort.  Their efforts (tower-trace / and tokio-trace) could be hugely helpful when they've matured a bit. 

I/O (IMHO) is best analyzed on bare-metal, and that means infrastructure/maintenance.  We can make that happen, but it's a level of effort thing.

If one of the goals is to pitch the platform and stimulate conversation, then an annual set of reproducible benchmarks/tests/scripts is useful.  e.g. here is how it was tested / hardware specifications etc. / scripts

LucioFranco:
Yeah, I believe h2 has a serious set of benchmarks and we plan on extending this. Tonic is mostly just a small crate on top of those main crates.

Issue #34, opened 2019-10-03
Consider an `IntoRequest` trait


hawkw:
### Crates

- `tonic`

### Motivation

Currently, Tonic's `Client` API requires users to write code like this:
```rust
    let request = tonic::Request::new(HelloRequest {
        name: "hello".into(),
    });

    let response = client.say_hello(request).await?;
```

The need to wrap the request struct in a `tonic::Request` is a minor ergonomics issue that could be surprising. This is admittedly a pretty minor papercut, but it seems like it would reduce some friction in the APIs that I suspect a new user is most likely to use (and make the basic examples seem simpler!)...

### Proposal

What do you think about changing the client API so that users could just write:
```rust
    let request = HelloRequest {
        name: "hello".into(),
    };

    let response = client.say_hello(request).await?;
```
We could add a trait like this:
```rust
pub trait IntoRequest {
    type ReqMessage;
    fn into_request(self) -> tonic::Request<Self::ReqMessage>
    where
        Self: Sized,
    ;
}
```
and generate impls for all the generated request messages, like
```rust
impl tonic::IntoRequest for #request {
    type ReqMessage = Self;
    fn into_request(self) -> tonic::Request<Self>
    where
        Self: Sized,
    {
        tonic::Request::new(self)
    }
}
```

Then, we could change the generated RPC methods on the client to be like this:
```rust
async fn #ident(&mut self, request: impl tonic::IntoRequest<ReqMessage = #request>) 
```
We could add an impl of `IntoRequest` for `Request<T>` like this:
```rust
impl<T> IntoRequest for Request<T> {
    type ReqMessage = T;
    fn into_request(self) ->Request<T>
    where
        Self: Sized,
    {
         self
    }
}
```
which would allow users could still construct `tonic::Request` types if needed, such as when manually adding headers etc.

Something similar could probably be done for response messages.

### Alternatives

We could also consider using `From` and `Into` here, rather than defining a new `IntoRequest` trait.  This is what I originally suggested in https://github.com/hyperium/tonic/issues/1#issuecomment-534328376. Ideally, I think it would be better to use the stdlib traits where possible, rather than defining a new one. However, @LucioFranco says that this causes some issues with type inference which a new trait could possibly avoid.

ChetanBhasin:
Hey! I'm fairly new to this project, but the library seems fantastic so far!

This looks like a nice first issue. I was hoping to take a crack at it sometime this week or over the weekend if a PR is welcome. 😃

[...]

PR #61, opened 2019-10-08
fix(client): Use `Stream` instead of `TrySteam` for client calls

alce:
## Motivation
Currently, client methods take `TryStreams` as inputs when it's not really necessary. This change makes client calls a bit simpler and potentially helps simplify a solution for #34.

## Solution
This PR is a POC that changes input stream bounds from `Stream<Item = Result<Message, Status>>` to `Stream<Item = Message>`. As is, client and servers work fine but interop tests need to be adjusted for this PR to be complete.


alce:
Updating interop was actually very easy. This is ready for review.

olix0r:
> Currently, client methods take TryStreams as inputs when it's not really necessary. 

Can you expand on that? Why isn't it necessary? How does a client get notified of a stream failure (with it's status/error message)?

alce:
Hi @olix0r, maybe the title is  misleading.

All this PR does is remove the need to wrap outgoing messages in `Result`s for client and bi-directional streams, responses do not change.

So instead of this:
```rust
client.method(stream::iter(vec![Ok(1), Ok(2), Ok(3)])).await?;
```
we can write this
```rust
client.method(stream::iter(vec![1, 2, 3])).await?
```

It is totally possible that I'm missing something but I don't understand the semantics of potentially sending an `Err(Status)` message as part of a streaming request. If you sneak in an Err(Status...) somewhere in the stream, the server just returns an `Unknown` error. I don't know what else it could do other than maybe echo the error back.

If the server tries to read all messages in the stream before returning a response, we don't even know how many messages, (or which ones) it managed to process.

olix0r:
Ah! I see: this is for _request streams_ and not _response streams_. That makes sense, then.

LucioFranco:
I think we might want to update this function to not take a result? https://github.com/hyperium/tonic/blob/master/tonic/src/codec/encode.rs#L32

alce:
I did update that one https://github.com/hyperium/tonic/pull/61/files#diff-3bc46617ab2d94a39c2f4f488aa3e803

LucioFranco:
@alce I must be tired I totally missed that! 😅

PR #66, opened 2019-10-10
feat: Add `IntoRequest` and `IntoStreamingRequest` traits

alce:
## Motivation
This PR implements `IntoRequest` and `IntoStreamingRequest` traits, as outlined in #34. At the moment, both unary and streaming requests work but interop crate needs to be updated for this PR to be complete. I'll wait for feedback and implement what is remaining if it's decided we should move forward with this.

## Solution
Implementation is almost exactly what is discussed in #34.


tonic/src/request.rs
```diff
@@ -1,4 +1,23 @@
 use crate::metadata::MetadataMap;
+use futures_core::Stream;
+
+/// A marker trait that should be implemented for all input messages.
+pub trait RequestMessage {}
```
LucioFranco: I don't think we need this marker trait since there is only ever one generic?

alce: You are right, it is not strictly necessary and a previous implementation did not have it. I decided to include it because:

- It sort of helps us constrain the other trait implementations. It just feels weird for IntoRequest or IntoStreamingRequest impl to exist for any type that is not an input to a rpc Method.
- I am toying around with an implementation that would let us call the streaming methods with iterators and having this bound sort of helps. Not sure if this will work of if it even makes sense.
- It makes code generation extremely simple, basically 1 line per input type.
- Triggered by the breakage in tonic-interop, I initially thought that it was going to be necessary to expose this marker for users to implement in certain scenarios, but that seems not to be necessary. I found the reason why interop is broken and has nothing to do with this.

None of these reasons hold much water though. In the end, however, this should be an internal implementation detail and as long as we are not changing the public api willy-nilly we should be able to iterate on the actual implementation.

LucioFranco: Yeah, those points make sense, my real motivation for removing it is that it's not actually required and just simplifies things. Happy either way.

alce:
I've fixed the issue with interop. This implementation seems to work well with different `.proto` setups. 

The only way I found to make it work without the `RequestMessage` trait is to generate `IntoRequest` implementations for all input types. I could not find a way to make a blanket implementation for T without the bound.

I'm happy to make that change if you prefer, it's easy.

LucioFranco:
@hawkw what do you think about having a `RequestMessage` trait to enforce the type that can be intoed?

LucioFranco:
and @alce sorry for the delay on this I just wanna see what eliza has to say since she came up with the original spec other wise this looks good!

tonic-build/src/client.rs
```diff
@@ -99,12 +99,14 @@ fn generate_unary(method: &Method, proto: &str, path: String) -> TokenStream {
     let (request, response) = crate::replace_wellknown(proto, &method);
 
     quote! {
-        pub async fn #ident(&mut self, request: tonic::Request<#request>)
-            -> Result<tonic::Response<#response>, tonic::Status> {
+        pub async fn #ident<R: tonic::IntoRequest<Message = #request>> (
+            &mut self,
+            request: R,
```
hawkw: Rather than using the `R` type parameter, this could just be
```rust
pub async fn #ident(&mut self, request: impl tonic::IntoRequest<Message = #request>)
     -> ...
```

tonic-build/src/client.rs
```diff
@@ -115,12 +117,14 @@ fn generate_server_streaming(method: &Method, proto: &str, path: String) -> Toke
     let (request, response) = crate::replace_wellknown(proto, &method);
 
     quote! {
-        pub async fn #ident(&mut self, request: tonic::Request<#request>)
-            -> Result<tonic::Response<tonic::codec::Streaming<#response>>, tonic::Status> {
+        pub async fn #ident<R: tonic::IntoRequest<Message = #request>>(
```
hawkw: Similarly, you could just use `impl tonic::IntoRequest` here

tonic-build/src/client.rs
```diff
@@ -131,14 +135,14 @@ fn generate_client_streaming(method: &Method, proto: &str, path: String) -> Toke
     let (request, response) = crate::replace_wellknown(proto, &method);
 
     quote! {
-        pub async fn #ident<S>(&mut self, request: tonic::Request<S>)
-            -> Result<tonic::Response<#response>, tonic::Status>
-            where S: Stream<Item = #request> + Send + 'static,
-        {
+        pub async fn #ident<S: tonic::IntoStreamingRequest<Message = #request>>(
+            &mut self,
+            request: S
```
hawkw: could use `impl tonic::IntoStreamingRequest<Message = #request>`

tonic/src/request.rs
hawkw: My 2¢ is that since the `Request` type itself is unconstrained, this shouldn't be added; if we _are_ going to constrain the `Message` type on `IntoRequest`, we should also constrain `Request`, and I'd consider adding that in a separate branch.

hawkw: Also, I think it's potentially very surprising to generate implementations of a `doc(hidden)` trait _in a user's codebase_ — I'd find it weird to run `cargo doc` on my project and see implementations of a trait that doesn't appear to exist.

alce: I removed this trait as suggested. IntoRequest for T is generated now instead, please let me know what you think.

LucioFranco:
@alce can you resolve the conflicts? I think CI may not be running because of that?

alce:
Goofed. Fixing...

tonic/src/request.rs
```diff
@@ -7,6 +8,23 @@ pub struct Request<T> {
...
+/// A trait that is implemented for all RPC request request types.
+pub trait IntoRequest {
+    /// The RPC request type
+    type Message;
+
+    /// Wrap `Message` in a `tonic::Request`
+    fn into_request(self) -> Request<Self::Message>;
+}
+
+#[doc(hidden)]
```
LucioFranco: why is this one doc hidden and the other is not?

alce: Because IntoRequest is explicitly generated and there are blanket implementations for this one. Should I expose this one as well?

LucioFranco: Yeah, I think it would make it less confusing no?

alce: Yep. Done.

alce:
I am having second thoughts about this PR.

 On the one hand, as Carl and Eliza pointed out, it makes writing client calls a lot nicer. Every time I write client code I moan about having to write Request::new() explicitly.

On the other hand, it just doesn't feel entirely right to me. I can't really put my finger on it but it makes the API sort of asymmetrical. Even if we were to implement something similar for Response::new() (which btw doesn't really bother me much).

Maybe if we had to call `.into_request()` and `.into_streaming_request()` it wouldn't feel so magical.

```rust
let message = HelloRequest{name: "hello".into()}
let response = client.say_hello(message.into_request())
```

Thoughts?

LucioFranco:
@alce I kinda agree with you, that actually might be better. What do you think @hawkw @carllerche ?

hawkw:
> Maybe if we had to call `.into_request()` and `.into_streaming_request()` explicitly it wouldn't feel so weird?.
> 
> ```rust
> let message = HelloRequest{name: "hello".into()}
> let response = client.say_hello(message.into_request())
> ```
> 
> Thoughts?

I don't really see much of a difference between making users write out `into_request` explicitly, and making them write `Request::new`? If we're going to require users to call the `into_request` method, I think it would be better to just close this PR and continue with the current state of things.

With that said, I don't think we should close this PR. I'd rather merge the code as-is, and call `into_request` implicitly. While having to call `Request::new` is not a _major_ issue, not having to does make the new user experience and basic examples much simpler, which I think is worth something. If the concern is  that
> [...] it makes the API sort of asymmetrical.

I think we should just make the same change on the server side and add an `IntoResponse` trait as well.

Just my two cents.

alce:
The comment about asymmetry was not so much about request vs response. The weirdness I get is  more about client vs server.

This client code
```rust
let point = Point { lat: 1, lon: 2 };
let response = client.get_feature(point).await?;
```
and this server's signature

```rust
async fn get_feature(&self, request: Request<Point>) -> Result<Response<...>;
```

are surprising, particularly when the client's function signature is not easily available, as is the case when following the default getting started.

But then, as I was writing this, I realized that:
 
* As things are now, you kinda need to hunt around for the generated code to get anything going beyond hello world, in which case people will see the sneaky IntoRequest in the client's signature.
* My gripe is, and always has been, with how deeply hidden the generated code is.

hawkw:
>   * As things are now, you kinda need to hunt around for the generated code to get anything going beyond hello world anyway, in which case people will see the sneaky IntoRequest in the client's signature.
> 
>   * My gripe is, and always has been, with how deeply hidden the generated code is.

@alce do you think we can replace the generated `IntoRequest` impls with blanket impls, like the ones @carllerche proposed in https://github.com/hyperium/tonic/issues/34#issuecomment-539296206? Then, the `impl IntoRequest for ...` would be in `tonic`'s docs, rather than in the generated code.

alce:
I couldn't figure out how to do it without a bound. That's one of the reasons why I had the `RequestMessage`  marker trait before, which was akin to Carl's `Message` trait in the comment. 

I will give it another go tomorrow to see if I can find another way.

LucioFranco:
I'm still not convinced that this indirection is worth it honestly. I think the `Request` type is a good line between too verbose and being just explicit enough.

alce:
@LucioFranco  I think I've found a more natural way to do it, I'll push the changes later today.

With regards to the so-called asymmetry I speak of, I think I'm splitting hairs and considering only my own workflow and bias. In a larger, more realistic context, this is almost irrelevant.

alce:
Here's a cleaner implementation that combines approaches discussed before. It introduces only one new trait and does not generate any code.

tonic/src/request.rs
```diff
@@ -7,6 +8,18 @@ pub struct Request<T> {
     message: T,
 }
 
+/// A trait that is implemented for all streaming RPC request types.
+pub trait IntoStreamingRequest {
```
hawkw: Should this be sealed to prevent it from being implemented downstream?

tonic-build/src/client.rs
```diff
@@ -115,12 +117,14 @@ fn generate_server_streaming(method: &Method, proto: &str, path: String) -> Toke
     let (request, response) = crate::replace_wellknown(proto, &method);
 
     quote! {
-        pub async fn #ident(&mut self, request: tonic::Request<#request>)
-            -> Result<tonic::Response<tonic::codec::Streaming<#response>>, tonic::Status> {
+        pub async fn #ident(
+            &mut self,
+            request: impl Into<tonic::Request<#request>>,
```
hawkw: I'm on the fence as to whether these should use `Into` or if we should still add an `IntoRequest` trait with blanket impls. I _think_ I prefer just using `Into`, but having an `IntoRequest` trait might be more consistent with `IntoStreamingRequest`...

tonic/src/request.rs
```diff
@@ -7,6 +8,18 @@ pub struct Request<T> {
     message: T,
 }
 
+/// A trait that is implemented for all streaming RPC request types.
```
hawkw: This doc comment should probably also describe what the trait is _used_ for, not just what types implement it. Something like
> Types implementing this trait may be used as an argument to generated client-streaming RPC methods

tonic-build/src/client.rs
alce: I was having trouble with overlapping trait implementations for an unbounded T and Request<T> but I _think_ this works as a replacement for From/Into, without generating any code:

```rust
pub trait IntoRequest<T> {
    fn into_request(self) -> Request<T>;
}

impl<T> IntoRequest<T> for T {
    fn into_request(self) ->  Request<Self> {
        Request::new(self)
    }
}

impl<T> IntoRequest<T> for Request<T> {
    fn into_request(self) -> Request<T> {
        self
    }
}
```
We could do that too, unless @LucioFranco is not happy?

hawkw: I could go either way on this, honestly.

LucioFranco: I'd prefer to have both `IntoRequest` and `IntoStreamingRequest` personally.

alce: Done

tonic/src/request.rs
```diff
@@ -7,6 +8,30 @@ pub struct Request<T> {
...
+    type Stream: Stream<Item = Self::Message> + Send + 'static;
+
+    /// The RPC request type
+    type Message;
+
+    /// Wrap the stream of messages in a `tonic::Request`
+    fn into_streaming_request(self) -> Request<Self::Stream>;
+}
+
+/// Conversion into a unary `Request`.
```
LucioFranco: Can we add a bit more docs here to explain how this works? I am thinking maybe an example that shows constructing a request and passing it and one with the message explaining the trait?

LucioFranco: Could use https://doc.rust-lang.org/stable/std/convert/trait.Into.html as inspiration

alce: Done. Docs are slightly better but I'm not sure they are clear enough.

tonic/src/request.rs
```diff
@@ -7,6 +8,84 @@ pub struct Request<T> {
...
+/// # pub struct Point {}
+/// # pub struct Client {}
+/// # impl Client {
+/// #   fn get_feature(&self, r: impl tonic::IntoRequest<Point>) {}
+/// # }
+/// # let client = Client {};
+/// use tonic::Request;
+///
+/// client.get_feature(Point {});
+/// client.get_feature(Request::new(Point {}));
```
LucioFranco: 😍

PR #74, opened 2019-10-15
Add rustfmt.toml with edition=2018

cpcloud:


LucioFranco:
@cpcloud hey! thanks for this, what was the reason again to add this?

cpcloud:
Oh because rustfmt is conservative about what edition it chooses and doesn't look at `Cargo.toml`, so it fails on some parts of the codebase that use the `async` keyword.

LucioFranco:
Fair enough!

PR #77, opened 2019-10-16
feat(transport): Add support client mTLS

jen20:
This commit adds a simple API for specifying the TLS certificate a gRPC client will present (via the same `Identity` wrapper as a server cert is configured). It also adds an API to specify which CA certificate client TLS certificates will be validated against for servers.

To demonstrate usage, a new example `tls_client_auth` is added. In order to implement this, new test certificates are added to the `data` directory of `tonic-examples`. Currently only one of the two added client certificates is used, the other is intended for future work but since it was issued by the same CA, it's helpful to add it now.

Some of the code for loading certificates and private keys previously existed in the `TlsAcceptor` class - this has been factored out to a new module so it can be reused by `TlsConnector` also.

Currently this is only implemented for `rustls` - an OpenSSL implementation will follow on this PR shortly.


tonic-examples/src/tls_client_auth/server.rs
```diff
@@ -0,0 +1,74 @@
...
+#[tonic::async_trait]
+impl pb::server::Echo for EchoServer {
+    async fn unary_echo(&self, request: Request<EchoRequest>) -> EchoResult<EchoResponse> {
+        let message = request.into_inner().message;
+        Ok(Response::new(EchoResponse { message }))
+    }
+
+    type ServerStreamingEchoStream = Stream;
+
+    async fn server_streaming_echo(
```
LucioFranco: btw to simplify this you can just drop all these unimplemented since codegen now provides a default impl that does this.

jen20: Oh, good to know - I'll do that.

jen20: I've pushed a new commit which removes the unnecessary implementations. It does seem that the associated types still need to be defined, however?

tonic-examples/Cargo.toml
```diff
@@ -44,8 +44,16 @@ path = "src/tls/client.rs"
...
+name = "tls-client-auth-server"
+path = "src/tls_client_auth/server.rs"
+
+[[bin]]
+name = "tls-client-auth-client"
+path = "src/tls_client_auth/client.rs"
+
 [dependencies]
-tonic = { path = "../tonic", features = ["rustls"] }
+tonic = { path = "../tonic", features = ["rustls", "openssl"] }
```
LucioFranco: whats the reason to use openssl here?

tonic/src/transport/service/tls.rs
```diff
@@ -220,50 +246,28 @@ impl TlsAcceptor {
...
-        let cert = {
-            let mut cert = std::io::Cursor::new(&identity.cert.pem[..]);
-            match pemfile::certs(&mut cert) {
-                Ok(certs) => certs,
-                Err(_) => return Err(Box::new(TlsError::CertificateParseError)),
-            }
-        };
+                let mut client_root_cert_store = tokio_rustls::rustls::RootCertStore::empty();
+                match client_root_cert_store.add_pem_file(&mut cert) {
+                    Err(_) => return Err(Box::new(TlsError::CertificateParseError)),
```
LucioFranco: do we want to attach the error message here?

jen20: I'll update this - we likely do want the error message.

tonic-examples/src/tls_client_auth/server.rs
LucioFranco: Yeah, we don't have default associated types yet unfortunetly

tonic/src/transport/service/tls.rs
jen20: Only kidding... the error type is `()`.

tonic-examples/Cargo.toml
jen20: Fixed.

jen20:
OK, I think these are all resolved!

PR #82, opened 2019-10-21
Make Streaming<T> Sync

### LINKED ISSUE #81: Consider making `Streaming<T>` Sync
### Crates

Main Crate

### Motivation

I'm trying to output a unidirectional tonic gRPC stream via SSE using warp. I'm using warp from https://github.com/seanmonstar/warp/pull/265 to be able to use it with async await. Changing warp to be compatible with async / await requires that all responses handed over to warp (and in turn handed over to hyper) to be `Sync`. However, the stream returned from a streaming gRPC response is not `Sync`.

### Proposal

Rustc complains that the `decoder` specifically is the part that is not sync: https://github.com/hyperium/tonic/blob/5d0a7955541509d2dbfdb9b689fb57cd2b842172/tonic/src/codec/decode.rs#L22

As far as I can see, `decoder` is not accessible from the outside, and thus it should be safe to mark the whole `Streaming` struct as `Sync`. I would be happy to file a PR for that.

### Alternatives

I don't see any, because AFAIK the `Sync` bound stems from a layer deep down inside `hyper`, and I guess it's unfeasible to change that somehow to remove it.

NeoLegends:
Safe because there are no `&self` methods on `Streaming<T>` that access any of its non `Sync` members.

Fixes #81
Closes #84

### DIFF
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -27,6 +27,7 @@ pub struct Streaming<T> {
     trailers: Option<MetadataMap>,
 }
 
+unsafe impl<T> Sync for Streaming<T> {}
 impl<T> Unpin for Streaming<T> {}
 
 #[derive(Debug)]


LucioFranco:
So I'm not sure this is the correct approach, we should instead add them via bounds.

NeoLegends:
Closes #84

LucioFranco:
done in #84 thank you for the help <3

PR #83, opened 2019-10-21
fix(codec): Properly decode partial DATA frames

### LINKED ISSUE #76: Decoding occasionally fails when streaming
### Version

Running from 1987a40.

### Platform

```
Linux nexus 5.3.1 #1-NixOS SMP Sat Sep 21 05:19:47 UTC 2019 x86_64 GNU/Linux
```

### Crates

Affects `tonic`

### Description

When running [these lines](https://github.com/hyperium/tonic/commit/1987a40586cf45eff867d755e7b7e2ac2b9b8046#diff-929f8d2640458ec49964beaf3ecd1f10R8-R27) I occasionally get an error message from `prost` informing me that a buffer underflow has occurred:

```
Error: Status { code: Internal, message: "failed to decode Protobuf message: buffer underflow" }
```
Sometimes this happens, other times it doesn't. It seems to be related to the number of bytes that I try to push through from the server to the client.

After digging around for quite some time, the failing requests all seem to be missing 3 additional bytes after the 5 byte header. When testing the `blob-client` with `10000` bytes a request ends up having a total length of `10005`. After the buffer is advanced 5 bytes, the length is now `10000`. The [`prost::encoding::bytes::merge`](https://github.com/danburkert/prost/blob/master/src/encoding.rs#L855) function then reads off 3 bytes to get the length of the byte array from the proto, which is `10000` and then [checks for underflow](https://github.com/danburkert/prost/blob/master/src/encoding.rs#L856-L858).

I'm stumped as to how these bytes are missing in some cases and are present in many others.

This prost issue (https://github.com/danburkert/prost/issues/98) suggests using `{encode|decode}_length_delimited`, but that didn't seem to have any effect when I swapped those calls in in `prost.rs`.

cpcloud:
Closes #76

TODO:

- [x] Add mock to reproduce in a unit test

## Motivation

Reproduced from #76:

> When receiving data by polling the body, it's possible to end up in a
> situation where the polling results in a buffer of size `N` where `N >=
> length`, but _also_ that `N < 5 + length`.
>
> Roughly the flow is like this:
>
> 1. Tonic polls for data, gets `N` bytes, where `length <= N < 5 + length`
> 1. Tonic sees that `N` bytes have been read, and if `N >= length`, _advances the buffer by 5 bytes
> 1. Advancing the buffer reduces the buffer length by 5, now we have that `length - 5 <= N < length`
> 1. Since we need `length` bytes for the proto, prost fails with a buffer underflow

## Solution

The solution that I've implemented here is to check that `self.buf`--the buffer
that holds the bytes for decoding--has enough data to decode both the tonic
header and the message. If the check returns `false`, wait for more data,
otherwise decode the bytes


tonic-examples/Cargo.toml
```diff
@@ -54,6 +62,8 @@ futures-preview = { version = "=0.3.0-alpha.19", default-features = false, featu
 async-stream = "0.1.2"
 http = "0.1"
 tower = "=0.3.0-alpha.2"
+env_logger = "*"
+log = "*"
```
cpcloud: I'll remove these.

tonic-examples/proto/blobservice/blobservice.proto
```diff
@@ -0,0 +1,29 @@
+// Copyright 2015 gRPC authors.
```
cpcloud: This isn't accurate. Should I remove the copyright notice entirely?

tonic-examples/src/blobservice/client.rs
```diff
@@ -0,0 +1,26 @@
...
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut client = BlobberClient::connect("http://[::1]:50051")?;
+    let nbytes = std::env::args().nth(1).unwrap().parse::<u64>().unwrap();
+    let request = tonic::Request::new(BlobRequest { nbytes });
+
+    let response = client.get_bytes(request).await?;
+    let mut inner = response.into_inner();
+    let mut i = 0_u64;
+    while let Some(_) = inner.try_next().await.map_err(|e| {
```
LucioFranco: this should use https://docs.rs/tonic/0.1.0-alpha.3/tonic/struct.Streaming.html#method.message

tonic/src/codec/decode.rs
```diff
@@ -32,7 +35,10 @@ impl<T> Unpin for Streaming<T> {}
 #[derive(Debug)]
 enum State {
     ReadHeader,
-    ReadBody { compression: bool, len: usize },
+    ReadBody {
+        is_compressed: bool,
+        message_length: usize,
```
LucioFranco: I think this follows rust a bit better :)

```suggestion
        message_len: usize,
```

tonic-examples/proto/blobservice/blobservice.proto
```diff
@@ -0,0 +1,29 @@
...
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
lovesegfault: This is also not accurate, I think. It's Apache OR MIT dual-licensed.

tonic/src/codec/mod.rs
```diff
@@ -3,8 +3,17 @@
 //! This module contains the generic `Codec` trait and a protobuf codec
 //! based on prost.
 
+// The mininum buffer size for codec data
+pub(crate) const BUFFER_SIZE: usize = 8 * 1024;
+
+// The number of bytes in the tonic header:
+// 1 * u8 for compression +
+// 1 * u32 for body length
+pub(crate) const HEADER_SIZE: usize = 5;
```
lovesegfault: If `mem::sizeof` is a const fn (I don't recall) couldn't you use that instead?

tonic-examples/proto/blobservice/blobservice.proto
lovesegfault: As an addendum, the right thing to do if license headers are desired is to use SPDX format ones, otherwise the LICENSE in the repo works just fine.

LucioFranco:
@cpcloud so if I understand correct the issue is that we may not have the initial 5 bytes from the buffer handed to us from hyper? Thus the `buf.advance(5)` fails?

cpcloud:
@LucioFranco No, the issue is that we'll receive the header + a partial body. For example:

```
> read 10005 bytes from the body
  > advance 5
  > 10000 left over, but the message length is actually 10003
```
We need to ensure that `self.buf` has enough of the body to decode the full message including the header.

tonic-examples/proto/blobservice/blobservice.proto
cpcloud: I'll just blow it away.

tonic/src/codec/mod.rs
cpcloud: Looks like it is: https://doc.rust-lang.org/std/mem/fn.size_of.html

LucioFranco:
Ok so after our discussion on discord I am happy with this solution. I hate to ask this but would it be possible to try and make this change with the smallest footprint possible which I think is just the second if statement in readbody. And instead of introducing a new example service can we add a test, there should already be tests for the codec. In the mock body simulate that the frames are split.

LucioFranco:
and thank you so much for figuring this out! It is super helpful, I'd like to get this change in before the next release this week. Let me know if you need anything.

cpcloud:
@LucioFranco Looks like the `decode` test never actually reads anything! :joy: Each `poll_data()` call is checking whether `0 > 10`.

cpcloud:
I'll add a fix in this PR.

tonic/src/codec/tests.rs
```diff
@@ -23,7 +23,8 @@ struct Msg {
 async fn decode() {
     let decoder = ProstDecoder::<Msg>::default();
 
-    let data = Vec::from(&[0u8; 1024][..]);
+    let data = Vec::from(&[0u8; 10000][..]);
```
LucioFranco: You can do this, which I just found out recently :)
```suggestion
    let data = vec![0u8; 10000];
```

cpcloud: Oh, sweet.

cpcloud:
Yeah!

PR #84, opened 2019-10-21
fix(codec): Enforce encoders/decoders are `Sync`

### LINKED ISSUE #81: Consider making `Streaming<T>` Sync

LucioFranco:
cc @NeoLegends

Closes #81


LucioFranco:
Yeah...this will be a bit harder than anticipated...

NeoLegends:
Yeah, this is the other option. I guess you're in the best position to know which bounds you will need and you will be able to satisfy in tonic.

For this implementation route we'd also want to add a test that ensures `Streaming` _stays_ sync (when adding new members).

LucioFranco:
@seanmonstar would you be able to take a look at this?

seanmonstar:
What I'm about to say sounds dumb: for trait objects, the order of additional auto traits matters, and most places write them as `dyn Foo + Send + Sync`. That's to say, `dyn Foo + Sync + Send` isn't the same thing. 😭

LucioFranco:
@seanmonstar all i can say is lol

NeoLegends:
Does this put #82 back on the table?

LucioFranco:
@seanmonstar can I get your +1 on this before we merge?

PR #95, opened 2019-10-23
fix(build): Fix missing argument in generate_connect

### LINKED ISSUE #94: tokio-build fails with "transport" feature disabled
### Version

0.1.0-alpha.4

### Platform

Linux 5.0.0-29-generic Ubuntu SMP Thu Sep 12 13:05:32 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

### Crates

tonic-build

### Description

tonic-build fails to compile with the "transport" feature disabled, throwing the following error:

```
error[E0061]: this function takes 0 parameters but 1 parameter was supplied
  --> /home/.../.cargo/registry/src/github.com-1ecc6299db9ec823/tonic-build-0.1.0-alpha.4/src/client.rs:10:19
   |
10 |     let connect = generate_connect(&service_ident);
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 0 parameters
...
69 | fn generate_connect() -> TokenStream {
   | ------------------------------------ defined here
```

I took a quick glance at the affected code, and it looks like it should be a quick fix :)

daniel5151:
## Motivation

Closes #94.

It's literally a single line edit, so might as just do it myself :smile: 

## Solution

Fixes `generate_connect`'s method signature when the "transport" feature is disabled.

### DIFF
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -66,7 +66,7 @@ fn generate_connect(service_ident: &syn::Ident) -> TokenStream {
 }
 
 #[cfg(not(feature = "transport"))]
-fn generate_connect() -> TokenStream {
+fn generate_connect(_service_ident: &syn::Ident) -> TokenStream {
     TokenStream::new()
 }
 


LucioFranco:
@daniel5151 is this something you would like released asap or can you live off of git for a bit?

daniel5151:
Nope, no rush!

On a somewhat unrelated note, I actually encountered this issue while tinkering with seeing if I could write an implementation of GrpcService that would work over unix sockets (to communicate with containerd). It's unlikely i'll peruse it though (I found a bodge that will get the job done), but I _am_ keeping an eye on #31 :eyes:

LucioFranco:
@daniel5151 I'd suggest using https://docs.rs/hyper/0.13.0-alpha.4/hyper/client/service/struct.Connect.html where the inner C type is a `tower_service::Service` that returns a UDS instead of a TCPStream.

daniel5151:
Yeah, that would be the Right Way to do it, through it turns into a bit of a yak-shave, since [hyperlocal](https://github.com/softprops/hyperlocal) hasn't been updated for 0.13 yet, so I'd have to port it forward.

Issue #93, opened 2019-10-23
Reduce dependencies


LucioFranco:
We should attempt to use feature flags to reduce dependencies that we don't use.

blittable:
Quick question here - feature flags to narrow the import scope of the upstream dependencies or features flags in the tonic code (e.g. put cfg attributes on handful of trace crate usages?)

I looked at the upstream dependencies, but didn't see any good opportunities to factor out features.

LucioFranco:
@blittable as of right now there are not many but that is because we pull in the whole world with the `transport` feature. I think we can revisit this issue once tokio/hyper get a stable release.

PR #97, opened 2019-10-27
failing unit test / tls feature import for unit test

blittable:
Was looking into #93 and saw a couple failing unit tests in the main Tonic crate -  on the import of ServerTlsConfig and ClientTlsConfig... added cfg attribute.


blittable:
tests to be run from the root directory work fine.

PR #99, opened 2019-10-28
feat(transport): Add service multiplexing/routing

### LINKED ISSUE #29: Server side service level Routing
Tracking issue for request Routing

Tonic should allow a single socket/server to serve multiple services. Routing is the way to get there.

Current progress: https://github.com/hyperium/tonic/blob/master/tonic-interop/src/bin/server.rs#L73

See also: https://github.com/tower-rs/tower-grpc/issues/2

LucioFranco:
This change introduces a new "router" built on top of
`transport::Server` that allows one to run multiple
gRPC services on the same socket.

### Example

```rust
Server::builder()
    .add_service(greeter)
    .add_service(echo)
    .serve(addr)
    .await?;
```

There is also a new `multiplex` example showcasing
server side service multiplexing and client side
service multiplexing.

BREAKING CHANGES: `Server::serve` is now crate private
and all services must be added via `Server::add_service`.
Codegen also returns just a `Service` now instead of a
`MakeService` pair.

Closes #29

cc @rlabrecque @seanmonstar @carllerche 

Signed-off-by: Lucio Franco luciofranco14@gmail.com


tonic-examples/src/tls_client_auth/client.rs
```diff
@@ -13,7 +13,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let client_key = tokio::fs::read("tonic-examples/data/tls/client1.key").await?;
     let client_identity = Identity::from_pem(client_cert, client_key);
 
-    let tls = ClientTlsConfig::with_openssl()
+    let tls = ClientTlsConfig::with_rustls()
```
rlabrecque: Was this supposed to change in this PR?

LucioFranco: Yes, but I think it was missed in another PR so I thought i might as well change this now. I want all examples with rustls :)

We really only need openssl for interop since the certs that grpc-go uses are too small for the min that rustls supports.

tonic/src/transport/server.rs
```diff
@@ -149,14 +164,26 @@ impl Server {
...
     }
 
-    /// Consume this [`Server`] creating a future that will execute the server
-    /// on [`tokio`]'s default executor.
-    pub async fn serve<M, S>(self, addr: SocketAddr, svc: M) -> Result<(), super::Error>
+    /// Create a router with the `S` typed service as the first service.
+    ///
+    /// This will clone the `Server` builder and create a router that will
+    /// route around different services.
+    pub fn add_service<S>(&mut self, svc: S) -> Router<S, Unimplemented>
```
hawkw: Why does this need to take `&mut self` if it immediately clones `self`?

tonic/src/transport/server.rs
```diff
@@ -421,3 +511,27 @@ where
         })
     }
 }
+
+#[derive(Default, Clone, Debug)]
+#[doc(hidden)]
+pub struct Unimplemented;
```
hawkw: This should probably have a single private field to protect against users accidentally constructing it...

tonic/src/transport/service/router.rs
```diff
@@ -0,0 +1,128 @@
...
+
+impl<A: Clone, B: Clone, Request> Clone for Routes<A, B, Request> {
+    fn clone(&self) -> Self {
+        Self {
+            routes: self.routes.clone(),
+        }
+    }
+}
+
+pub struct Or<A, B, Request> {
```
hawkw: Should probably either have docs or be `doc(hidden)`...

tonic/src/transport/service/router.rs
```diff
@@ -0,0 +1,128 @@
...
+            predicate: self.predicate.clone(),
+            a: self.a.clone(),
+            b: self.b.clone(),
+        }
+    }
+}
+
+impl<A, B, Request> fmt::Debug for Or<A, B, Request> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Or")
```
hawkw: might be better to format as `"Or { ... }` so it's still clear that it's a struct, a lot of types in std do that...

tonic/src/transport/server.rs
```diff
@@ -421,3 +511,27 @@ where
...
+impl Service<Request<Body>> for Unimplemented {
+    type Response = Response<BoxBody>;
+    type Error = crate::Error;
+    type Future = future::Ready<Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Ok(()).into()
+    }
+
+    fn call(&mut self, _req: Request<Body>) -> Self::Future {
```
hawkw: do you think this should record a tracing event as well? that way, users get some visibility into when requests hit the unimplemented route (in logs, metrics, etc...)

tonic/src/transport/service/router.rs
```diff
@@ -0,0 +1,128 @@
...
+    ) -> Self {
+        let routes = Or::new(predicate, a, b);
+        Self { routes }
+    }
+}
+
+impl<A, B, Request> Routes<A, B, Request> {
+    pub(crate) fn push<C>(
+        self,
+        predicate: impl Fn(&Request) -> bool + Send + Sync + 'static,
```
hawkw: It might be worth considering a `Predicate` trait so that users can add more complex filtering? Probably not in this branch though.

tonic/src/transport/server.rs
LucioFranco: Because you can create multiple routers from one `Server` config was the idea. It takes a &mut because it follows like the others.

tonic/src/transport/server.rs
LucioFranco: Ummm probably a good idea, but I'd like to add tracing in another PR.

tonic/src/transport/service/router.rs
LucioFranco: Yeah, once we get a stable release of tokio, im gonna update a bunch of the tower services and will impl this there.

PR #105, opened 2019-10-31
chore: Add `cargo-hack` to check features

LucioFranco:
r? @taiki-e this crate is awesome :) 

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -29,7 +29,7 @@ jobs:
     - name: Check fmt
       run: cargo fmt -- --check
     - name: Check features
-      run: cargo hack check --all --each-feature
+      run: cargo hack check --all --each-feature --no-dev-deps
     - name: Check all targets
       run: cargo check --all --all-targets --all-features
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,18 +24,14 @@ jobs:
     - uses: actions/checkout@master
     - name: Install rustfmt
       run: rustup component add rustfmt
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -24,18 +24,14 @@ jobs:
...
-    - name: Check with transport no tls
-      run: cargo check -p tonic
-    - name: Check with transport w/ openssl
-      run: cargo check -p tonic --features openssl
-    - name: Check with transport w/ rustls
-      run: cargo check -p tonic --features rustls
+    - name: Check features
+      run: cargo hack check --all --each-feature
+    - name: Check all targets
+      run: cargo check --all --all-targets --all-features
```
LucioFranco: I need to add the `all-targets` or else it will miss benchmarks, do you think there might be a better solution @taiki-e?

taiki-e: Hmm, I'm not sure if there is another good solution. (afaik, `--all-targets` is basically [the same as `--lib --bins --examples --tests --benches`](https://github.com/rust-lang/cargo/blob/6a7f505a185b000e38bdad64392098e0b2e50802/src/cargo/ops/cargo_compile.rs#L551-L560), and note that the document is not checked (https://github.com/rust-lang/cargo/issues/6669)

LucioFranco: if we run it last it should be fine I can wait till we fine a better solution.

LucioFranco:
@taiki-e thank you so much for this crate, its what I very much wanted but had no clue how to express what I wanted. <3

Issue #101, opened 2019-10-29
Usability with TLS trust anchors


xmclark:
## Bug Report

This bug reports some usability issues with TLS in `tonic`. 

### Version
tonic = "0.1.0-alpha.3"
tonic-build = "0.1.0-alpha.3"

### Platform

Windows 64-bit 

### Description

When building a grpc client with `tonic` and `rustls` flag, without specifying a certificate in the builder, I expected `tonic` to find my server's certificate in the root store, but instead I get some pretty cryptic error messages out of `webpki` crate.

I did some digging, and even opened an issue on `webpki`. The short of it: Tonic is able to build a full certificate chain. That's about all I know about that topic! I really know nothing about it. What fixed it for me was adding a line like this when creating my client config:

`tls.root_store.add_server_trust_anchors(&webpki_roots::TLS_SERVER_ROOTS);`

I don't know exactly what this line does, but it does fix my problem. It does require a new dependency on the `webpki_roots` crate. 

I could set the  trust anchors manually, but it would circumvent a lot of the boilerplate tonic provides around `rustls` and `ClientConfig`. 

---

Does it make sense to add this behavior and dependency to `tonic`? My first reaction is no, but this seems like something others might trip on, and it would be nice if `tonic` exposed this for free for users like me that need it. Perhaps behind a feature flag? I also don't know if OpenSSL operates the same way and would take advantage of the feature in the same way. I wasn't able to try it because OpenSSL linking is a bit awkward on windows. 

I'm happy to PR a change in!

LucioFranco:
This is a really good question! So most gRPC examples I've seen have their own ca roots so I have not really seen users require this but I think its something we might be able to support. I'm not opposed to adding an additional feature flag around adding default roots. So I believe the `webpki_roots::TLS_SERVER_ROOTS` is actually just the mozilla one, where as I believe with openssl you can probe the system for its certs. I'm not sure if we should expose this as a builder since they can vary? I'm not the biggest expert on TLS so not sure what the right path is.

@jen20 any thoughts?

jen20:
I think we should definitely expose a way to make this simpler - I'll take a look into some of the options and have a think about what could be done.

xmclark:
I spent some time this week to learn a bit about TLS and certificate trust chains. I found the docs at Let's Encrypt to be super useful!
https://letsencrypt.org/certificates/

[...]

PR #114, opened 2019-11-03
feat(transport): Add system root anchors for TLS

jen20:
As per #101, it is sometimes desirable to use standard web PKI roots for gRPC clients. This commit adds a method to `ClientTlsConfig` to allow this. The behaviour differs per TLS library:

- OpenSSL uses `openssl-probe` to search the system for roots and add them.
- ~Rustls adds the Mozilla-supplied roots from the `webpki-roots` crate.~

~This is not feature flagged, as there appears to be no convenient way to gate a dependency on multiple conditions.~

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -34,6 +34,8 @@ transport = [
 ]
 openssl = ["openssl1", "tokio-openssl", "tls"]
 rustls = ["tokio-rustls", "tls"]
+openssl-roots = ["openssl-probe"]
+rustls-roots = ["rustls-native-certs"]
 tls = []
 
 [[bench]]
@@ -73,9 +75,11 @@ tower-load = { version = "=0.3.0-alpha.2", optional = true }
 # openssl
 tokio-openssl = { version = "=0.4.0-alpha.6", optional = true }
 openssl1 = { package = "openssl", version = "0.10", optional = true }
+openssl-probe = { version = "0.1", optional = true }
 
 # rustls
 tokio-rustls = { version = "=0.12.0-alpha.5", optional = true }
+rustls-native-certs = { version = "0.1", optional = true }
 
 [dev-dependencies]
 static_assertions = "1.0"
--- a/tonic/src/transport/endpoint.rs
+++ b/tonic/src/transport/endpoint.rs
@@ -259,18 +259,27 @@ impl ClientTlsConfig {
     }
 
     /// Sets the domain name against which to verify the server's TLS certificate.
+    ///
+    /// This has no effect if `rustls_client_config` or `openssl_connector` is used to configure
+    /// Rustls or OpenSSL respectively.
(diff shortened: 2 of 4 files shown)


tonic/Cargo.toml
```diff
@@ -32,8 +32,8 @@ transport = [
     "tower-balance",
     "tower-load",
 ]
-openssl = ["openssl1", "tokio-openssl", "tls"]
-rustls = ["tokio-rustls", "tls"]
+openssl = ["openssl1", "tokio-openssl", "openssl-probe", "tls"]
+rustls = ["tokio-rustls", "tls", "webpki-roots"]
```
LucioFranco: I think this might make more sense? So that users could enable it like so: `["rustls", "rustls-roots"]`. I would probably turn this off by default? Not sure.
```toml
rustls-roots = ["webpki-roots"]
openssl-roots = ["openssl-probe"]
```

jen20: That seems like a reasonable way of dealing with it actually, I can use the same for client identity. Something of a shame Cargo doesn't allow for dependencies iff two features are specified at the same time.

LucioFranco: Yeah that part is unfortunate but I think if we document it, its an easy cost to pay.

tonic/src/transport/endpoint.rs
```diff
@@ -213,6 +213,9 @@ pub struct ClientTlsConfig {
     domain: Option<String>,
     cert: Option<Certificate>,
     identity: Option<Identity>,
+    #[cfg(feature = "openssl-roots")]
```
LucioFranco: I think in this case you want `cfg(or(feature = "opnessl-roots", feature = "rustls-roots"))`

tonic/src/transport/endpoint.rs
```diff
@@ -259,18 +265,47 @@ impl ClientTlsConfig {
     }
 
     /// Sets the domain name against which to verify the server's TLS certificate.
+    ///
+    /// This has no effect if `rustls_client_config` or `openssl_connector` is used to configure
```
LucioFranco: so I wonder if there is a way we can make this actually work? From what I can tell everything just takes a &mut self so you can in theory use the passed config and just apply these changes to it. If they don't pass a config then its just the base version that we apply changes too?

jen20: As discussed I've opened #116 to discuss this further since it's likely a separate change.

Adding these comments documents the current behaviour, which in the first design iteration of this PR was important. We could remove the docs now we don't add any new builder methods, but I don't think the extra clarity harms anything.

jen20:
I've updated this in light of the new `rustls-native-certs` library for `rustls`, so the behaviour is now much more similar between the supported TLS libraries.

jen20:
Fixes #101.

jen20:
Docs for the new feature flags are now added in `tonic/src/lib.rs`.

PR #120, opened 2019-11-06
feat(transport): Enable TCP_NODELAY.

dominiquelefevre:
The combination of the Nagle algorithm on a client, and delayed ack on a server
can introduce up to 200ms latency if a small gRPC messages is sent with multiple
writes.

Enable TCP_NODELAY to make sure that neither client, nor server buffer their
messages for too long when exchaning small requests.

### DIFF
--- a/tonic/src/transport/server.rs
+++ b/tonic/src/transport/server.rs
@@ -427,7 +427,8 @@ struct TcpIncoming {
 
 impl TcpIncoming {
     fn bind(addr: SocketAddr) -> Result<Self, crate::Error> {
-        let inner = conn::AddrIncoming::bind(&addr).map_err(Box::new)?;
+        let mut inner = conn::AddrIncoming::bind(&addr).map_err(Box::new)?;
+        inner.set_nodelay(true);
 
         Ok(Self { inner })
     }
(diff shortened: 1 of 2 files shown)


LucioFranco:
@dominiquelefevre I think this is a good idea to add but I'd like to look into what other grpc libraries do as defaults. I am curious if grpc-go's http2 impl sets nodelay by default, same with netty?

dominiquelefevre:
grpc-go sets nodelay by default, and I do not know of any way to disable that.

LucioFranco:
@dominiquelefevre ok that makes sense, I am happy to merge this but maybe in a follow up PR we should probably provide this as a config option as well.

dominiquelefevre:
@LucioFranco, I am opposed to a config option.

TCP_NODELAY is required for services that exchange small requests. If requests are big, sending one more incomplete TCP packet with a http request prologue introduces no noticeable overhead, thus it makes no sense to disable TCP_NODELAY in such scenarios. So, just keep the API minimal.

LucioFranco:
@dominiquelefevre no I agree, what I mean't was to enable it by default and let users disable it. Not a high priority and we don't have to do anything until someone needs it. Thanks for all the help! :)

PR #121, opened 2019-11-07
fix(build): Prevent duplicated client/server generated code

steele:
## Motivation
The tonic-build process collects up RPC services as they are provided by prost, before writing them out as part of the finalization step for a given protocol buffer package.

In the case of imported protocol buffer packages, there may be RPC services included by import in addition to those in the top-level package. Therefore it is necessary to make sure each set of client/server services gathered by tonic-build is cleared after the finalization process for a given protocol buffer package, otherwise they will be incorrectly aggregated as the generation process proceeds through the subsequent packages.

This incorrect behaviour is currently observed in tonic-build.

## Solution
Reset the `TokenStream`s that capture the services for client and server use-cases after the finalization process has written out the generated code.


steele:
> Thanks for this! I think you may need to rebase against master since we are updating our toolchain to `stable`.

Should be rebased to master that uses the stable toolchain now

steele:
👍 thanks @LucioFranco!

rlabrecque:
Nice work!

PR #124, opened 2019-11-08
Use Stable Rust in CI

jen20:
This PR updates GitHub Actions to use stable Rust. It is complementary to #122 which updates the documentation around this.


jen20:
I've updated this now that #122 is merged.

PR #125, opened 2019-11-09
Use `rustls` for interop tests

jen20:
This commit changes the interop tests to use rustls instead of openssl. Apparently in the past there was some issue with this, but it seems to work OK (locally) to me.


jen20:
Evidently that is not actually the case, guess I had stale binaries or something.

jen20:
The reason for switching out the certificates here is that the upstream repo (from which the current set of certs originated) has 1024-bit keys, which `rustls` does not support.

PR #132, opened 2019-11-10
fix(transport): Update builders to move self

### LINKED ISSUE #115: Move builders to be move by value
The `http` crate with its `0.2` release will move its builders to be from `&mut self` to `self`. I want to consider following suit with `tonic`. Mostly, I wanted to open this to start a discussion.

cc @jen20

alce:
This patch changes builders to take self by value.

closes #115


jen20:
Hi @acle! Looks like there are a few conflicts after some changes upstream, might be worth rebasing this onto `master`?

PR #137, opened 2019-11-13
feat: Add `Status` constructors

alce:
This patch adds shortcut constructors for all Status codes as well as documentation for all Code variants.

`From<tokio::timer::timeout::Elapsed>` for Status is also implemented, which is useful for the in-progress revamp of request timeouts.

Edit: the `From<Elapsed> for Status` does not compile when the transport feature is disabled. Without it, there is no tokio available. I removed the implementation for now.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -9,6 +9,7 @@ const GRPC_STATUS_MESSAGE_HEADER: &str = "grpc-message";
 const GRPC_STATUS_DETAILS_HEADER: &str = "grpc-status-details-bin";
 
 /// A gRPC status describing the result of an RPC call.
+///
 /// Values can be created using the `new` function or one of the specialized
 /// associated functions.
 /// ```rust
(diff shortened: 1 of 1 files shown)


tonic/src/status.rs
```diff
@@ -8,7 +8,19 @@ const GRPC_STATUS_HEADER_CODE: &str = "grpc-status";
 const GRPC_STATUS_MESSAGE_HEADER: &str = "grpc-message";
 const GRPC_STATUS_DETAILS_HEADER: &str = "grpc-status-details-bin";
 
+// Standard status codes
```
LucioFranco: do we need this comment?

tonic/src/status.rs
```diff
@@ -8,7 +8,19 @@ const GRPC_STATUS_HEADER_CODE: &str = "grpc-status";
...
+
 /// A gRPC status describing the result of an RPC call.
+/// Values can be created using the `new` function or one of the specialized
```
LucioFranco:
```suggestion

/// Values can be created using the `new` function or one of the specialized
```

tonic/src/status.rs
```diff
@@ -62,13 +106,140 @@ impl Status {
...
+        Status::new(Code::Ok, message)
+    }
+
+    /// The operation was cancelled (typically by the caller).
+    pub fn cancelled(message: impl Into<String>) -> Status {
+        Status::new(Code::Cancelled, message)
+    }
+
+    /// Unknown error. An example of where this error may be returned is if a
+    /// Status value received from another address space belongs to an error-space
```
LucioFranco:
```suggestion
    /// `Status` value received from another address space belongs to an error-space
```

LucioFranco:
@alce I think not having the elasped one is fine for now.

PR #138, opened 2019-11-13
fix: Sanitize custom metadata

alce:
## Motivation

Tonic's `MetadataMap` prevents  entries with  invalid or reserved HTTP2 header names to be inserted but does not disallow headers reserved by the gRPC protocol.

The metadata map is then copied verbatim to HTTP headers and written the wire. Some headers may be updated by inserting new values but some will be written as-is.

## Solution

This patch prevents a small set of headers to be written when sending requests or responses. It does not panic or returns errors if entries with these names are added to the metadata map, it just won't copy them to http request or responses.

### DIFF
--- a/tonic/src/metadata/map.rs
+++ b/tonic/src/metadata/map.rs
@@ -197,6 +197,18 @@ pub struct OccupiedEntry<'a, VE: ValueEncoding> {
 // ===== impl MetadataMap =====
 
 impl MetadataMap {
+    // Headers reserved by the gRPC protocol.
+    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 8] = [
+        "te",
+        "user-agent",
+        "content-type",
+        "grpc-timeout",
+        "grpc-message",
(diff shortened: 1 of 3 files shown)


tonic/src/metadata/map.rs
```diff
@@ -237,6 +249,29 @@ impl MetadataMap {
...
+    /// # use tonic::metadata::*;
+    /// let mut map = MetadataMap::new();
+    /// map.insert("grpc-timeout", "1S".parse().unwrap());
+    /// map.insert("x-host", "example.com".parse().unwrap());
+    ///
+    /// let http_map = map.into_sanitized_headers();
+    ///
+    /// assert!(http_map.get("grpc-timeout").is_none());
+    /// ```
+    pub fn into_sanitized_headers(mut self) -> http::HeaderMap {
```
LucioFranco: I think we should probably make this `pub(crate)` for now? Like we do with into http in request/response.

PR #140, opened 2019-11-14
Enable multiple services in same package across multiple files

jpgneves:
## Motivation

Prior to this change, for each _file_ containing one or more
services, the code generator would generate with the following
modules:

```
pub mod client {
 //...
}

pub mod server {
 //...
}
```

While this works fine if all the services in the same protobuf
package are declared in a single file, this breaks horribly if
multiple files belonging to the same protobuf package declare
services. In that case, the code generator would generate several
modules with the same name in the same `.rs` file, which is
invalid.

For instance, given two files `foo.proto` and `bar.proto` like:

```
// foo.proto
package mypackage;
service Foo {
}

// bar.proto
package mypackage;
service Bar {
}
```

The generated code would result in a `mypackage.rs` such as:

```
// Generated from foo.proto
pub mod client {
  //...
}

pub mod server {
  //...
}

// Generated from bar.proto
pub mod client {
  //...
}

pub mod server {
  //...
}
```

One codebase where this scenario could be found is in [Open Match](https://github.com/googleforgames/open-match/tree/master/api)

## Solution

This change makes it so the name of the service is prepended to
the generated module, and therefore we avoid module name collisions.

After this change, given the two proto files mentioned previously,
we will generate the following modules inside `mypackage.rs`:

```
pub mod foo_client {
 //...
}

pub mod foo_server {
 //...
}

pub mod bar_client {
 //...
}

pub mod bar_server {
 //...
}
```

### DIFF
--- a/tests/included_service/src/lib.rs
+++ b/tests/included_service/src/lib.rs
@@ -5,4 +5,4 @@ pub mod pb {
 // Ensure that an RPC service, defined before including a file that defines
 // another service in a different protocol buffer package, is not incorrectly
 // cleared from the context of its package.
-type _Test = dyn pb::server::TopService;
+type _Test = dyn pb::topservice_server::TopService;
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -5,44 +5,51 @@ use quote::{format_ident, quote};
 
 pub(crate) fn generate(service: &Service, proto: &str) -> TokenStream {
     let service_ident = quote::format_ident!("{}Client", service.name);
+    let client_mod = quote::format_ident!("{}_client", service.name.to_ascii_lowercase());
     let methods = generate_methods(service, proto);
 
     let connect = generate_connect(&service_ident);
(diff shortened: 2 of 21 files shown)


LucioFranco:
Hi @jpgneves thanks for this. This seems mostly good to me but I'd like to see what @jen20 thinks.

alce:
LGTM. I tried this patch on a server with multiple services. Everything worked out of the box with minor changes and gives me more options to organize proto definitions/modules.

jen20:
I think this is a solid improvement, I actually hit this case the other day.

LucioFranco:
Hey @jpgneves sorry for going silent on this, I'd like to get this merged, would you mind fixing the conflicts, and then we can merge it? :)

LucioFranco:
Thank you for this contribution! I merged this in #173!

jpgneves:
Apologies for the radio silence myself! Glad to see this in! Thank you! :tada:

PR #143, opened 2019-11-16
remove VecDeque from examples

alce:
The futures crate removed the Stream implementation for VecDeque in 0.3, which was used as a streaming response type in some examples.

This PR replaces VecDeque with a dummy Stream implementation for examples that just need a Stream type but don't actually use it.


tonic-examples/src/authentication/server.rs
```diff
@@ -3,12 +3,12 @@ pub mod pb {
...
 
 use pb::{EchoRequest, EchoResponse};
-use std::collections::VecDeque;
 use tonic::{body::BoxBody, transport::Server, Request, Response, Status, Streaming};
+use tonic_examples::ResponseStream;
 use tower::Service;
 
 type EchoResult<T> = Result<Response<T>, Status>;
-type Stream = VecDeque<Result<EchoResponse, Status>>;
+type Stream = ResponseStream<Result<EchoResponse, Status>>;
```
LucioFranco: in this case we should just resort to `Pin<Box<dyn Stream<...>>>` this way we don't need to implement a dummy one.

jen20:
Thanks @alce!

PR #145, opened 2019-11-21
feat: expose tcp_nodelay for clients and servers

### LINKED ISSUE #119: Expose `set_nodelay` for server and channel
We should expose the hyper `set_nodelay` option as a config option for the transport channel and server.

alce:
closes #119

### DIFF
--- a/tonic-examples/Cargo.toml
+++ b/tonic-examples/Cargo.toml
@@ -80,7 +80,7 @@ tower = { git = "https://github.com/tower-rs/tower" }
 # Required for routeguide
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-rand = "0.7.2"
+rand = "0.6"
 
 # Required for wellknown types
 prost-types = "0.5"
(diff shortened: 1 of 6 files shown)


deny.toml
```diff
@@ -16,6 +16,7 @@ deny = [
 ]
 skip = [
     { name = "crossbeam-utils", version = "=0.6.6" },
+    { name = "crossbeam-queue", version = "=0.1.2" },
```
LucioFranco: can we back this out so that #153 contains the change.

LucioFranco:
Thanks for this

LucioFranco:
if we can rebase this with master then we can merge this thanks!

tonic-examples/Cargo.toml
```diff
@@ -80,7 +80,7 @@ tower = { git = "https://github.com/tower-rs/tower" }
 # Required for routeguide
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-rand = "0.7.2"
+rand = "0.6"
```
LucioFranco: why rand 0.6?

alce: Sorry, didn't see this. Because at the time I wrote this, some other dependency (some tokio sub-crate, I think) wall pulling in 0.6, so I changed it to match.

alce: I guess once 0.1 is close to finished we can sweep through all deps and remove what is not essential.

PR #147, opened 2019-11-21
Use a new prost ServiceGenerator finalizer to fix duplicate client and server definitions

### LINKED ISSUE #38: Client/Server /can be/ duplicated into their include files
### Version

master: afa9d9d236b7ca96d70d696ab38aadac7cdf41ff

### Platform

Linux RILEY-LT 4.19.43-microsoft-standard #1 SMP Mon May 20 19:35:22 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

### Crates

cargo-build

### Description

When using includes it seems like the client/server are generated into the include file output. This only seems to happen under certain conditions, I believe related to compiling multiple protos at once via `tonic_build::configure().compile(&["proto/health.proto", "proto/test.proto"], &["proto"]).unwrap();`

Repro here:

https://github.com/rlabrecque/tonic_include_repro

Specifically see: https://github.com/rlabrecque/tonic_include_repro/blob/master/src/proto/google.protobuf.rs

parasyte:
Note: This PR is in draft mode until the dependent PR in `prost` is accepted and released: https://github.com/danburkert/prost/pull/241

## Motivation

Fixes #38

This change allows `tonic-build` to properly group all services from multiple `.proto` files into a single package namespace.

## Solution

Defer code generation to the end of all package processing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "tonic-examples",
     "tonic-interop",
 
+    "tests/complex_package_layout",
     "tests/included_service",
     "tests/same_name",
     "tests/wellknown",
--- a/deny.toml
+++ b/deny.toml
@@ -16,6 +16,7 @@ deny = [
 ]
 skip = [
     { name = "crossbeam-utils", version = "=0.6.6" },
+    { name = "crossbeam-queue", version = "=0.2.0" },
 ]
 skip-tree = [
     { name = "rand", version = "=0.6.5" },
--- a/tests/complex_package_layout/Cargo.toml
+++ b/tests/complex_package_layout/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "complex_package_layout"
+version = "0.1.0"
+authors = ["Jay Oster <jay@kodewerx.org>"]
+edition = "2018"
+publish = false
+license = "MIT"
+
+[dependencies]
+bytes = "0.4"
+prost = "0.5"
+tonic = { path = "../../tonic" }
+
+[build-dependencies]
+tonic-build = { path = "../../tonic-build" }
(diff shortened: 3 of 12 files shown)


parasyte:
Ah jeez, failures in cargo-deny and tests that I didn't touch at all... Maybe these will get sorted in master in the meantime and this PR will be fixed by merge or rebase! 😉

parasyte:
I'm not happy with the new requirement that user code has to include all packages explicitly. Go doesn't have this problem because it will happily pull all packages into scope from a single `import` directive.

I don't see an obvious way to auto-include packages within the generated source. The conflict is that imports in protobuf are "file level" and includes in tonic-build are "package level".

parasyte:
I see, this is actually a bug! The client and server modules are being output to the wrong package. I will be addressing this tomorrow.

alce:
@parasyte the cargo-deny is failing in part because of a duplicated version of crossbeam-queue, which is addressed here #145 and in part because you are referencing a different version of prost-build.

We'll need #145 to be merged first, or at least the commits that fix the crossbeam duplication and then fix the prost-build issue here

parasyte:
Ok, we're good here. All tests are passing, just waiting for `prost-build` 0.5.1 and @alce's `crossbeam-queue` patch for the cargo deny checks to pass.

jen20:
@parasyte I'll look at resolving the issue with cargo-deny shortly, it's affected quite a few of our open PRs right now.

LucioFranco:
This looks good! Once prost gets a publish out we can merge this! Thanks!

parasyte:
That sounds great. I'll keep the PR as a draft until I can update `Cargo.toml` with whatever prost version gets released.

parasyte:
Looks like #173 fixes this by namespacing the `client/server` modules separately.

LucioFranco:
@parasyte yup, thank you for this contribution sorry we couldn't get it in, but also looks like prost is a bit slow on upgrading anyways. I am gonna close this as it seems its been solved for you!

parasyte:
I haven't actually used the new code yet, but I am interested in trying it.

PR #148, opened 2019-11-21
Make `Endpoint` connectors swappable

akshayknarayan:
## Motivation

As discussed in https://github.com/hyperium/tonic/issues/136, it would be useful if tonic supported other transports other than TCP.
The Server side is relatively straightforward to do, like so:
```rust
let fut = hyper::Server::builder(/* impl of hyper::server::accept::Accept */).serve(hyper::service::make_service_fn(move |_| {
  async move { Ok::<_, hyper::Error>(/* tonic rpc server type */ ) }
}));
```

The client side is what this PR addresses.

## Solution

Change the `connect()` API to lift up the use of `connector()` to `Endpoint::connect()`, allowing users to provide their own implementations (for example, Unix-domain sockets). Any type which impls `tower_make::MakeConnection` is suitable.

To avoid breaking the default case of HTTP(S), introduce `connect_with_connector()` and retain `connect()`, which creates the default connector according to the activated feature gate and passes it to `connect_with_connector()`.


alce:
@akshayknarayan cargo-deny check is failing in part because hyper-unix-connector depends in hyperlocal, which in turn depends on hyper 0.12. Tonic depends on hyper 0.13.0-alpha.4 and cargo check is banning the duplicate dependency.

tonic-build/src/client.rs
```diff
@@ -62,6 +62,21 @@ fn generate_connect(service_ident: &syn::Ident) -> TokenStream {
                 let conn = tonic::transport::Endpoint::new(dst)?.connect().await?;
                 Ok(Self::new(conn))
             }
+
+            /// Attempt to create a new client by connecting to a given endpoint using a custom
+            /// connector.
+            pub async fn connect_with_connector<C, D>(dst: D, connector: C) -> Result<Self, tonic::transport::Error>
+            where
+                C: tonic::transport::MakeConnection<http::Uri> + Send + 'static,
```
LucioFranco: can we clean up some of these imports these seems a bit verbose :smile:

tonic/src/transport/endpoint.rs
```diff
@@ -157,7 +157,54 @@ impl Endpoint {
...
+    /// # Example with non-default Connector
+    /// ```rust
+    /// // Use for unix-domain sockets
+    /// use hyper_unix_connector::UnixClient;
+    /// use tonic::transport::Endpoint;
+    ///
+    /// let endpoint = Endpoint::from_static("http://example.com");
+    /// endpoint.connect_with_connector(UnixClient); //.await
+    /// ```
+    pub async fn connect_with_connector<C>(&self, connector: C) -> Result<Channel, super::Error>
```
LucioFranco: so I think it makes more sense for the endpoint to contain a connector option that by default uses the built in tonic one. So instead of providing it at connect time its provided with the endpoint. This means in a load balance list of endpoints they can each have their own type of connector. This should also simplify a lot of the code too.

jen20:
It looks like we have the same issue here as in #151 with `cargo-deny` picking up an existing dependency as bad.

tonic/src/transport/endpoint.rs
akshayknarayan: I have poked around at this idea a bit, and it probably involves making several things `pub`, like `transport::service::Connector`, which in turn would require exposing `transport::service::io::BoxedIo`, etc. 

This is because having `Endpoint<C: MakeConnection<...>>` means we will have
```rust
pub fn tls_config(self, tls_config: ClientTlsConfig) -> Endpoint<transport::service::Connector> {
```

and `transport::service::Connector` is currently `pub(crate)`.

What is your plan regarding the `tls` feature? Things might be simpler if it was in its own crate - then usage would just be

```rust
Endpoint::from(...).connector(/* tls connector from the external crate */).connect().await
```

with a plain `HTTPConnector` as the default.

LucioFranco: You can use a `Box<dyn Connector>` type to avoid the generic for now.

LucioFranco:
Hey! I opened #184 which adds this and a UDS example. Thank you for contributing this, I am gonna close this PR <3! Let me know if the other one doesn't meet your needs!

PR #151, opened 2019-11-22
feat(transport): Expose tcp keepalive to clients & servers

zenria:
## Motivation

Long running tcp connections (grpc streaming), with very low activity may be silently disconnected if done though firewalls or NAT routers. TCP keepalive fixes this.

## Solution

Expose hyper tcp_keepalive api on `Endpoint` and `Server`.


jen20:
Hi @zenria! Thanks for opening up a pull request, we much appreciate the contribution! It looks like your PR is fine, but we need to upgrade some other dependencies to get the build green before merging it. I'll try to look at that sometime either today or tomorrow.

LucioFranco:
Once we get deny working which I think you just need to merge master we can merge this! Thanks!

zenria:
I've updated my PR so it can merge with current master

PR #157, opened 2019-11-26
Initial thoughts for parent propagation (WIP)

### LINKED ISSUE #156: Default span propagation for tracing in builders
Following on from https://github.com/jtescher/opentelemetry-rust/issues/17 

It would be useful if the server and client builders provided a default implementing for propagating parent spans via the `grpc-trace-bin` header. Currently I have something similar to this for a service I'm working on using https://github.com/census-instrumentation/opencensus-go/blob/17d7955af9d42886455ce010dd46878208041a58/plugin/ocgrpc/trace_common.go as a reference for how it's typically done in other languages:

```Rust
Server::builder()
    .interceptor_fn(|svc, req| {
        let grpc_trace = &req.headers()["grpc-trace-bin"];
        let bp = BinaryPropagator::new();
        let span_state = bp.from_bytes(grpc_trace.as_bytes().into().span_id();
        let parent = match span_state {
            0 => None,
            e @ _ => Some(Id::from_u64(e)),
        };
        let root = match req.uri().path() {
            "/package.service/method" => span!(parent: parent, Level::INFO, "package.service.method"),
            _ => span!(parent: parent, Level::INFO, "package.service.unknown"),
        };
        let _enter = root.enter();
        svc.call(req)
    })
    .add_service(service).serve(addr).await?;
```

The ugly match is because the span name needs a static lifetime so if there was also a static string representing the `package.service.name` identifier in the generated grpc code that would be useful as well.

xd009642:
## Motivation

Very initial sketch of what a grpc parent propagation may look like. Will hopefully fix #156 eventually

## Solution

Use the interception function to create a parent span. Unfortunately the span binary can differ from tracing system to tracing system so I've kept this initially very very general. No solution for good span names, very WIP.

### DIFF
--- a/tonic/src/transport/server.rs
+++ b/tonic/src/transport/server.rs
@@ -6,7 +6,7 @@ use super::{service::TlsAcceptor, tls::Identity, Certificate};
 use crate::body::BoxBody;
 use futures_core::Stream;
 use futures_util::{future, ready, try_future::MapErr, TryFutureExt, TryStreamExt};
-use http::{Request, Response};
+use http::{HeaderValue, Request, Response};
 use hyper::{
     server::{accept::Accept, conn},
     Body,
@@ -29,6 +29,8 @@ use tower::{
 };
 #[cfg(feature = "tls")]
 use tracing::error;
+
+use tracing::{span, Id, Level};
 
 type BoxService = tower::util::BoxService<Request<Body>, Response<BoxBody>, crate::Error>;
 type Interceptor = Arc<dyn Layer<BoxService, Service = BoxService> + Send + Sync + 'static>;
@@ -171,6 +173,24 @@ impl Server {
             interceptor: Some(Arc::new(layer)),
             ..self
         }
+    }
+
+    /// Setup propagation of trace parents requires a function that
+    /// can turn the `grpc-trace-bin` header into a parent span Id
+    pub fn propagate_trace_parents<F>(self, f: F) -> Self
+    where
+        F: Fn(&HeaderValue) -> Option<Id> + Send + Sync + 'static,
+    {
+        self.interceptor_fn(move |svc, req| {
+            if let Some(grpc_trace) = req.headers().get("grpc-trace-bin") {
+                let span_id = f(grpc_trace);
+                let root = span!(parent: span_id, Level::INFO, "request");
+                let _enter = root.enter();
+                svc.call(req)
+            } else {
+                svc.call(req)
+            }
+        })
     }
 
     /// Create a router with the `S` typed service as the first service.


xd009642:
Also it would be super helpful if there was a way to put non-static strings into span names though I do have a vague inkling on why that's not possible

tonic/src/transport/server.rs
```diff
@@ -173,6 +175,24 @@ impl Server {
...
+    pub fn propagate_trace_parents<F>(self, f: F) -> Self
+    where
+        F: Fn(&HeaderValue) -> Option<Id> + Send + Sync + 'static,
+    {
+        self.interceptor_fn(move |svc, req| {
+            if let Some(grpc_trace) = req.headers().get("grpc-trace-bin") {
+                let span_id = f(grpc_trace);
+                let root = span!(parent: span_id, Level::INFO, "request");
+                let _enter = root.enter();
+                svc.call(req)
```
davidbarsky: This won't be propagated beyond this span, unfortunately. `_enter` would dropped at the end of this scope. You'd need something like: https://github.com/tokio-rs/tracing/blob/master/tracing-tower/src/service_span.rs, which hasn't been updated to the releases of Tower using `std::future::Future`. That might happen this week or next week...?

xd009642: Probably a noob question but is it possible to use `tracing-futures` to do this by `svc.call(req).instrument(root)` ?

davidbarsky: I'm not 100% sure, but I think the instrumentation needs to be on the Service-level. Give it a shot?

hawkw: > Probably a noob question but is it possible to use `tracing-futures` to do this by `svc.call(req).instrument(root)` ?

Yeah, that will work fine!

(@davidbarsky, this is all that `tracing-tower` does :P)

hawkw:
> Also it would be super helpful if there was a way to put non-static strings into span names though I do have a vague inkling on why that's not possible

Span _names_ have to be string literals, but you can put an arbitrary non-static string in a span _field_, is that sufficient?

For example:
```rust
let mut my_string = String::new();
my_string.push_str("this string");
my_string.push_str(" is constructed");
my_string.push_str(" dynamically");
span!(Level::INFO, "my_span_name", my_string)
```

tonic/src/transport/server.rs
hawkw: I think allowing users to allow their own span is probably a good idea.

davidbarsky:
@hawkw 

> Yeah, that will work fine!
> (@davidbarsky, this is all that tracing-tower does :P)

oops lol

PR #169, opened 2019-12-09
feat(transport): Add server graceful shutdown

xd009642:
Some error messages, one about hyper::server::Builder not existing and another about lack of clone on futures. Just putting this up as an initial WIP to make sure I'm on the right track :)

## Motivation

Recommended good practice for gRPC so should be exposed

## Solution
Expose it via builder pattern

### DIFF
--- a/tonic/src/transport/server.rs
+++ b/tonic/src/transport/server.rs
@@ -56,7 +56,7 @@ pub struct Server {
     init_connection_window_size: Option<u32>,
     max_concurrent_streams: Option<u32>,
     tcp_keepalive: Option<Duration>,
-    shutdown: Option<Box<dyn Future<Output=()>>>,
+    shutdown: Option<Arc<dyn Future<Output = ()>>>,
     tcp_nodelay: bool,
 }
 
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server.rs
```diff
@@ -110,6 +111,14 @@ impl Server {
         }
     }
 
+    /// Set the graceful shutdown thing
+    pub fn graceful_shutdown(self, shutdown: Box<dyn Future<Output=()>>) -> Self {
```
LucioFranco: I'd probably make this `impl Future<Output = ()>` and then box it internally.

tonic/src/transport/server.rs
```diff
@@ -268,15 +289,19 @@ impl Server {
             concurrency_limit,
             // timeout,
         };
```
LucioFranco:
```suggestion
        };

```

LucioFranco:
hmm looks like there are some compile errors?

xd009642:
Yeah `hyper::server::builder` isn't found in `hyper::server` but that's not something I changed...

xd009642:
Oh wait I did change something! Sorting it now

xd009642:
@LucioFranco did you have a chance to look at my type based woes?

LucioFranco:
Looks like there are a couple of CI failures, mind fixing them, then we can merge this.

xd009642:
Yeah but I'm not sure quite how, my attempt to make serve_common to avoid repeating code leads to ```type inside `async` object must be known in this context``` because I think the acceptor type isn't known from the calling context. I could just duplicate the code across the two functions but I was hoping to do something cleaner than that

PR #170, opened 2019-12-10
feat(build): Decouple codgen from `prost`

Perlmint:
## Motivation

Prepare using tonic with other protocol buffer implement or flatbuffers.

## Solution

Add trait to provide needed info to the code generator.
And, introduce new features to make using prost optional.


tonic-build/src/client.rs
```diff
@@ -95,71 +99,91 @@ fn generate_methods(service: &Service, proto: &str) -> TokenStream {
...
+    let (request, response) = method.request_response_name(context);
 
     quote! {
         pub async fn #ident(
             &mut self,
             request: impl tonic::IntoRequest<#request>,
         ) -> Result<tonic::Response<#response>, tonic::Status> {
            self.ready().await?;
-           let codec = tonic::codec::ProstCodec::default();
+           let codec = tonic::codec::#codec_name::default();
```
LucioFranco: I was thinking we should actually make this
```rust
let codec = #codec_path::default();
```
so that you can use any codec from any path include crates that are not tonic.

tonic-build/src/lib.rs
```diff
@@ -304,32 +227,14 @@ fn generate_doc_comment(comment: &str) -> TokenStream {
...
 // Generate a larger doc comment composed of many lines of doc comments
-fn generate_doc_comments<T: AsRef<str>>(comments: &[T]) -> TokenStream {
+fn generate_doc_comments<'a, T: AsRef<str> + 'a, I: Iterator<Item = &'a T>>(
+    comments: I,
+) -> TokenStream {
     let mut stream = TokenStream::new();
 
-    for comment in comments {
-        stream.extend(generate_doc_comment(comment.as_ref()));
+    for ref comment in comments {
```
LucioFranco:
```suggestion
    for comment in &comments {
```

tonic-build/src/prost.rs
```diff
@@ -0,0 +1,191 @@
...
+        self.comments.leading.iter()
+    }
+}
+
+pub(crate) struct ProstContext {
+    pub(crate) proto_path: String,
+}
+
+impl schema::Context for ProstContext {
+    const CODEC_NAME: &'static str = "ProstCodec";
```
LucioFranco: so we probably want to make this non-const so that users can set it via a builder fn possibly. This path to this codec may not always be static.

tonic-build/src/schema.rs
```diff
@@ -0,0 +1,32 @@
...
+    type Context: Context + Sized + 'a;
+
+    fn name(&self) -> &str;
+    fn package(&self) -> &str;
+    fn identifier(&self) -> &str;
+    fn methods(&'a self) -> Self::MethodIterator;
+}
+
+pub(crate) trait Method<'a>: Commentable<'a> {
+    type Context: Context + Sized + 'a;
```
LucioFranco: You don't need to add sized here iirc since its always included in a trait bound.

tonic-build/src/lib.rs
Perlmint: Actually `comments` is an iterator. So, your suggestion is not appliable straightforward. I changed the type of comments to IntoIterator.

tonic-build/src/schema.rs
Perlmint: I didn't specify Sized constraint. But, it is useless, unused constraint. So I removed it.

Perlmint:
@LucioFranco Can you check some added commits?

Which way is preferred here to fix [this check failure](https://github.com/hyperium/tonic/pull/170/checks?check_run_id=343007356)? Adding _ prefix or disabling the function by feature gate?

LucioFranco:
@Perlmint looks like there are a couple conflicts, I know some of the build code change, would you be able to update this? Then I can take another look. Thanks!

Perlmint:
I rebased my commits. But, It looks like I should add more commits - make types that needed to describe schema public. After pushing the commits, I'll notify.

Perlmint:
@LucioFranco Fix all checks and make possible implementing the service generator out of tonic-build.

LucioFranco:
Sorry for the delay on this, I am gonna take a longer look at this once `tonic` 0.1 is out because I would like to push this to 0.2 since there are still somethings to work out. Thanks for your patience.

Perlmint:
I got it. When you have time to review this, please notify me. I'll update soon :)

magnet:
@Perlmint Any plan to rebase this on master? :) If not I can give it a shot, but I won't be able to add commits to this MR

Perlmint:
@magnet Oh, I'm just waiting for that @LucioFranco is ready to review this. If you need, I'll do it soon. :)

LucioFranco:
@Perlmint yes, sorry for the delay! We should get the ball rolling on this, looks like fbs support is almost ready :) Thanks for the patience, if we can rebase this against master then I think we can move forward.

Perlmint:
@magnet rebased!

LucioFranco:
Great thank you! @magnet let me know if you have any issues, I plan on taking a longer look at all of this soon.

PR #85, opened 2019-10-22
Allow access to peer identity in a gRPC Server Interceptor

jen20:
This is an initial work in progress of a mechanism to access the TLS identity of the client in a server interceptor.

Issue #71, opened 2019-10-12
transport: expose connection metadata


domodwyer:
### Crates

* tonic

### Motivation

Both the C++ gRPC client, and the Go client (and probably more) expose the peer address and other connection metadata such as the mTLS certificate used by the client - these features combined are great for certificate-based access control and request source filtering / audit logs / metrics / etc. 

It looks like there was [a brief discussion] for the C++ implementation on how to expose the peer address in a way that could be used across multiple transports (both IP and unix domain sockets are particularly common) that would be wise to take into account too - they ended up returning a string URI for each transport. The Go implementation handles this nicely using the [net.Addr] abstraction in the standard library that covers various transports.

### Proposal

Add a `peer`-like field/method to the Request struct.

Perhaps we would express the different transports as enum variants with their associated addresses? Something like:

```rust
enum PeerAddr {
    Tcp(std::net::SocketAddr),
    Unix(std::os::unix::net::SocketAddr),
}
```

I'm not too sure what is best for the mTLS certificate as #47 points out (and #48 fixes), attempting to abstract away the different TLS types for each backend into a common type is probably not the best direction to take - that said, maybe each TLS implementation can have a method on the Request that returns their native certificate type, and place them behind feature flags in the same way the `with_rustls`/`with_openssl` methods are.

[C++ gRPC client]: https://grpc.github.io/grpc/cpp/classgrpc__impl_1_1_server_context.html#a7466eb08da322005c49e10303b214afc
[Go client]: https://godoc.org/google.golang.org/grpc/peer
[a discussion]: https://github.com/grpc/grpc/issues/1132
[net.Addr]: https://golang.org/pkg/net/#Addr

### Alternatives

As noted above, the C++ implementation returns a string URI and leaves the caller to parse it according to the expected value. This requires [unnecessarily verbose] code for the caller to extract the address, I hope we'd be able to do something a little nicer!

[C++ gRPC client]: https://grpc.github.io/grpc/cpp/classgrpc__impl_1_1_server_context.html#a7466eb08da322005c49e10303b214afc
[Go client]: https://godoc.org/google.golang.org/grpc/peer
[a brief discussion]: https://github.com/grpc/grpc/issues/1132
[net.Addr]: https://golang.org/pkg/net/#Addr
[unnecessarily verbose]: https://github.com/grpc/grpc/issues/5693

LucioFranco:
[...]

PR #171, opened 2019-12-11
Attach a context to requests

Farenjihn:
## Motivation

Being able to access various things (ip address of the peer or other things) in server interceptors / future middleware implementations in tonic. This is mainly for logging, access control, etc...

This would solve #71.

I need this feature so I went ahead and implemented it in the way described below. It may not fit with what tonic intends to provide in the long term but since it's done, I may as well create this PR.

The implementation has some overlap with #85, since it's inspired by it, but tries to be a bit more flexible.

## Solution
Create a `Ctx` struct which is populated according to configuration passed to the `Server` builder. This struct is added to the `Request`'s extensions so it becomes available to interceptors.

Here's what the current usage looks like:

```rust
let mut builder = Server::builder();
builder
    .with_ctx_field(CtxField::PeerAddr) // Add the PeerAddr field in the Ctx struct
    .interceptor_fn(|svc, req| {
        let ctx: &Ctx = req.extensions().get().unwrap();
        // do something
    })
    .add_service(/* service */)
    .serve(/* addr */)
    .await?;
```
The context struct can be extended with more fields, which can also be gated behind features. This PR only adds a field to read the peer's ip address.

## Open questions
Provided the current implementation is acceptable for merging:

1. Would it make sense to do something similar for clients ?
2. @jen20 If this gets merged, how should we proceed for #85 ?

Let me know what you think.

### DIFF
--- a/tonic-examples/Cargo.toml
+++ b/tonic-examples/Cargo.toml
@@ -66,6 +66,14 @@ path = "src/multiplex/client.rs"
 name = "gcp-client"
 path = "src/gcp/client.rs"
 
+[[bin]]
+name = "context-client"
+path = "src/context/client.rs"
+
+[[bin]]
+name = "context-server"
+path = "src/context/server.rs"
+
 [dependencies]
 tonic = { path = "../tonic", features = ["tls"] }
 bytes = "0.4"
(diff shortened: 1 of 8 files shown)


LucioFranco:
Hey! Thank you for this! I have another proposal #177 and we've been working on this in #85. I will see if I can get this done before 0.1 release so you can achieve the same thing. I'd like to see the server do this implicitly instead of needing to pass a `Ctx`.

Farenjihn:
Closing since there is a different plan

LucioFranco:
Thank you <3

PR #179, opened 2019-12-12
chore(examples): add helloworld-client-blocking

bmwill:
Add an example of using a tonic client in a synchronous environment.


bmwill:
Not quite sure why the checks/tests on ubuntu are failing, or if that's expected?

bmwill:
Oh this is probably due to #173. Let me rebase and fix that

PR #201, opened 2019-12-20
chore(ci): Switch to use GitHub Action for cargo-deny

repi:
This upgrades CI from manually downloading cargo-deny to use our new [cargo-deny GitHub Action](https://github.com/EmbarkStudios/cargo-deny-action). 

Includes a lot of fixes, upgrades a new features as it brings cargo-deny from 0.4.0 to 0.5.2. List of changes:
https://github.com/EmbarkStudios/cargo-deny/releases

Biggest one here is that it will now also check the RUSTSEC security advisory database and will (by default) fail builds if any security advisories are detected for used crates

I was thinking first that we should wait until we have a v1 of the action and cargo-deny before using it, but as `tonic` is one of the first non-Embark repos using `cargo-deny` I think we can switch to the action here - but monitor how it works. Can always pin the action to an exact version also and/or disable advisories or other features.

cc @Jake-Shadle @LucioFranco

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,21 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
-    - name: download cargo-deny
-      shell: bash
-      env:
-        DVS: "0.4.0"
-        DREPO: EmbarkStudios/cargo-deny
-        TARGET: x86_64-unknown-linux-musl
-      run: |
-        temp_archive=$(mktemp --suffix=.tar.gz)
-        curl -L --output "$temp_archive" https://github.com/$DREPO/releases/download/$DVS/cargo-deny-$DVS-$TARGET.tar.gz
-
-        tar -xzvf "$temp_archive" -C . --strip-components=1 --wildcards "*/cargo-deny"
-    - name: cargo-deny check licenses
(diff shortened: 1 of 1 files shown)


repi:
Great! 

And this passed quickly, only one advisory warning, but no failures.

```
warning[RUSTSEC-2019-0031]: spin is no longer actively maintained

     ┌── /github/workspace/Cargo.lock:149:1 ───
     │
 149 │ spin 0.5.2 registry+https://github.com/rust-lang/crates.io-index
     │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ unmaintained advisory detected
     │
     = The author of the `spin` crate does not have time or interest to maintain it.
       
       Consider [`lock_api`](https://crates.io/crates/lock_api) (a subproject of
       `parking_lot`) as an alternative which also supports `no_std` environments.
     = URL: https://github.com/mvdnes/spin-rs/commit/7516c80
     = spin v0.5.2
       ├── lazy_static v1.4.0
       │   ├── bstr v0.2.8
       │   │   └── csv v1.1.1
       │   │       └── criterion v0.3.0
```

LucioFranco:
@repi yeah, looks like a criterion dep so not a big deal, well prob move off that soon.

Issue #205, opened 2019-12-22
UDS example does not build on Windows


repi:
Find a Windows equivalent of the below or try and disable the example on Windows?

```
error[E0432]: unresolved import `tokio::net::UnixStream`
  --> examples\src/uds/client.rs:10:5
   |
10 | use tokio::net::UnixStream;
   |     ^^^^^^^^^^^^^^^^^^^^^^ no `UnixStream` in `net`

error[E0432]: unresolved import `tokio::net::UnixListener`
 --> examples\src/uds/server.rs:9:5
  |
9 |     net::UnixListener,
  |     ^^^^^------------
  |     |    |
  |     |    help: a similar name exists in the module: `TcpListener`
  |     no `UnixListener` in `net`
```

LucioFranco:
We should just feature gate it, since I do not know of a similar thing for mio and windows. https://docs.rs/mio-uds/0.6.7/src/mio_uds/lib.rs.html#3

messense:
@LucioFranco https://github.com/Azure/mio-uds-windows

alce:
@messense crate looks good, unfortunately it has some outdated deps that overlap with Tonic's and I don't think it'll be worth the hassle, what do you think?

If you all agree I can try to figure out how to hide it from Windows

messense:
I think hiding it from Windows is fine.

alce:
I guess this was resolved on https://github.com/hyperium/tonic/pull/206. 

@LucioFranco  can we close this?

LucioFranco:
Yes! thank you.

PR #206, opened 2019-12-22
Run CI checks & tests on Windows and Mac

repi:
## Motivation

Ran into some build errors on Windows in one of the examples (#205) and realized the GitHub CI action is set up only check and test the project on Linux, which is likely why it wasn't caught.

## Solution

Enable CI checks and tests on all supported targets.

Oh and this will likely fail until #205 is fixed


LucioFranco:
Looks like we just need to feature gate the uds example to `cfg(unix)` and it should work!

LucioFranco:
Thanks @repi for starting this!!

PR #207, opened 2019-12-22
chore: Add decode benchmarks

alce:
This PR adds preliminary decoding benchmarks. `Streaming::new_request` is temporary exposed while we figure out how to better organize benchmarks.


tonic/benches/decode.rs
```diff
@@ -0,0 +1,160 @@
...
+}
+
+impl MockDecoder {
+    fn new(message_size: usize) -> Self {
+        MockDecoder { message_size }
+    }
+}
+
+impl Decoder for MockDecoder {
+    type Item = Vec<u8>;
```
LucioFranco: we should make this a `bytes` and use `BytesMut::freeze`

LucioFranco:
@alce is this ready for review?

alce:
ready

PR #217, opened 2020-01-04
fix(transport): Improve `Error` type

### LINKED ISSUE #213: Error Display implementation contains source display impl as well
## Feature Request

When printing a connection refused error and all its source errors you currently end up with:
```
Error: Client: error trying to connect: tcp connect error: Connection refused (os error 111)
Caused by: error trying to connect: tcp connect error: Connection refused (os error 111)
Caused by: tcp connect error: Connection refused (os error 111)
Caused by: Connection refused (os error 111)
```
That is very redundant and not very nice to read. A good rule is to not include the `Display` implementation of your source error in your error. Or if you do, don't return it from `source()`. Never do both.

I created a forum post on this topic a while back: https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362

I see the stack of errors here is not only `tonic` specific, but the problem exists in `hyper` as well.

### Proposal

Change some error type implementations. Especially if there are errors which are basically just combinations of sub-errors, then make them "transparent". Meaning their `Display` impl just proxies to the `Display` of the sub-error and `source()` become something like:
```rust
fn source() -> Option<dyn Error> {
    match self {
        Io(error) => error.source(),
        Something(error) => error.source(),
        ...
    }
}
```

faern:
## Motivation

Fixes #213 

## Solution

Makes the `tonic::transport::Error` type much simpler. It just forwards what's in the source error now. It was previously opaque anyway so a user of the library could not match on the `ErrorKind` or anything like that anyway.

However, this just solves the problem of the error not having redundant information in it. The library still uses a lot of opaque `Box<dyn Error>` internally. I personally think it would be nicer if `tonic` exposed a proper `hyper::Error` where the error comes from `hyper` etc. But I did not feel that was in scope for getting this part done.

### DIFF
--- a/tonic/src/transport/error.rs
+++ b/tonic/src/transport/error.rs
@@ -1,57 +1,23 @@
 use std::{error, fmt};
 
 /// Error's that originate from the client or server;
-pub struct Error {
-    kind: ErrorKind,
-    source: Option<crate::Error>,
-}
+#[derive(Debug)]
+pub struct Error(crate::Error);
 
 impl Error {
-    pub(crate) fn from_source(kind: ErrorKind, source: crate::Error) -> Self {
-        Self {
-            kind,
-            source: Some(source),
-        }
-    }
-}
-
-#[derive(Debug)]
-pub(crate) enum ErrorKind {
-    Client,
-    Server,
-}
-
-impl fmt::Debug for Error {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut f = f.debug_tuple("Error");
-        f.field(&self.kind);
-        if let Some(source) = &self.source {
-            f.field(source);
-        }
-        f.finish()
+    pub(crate) fn from_source(source: impl Into<crate::Error>) -> Self {
+        Self(source.into())
     }
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Some(source) = &self.source {
-            write!(f, "{}: {}", self.kind, source)
-        } else {
-            write!(f, "{}", self.kind)
-        }
+        self.0.fmt(f)
     }
 }
 
 impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
-        self.source
-            .as_ref()
-            .map(|e| &**e as &(dyn error::Error + 'static))
-    }
-}
-
-impl fmt::Display for ErrorKind {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{:?}", self)
+        self.0.source()
     }
 }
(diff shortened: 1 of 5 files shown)


tonic/src/transport/error.rs
```diff
@@ -1,57 +1,23 @@
...
 }
 
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Some(source) = &self.source {
-            write!(f, "{}: {}", self.kind, source)
-        } else {
-            write!(f, "{}", self.kind)
-        }
+        self.0.fmt(f)
```
LucioFranco: Would it make sense to add something here related to this being a `transport` error? Since basically we are wrapping any boxed error it might be nice to ensure that this type is know to be coming from the transport layer?

LucioFranco:
This looks great! Thank you! Looks like you just need to run `cargo fmt` and CI will be happy :)

tonic/src/transport/error.rs
faern: That breaks entire point of the error just being a proxy for the underlying error.

This error is only ever returned back from the transport module. So any user must already know it's a transport error. No?

LucioFranco: Well, I guess you can use the type system to check what error this. looks like this is what hyper does for debug https://github.com/hyperium/hyper/blob/master/src/error.rs#L322

and it seems it does what you did here for display too.

faern: I'm really not a fan of how `hyper` designed their errors :)

LucioFranco: Can you explain why? I personally have not had an issue with it but not sure what others have experienced either :)

faern: No you can't really use the typesystem for anything with any of these opaque errors. All their `Kind`/`ErrorKind` enums are private :(

faern: The only thing they ever allow you to do currently is to print them. And even then they don't look good because they have a ton of duplicated info in each layer.

LucioFranco: You should be able to downcast them https://docs.rs/hyper/0.13.1/hyper/error/struct.Error.html#method.into_cause for std errors and h2 but beyond that yeah its a bit tough, the problem is adding more error types becomes a breaking change. Now that we have the non exhaustive it should be better but I guess rusts error handling is still a bit in flux. Do you have an error type that you like as a good example?

faern: I think we have currently, with this PR, made it impossible to downcast the error to an underlying hyper error. Our new tonic error can't be downcasted because it's a newtype wrapping the hyper error. And calling `source()` on it will not yield the hyper error but rather the source of it.

That being said, I have personally never downcasted an error nor felt the need to do so.

No I don't have an example of a well written error type. I do know of some concrete pain points in, what I consider, bad errors. But I don't have a manual for nice ones. If I had I would contribute it to the official API guidelines, a document I really think is lacking in the error area.

If you make it possible to destructure the error, it's probably correct to break the API when new errors are added? Because if you have such errors you tell the user what can happen and encourage them to handle the different error cases. Which means your API indeed changes if you add a new error. Users should then get compilation errors because there will be new types of errors they are expected to handle.

PR #223, opened 2020-01-10
feat(build): Add extern_path config support

adamhjk:
Adds support for prost-build's "extern_path" feature. This allows you to
reference another prost-generated protobuf, including any traits that
may have been defined for it, from another crate or location.

## Motivation

When using protobuf definitions not only defined in, but also extended by, another crate - you need to use prost's `extern_path` feature, so that the generated code references your extended external definitions, rather than creating new ones.

## Solution

Add `extern_path` to `tonic-build`.

There are no tests, as there were none to begin with. It does Work On My Machine, though.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -76,6 +76,7 @@ mod server;
 pub struct Builder {
     build_client: bool,
     build_server: bool,
+    extern_path: Vec<(String, String)>,
     field_attributes: Vec<(String, String)>,
     type_attributes: Vec<(String, String)>,
     out_dir: Option<PathBuf>,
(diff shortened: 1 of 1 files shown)


tonic-build/src/lib.rs
```diff
@@ -111,6 +112,17 @@ impl Builder {
         self
     }
 
+    /// Declare externally provided Protobuf package or type.
+    ///
+    /// Passed directly to `prost_build::Config.extern_path`.
+    pub fn extern_path<T: AsRef<str>>(mut self, proto_path: T, rust_path: T) -> Self {
```
LucioFranco: quick change, can you change both params to be `proto_path: impl AsRef<str>`

PR #228, opened 2020-01-11
feat(transport): Add server side peer cert support

LucioFranco:
This finishes up #85


jen20:
This looks good to me - I’d come to the conclusion that actual parsing of the certificate should be handled by an interceptor in the client also.

LucioFranco:
@jen20 right I suspect at that point a user can bring in what ever parsing they want, instead of us providing it.

PR #232, opened 2020-01-14
feat: Add gRPC interceptors

LucioFranco:
This change introduces proper gRPC interceptors that are avilable
regardless of the transport used. Each codegen service now produces an
additional method called `with_interceptor` that accepts a
`Interceptor`.

All examples have been updated to use this new style and interop has a
custom `tower::Service` middleware to echo the headers. There is also a
new `interceptor` example that shows basic usage.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -86,27 +86,37 @@ path = "src/uds/client.rs"
 name = "uds-server"
 path = "src/uds/server.rs"
 
+[[bin]]
+name = "interceptor-client"
+path = "src/interceptor/client.rs"
+
+[[bin]]
+name = "interceptor-server"
+path = "src/interceptor/server.rs"
+
 [dependencies]
 tonic = { path = "../tonic", features = ["tls"] }
 prost = "0.6"
-
-tokio = { version = "0.2", features = ["rt-threaded", "time", "stream", "fs", "macros", "uds"] }
-futures = { version = "0.3", default-features = false, features = ["alloc"]}
+tokio = { version = "0.2", features = [
+    "rt-threaded",
+    "time",
+    "stream",
+    "fs",
+    "macros",
+    "uds"
+] }
+futures = { version = "0.3", default-features = false, features = ["alloc"] }
 async-stream = "0.2"
-http = "0.2"
-tower = "0.3" 
-
+tower = "0.3"
 # Required for routeguide
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 rand = "0.7"
-
 # Tracing
 tracing = "0.1"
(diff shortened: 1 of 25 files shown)


examples/src/interceptor/client.rs
```diff
@@ -0,0 +1,34 @@
...
+    let response = client.say_hello(request).await?;
+
+    println!("RESPONSE={:?}", response);
+
+    Ok(())
+}
+
+/// This function will get called on each outbound request. Returning a
+/// `Status` here will cancel the request and have that status returned to
+/// the client.
```
alce: It seems this comment should go on server.rs?

examples/src/interceptor/server.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    println!("GreeterServer listening on {}", addr);
+
+    Server::builder().add_service(svc).serve(addr).await?;
+
+    Ok(())
+}
+
+/// This function will get called on each inbound request, if a `Status`
+/// is returned, it will cancel the request and return that status.
```
alce: Looks like the comment for the client's intercept fn?

examples/src/interceptor/client.rs
LucioFranco: No I think it makes sense because the client here is the actual generated client struct.

examples/Cargo.toml
```diff
@@ -86,27 +86,37 @@ path = "src/uds/client.rs"
...
-
-tokio = { version = "0.2", features = ["rt-threaded", "time", "stream", "fs", "macros", "uds"] }
-futures = { version = "0.3", default-features = false, features = ["alloc"]}
+tokio = { version = "0.2", features = [
+    "rt-threaded",
+    "time",
+    "stream",
+    "fs",
+    "macros",
+    "uds"
```
alce: nit. Newlines on inline tables are weird. Maybe even not officially supported?

LucioFranco: I had a formatter plugin do this but ill revert 👍

tonic-build/src/client.rs
```diff
@@ -34,6 +34,11 @@ pub(crate) fn generate(service: &Service, proto: &str) -> TokenStream {
                     Self { inner }
                 }
 
+                pub fn with_interceptor(inner: T, interceptor: impl Into<tonic::Interceptor>) -> Self {
+                    let inner = tonic::client::Grpc::with_interceptor(inner, interceptor);
+                    Self { inner }
+                }
+
```
alce: Do you think it would be worth it to use a builder-like pattern instead of `whit_interceptor` constructor? Same for the server.

LucioFranco: Since there is only one config option as of now, I wanted to keep this simple, I think if we add another option in the future we go with a builder.

PR #244, opened 2020-01-18
fix(transport): Use Uri host if no domain for tls

### LINKED ISSUE #239: Endpoint.connect() with TLS fails with InvalidDNSNameError
### Version

tonic v0.1.0
tonic-build v0.1.0

### Platform

Darwin x86_64

### Crates

webpki v0.21.0

### Description

Creating a channel by calling `connect()` on an `Endpoint` configured with a TLS config fails with the error

        Error(hyper::Error(Connect, InvalidDNSNameError))

unless the `domain_name` option on the `ClientTlsConfig` is set. Contrary to the docs, the `domain_name` option is used (and required) even when the `rustls_client_config` option is set on the config.

The inner `InvalidDNSNameError` comes from the webpki crate, returned at this line: https://github.com/hyperium/tonic/blob/5fc6762435494d8df023bea8e35a5d20d81f2f3b/tonic/src/transport/service/tls.rs#L88

When the `ClientTlsConfig` builds a `TlsConnector`, the entire connection URI string is used as the domain name if the domain name has not been set on the config. https://github.com/hyperium/tonic/blob/5fc6762435494d8df023bea8e35a5d20d81f2f3b/tonic/src/transport/channel/tls.rs#L83-L86

I would expect the `domain_name` option to actually be optional, and for connections to succeed in the common case when the host component of the connection URI matches the domain name for TLS SNI.

LucioFranco:
Fixes #239

### DIFF
--- a/tonic/src/transport/channel/tls.rs
+++ b/tonic/src/transport/channel/tls.rs
@@ -1,6 +1,7 @@
 use crate::transport::{
     service::TlsConnector,
     tls::{Certificate, Identity},
+    Error,
 };
 use http::Uri;
 use std::fmt;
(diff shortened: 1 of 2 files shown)


Nokel81:
Tested, this PR does seem to work as intended.

LucioFranco:
@Nokel81 great, thanks for the update, ill get a release of this out shortly.

PR #251, opened 2020-01-25
chore(transport): Upgrade rustls and base64

repi:


jen20:
Hi @repi! Thanks for the pull request here. The CI failure you're seeing is not actually a result of your change, we'll need to either tweak the cargo-deny settings to avoid removing our dependency on `spin` or take one of the other actions. In the interests of getting your PR merged as soon as possible, I think the first option is the best for now, I'll look at this shortly.

repi:
I was going to PR `rustls` also about upgrading to `base64` 0.11, so we can wait with this one until `rustls` has upgraded so one doesn't bring in multiple versions of it

LucioFranco:
@repi thanks! Yeah, once rustls gets a publish on crates io lets do both of them in here and then make deny happy. :)

repi:
@LucioFranco new rustls 0.17 out now that uses this

LucioFranco:
@repi perfect, I am going to coordinate this with a new 0.2 release of tonic since we expose rustls in our public API.

Looks like we need to bump rustls' version to make deny happy too :)

LucioFranco:
@repi friendly ping :) Seems like we can finally merge this and update rustls?

repi:
@LucioFranco there, tested upgrading rustls crates as well now!

PR #257, opened 2020-02-03
Allow user to disable server certificate validation.

YoEight:
## Motivation
Currently, if you want to disable server certificate validation, you have to manually set a `rustls` configuration. Unfortunately, it also overrides default tonic configuration like ALPN protocols.

## Solution
This PR exposes `danger_accept_invalid_certs` in `ClientTlsConfig`. That function while still preserving `tonic` default `tls` configuration, disables server certificate validation. Because it introduces a security vulnerability, we gated that function behind `tls-dangerous` feature flag.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -75,8 +75,8 @@ tracing-futures = { version = "0.2", optional = true }
 # rustls
 tokio-rustls = { version = "0.12", optional = true }
 rustls-native-certs = { version = "0.1", optional = true }
-rustls = { version = "^0.16", features = ["dangerous_configuration"], optional = true }
-webpki = { version = "^0.21", optional = true }
+rustls = { version = "0.16", features = ["dangerous_configuration"], optional = true }
+webpki = { version = "0.21", optional = true }
 
 [dev-dependencies]
 tokio = { version = "0.2", features = ["rt-core", "macros"] }
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -35,6 +35,7 @@ transport = [
 ]
 tls = ["transport", "tokio-rustls"]
 tls-roots = ["tls", "rustls-native-certs"]
+tls-dangerous = ["rustls", "webpki"]
 
 # [[bench]]
 # name = "bench_main"
@@ -42,7 +43,7 @@ tls-roots = ["tls", "rustls-native-certs"]
 
 [dependencies]
 bytes = "0.5"
-futures-core = { version = "0.3", default-features = false } 
+futures-core = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
 tracing = "0.1"
 http = "0.2"
@@ -74,6 +75,8 @@ tracing-futures = { version = "0.2", optional = true }
 # rustls
 tokio-rustls = { version = "0.12", optional = true }
 rustls-native-certs = { version = "0.1", optional = true }
+rustls = { version = "^0.16", features = ["dangerous_configuration"], optional = true }
+webpki = { version = "^0.21", optional = true }
 
 [dev-dependencies]
 tokio = { version = "0.2", features = ["rt-core", "macros"] }
@@ -88,4 +91,3 @@ rustdoc-args = ["--cfg", "docsrs"]
 [[bench]]
 name = "decode"
 harness = false
-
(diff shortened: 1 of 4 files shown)


tonic/Cargo.toml
```diff
@@ -35,14 +35,15 @@ transport = [
 ]
 tls = ["transport", "tokio-rustls"]
 tls-roots = ["tls", "rustls-native-certs"]
+tls-dangerous = ["rustls", "webpki"]
```
LucioFranco: This should only have to enable `"rustls/dangerous_configuration"` then we leave it out of the default. So this is an opt-in feature that will only enable that specific rustls feature if you opt in.

tonic/src/transport/channel/tls.rs
```diff
@@ -15,6 +15,7 @@ pub struct ClientTlsConfig {
     cert: Option<Certificate>,
     identity: Option<Identity>,
     rustls_raw: Option<tokio_rustls::rustls::ClientConfig>,
+    _certs_validation: CertsValidation,
```
LucioFranco: This should probably be a bool and should be feature gated.

```rust
#[cfg(feature = "tls-dangerous")]
cert_validation: bool
```

tonic/src/transport/service/tls.rs
```diff
@@ -32,6 +32,14 @@ enum TlsError {
     PrivateKeyParseError,
 }
 
+/// If we do certificats validation.
+#[derive(Copy, Clone, Debug)]
+pub(crate) enum CertsValidation {
```
LucioFranco: if we use a bool we can get rid of this.

tonic/Cargo.toml
```diff
@@ -74,6 +75,8 @@ tracing-futures = { version = "0.2", optional = true }
 # rustls
 tokio-rustls = { version = "0.12", optional = true }
 rustls-native-certs = { version = "0.1", optional = true }
+rustls = { version = "^0.16", features = ["dangerous_configuration"], optional = true }
```
LucioFranco: Do we need the `^` here?

tonic/src/transport/channel/tls.rs
```diff
@@ -15,6 +15,7 @@ pub struct ClientTlsConfig {
     cert: Option<Certificate>,
     identity: Option<Identity>,
     rustls_raw: Option<tokio_rustls::rustls::ClientConfig>,
+    _certs_validation: bool,
```
LucioFranco: I think we can remove the `_` prefix here?

tonic/Cargo.toml
YoEight: I didn't know feature flags were transitive, I'm looking into it.

tonic/Cargo.toml
YoEight: My bad, I thought that `0.16` meant `[0.16 - 0.17[` while `^0.16` was `>= 0.16`

tonic/src/transport/channel/tls.rs
YoEight: Problem is we only use that field when our new feature flag is enabled, resulting in a rustc warning when the feature flag is off.

tonic/src/transport/channel/tls.rs
YoEight: It makes it really difficult to put that macro here considering I added a new parameter to `new_with_rustls_cert`.

tonic/Cargo.toml
LucioFranco: Honestly, I don't remember mostly just want to keep them similar to the others :smile:

tonic/src/transport/channel/tls.rs
LucioFranco: So I think the right option here is to just add `#[cfg(feature = "tls-dangerous")]` for each occurrence of this. So its impossible to even set if that feature flag is not set, what do you think?

tonic/Cargo.toml
YoEight: After thinking about it, I don't understand what you want me to do

LucioFranco: It should be possible to set `tls-dangerous` to be like:

```toml
tls-dangerous = ["rustls/dangerous-config", "webpki"]
```

Since, rustls is already included with `tokio-rustls` this feature set should work.

Let me know if that makes sense!

LucioFranco:
I am going to close this for now since its a bit old now and if we can avoid having to expose something like this then I think that is better since it is dangerous :D lmk if you have anyquestions.

LucioFranco:
I would suggest for now to go around the transport module and use hyper-rustls + hyper directly.

YoEight:
@LucioFranco could you elaborate on this, specifically in the light of 0.7? Thanks for your time.

LucioFranco:
https://github.com/hyperium/tonic/pull/968 here is a PR that adds an example for this.

Issue #260, opened 2020-02-06
tonic-build fails when service in/out objects are extern_path'd


cthulhua:
### Version

tonic 0.1.1

### Platform

Darwin Dannys-Work-MacBook-Pro.local 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64

### Crates

tonic-build

### Description

The `replace_wellknown` function in tonic-build/src/lib.rs fails when any objects referred to in service method definitions have been replaced (via tonic/prost's extern_path functionality). These paths are absolute paths and should not be replace with a relative path rooted at super.

This failure results in a build failure (panic during build.rs execution):
```
   Compiling my_application v0.1.0 (/Users/danny/tonic_extern_path/my_application)
error: failed to run custom build command for `my_application v0.1.0 (/Users/danny/tonic_extern_path/my_application)`

Caused by:
  process didn't exit successfully: `/Users/danny/tonic_extern_path/target/debug/build/my_application-807cb5f2cea6c0e3/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("expected identifier")', /Users/danny/.cargo/registry/src/github.com-1ecc6299db9ec823/tonic-build-0.1.1/src/lib.rs:332:9
stack backtrace:
   0:        0x10b6f6135 - backtrace::backtrace::libunwind::trace::h2645f0921d177f94
                               at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1:        0x10b6f6135 - backtrace::backtrace::trace_unsynchronized::h2d634d751cc39c09
                               at /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2:        0x10b6f6135 - std::sys_common::backtrace::_print_fmt::hbf6c15b03ac1eab6
                               at src/libstd/sys_common/backtrace.rs:77
   3:        0x10b6f6135 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h0e1e1d13ef39c2d8
                               at src/libstd/sys_common/backtrace.rs:59
   4:        0x10b712e6d - core::fmt::write::h1fc31e3d3e8cc0fb
[...]

PR #261, opened 2020-02-06
fix(build): Don't replace extern_paths

cthulhua:
tonic-build will replace all paths except for google well known types
with one rooted at the super module. For paths that have already been
replaced with a fully qualified path via the extern_path config option,
(e.g. "::uuid::Uuid"), this results in an invalid path
(e.g. "super::::uuid::Uuid"), and a build failure. These paths should
also be excluded when prefixing relative modules with super.

## Motivation

See commit message, and #260 

When extern_path'd are used in service method definitions, prost-build panics.

See https://github.com/cthulhua/tonic_extern_path/tree/master for a repro of the issue (it works when build against this branch).

## Solution

Similar to how google well known types are handled, in the event an absolute path would be made relative, leave it absolute.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,4 +9,6 @@ members = [
     "tests/included_service",
     "tests/same_name",
     "tests/wellknown",
+    "tests/extern_path/uuid",
+    "tests/extern_path/my_application"
 ]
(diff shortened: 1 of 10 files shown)


cthulhua:
> Sorry for the delay in review! I think this makes sense, though we should add this somewhere to the docs. I think under extern_path state that you need to prefix them with `::`. Otherwise LGTM

Thanks, I've added to the docstring for extern_path clarifying the need for fully qualified paths

cthulhua:
Sorry about the delay; been caught up in quite a bit recently. I adapted my repro crates into tests; i just assert that a type generated from a service uses an external crate.

PR #272, opened 2020-02-21
initial cut of RouterService

cthulhua:


LucioFranco:
This looks pretty good! I think we should just expose it and I think that should work!

alce:
Hi, this looks very useful. 

Unless I'm mistaken, could we add something like this to the Router's impl instead of the `service` free function and it should be good to go?

```rust
    pub fn into_service(self) -> RouterService<A, B> {
        RouterService { router: self }
    }
```

a1ph:
Should it be closed as #419 is merged.

PR #275, opened 2020-02-28
Add optional service methods

hlb8122:
## Motivation
https://github.com/hyperium/tonic/issues/265

## Solution
This is currently an incomplete solution - it is the minimum required to demonstrate this approach and allow for a decision.

It mirrors closely the way `optional_layer` method works.

## Considerations
* Is this zero-cost?
* Should `NamedService` be implemented on `Either` like this? Does this break anything?
* There's an `Unimplemented` service which seems to do the same thing as `EmptyService`, perhaps we should use that instead?

## TODO
After deciding whether this approach is feasible the following needs to be done:

- [x] Implement `add_optional_service` onto the `Router` in a similar manner
- [x] Documentation

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -240,6 +240,10 @@ impl Server {
     ///
     /// This will clone the `Server` builder and create a router that will
     /// route around different services.
+    ///
+    /// # Note
+    /// Even when the argument given is `None` this will capture *all* requests to this service name.
+    /// As a result, one cannot use this to toggle between two indentically named implementations.
     pub fn add_optional_service<S>(
         &mut self,
         svc: Option<S>,
@@ -365,6 +369,42 @@ where
             let path = req.uri().path();
 
             path.starts_with(&svc_route)
+        };
+        let routes = routes.push(pred, svc);
+
+        Router { server, routes }
+    }
+
+    /// Add a new optional service to this router.
+    ///
+    /// # Note
+    /// Even when the argument given is `None` this will capture *all* requests to this service name.
+    /// As a result, one cannot use this to toggle between two indentically named implementations.
+    pub fn add_optional_service<S>(
+        self,
+        svc: Option<S>,
+    ) -> Router<Either<S, Unimplemented>, Or<A, B, Request<Body>>>
+    where
+        S: Service<Request<Body>, Response = Response<BoxBody>>
+            + NamedService
+            + Clone
+            + Send
+            + 'static,
+        S::Future: Send + 'static,
+        S::Error: Into<crate::Error> + Send,
+    {
+        let Self { routes, server } = self;
+
+        let svc_name = <S as NamedService>::NAME;
+        let svc_route = format!("/{}", svc_name);
+        let pred = move |req: &Request<Body>| {
+            let path = req.uri().path();
+
+            path.starts_with(&svc_route)
+        };
+        let svc = match svc {
+            Some(some) => Either::A(some),
+            None => Either::B(Unimplemented::default()),
         };
         let routes = routes.push(pred, svc);
 
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -243,7 +243,7 @@ impl Server {
     pub fn add_optional_service<S>(
         &mut self,
         svc: Option<S>,
-    ) -> Router<Either<S, EmptyService>, Unimplemented>
+    ) -> Router<Either<S, Unimplemented>, Unimplemented>
     where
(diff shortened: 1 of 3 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -80,6 +81,10 @@ pub trait NamedService {
     const NAME: &'static str;
 }
 
+impl<S: NamedService, T> NamedService for Either<S, T> {
+    const NAME: &'static str = S::NAME;
```
LucioFranco: This will only match if the `S` service name matches but doesn't work for T?

hlb8122: Yeah.

A better alternative here would probably be:
`impl <S:NamedService> NamedService for Either<S, Empty Service>`.

LucioFranco: I think that having a `const` here limits us terribly because it needs to evaluate the service name at compile time while Either is a runtime enum trick. Because of that it doesn't really match very well.

I want to say warp has the same issue? There is one possible solution and that may be boxing things but I am not exactly sure how that would work with the const since you'd still need to define the `NAME` type at compile time even if you want to trait object.

hlb8122: @LucioFranco Sorry for the inactivity and sorry I don't follow.

I've added an the following example which toggles a service based on a command line argument.

```rust
    let args: Vec<String> = env::args().collect();
    let enabled = args.get(1) == Some(&"enabled".to_string());

    let addr = "[::1]:50051".parse().unwrap();

    let optional_service = if enabled {
        println!("MyGreeter enabled");
        Some(GreeterServer::new(greeter))
    } else {
        println!("MyGreeter disabled");
        None
    };

    println!("GreeterServer listening on {}", addr);

    Server::builder()
        .add_optional_service(optional_service)
        .serve(addr)
        .await?;
```
and it seems to work fine.

LucioFranco: Try adding a service after, I think that will cause the issue I was talking about.

hlb8122: So I just gave it another go:
```rust
#[tokio::main]
    let args: Vec<String> = env::args().collect();
    let enabled = args.get(1) == Some(&"enabled".to_string());

    // Set-up services
    let switch_service = if enabled {
        println!("Switch enabled");
        Some(SwitchableServer::new(MySwitch::default()))
    } else {
        println!("Switch disabled");
        None
    };
    let persistent_service = PersistentServer::new(MyPersistent::default());

    // Build server
    let addr = "[::1]:50051".parse().unwrap();
    Server::builder()
        .add_optional_service(switch_service)
        .add_service(persistent_service)
        .serve(addr)
        .await?;

    Ok(())
}
```
and still seems to work fine.

Screenshot from 2020-07-01 21-22-15

LucioFranco: Right, what I don't think will work is if the Either enum is on the `T` generic, since that doesn't even have a bound? I can try to take a look tomorrow but in my head the `T` service here should not work.

hlb8122: @LucioFranco okey. If you give it the green light I'll go ahead and add docs (tests) and the method on the `Router`.

tonic/src/transport/server/mod.rs
```diff
@@ -221,6 +226,30 @@ impl Server {
...
+            + NamedService
+            + Clone
+            + Send
+            + 'static,
+        S::Future: Send + 'static,
+        S::Error: Into<crate::Error> + Send,
+    {
+        let svc = match svc {
+            Some(some) => Either::A(some),
+            None => Either::B(EmptyService),
```
LucioFranco, line 248: I guess I missed this before but this is where my brain was breaking. Is there a reason we can't reuse the `Unimplemented` service here?

tonic/src/transport/server/mod.rs
LucioFranco: So another issue I see here (that may be totally minor and we may want to document) is that this service will catch the services with this name. So doing funky stuff around enabling two variations of the same service will cause issues.

hlb8122: @LucioFranco Yeah I thought this might be the case, which is why I made two explicitly different services [here](https://github.com/hyperium/tonic/pull/275#discussion_r448596956).

I'll try paraphrase the phenomena in the documentation as best I can.

tonic/src/transport/server/mod.rs
hlb8122: I was unsure and was hoping you could tell me. If you see no problem with it then I'll go ahead.

PR #280, opened 2020-03-02
Grpc proxy

john-hern:
Just trying to get feedback from this approach. This is an attempt to hook the Svc pipeline in Tonic to expose a proxy for GRPC-WEB. It's by all means not a finished product. I'm interested in this approach verse how I'm approaching it in this other project: 

https://github.com/john-hern/tonic_grpc_web

In either approach some changes to Tonic are required. The "Wrapped" approach in tonic_grpc_web, would require allowing http1 requests via hyper, or some way to configure. Additionally, since services are resolved via naming, it may be required to be a layer, so that might have to be exposed.  Thoughts?


LucioFranco:
@john-hern hey! looks like there was a bunch of extra stuff included in the diff, would you be able to remove those?

I also suggest instead of chaging the gitignore within the project setting up a global gitignore with `.vscode` in there https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

For example my global gitignore contains:
```
.vscode/*
```

john-hern:
Okay, let me clean it up. Sorry, not used to work on shared projects.

LucioFranco:
@john-hern no worries :)

PR #287, opened 2020-03-08
fix(build): Remove debug println!

alce:
Remove unnecessary println! from tonic-build:

`out: Output { status: ExitStatus(ExitStatus(0)), stdout: "", stderr: "" }`

can get a bit noisy and I think it serves no purpose.

Also, fixed a couple of typos, sorry to mix things up but such small edits...hope you don't mind.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -92,7 +92,6 @@ pub fn fmt(out_dir: &str) {
             .output()
             .unwrap();
 
-        println!("out: {:?}", out);
         assert!(out.status.success());
     }
 }
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -66,7 +66,9 @@ pub mod prost;
 pub mod schema;
 
 #[cfg(feature = "rustfmt")]
-use std::process::Command;
+use std::io::{self, Write};
+#[cfg(feature = "rustfmt")]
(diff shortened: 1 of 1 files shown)


tonic-build/src/lib.rs
```diff
@@ -92,7 +92,6 @@ pub fn fmt(out_dir: &str) {
             .output()
             .unwrap();
 
-        println!("out: {:?}", out);
```
LucioFranco: I think this was originally here to give you the context on the error, maybe it just makes sense to actually destructure this error and print something nice when it fails so the user can figure out what went wrong, what do you think?

tonic-build/src/lib.rs
```diff
@@ -83,16 +83,18 @@ pub fn fmt(out_dir: &str) {
...
             .arg("2018")
             .arg(format!("{}/{}", out_dir, file))
-            .output()
-            .unwrap();
+            .output();
 
-        assert!(out.status.success());
+        match result {
+            Err(e) => println!("error running rustfmt: {:?}", e),
+            Ok(out) => assert!(out.status.success()),
```
LucioFranco: I think it may also make sense to remove this assert :) and give a good error message if its not successful.

alce: Sounds good. Just pushed another commit that forwards rustfmt's stderr and exit status in case of failure.

PR #288, opened 2020-03-08
Grpc-web proxy

john-hern:
Okay, here is a cleaned up version. Pulled the website out I used to test GRPC-Web JS client with the code. Still have a lot of work to do, but this is likely the minimal amount of changes to Tonic I could see in order to get the proxy at the right layer in the stack. I still have a bit of work to do. Hopefully the approach here is clear.


LucioFranco:
@john-hern sorry for the delay on this! I took a quick look, it looks like you're heading in the right direction. I have yet to read through more of the grpc-web spec but would you be able to explain a bit more the use case and how you see this fitting into tonic? That and getting CI to pass and I think we can move forward!

LucioFranco:
Superseded by https://github.com/hyperium/tonic/pull/455

PR #302, opened 2020-03-27
chore(build): Refactor codegen traits

LucioFranco:

### DIFF
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -1,6 +1,6 @@
 fn main() {
-    tonic_build::prost::compile_protos("proto/helloworld/helloworld.proto").unwrap();
-    tonic_build::prost::compile_protos("proto/routeguide/route_guide.proto").unwrap();
-    tonic_build::prost::compile_protos("proto/echo/echo.proto").unwrap();
-    tonic_build::prost::compile_protos("proto/google/pubsub/pubsub.proto").unwrap();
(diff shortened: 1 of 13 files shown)


Skareeg:
I don't mean to intrude here, but why was build.rs code changed back from tonic_build::prost::compile_protos to tonic_build::compile_protos? That is what the file originally was...

LucioFranco:
We just re-export the prost version, since that is the default we want users to use. If you disable the feature you need to go through lower-level functions to get the functionality.

LucioFranco:
Right now, we don't have fbs support yet but that has not been mapped out yet! Likely, for the first phase of fbs we will expose a fbs compile_protos and still leave the prost one as a default until we find a more ergonomic method.

The plan though is to still keep prost as the main proto spec.

msinojia:
> Right now, we don't have fbs support yet but that has not been mapped out yet! Likely, for the first phase of fbs we will expose a fbs compile_protos and still leave the prost one as a default until we find a more ergonomic method.
> 
> The plan though is to still keep prost as the main proto spec.

Since it has been a few months, I wanted to stay up-to date. Is the flatbuffers support out yet, or would it be in the near future?

LucioFranco:
@msinojia doesn't seem like the work has been completed on the fbs implementation and some of this work has stalled. I don't think we currently have a clear plan for it until someone is ready to drive it. I likely won't have time since there are some other priorities for tokio that need to be done and we don't require fbs at work.

Issue #135, opened 2019-11-12
Implement gRPC standard health checking service


jen20:
### Motivation

gRPC defines a standard health checking service, [here](https://github.com/grpc/grpc/blob/master/doc/health-checking.md). This is a somewhat common requirement for services.

### Proposal

Implement a crate which allows this to be implemented via a simple trait and exposed via the Tonic `add_service` function.

### Alternatives

Implement this as a crate outside tonic. This _may_ be a better option, but I don't think there is a better place to track it than here regardless of which path we take implementing it.

LucioFranco:
@jen20 separate crate makes sense 👍

xd009642:
Has there been any progress on this? Just wondering before I implement my own :smile:

jen20:
@xd009642 I haven't made any yet, but do need to at some point. If you do it before Saturday you'll likely beat me to it though :-)

xd009642:
I think I'll be starting on Monday so you may still beat me :smile:

zakhenry:
I’ve done the other end of this (client health checking) using tonic - https://github.com/zakhenry/nightingale It might make sense for this lib to be in there same repo? I’d be happy to contribute 😁

rlabrecque:
Here's something for someone to start on, I hacked this together quickly a few weeks back, I don't think it quite follows the spec, and is kinda gross, but it was good enough for my needs initially.

```
pub trait NamedService {
    fn get_name() -> String;
}
```

```
use grpc::health::v1::{
    health_check_response::ServingStatus,
    server::{Health, HealthServer},
    HealthCheckRequest, HealthCheckResponse,
};
use std::collections::VecDeque;
use tonic::{transport::ServiceName, Code, Request, Response, Status};
#[allow(unused_imports)]
use tracing::{debug, error, info, trace, warn};

type HealthServiceResult<T> = Result<Response<T>, Status>;
type Stream = VecDeque<Result<HealthCheckResponse, Status>>;

#[derive(Clone, Debug)]
pub struct HealthService {
    services: Vec<String>,
}

impl HealthService {
    pub fn new(service_names: Vec<String>) -> HealthServer<Self> {
        let mut health_service = HealthService {
            services: service_names,
        };

        health_service.services.push(Self::get_name());
        health_service.services.push(String::new()); // Empty string to match against for "any service"

        HealthServer::new(health_service)
    }
}

impl NamedService for HealthService {
    fn get_name() -> String {
[...]

PR #304, opened 2020-03-27
feat(health): Add tonic-health server impl

jen20:
This commit adds a new crate `tonic-health` which implements the [standard GRPC Health Checking][checking] protocol, as per #135.

Currently there is only a server implementation, though others have alluded in the discussion in #135 that client implementations exist which could also be imported as necessary.

A example server has also been added - once the client work is done a client for this should be added also.

Some further work that needs doing:

- [ ] Clean up `Cargo.toml` metadata
- [ ] Add a `transport` feature which unlocks the various methods which use `NamedService`
- [ ] Add some tests of both `Check` and `Watch`
- [ ] Consider whether to write or import a client as part of this PR or separately.

[checking]: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

### DIFF
--- a/tonic-health/Cargo.toml
+++ b/tonic-health/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "tonic-health"
+version = "0.1.0"
+authors = ["James Nugent <james@jen20.com>"]
+edition = "2018"
+
+[dependencies]
+tokio = { version = "0.2", features = ["sync"] }
+tonic = { path = "../tonic" }
+bytes = "0.5"
+prost = "0.6"
+
+[dev-dependencies]
+tokio = { version = "0.2", features = ["rt-core", "macros"]}
+
+[build-dependencies]
+tonic-build = { path = "../tonic-build" }
--- a/tonic-health/src/server.rs
+++ b/tonic-health/src/server.rs
@@ -1,0 +1,166 @@
+use crate::proto::health_server::{Health, HealthServer};
+use crate::proto::{HealthCheckRequest, HealthCheckResponse};
+use crate::ServingStatus;
+use std::collections::HashMap;
+use std::sync::Arc;
+use tokio::sync::{mpsc, watch, RwLock};
+use tonic::{transport::NamedService, Request, Response, Status};
+
+/// Creates a `HealthReporter` and a linked `HealthServer` pair. Together,
+/// these types can be used to serve the gRPC Health Checking service.
+///
+/// A `HealthReporter` is used to update the state of gRPC services.
+///
+/// A `HealthServer` is a Tonic gRPC server for the `grpc.health.v1.Health`,
+/// which can be added to a Tonic runtime using `add_service` on the runtime
+/// builder.
+pub fn health_reporter() -> (HealthReporter, HealthServer<impl Health>) {
+    let reporter = HealthReporter::new();
+    let service = HealthService::new(reporter.statuses.clone());
+    let server = HealthServer::new(service);
+
+    (reporter, server)
+}
+
+type StatusPair = (watch::Sender<ServingStatus>, watch::Receiver<ServingStatus>);
+
+/// A handle providing methods to update the health status of gRPC services. A
+/// `HealthReporter` is connected to a `HealthServer` which serves the statuses
+/// over the `grpc.health.v1.Health` service.
+#[derive(Clone, Debug)]
+pub struct HealthReporter {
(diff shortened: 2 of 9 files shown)


tonic-health/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+
+
+[dependencies]
+async-stream = "0.2"
+tokio = { version = "0.2", features = ["sync", "stream"] }
+tonic = { path = "../tonic", features = ["codegen", "data-prost", "transport"] }
+bytes = "0.5"
+prost = "0.6"
+
+
```
LucioFranco:
```suggestion
```

tonic-health/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+[dev-dependencies]
+tokio = { version = "0.2", features = ["rt-core", "macros"]}
+
+
```
LucioFranco:
```suggestion
```

tonic-health/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+license = "MIT"
+repository = "https://github.com/hyperium/tonic"
+homepage = "https://github.com/hyperium/tonic"
+description = """
+Health Checking module of `tonic` gRPC implementation.
+"""
+readme = "README.md"
+categories = ["network-programming", "asynchronous"]
+keywords = ["rpc", "grpc", "async", "healthcheck"]
+
```
LucioFranco:
```suggestion
```

tonic-health/src/server.rs
```diff
@@ -0,0 +1,159 @@
+use crate::proto::health_server::{Health, HealthServer};
+use crate::proto::{HealthCheckRequest, HealthCheckResponse};
+use crate::ServingStatus;
+use std::collections::HashMap;
+use std::pin::Pin;
+use std::sync::Arc;
+use tokio::stream::Stream;
+use tokio::sync::{watch, RwLock};
+use tonic::{transport::NamedService, Request, Response, Status};
```
LucioFranco: We should probably feature gate transport stuff behind a `transport` feature?

jen20:
I think we should address a client here in a separate PR, so this looks ready for re-review (at least the added tests) - everything else has been addressed.

LucioFranco:
@jen20 merge when you're ready!

jen20:
Thanks for reviewing @LucioFranco!

jcramb:
Is the tonic-health checking expected to be working right now because using the example provided only results in `Error: Status { code: Unimplemented }` being returned for every connection. 

For reference I used the example `health` server and then compiled my own client invoking the health check in the following way,

```
let mut client = pb::health_client::HealthClient::new(channel);
let req = pb::HealthCheckRequest {
    service: "",
};
let resp = client.check(req).await?.into_inner();
println!("{}", resp.status);
```

This all compiled and connected successfully returning the error stated above. Am I using this incorrectly or do you have any idea on how to fix this? I've even printed out the service name as used by the tonic-health crate `helloworld.Greeter` and set this explicitly but it still returns the same result.

jen20:
Hmm, that's curious - it is indeed expected to be working! I'll take a look and see if I can reproduce this...

jen20:
I've just tried this on the current `master` of `tonic` making use of `grpcurl`. Some of the paths are specific to my checkout directory, and I had to change the port of the server owing to a conflict on my machine, but all seems to be well overall:

```console
$ cargo run --bin health-server
   Compiling examples v0.1.0 (/Users/James/Code/rust/tonic/examples)
    Finished dev [unoptimized + debuginfo] target(s) in 4.89s
     Running `/Users/James/Code/rust/tonic/target/debug/health-server`
HealthServer + GreeterServer listening on [::1]:50057
```

```console
$ grpcurl -plaintext \
        -proto=/Users/James/Code/rust/tonic/tonic-health/proto/health.proto \
        -import-path=/Users/James/Code/rust/tonic/tonic-health/proto \
        -format=json \
        -d='{"service": "helloworld.Greeter"}' \
        '[::1]:50057' \
        grpc.health.v1.Health.Check
{
  "status": "SERVING"
}
```

If I try a `service` which is not registered (including `""`, which is not special cased) I get a `NotFound` status code rather than an `Unimplemented` status code.

If you have `grpcurl` available (it's in Homebrew on macOS, but not sure on other platforms), could you try a similar command line modified for your checkout directories to try to narrow down what is going on here, @jcramb?

PR #306, opened 2020-03-28
fix(build): Server service uses generic body bound

### LINKED ISSUE #281: Generation of server code independent of tonic transport
### Crates

tonic-build

### Motivation

In version 0.1.1, disabling feature `transport` in both tonic and tonic-build results in generated server code that does not compile due to an unconditional dependency on `tonic::codegen::HyperBody`.
Such coupling is not present in the client code, where it is possible to use the generic client with other `Service` impls, not necessarily dependent on `hyper`.

### Proposal

The generated `Service` impl should be generic over the body type, with suitable bounds that are implemented for `hyper::Body`.

bnjjj:
## Motivation

close #281 

## Solution

I choose to use `BoxBody`as a generic body when the transport feature is disabled. Let me know if it's a good choice.

### DIFF
--- a/tonic-build/src/server.rs
+++ b/tonic-build/src/server.rs
@@ -49,7 +49,12 @@ pub fn generate<'a, T: Service<'a>>(service: &'a T, context: &T::Context) -> Tok
                 }
             }
 
-            impl<T: #server_trait> Service<http::Request<RequestBody>> for #server_service<T> {
+            impl<T, B> Service<http::Request<B>> for #server_service<T>
+                where
+                    T: #server_trait,
+                    B: HttpBody + Send + Sync + 'static,
+                    B::Error: Into<StdError> + Send + 'static,
+            {
                 type Response = http::Response<tonic::body::BoxBody>;
                 type Error = Never;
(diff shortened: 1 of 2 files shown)


tonic-build/src/server.rs
```diff
@@ -49,7 +49,7 @@ pub fn generate<'a, T: Service<'a>>(service: &'a T, context: &T::Context) -> Tok
                 }
             }
 
-            impl<T: #server_trait> Service<http::Request<HyperBody>> for #server_service<T> {
+            impl<T: #server_trait> Service<http::Request<RequestBody>> for #server_service<T> {
```
LucioFranco: So instead of making this accept a BoxBody we actually want to make this generic over it. So we can do something like:

```rust
impl<T, B> Service<http::Request<B>> for #server_service<T>
where
	T: #server_trait,
	B: HttpBody + Send + Sync + 'static,
	B::Error: Into<StdError> + Send + 'static,
{ ... }
```

bnjjj: Updated ! Thanks a lot for your help

Issue #292, opened 2020-03-11
Add ability to set `Status` details


LucioFranco:
Reference issue https://github.com/tower-rs/tower-grpc/issues/136

fluxxu:
Looks like there is a PR for this in the old repo by @GeorgeHahn:
https://github.com/tower-rs/tower-grpc/pull/191
I can help to transfer it to this repository if needed

GeorgeHahn:
That PR was the bare minimum to get messages into the `Status` details; it could use some API love. This would be great to have though!

PR #308, opened 2020-03-29
feat: Add Status with Details Constructor

SkamDart:
Changes are just a Copy pasta of [this pr](https://github.com/tower-rs/tower-grpc/pull/191) in tower-rs.

## Motivation
Reference #292 for motivation. 

## Solution
The prescribed solution exposes two constructors with `Status::with_details` and `Status::with_raw_details` that allow the caller to construct a Status with a `Code`, `String`, and `Bytes` or a `Message` to be converted into bytes.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -1,6 +1,7 @@
 use bytes::Bytes;
 use http::header::{HeaderMap, HeaderValue};
 use percent_encoding::{percent_decode, percent_encode, AsciiSet, CONTROLS};
+use prost::Message;
 use std::{borrow::Cow, error::Error, fmt};
 use tracing::{debug, trace, warn};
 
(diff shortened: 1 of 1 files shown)


SkamDart:
Where should I add the tests? Wasn't quite sure where to place them as they require a proto fixture.

tonic/src/status.rs
```diff
@@ -401,6 +402,24 @@ impl Status {
...
+    pub fn with_raw_details(code: Code, message: impl Into<String>, details: Bytes) -> Status {
+        Status {
+            code,
+            message: message.into(),
+            details: details,
+        }
+    }
+
+    /// Create a new `Status` with the associated code, message, and protobuf details field.
+    pub fn with_details(code: Code, message: impl Into<String>, details: impl Message) -> Status {
```
LucioFranco: I actually believe we should not include a prost based one here. I think we can just have the one above that takes a Bytes and we remove the impl Into<String>. We can also rename the one above to `with_details`. then let the user choose if they want to encode via prost or something else.

SkamDart: Agreed. Just copy pasta from the PR I inherited. I'm fairly new to Rust so I appreciate your patience. Is this `cargo check --no-default-features` CI check to enforce that there's no unnecessary coupling of external crates to the core library?

PR #313, opened 2020-03-31
chore: Update blocking example drop order

Sushisource:
## Motivation
Drop order in blocking example is actually the opposite of what the comment mentions

## Solution
Switch the order. According to https://github.com/rust-lang/rfcs/blob/master/text/1857-stabilize-drop-order.md#tuples-structs-and-enum-variants Rust structs drop their fields in decl order (unlike C++ which is I imagine what tripped things up here). I just learned this recently myself!


LucioFranco:
This seems good, @Sushisource can you rebase against master to get CI fixed and then we can merge!

PR #326, opened 2020-04-07
feat(transport): Add AsRef impl for Certificate

IvanVTBTL:
## Motivation

Currently there is no way to examine certificates obtained by peer_certs functions

## Solution

Introduced functions enable consumer to parse and examine provided DER certificates


IvanVTBTL:
@LucioFranco pinging as suggested on discord

PR #341, opened 2020-05-05
feat(transport): Dynamic load balancing

dawid-nowak:
Attempt to add a feature to Tonic so client side load balancing can be done over a
group of endpoints that change dynamically over time.

## Motivation
An external service which is responsible for discovering new endpoints
and monitoring their health could make a decision to add/remove an
endpoint from the load balancing group in Tonic client. At the moment, the tonic only allows to load balance across endpoints from a list.

## Solution
The solution is to expose a new method on Channel interface called balance_with_manager which will take a trait EventManager defined in Transport. EventManager will hold a lists of endpoint which need to be added/removed from balance service in tower. 
The approach is very similar to existing Channel::balance_list.

##Comments
Comments on how to improve it are welcome :)


examples/Cargo.toml
```diff
@@ -116,14 +124,14 @@ prost = "0.6"
 tokio = { version = "0.2", features = ["rt-threaded", "time", "stream", "fs", "macros", "uds"] }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
 async-stream = "0.2"
-tower = "0.3"
+tower = { version="0.3"}
```
LucioFranco, line 127:
```suggestion
tower = "0.3"
```

LucioFranco:
This seems like an interesting start, I am a bit cautious about adding more complex types instead of just exposing the ability to pass in your own discover? I'd rather expose some sort of discover or channel hook than a endpoint manager. What do you think?

dawid-nowak:
I agree that passing discover would be cleaner/better. 

For time being I just wanted to keep it as close to existing implementation.  I looked into passing discover directly as there already is a method **balance** in **channel::mod.rs** that does exactly that.   
The problem is that method takes **Connection** which is not public. The code in  **balance_list** and **ServiceList** which implements **Discover** converts public **Endpoint** to private **Connection**. 

If we wanted to expose **Discover** directly we would need to make **Connection** public or seek some other solution which converts from **Endpoint** to **Connection**.

Ultimately, I am not entirely sure how to solve this conundrum :)

LucioFranco:
@dawid-nowak could we have a `(Key, Endpoint)` pair? That is sent over the channel via the enum https://docs.rs/tower-discover/0.3.0/tower_discover/enum.Change.html. We can then use endpoint which is public then internally translate it to a connection? That should work right?

dawid-nowak:
I think we are in agreement that ideally that would be the case. 

Expose Discover that produces the stream of Change events. 
I had a dig around and I think the main obstacle is that **tower::Discover** trait takes **Service** type. At the moment **Endpoint** is not a service so I am not sure how would that work. 

Is that correct ?

LucioFranco:
@dawid-nowak yeah, so I think instead of exposing a `D: Discover` we just assume that the user always wants to configure the tonic clients via `Endpoint`. This way we can just have a `channel::<Change<SomeKey, Endpoint>>` we can then pass out the `tx` side and let the user push change events.

This way we can implement discover ourselves by taking the stream of the channel change events and mapping them to our internal connection type. So internally we would implement a new discover type similar to `ServiceList` but that wraps a the `rx` end of a channel of change enum where the value is `Endpoint`.

Hopefully that makes a bit of sense, I am totally happy to sketch some thing out a bit more :)

dawid-nowak:
yeah, that makes sense, 
will need to figure out how to map those streams :)  so if you have examples handy that would speed up the process

LucioFranco:
basically I think we can do something similar to this https://docs.rs/tonic/0.2.1/tonic/transport/channel/struct.Channel.html#method.balance_list. Where we have:

```rust
fn balance_channel::<K>() -> (Sender<Change<K, Endpoint>>, Self)
```

This then would allow the user to:

```rust
let (tx, channel) = Channel::balance_channel();

tx.send(("some_key", some_endpoint)).await.unwrap();

channel.call(...).await;
```

We then implement a new discover type that is similar to `ServiceList` but uses a channel. We can then do something very similar to this https://github.com/hyperium/tonic/blob/master/tonic/src/transport/channel/mod.rs#L115

dawid-nowak:
Take 2. I think we are getting closer.

tonic/src/transport/channel/mod.rs
```diff
@@ -103,18 +106,23 @@ impl Channel {
...
-        let discover = ServiceList::new(list);
+        Self::balance_channel(tokio::stream::iter(list))
+    }
 
-        Self::balance(discover, buffer_size)
+    /// Balance a list of [`Endpoint`]'s.
+    ///
+    /// This creates a [`Channel`] that will listen to a stream of change events and will add or remove provided endpoints.
+    pub fn balance_channel<K>(
+        changes: impl Stream<Item = Change<K, Endpoint>> + Unpin + Send + 'static,
```
LucioFranco, line 119: So for the sake of simplicity I think this should take a `size: usize` and return the `tx` end of the channel. For the `balance_list` fn we can just iterate through all the endpoints and push them into the channel.

tonic/src/transport/channel/mod.rs
```diff
@@ -103,18 +106,23 @@ impl Channel {
...
+    ) -> Self
+    where
+        K: Hash + Eq + Send + Clone + Unpin + 'static,
```
LucioFranco, line 122: `K` doesn't need `Unpin`

tonic/src/transport/service/discover.rs
```diff
@@ -1,34 +1,36 @@
...
+use tokio::stream::Stream;
 use tower::discover::{Change, Discover};
 
-pub(crate) struct ServiceList {
-    list: VecDeque<Endpoint>,
-    connecting:
-        Option<Pin<Box<dyn Future<Output = Result<Connection, crate::Error>> + Send + 'static>>>,
-    i: usize,
+pub(crate) struct DynamicServiceStream<K: Hash + Eq + Clone + Unpin> {
+    changes: Box<dyn Stream<Item = Change<K, Endpoint>> + Unpin + Send + 'static>,
```
LucioFranco, line 14: if we make the change above as proposed then this can just be the concrete `mpsc::Reciever` type which should simplify this impl!

tonic/src/transport/channel/mod.rs
dawid-nowak: done

tonic/src/transport/channel/mod.rs
dawid-nowak: i will leave as is for time being, not sure how to reolve it just yet

tonic/src/transport/service/discover.rs
dawid-nowak: done

examples/Cargo.toml
```diff
@@ -42,6 +42,14 @@ path = "src/load_balance/client.rs"
 name = "load-balance-server"
 path = "src/load_balance/server.rs"
 
+[[bin]]
+name = "load-balance-client-discovery"
```
LucioFranco, line 46: What do you think about naming this `dynamic-load-balance`? Might be a bit more clear

tonic/src/transport/channel/mod.rs
```diff
@@ -9,24 +9,26 @@ pub use endpoint::Endpoint;
...
-use super::service::{Connection, ServiceList};
+use super::service::{Connection, DynamicServiceStream};
 use crate::{body::BoxBody, client::GrpcService};
 use bytes::Bytes;
 use http::{
     uri::{InvalidUri, Uri},
     Request, Response,
 };
 use hyper::client::connect::Connection as HyperConnection;
+use std::hash::Hash;
```
LucioFranco, line 20: minor nit: can we move this import into the one below?

tonic/src/transport/channel/mod.rs
```diff
@@ -104,17 +106,26 @@ impl Channel {
...
-        let list = list.collect::<Vec<_>>();
-
-        let buffer_size = list
-            .iter()
-            .next()
-            .and_then(|e| e.buffer_size)
-            .unwrap_or(DEFAULT_BUFFER_SIZE);
+        let (channel, mut tx) = Self::balance_channel(DEFAULT_BUFFER_SIZE);
+        list.for_each(|endpoint| {
+            let _res = tx.try_send(Change::Insert(endpoint.uri.clone(), endpoint));
```
LucioFranco, line 111: _probably_ should just unwrap here if there is an error we want to know because that would mean there is a bug in this code!

tonic/src/transport/channel/mod.rs
```diff
@@ -104,17 +106,26 @@ impl Channel {
...
+        channel
+    }
 
-        Self::balance(discover, buffer_size)
+    /// Balance a list of [`Endpoint`]'s.
+    ///
+    /// This creates a [`Channel`] that will listen to a stream of change events and will add or remove provided endpoints.
+    pub fn balance_channel<K>(
+        capacity: usize,
+    ) -> (Self, tokio::sync::mpsc::Sender<Change<K, Endpoint>>)
```
LucioFranco, line 122: might want to just import this type, its making this signature a bit long :smile:

tonic/src/transport/service/discover.rs
```diff
@@ -1,34 +1,34 @@
 use super::super::service;
 use super::connection::Connection;
 use crate::transport::Endpoint;
+use std::hash::Hash;
```
LucioFranco, line 4: same here

tonic/src/transport/service/discover.rs
```diff
@@ -1,34 +1,34 @@
...
+use tokio::stream::Stream;
 use tower::discover::{Change, Discover};
 
-pub(crate) struct ServiceList {
-    list: VecDeque<Endpoint>,
-    connecting:
-        Option<Pin<Box<dyn Future<Output = Result<Connection, crate::Error>> + Send + 'static>>>,
-    i: usize,
+pub(crate) struct DynamicServiceStream<K: Hash + Eq + Clone> {
+    changes: tokio::sync::mpsc::Receiver<Change<K, Endpoint>>,
```
LucioFranco, line 14: same here can we import this type?

examples/src/load_balance_with_discovery/client.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use std::sync::Arc;
+
+use std::sync::atomic::{AtomicBool, Ordering::SeqCst};
+use tokio::time::timeout;
+use tower::discover::Change;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let e1 = Endpoint::from_static("http://[::1]:50051").timeout(std::time::Duration::from_secs(1));
```
LucioFranco, line 18: why the timeout here?

examples/Cargo.toml
dawid-nowak: done

tonic/src/transport/channel/mod.rs
dawid-nowak: done

tonic/src/transport/channel/mod.rs
dawid-nowak: done

tonic/src/transport/channel/mod.rs
dawid-nowak: done

tonic/src/transport/service/discover.rs
dawid-nowak: done

tonic/src/transport/service/discover.rs
dawid-nowak: done

examples/src/load_balance_with_discovery/client.rs
dawid-nowak: just me playing around, removed

dawid-nowak:
:metal:

PR #345, opened 2020-05-07
fix: base64 encode details header

### LINKED ISSUE #342: Panic when returning Status with arbitrary Bytes in details
### Version
0.2.0

### Platform
Mac OS Catalina

### Description
I was trying to return a Status with details from an unary RPC call looking like this
```
Err(Status::with_details(
  Code::ResourceExhausted,
  "Too many requests",
  Bytes::from_static(b"hello"),
))
```
It panicked with an error:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Status { code: Internal, message: "Couldn\'t serialize non-text grpc status header" }', /Users/<usr>/.cargo/registry/src/github.com-1ecc6299db9ec823/tonic-0.2.0/src/server/grpc.rs:225:9
```

Upon discussing with @LucioFranco in Discord:
- This is probably due to [this validation](https://docs.rs/http/0.2.1/src/http/header/value.rs.html#559)
- Since the bytes are going to a `*-bin` header, they should probably be base64-serialized according to [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).

LucioFranco:
Closes #342

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,5 +14,6 @@ members = [
     "tests/wellknown",
     "tests/extern_path/uuid",
     "tests/ambiguous_methods",
-    "tests/extern_path/my_application"
+    "tests/extern_path/my_application",
+    "tests/integration_tests"
 ]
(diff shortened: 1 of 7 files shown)


dvshur:
Could you try it with `Bytes::from_static(&[1])`?

LucioFranco:
@dvshur should be fixed now.

PR #347, opened 2020-05-07
fix(build): Make generated server service public

### LINKED ISSUE #344: Why doc(hidden) on impl Service?
It seems that service is generated with `doc(hidden)` attribute (introduced in https://github.com/hyperium/tonic/commit/cd9133a41cc6bedf23f64eb44b2b2f9bf93925e4):

https://github.com/hyperium/tonic/blob/68d2ab3339f1cc8bf4cf90c6c9fb37dbae251425/tonic-build/src/server.rs#L34-L37

This type is intended for direct using, e.g. in helloworld:

https://github.com/hyperium/tonic/blob/68d2ab3339f1cc8bf4cf90c6c9fb37dbae251425/examples/src/helloworld/server.rs#L36

It seems that @carllerche also had this question: https://github.com/hyperium/tonic/issues/1#issuecomment-533300029, but it left without answer.

l4l:
resolves #344

### DIFF
--- a/tonic-build/src/server.rs
+++ b/tonic-build/src/server.rs
@@ -31,7 +31,6 @@ pub fn generate<T: Service>(service: &T, proto_path: &str) -> TokenStream {
 
             #service_doc
             #[derive(Debug)]
-            #[doc(hidden)]
             pub struct #server_service<T: #server_trait> {
                 inner: _Inner<T>,
             }


Michael-F-Bryan:
Thanks for fixing this @l4l. Do you know when the next release will be coming out?

LucioFranco:
@Michael-F-Bryan I can get one out this week.

Issue #343, opened 2020-05-05
Custom metadata on error responses


dvshur:
### Motivation
Currently it seems only possible to return some custom info with a non-zero gRPC status in one predefined header:
`const GRPC_STATUS_DETAILS_HEADER: &str = "grpc-status-details-bin";`

```
#[derive(Clone)]
pub struct Status {
    /// The gRPC status code, found in the `grpc-status` header.
    code: Code,
    /// A relevant error message, found in the `grpc-message` header.
    message: String,
    /// Binary opaque details, found in the `grpc-status-details-bin` header.
    details: Bytes,
}
```

What if I want to use a different header name? What if I want to return more than one header/trailer on error response? gRPC spec does support that possibility. Libraries in other languages that I've worked with (Go, Scala) also don't have that restriction. We actually use a different header name in production in our company, so it's a quite real use case.

### Proposal
Min: allow `details` from `Status` to go to a header with a custom name.
Max: allow any number of custom metadata key/values, ASCII or binary (via `*-bin` headers), on non-zero gRPC statuses.

LucioFranco:
I'd like to expose a similar ability to get access to a metadata map like https://docs.rs/tonic/0.2.0/tonic/struct.Response.html#method.metadata

This way you can add your own custom metadata values.

avantgardnerio:
Sorry to comment on an old issue, but it appears that isn't possible because the only way to create metadata is through `Response`:

https://github.com/hyperium/tonic/blob/9ea03c2c44ad6bf76e4141e6fb40870b9b30ab63/tonic/src/response.rs#L26

and that calls a `Metadata` constructor with `0` capacity:

https://github.com/hyperium/tonic/blob/9ea03c2c44ad6bf76e4141e6fb40870b9b30ab63/tonic/src/metadata/map.rs#L227

Resulting in:

```
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0'
```

At runtime. Is there a way at present to achieve this? A quick googling didn't reveal any currently open issues around this behavior. I can open one if that's the correct next step.

LucioFranco:
@avantgardnerio if you're still seeing this, lets open another issue and really what would be helpful would be to recreate this in a test in map.rs (there are already a bunch of tests). Thanks!

avantgardnerio:
I am still seeing it, and I will PR a test, thank you for your prompt response!

avantgardnerio:
[...]

PR #348, opened 2020-05-10
feat: Add metadata to error responses

dvshur:
## Motivation
Issue #343 as well as some Discord discussions around May 5th—7th 2020 contain motivation for the PR. 

## Solution
- Added `metadata` field to the `Status` struct, like in the `Response` struct
    - If the provided metadata contains any headers with names reserved either by the gRPC spec or by `Status` fields (e.g. `grpc-status-details-bin`), they will be ignored.
- Updated existing `Status` constructors and conversions to support `metadata`
- Added `Status` constructors with `metadata` only or also with `details`
- Updated `Debug` and `Display` implementations for `Status` to support the new field
- Added an integration test for the functionality
- This change should be backward compatible

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -521,10 +521,10 @@ impl fmt::Display for Status {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "grpc-status: {:?}, grpc-message: {:?}, grpc-status-details-bin: {:?}, metadata: {:?}",
+            "status: {:?}, message: {:?}, details: {:?}, metadata: {:?}",
             self.code(),
             self.message(),
-            base64::encode_config(self.details(), base64::STANDARD_NO_PAD),
+            self.details(),
             self.metadata(),
         )
     }
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -481,6 +481,10 @@ impl fmt::Debug for Status {
             builder.field("details", &self.details);
         }
 
+        if !self.metadata.is_empty() {
+            builder.field("metadata", &self.metadata);
+        }
+
         builder.finish()
     }
 }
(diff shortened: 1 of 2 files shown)


tonic/src/status.rs
```diff
@@ -470,9 +521,11 @@ impl fmt::Display for Status {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "grpc-status: {:?}, grpc-message: {:?}",
+            "grpc-status: {:?}, grpc-message: {:?}, grpc-status-details-bin: {:?}, metadata: {:?}",
             self.code(),
-            self.message()
+            self.message(),
+            base64::encode_config(self.details(), base64::STANDARD_NO_PAD),
```
LucioFranco, line 527: I don't think we need to base64 here since from the user's pov they don't really care how it is encoded on the wire but only how they passed it in.

dvshur: My motivation for opting for base 64 here was this: in your previous message you used header names, not struct field names. E.g. `grpc-status: {:?}`, `grpc-message: {:?}`. This is what will be sent to the wire, not what user provided to the struct itself.

If this message has to do more with user perspective, then maybe we should use struct field names instead of header names everywhere? Like `status: {:?}, message: {:?}, details: {:?}, metadata: {:?}"`?

dvshur:
Are you good with the constrictor naming? `with_details_and_metadata`? It's a little bit long, but explicit :)

tonic/src/status.rs
LucioFranco: Yeah, that seems like a totally fine thing to do! I think I like that!

LucioFranco:
@dvshur yeah, I think for now it is fine, I suspect in the future we might refactor but its a code path that is not hit often enough to really require good ergonomics.

tonic/src/status.rs
dvshur: Done

PR #354, opened 2020-05-16
examples: update to `tracing` 0.1.16, use `#[instrument]`

hawkw:
Now that `tracing-attributes`'s `#[instrument]` macro plays nicely with
`async-trait`, we can update the tracing example to use `instrument`.
This lets us simplify the events emitted in the example.

Signed-off-by: Eliza Weisman <eliza@buoyant.io>


LucioFranco:
@hawkw oops conflict!

Issue #75, opened 2019-10-15
Support deadlines with the grpc-timeout header


thomasetter:
The [official protocol definition](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) mentions the `"grpc-timeout"` header which is a way for the client to tell the server what deadline it has.
As I understand it, the timeout should be converted into a deadline as soon as the header is read (before the rest of the payload).

Using this, when the server is overloaded, requests can just be dropped with `DEADLINE_EXCEEDED` or `CANCELLED` (not sure which one should be picked on the server side) if it takes too long until they are processed.
E.g. in C++, There is the [`ServerContext::IsCancelled()`](https://grpc.github.io/grpc/cpp/classgrpc__impl_1_1_server_context.html#ac0fe2a9425b18b724f910adbe0e3c17a) function which can be used to exit early: [Blog post on gRPC deadlines](https://grpc.io/blog/deadlines/)

How can this be achieved with tonic?
Are there features in tower/hyper which could be used?

LucioFranco:
Hi Yes, this should be possible by implementing a custom service layer that reads the initial headers and wraps its inner future with a timeout.

ParkMyCar:
I think this would be pretty interesting to build, if no one else is already working on it, I can try to put up a PR in a bit?

Aside from timeouts, there's also [cancelation](https://grpc.io/docs/guides/concepts/#rpc-termination). Is there a story around this yet in tonic?

LucioFranco:
I think the right method would be to create a version of https://github.com/tower-rs/tower/tree/master/tower/src/timeout that gets the timeout from a header rather than a static value.

Maher4Ever:
@LucioFranco Hi, is there an update regarding cancellation support in `tonic`?

LucioFranco:
No updates from my end yet, its easy to support cancellation by dropping the task but we currently do not parse the headers for a timeout automatically.

Sushisource:
Huge +1 here. In particular I need support for setting them on the client. Adding them manually appears to not work, since it looks like all the `GRPC_RESERVED_HEADERS` are silently stripped (which isn't so fun) as mentioned in #516. If there's anything I can do do help @LucioFranco I am happy to do so.

liufuyang:
I feel I need to set deadline or timeout on client as well, otherwise my Rust client cannot talk to our Java gRPC servers as they are build to require having deadline specified.

PR #362, opened 2020-05-26
Support deadlines with "grpc-timeout" header

ParkMyCar:
## Motivation
This PR fixes issue #75, which is to implement support for the `"grpc-timeout"` header in tonic. Setting a timeout/deadline is an official part of the gRPC spec, and is a [recommended](https://grpc.io/blog/deadlines/) practice.

#### Note
This code needs to be cleaned up a bit before merging, e.g. adding a `set_duration(...)` function to the generated client code, instead of making `client::Grpc.inner` public (and more). I put this PR up in this state to get some early feedback on the design of the timeout implementation.

### Server
Requests are timed out by attempting to read a `"grpc-timeout"` value from the `MetadataMap` of a request. If we find a value for a timeout, then we wrap inner service call with a `tokio::time::Timeout`, using the duration we read previously.

### Client
Taking inspiration from other [gRPC clients](https://grpc.io/blog/deadlines/#c) we add a function to `client::Grpc` called `set_duration(deadline: Duration)` which sets a duration value on the inner struct. When any request is made with this client, we insert this duration value in the `"grpc-timeout"` header.

## Testing
To make sure this works, and to prevent regressions, I added a `cancelation_on_timeout` test which sets a deadline on the client of 500ms, but the handler on the server pauses for 1,000ms before returning a value. We assert this request returns with the expected message and code.

#### Note
As part of the cleanup I'd also like to add a benchmark test to assert the timely cleanup of assets on timeout. e.g. in the current test, our request should only take 10s (ideally 1s) of milliseconds longer than the deadline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ target/
 **/*.rs.bk
 Cargo.lock
 tags
+.DS_Store
--- a/tests/integration_tests/proto/test.proto
+++ b/tests/integration_tests/proto/test.proto
@@ -7,4 +7,6 @@ service Test {
 }
 
 message Input {}
-message Output {}
+message Output {
+  string message = 1;
+}
--- a/tests/integration_tests/tests/status.rs
+++ b/tests/integration_tests/tests/status.rs
@@ -160,7 +160,6 @@ async fn cancelation_on_timeout() {
         .unary_call(Request::new(Input {}))
         .await
         .unwrap_err();
-    dbg!(&err);
 
     assert_eq!(err.message(), "request timed out!");
     assert_eq!(err.code(), Code::Cancelled);
--- a/tests/integration_tests/tests/status.rs
+++ b/tests/integration_tests/tests/status.rs
@@ -117,3 +117,55 @@ async fn status_with_metadata() {
 
     jh.await.unwrap();
 }
+
+#[tokio::test]
+async fn cancelation_on_timeout() {
+    struct Svc;
+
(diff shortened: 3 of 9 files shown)


evoxmusic:
Well done @ParkMyCar for this PR. It was a missing feature and luckily you made it 🙏

LucioFranco:
Hey @ParkMyCar I think this is a fine approach but I'd like to suggest a different one. The goal with the main `client/grpc.rs` and `server/grpc.rs` modules was to not have it 100% tied to tokio. This is to allow other runtimes to use it. By bringing in the timeout mechanics we unfortunately bind users to tokio. The solution I see is to instead of implementing this at the root level of the core grpc mechanisms support this at the transport level. From here we can implement a tower service similar to the current `Timeout` one but instead pulls the timeout from a header rather than a static value (or maybe both!). What do you think?

ParkMyCar:
@LucioFranco, totally makes sense, thanks for the feedback!
I’m on vacation right now, but within the next week or so I’ll update the branch with an implementation at the Transport level

ParkMyCar:
Just a heads up, I still intend on updating this PR, just haven't had time in the past couple weeks. If someone else is interested they should feel free 🙂

LucioFranco:
@ParkMyCar thanks for the update! No rush :)

ParkMyCar:
Opened a new PR #516, because I previously deleted the fork used to create this one and couldn't figure out how to update this one with a new fork, sorry if this causes unnecessary spam!

PR #368, opened 2020-06-07
chore(docs): fix incorrect proto path in helloworld tutorial

nicklaw5:
This change reverts #366. The `proto/` directory, as described in the guide, is created under the root of the crate. Not under `src/`.

I've also included a fix to the `grpcurl` endpoint.


zancas:
When will this merge?

LucioFranco:
sorry for the delay, been on vacation :)

Issue #370, opened 2020-06-15
allow get uri from transport::Endpoint


Sherlock-Holo:
### Crates

tonic

### Motivation

I am using `tonic` and stored the `Endpoint`, ensure I can reconnect GRPC server directly, however sometimes I need to get the uri in `Endpoint` to print some log.

### Proposal

add a new method `get_uri(&self) -> &Uri` for `Endpoint`, choose `&Uri` not `Uri` because it can reduce clone if the user doesn't want the ownership.

an implement is 
```rust
fn get_uri(&self) -> &Uri {
    &self.uri
}
```

### Alternatives

I have tried to create wrapper type `Endpoint(Uri, Endpoint)` to store the uri, but it needs a clone of the uri, adding the new method can reduce uselessly memory allocate

LucioFranco:
I believe the debug/display implementation should show the uri but it should be possible to expose that. I'd accept a PR for this :)

Sherlock-Holo:
`Endpoint` only implementd `Debug` but it doesn't show the uri, I will try to create a PR to add this feature :)

PR #371, opened 2020-06-16
feat: add new method get_uri for Endpoint

Sherlock-Holo:
This PR adds a new method `get_uri` for `Endpoint`, to allow user to get the `&Uri` from an `Endpoint`

## Motivation

I am using `tonic` and stored the `Endpoint`, ensure I can reconnect GRPC server directly, however sometimes I need to get the uri in `Endpoint` to print some log.

One way is to create a wrapper `struct Endpoint(Uri, tonic::transport::Endpoint)`, but it will allocate a new `Uri`.

I think the better way is let `Endpoint` can return the `Uri` reference which is in the `Endpoint`, user can use it to print logs or do something else, without unnecessary allocating.

related issue is #370

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -226,6 +226,19 @@ impl Endpoint {
 
         Channel::connect(connector, self.clone()).await
     }
+
+    /// Get the endpoint uri.
+    ///
+    /// ```
+    /// # use tonic::transport::Endpoint;
+    /// # use http::Uri;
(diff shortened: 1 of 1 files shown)


Sherlock-Holo:
the failed test makes me confused, I don't modify related codes.

Is there anyone who could tell me how to solve this problem? :(

LucioFranco:
The error seems like a race condition with the OS letting go of the previous socket to allow the new one to bind. I reran the test to see if that fixes it.

Sherlock-Holo:
the test is passed, thanks for your help :)

PR #374, opened 2020-06-18
Use direction in error reporting during decode

snawaz:
It is not a bug-fix or a new feature as such. Just being more explicit and loud in reporting error using the communication direction. 

## Motivation

We encountered this error message today:

```rust
(Status { code: Internal, message: "Unexpected compression flag: 123" }
```
It's not immediately obvious as to whether this error-scenario was encountered while _sending_ the request or while _receiving_ the response. I think adding that information in the message would be highly helpful to figure out the actual source causing the error. 

## Solution

It's pretty simple. I used `self.direction` to detect the error-scenario (as in, sending or receiving?) and created an error message accordingly.

Please let me know if this little improvement is acceptable in this form, or I missed out anything important. I'd like to make changes as required.

### DIFF
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -168,10 +168,15 @@ impl<T> Streaming<T> {
                 }
                 f => {
                     trace!("unexpected compression flag");
-                    return Err(Status::new(
-                        Code::Internal,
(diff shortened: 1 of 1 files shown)


snawaz:
`CI / cargo-deny check `FAILED for no obvious reason. I think re-run should work? I don't see any re-run option though.

PR #375, opened 2020-06-18
Add tower server example

belak:
Note that this also adds `impl From<Status> for http::Response<BoxBody>`
in order to make writing middleware which can return an error much
easier.

## Motivation

This is based off the example from https://gist.github.com/LucioFranco/9d5cab419797f54fc7b96bc409d8c6e8. It has been expanded into something a little more complex which requires specifying the Request/Response types but allows for intercepting metadata values with an async context. This allows lookups to happen to verify things like auth tokens.

## Solution

Most of the code here is the example, but it also replaces `Grpc::map_status` with an `impl From<Status> for http::Response<tonic::body::BoxBody>` in order to make it easier to write middleware which can return an error.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -67,6 +67,10 @@ name = "tls-client-auth-client"
 path = "src/tls_client_auth/client.rs"
 
 [[bin]]
+name = "tower-server"
+path = "src/tower/server.rs"
+
+[[bin]]
 name = "multiplex-server"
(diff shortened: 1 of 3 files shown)


belak:
> Just for now, can we back out the status change and the auth header check and just leave a comment. I'd like to tackle the status change in another PR and think about it a bit more.

Done. I've also opened #376 for discussion around adding some method to convert from Status to an http::Response.

PR #382, opened 2020-06-26
Allow empty packages - fixes #381

### LINKED ISSUE #381: If prost allows empty packages, tonic generates wrong endpoints
### Version

0.2.1

### Platform

Linux 5.5.11-200.fc31.x86_64 #1 SMP Mon Mar 23 17:32:43 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
### Crates

tonic-build

### Description

I have a fork of prost at https://github.com/vivint-smarthome/prost/tree/prost-no-package that allows for packages to not be specified because we really want to use tonic/prost but we have pre-defined protos without a package and adding a package is a breaking change for grpc. Tonic-build generates endpoints that begin with `/.<ServiceName>/<MethodName>` which is incorrect. The correct output should be `/<ServiceName>/<MethodName>`.

I'll submit a pull request to fix this.

tylerhawkes:
## Motivation

If prost allows for no packages then tonic generates bad endpoints for both clients and servers. See https://github.com/hyperium/tonic/issues/381.

## Solution

The solution is to make the `.` between the package name and service name dependent on whether or not the package is empty.

### DIFF
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -92,8 +92,9 @@ fn generate_methods<T: Service>(service: &T, proto_path: &str) -> TokenStream {
 
     for method in service.methods() {
         let path = format!(
-            "/{}.{}/{}",
+            "/{}{}{}/{}",
             service.package(),
+            if service.package().is_empty() {""} else {"."},
             service.identifier(),
             method.identifier()
         );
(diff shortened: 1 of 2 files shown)


LucioFranco:
Looks like you just need to run rustfmt and we can merge this thanks!

tylerhawkes:
Should be good to go now.

PR #385, opened 2020-07-02
Propagate errors in tls_config instead of unwrap/panic

repi:
## Motivation

We ran into `tls_connector` failing and causing our app to panic and shutdown as it seems there wasn't any way to avoid panicking in `tls_config`.

## Solution

So after talking to @LucioFranco briefly `tls_config` now returns a `Result` instead and propagates errors to the caller, where they have to be handled. 

This is a bit of an intrusive change to the API here. could alternatively introduce an additional `tls_config` function that does the error handling and keep the existing unwrapping one, to allow it to become an additive opt-in minor semver change instead?

### DIFF
--- a/examples/src/gcp/client.rs
+++ b/examples/src/gcp/client.rs
@@ -32,7 +32,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .domain_name("pubsub.googleapis.com");
 
     let channel = Channel::from_static(ENDPOINT)
-        .tls_config(tls_config)
+        .tls_config(tls_config)?
         .connect()
         .await?;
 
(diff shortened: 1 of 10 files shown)


LucioFranco:
This seems :+1: to me @jen20 thoughts?

LucioFranco:
@repi looks like a CI failure https://github.com/hyperium/tonic/pull/385/checks?check_run_id=831660623#step:7:139

jen20:
Sorry, I didn't see this until now - this looks great to me.

fujiapple852:
Note that Cargo's [modified SemVer logic](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) treats 0.2.2 as a compatible update from 0.2.1 and thus this release of tonic causes a break.  Trivial to fix but perhaps yanking and releasing as 0.3.0 would be better?

LucioFranco:
@fujiapple852 you're right....._sigh_ will do that today.

LucioFranco:
Getting a `0.3.0` release out right now. @fujiapple852 thanks for pointing this out!

PR #392, opened 2020-07-09
Add Endpoint::connect_lazy method

### LINKED ISSUE #167: Construct a Client that will lazily establish a connection
## Feature Request

Construct a Client that will lazily establish a connection

### Motivation

Sometimes it would be useful to be able to construct a client and defer establishing a connection with a server until you make a request.

### Proposal

Today when creating a client the connection with the server must be established before a client struct is returned from the `::connect()` function. Perhaps another constructor could be added which lazily connects and defers actually establishing the connection until a request is made.

Its expected (at least to me) that a client already needs to know and understand how to reconnect to a server in the event of a broken connection so this shouldn't be much more difficult than handling that case.

cc: @LucioFranco

danburkert:
## Motivation

Adds a 'lazy' constructor which will not attempt to connect to the
endpoint until first use. This is useful in situations where the
connection is created when the remote service may temporarily not be
responding, such as server startup.

Fixes #167

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -214,6 +214,25 @@ impl Endpoint {
         Channel::connect(connector, self.clone()).await
     }
 
+    /// Create a channel from this config.
+    ///
+    /// The channel returned by this method does not attempt to connect to the endpoint until first
+    /// use.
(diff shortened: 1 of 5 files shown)


danburkert:
I've tested this locally in my application, and seems to work well.

PR #397, opened 2020-07-11
Bump packages to latest versions

belak:
## Motivation

I've been debugging some issues with tls-roots not working properly on macOS, so I figured I'd bump packages to the latest versions just in case something was fixed there.

## Solution

I used cargo-outdated to check for the latest versions and updated them.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -43,11 +43,11 @@ prost = ["prost1", "prost-derive"]
 
 [dependencies]
 bytes = "0.5"
-futures-core = { version = "0.3", default-features = false } 
+futures-core = { version = "0.3", default-features = false }
 futures-util = { version = "0.3", default-features = false }
 tracing = "0.1"
 http = "0.2"
-base64 = "0.11"
+base64 = "0.12"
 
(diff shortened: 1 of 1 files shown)


jen20:
Hi @belak! Thanks for opening a pull request - this looks good to me and has no issue with any of our checks, I'll go ahead and merge it. Did it resolve your issue with TLS roots on macOS?

belak:
Unfortunately no... it also only seems to be on some computers as well, so I'm not super sure what to try next. Thanks for the quick response!

One thing I have noticed is that this does change portions of the `tonic::transport::ClientTlsConfig` API (as tokio-rustls was updated). Is that an issue with version compatibility, or is that fine?

belak:
Well, I've narrowed it down to an old self-signed cert I had on this laptop, causing some rustls-native-certs to fail with an error `transport error: BadDER`.

There's an example here which allows usage of partial results returned from rustls-native-certs, but I'm not sure if that should be the default or now. https://github.com/ctz/hyper-rustls/commit/d8a025de8d5bc9a3396857f8d02a22ef63ea30b6

Issue #410, opened 2020-07-23
Allow for easier routing between http server and multiple grpc servers 


xd009642:
I have two gRPC services in my server, my main server and the health check. I'm also adding a HTTP endpoint via warp. So I tried to do this and found things didn't quite work. Assuming I'm not doing anything egregiously wrong, there might be some changes to tonic-health and #272 or an equivalent may be required.

I've created an example repository here: https://github.com/xd009642/tonic-route-warp-example 

In the mean time I'll start hosting two servers listening at different ports but it would be nice if this would work :smile:

LucioFranco:
Yeah, that PR is related we just need to find a way to return an easy to use service from the router.

xd009642:
is there anything stopping #272 being merged? I saw there was one comment adding something but otherwise (at least to my untrained) eyes it looked pretty solid

LucioFranco:
@xd009642 we just need to address https://github.com/hyperium/tonic/pull/272#issuecomment-612143535

alce:
I believe this was resolved on https://github.com/hyperium/tonic/pull/419. Please re-open if there are further questions.

PR #419, opened 2020-07-30
Support routing between HTTP server and multiple gRPC services

tjtelan:
## Motivation
Coincidentally I was going to open an issue for this after spending some time on my own trying to accomplish this, but #410 was on the first page.

I wanted to help merge #272 since it appeared to be slightly stalled. I have a need for the functionality of running an HTTP server and a gRPC server with multiple services on the same port. 

## Solution
The major work appeared to be complete and was appeared to be an acceptable implementation. The finishing code was actually provided in [#272 (comment)](https://github.com/hyperium/tonic/pull/272#issuecomment-612143535) so I just barely provided work. It was a successful group effort overall.

I rebased the PR branch w/ respect to the original author @cthulhua (with a minor change the import of `futures_util::Future::Either`) and moved the `service` function into the Router impl as per the original feedback.

## Testing
Additionally, I added a new example `hyper_warp_multiplex`, a mix of the `hyper_warp` and `multiplex` examples.

- `cargo run --bin hyper-warp-multiplex-server`
- `cargo run --bin hyper-warp-multiplex-client`

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -25,7 +25,7 @@ use super::service::{Or, Routes, ServerIo, ServiceBuilderExt};
 use crate::{body::BoxBody, request::ConnectionInfo};
 use futures_core::Stream;
 use futures_util::{
-    future::{self, MapErr},
+    future::{self, Either, MapErr},
     TryFutureExt,
 };
 use http::{HeaderMap, Request, Response};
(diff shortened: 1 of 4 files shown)


cthulhua:
Thanks for taking this over; I’ve had a wild few months and haven’t been able to close this out

olanod:
I also ran into the need of having the gRPC server alongside a warp HTTP server on the same port(might even need websockets?) and wanted to know how feasible it is and if it would require work only on tonic's side.

tonic/src/transport/server/mod.rs
```diff
@@ -78,6 +78,38 @@ pub struct Router<A, B> {
     routes: Routes<A, B, Request<Body>>,
 }
 
+#[derive(Debug)]
+/// A service made from a router
```
LucioFranco, line 82:
```suggestion
/// A service that is produced from a Tonic `Router`.
///
/// This service implementation will route between multiple Tonic
/// gRPC endpoints and can be consumed with the rest of the `tower`
/// ecosystem.
#[derive(Debug)]
```

tonic/src/transport/server/mod.rs
```diff
@@ -476,6 +508,11 @@ where
             .serve_with_shutdown(self.routes, incoming, Some(signal))
             .await
     }
+
+    /// Create a tower service out of a router
```
LucioFranco, line 512:
```suggestion
    /// Create a tower service out of a router.
```

PR #434, opened 2020-08-19
Add content-type for Unimplemented

entropitor:
## Motivation
According to the spec content-type should be set

## Solution
Add the application/grpc for `Unimplemented`

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -592,6 +592,7 @@ impl Service<Request<Body>> for Unimplemented {
             http::Response::builder()
                 .status(200)
                 .header("grpc-status", "12")
+                .header("content-type", "application/grpc+proto")
                 .body(BoxBody::empty())
                 .unwrap(),
         )


tonic/src/transport/server/mod.rs
```diff
@@ -592,6 +592,7 @@ impl Service<Request<Body>> for Unimplemented {
             http::Response::builder()
                 .status(200)
                 .header("grpc-status", "12")
+                .header("content-type", "application/grpc+proto")
```
LucioFranco, line 595:
```suggestion
                .header("content-type", "application/grpc")
```

The spec is actually incorrect :( so we need to remove that part, otherwise this is great!

entropitor:
Updated!

edrevo:
Do we have any tests that cover this? I am compiling against the latest commit in master (40e64b5) and I'm getting a `HTTP status code 200; transport: missing content-type field` error from grpc-go when an unimplemented method is being invoked.

Here's a snippet of my Cargo.toml:

```toml
tonic = { git = "https://github.com/hyperium/tonic/" }
```

And the corresponding snippet from Cargo.lock:

```
[[package]]
name = "tonic"
version = "0.3.1"
source = "git+https://github.com/hyperium/tonic/#40e64b53a74bab4da5eb8a13a0ab95b872b8ac98"
dependencies = [
 "async-stream",
 "async-trait",
 "base64",
 "bytes",
 "futures-core",
 "futures-util",
 "http",
 "http-body",
 "hyper",
 "percent-encoding",
 "pin-project",
 "prost",
 "prost-derive",
 "tokio",
 "tokio-util",
 "tower",
 "tower-balance",
 "tower-load",
 "tower-make",
 "tower-service",
 "tracing",
 "tracing-futures",
]
```

LucioFranco:
@edrevo I don't see anything wrong with this PR but we don't have tests for this specific one. Do you by chance have an output of the h2 frames between that and grpc-go, it would be good to see what is actually happening on the wire.

edrevo:
Is there any easy way of getting the actual frames? I've set the log level to trace, but the actual frame content doesn't seem to be logged: https://gist.github.com/edrevo/853fe19a3593cfd724fa779332310c0a

edrevo:
Found it! https://github.com/hyperium/tonic/pull/441

PR #435, opened 2020-08-20
tonic-health: define explicit tonic-build features

ernestas-poskus:
Explicitly defining tonic-health, tonic-build features, note that rustfmt got removed. The reason for removal is an unnecessary build dependency in musl nightly environments.


LucioFranco:
I'm not following why rustfmt is an extra dependency? If so we could also likely just bake the files directly into the crate instead of building it each time.

ernestas-poskus:
prebuilding sounds even better :+1:

PR #438, opened 2020-08-29
fix(transport) Do not panic when building and Endpoint with an invali…

alce:
Fixes https://github.com/hyperium/tonic/issues/437


LucioFranco:
This is a breaking change and we will need to release it as so.

PR #440, opened 2020-08-29
chore: Update interop go binaries

alce:
Update go interop binaries and add a simple script to update them once in a while.

closes https://github.com/hyperium/tonic/issues/415


interop/update_binaries.sh
```diff
@@ -0,0 +1,33 @@
+set -e
```
LucioFranco, line 1: Bless you for this script 🙏 I was really not looking forward to updating these haha

PR #441, opened 2020-08-30
fix(build): Add content-type for generated unimplemented service

edrevo:
## Motivation

According to the spec content-type should be set. Related to https://github.com/hyperium/tonic/pull/434

## Solution

Add the application/grpc content-type header.


wllenyj:
@LucioFranco When will a new version be released with this patch. i can only use cargo patch to github master to solve this problem now.

Thanks a lot.

LucioFranco:
New version is hopefully planned for this week, we got a new version of tower out last week which was a blocker.

PR #449, opened 2020-09-05
fix(transport): Fix wrong doc for domain

linxiulei:
## Motivation

Both my test and function 'tls_connector' indicated 'domain_name' works for using `rustls_client_config` or not


alce:
Thanks!

PR #451, opened 2020-09-07
BREAKING CHANGE: consume endpoint while connecting

### LINKED ISSUE #450: Endpoint should be able to accept String instead of &'static str
`&'static` str isn't programatically useful, we always generate Url from some places like Config file then pass it to tonic. It's not convinient to use &'static str

GopherJ:
## Motivation

Fix #450, makes it `connect` method consume endpoint to avoid `lifetime issue`, it makes this piece of code being able to compile

https://user-images.githubusercontent.com/33961674/92327716-f321f300-f05b-11ea-8c42-4e2fb6234fcf.png

## Solution

change `&self` to `self`

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -203,7 +203,7 @@ impl Endpoint {
     }
 
     /// Create a channel from this config.
-    pub async fn connect(&self) -> Result<Channel, Error> {
+    pub async fn connect(self) -> Result<Channel, Error> {
         let mut http = hyper::client::connect::HttpConnector::new();
         http.enforce_http(false);
         http.set_nodelay(self.tcp_nodelay);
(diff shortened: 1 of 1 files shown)


alce (commented):
It is not necessary to clone self anymore

GopherJ:
@alce Hi, my purpose is to be able to keep the channel around and generate a grpc client while necessary. I'm actually not sure if we need to clone and keep `&self` because I don't need to store `Endpoint`, I need `Channel` though

alce:
Clients are cheap to clone. Maybe instead of trying to hold on to the channel future and connect whenever you need a client just clone the client itself?

GopherJ:
Hi, I'm cloning Channel because it's even cheaper than Client and I can achieve at the same time multiplexing

I keep a shared channel and everytime when I need to call a service, I clone then the channel and build a client to do the call.  The health of the channel is checked every 10seconds

Am I doing the right thing?

alce:
Yeah it makes sense to do it that way if you have multiple clients and want to share the channel. The problem is, I think, the interaction between the actor implementation you are using and how tonic is implemented.

I do not know how you can make them play well together but I don't think the solution is to change tonic's Endpoint implementation as you propose here, but let's see what @LucioFranco thinks.

LucioFranco:
These borrow for a reaosn, you likely want to wrap the async connect call in your own `async move {}` block to allow it to be `'static`.

Feel free to ping me on discord if the above solution doesn't work for you but I am going to close this PR since we don't want it to move like what @alce said. Thanks for the PR though :)

PR #457, opened 2020-09-18
feat(transport): add user-agent header to client requests.

### LINKED ISSUE #453: Support gRPC channel options
I also work with gRPC in Go and Python, and both have a way to use the channel options to do a number of things, in my specific case I'd like to set the `user-agent` header, which cannot be done via metadata.

In Python it looks like this:

```
connect = 'localhost:50041'
user_agent = "monkey 1.0.0"
options = (
    ('grpc.primary_user_agent', user_agent),
)
channel = grpc.insecure_channel(connect, options=options)
```

In Go, it looks like this:
```
endpoint = 'localhost:50041'
user_agent := "monkey 1.0.0"
var options []grpc.DialOption
options = append(options, grpc.WithInsecure())
options = append(options, grpc.WithUserAgent(user_agent)
grpc.Dial(endpoint, options)
```

I cannot find a way to do this with Tonic - am I missing this?  Also, I unfortunately don't know anything about Hyper or Tower, so I don't know where to start implementing it.

alce:
This PR adds a  user-agent header to client requests. The default value is `tonic/x.x.x`. The value can be configured through the `Channel` builder.

```rust
Endpoint::from_static("...").user_agent("Greeter 1.1")?.connect();
```

The value passed to the `user_agent` method will be prepended to the default, in this case the full value will be: `Greeter 1.1 tonic/0.3.1`.

The header is added by a new `UserAgent` layer in the client's service stack. While this is not the most straight forward implementation, it seemed to be the cleanest and safest one. It also aligns better with other client implementations (grpc-go clients, in particular).

**Alternatives**
* Remove `user-agent` from the list of grpc reserved headers.
* Do not filter out the `user-agent` header in tonic <-> http request conversions.
* Sniff the request after passing it to the interceptor fn to see if user-agent was set, pluck it out and add it manually after mapping the request.
* Move the optional header  from the `Endpoint` to the `client::Grpc` struct and set the header after converting the request.
* Don't set user-agent headers at all.

...and maybe others. 
  
fixes #453

### DIFF
--- a/tests/integration_tests/tests/user_agent.rs
+++ b/tests/integration_tests/tests/user_agent.rs
@@ -2,8 +2,7 @@ use futures_util::FutureExt;
 use integration_tests::pb::{test_client, test_server, Input, Output};
 use std::time::Duration;
 use tokio::sync::oneshot;
-use tonic::transport::Endpoint;
-use tonic::{transport::Server, Request, Response, Status};
+use tonic::{transport::Endpoint, transport::Server, Request, Response, Status};
 
 #[tokio::test]
 async fn writes_user_agent_header() {
--- a/tests/integration_tests/tests/user_agent.rs
+++ b/tests/integration_tests/tests/user_agent.rs
@@ -1,0 +1,53 @@
+use futures_util::FutureExt;
+use integration_tests::pb::{test_client, test_server, Input, Output};
+use std::time::Duration;
+use tokio::sync::oneshot;
+use tonic::transport::Endpoint;
+use tonic::{transport::Server, Request, Response, Status};
+
+#[tokio::test]
+async fn writes_user_agent_header() {
+    struct Svc;
+
+    #[tonic::async_trait]
+    impl test_server::Test for Svc {
+        async fn unary_call(&self, req: Request<Input>) -> Result<Response<Output>, Status> {
+            match req.metadata().get("user-agent") {
+                Some(_) => Ok(Response::new(Output {})),
+                None => Err(Status::internal("user-agent header is missing")),
+            }
+        }
+    }
+
+    let svc = test_server::TestServer::new(Svc);
+
+    let (tx, rx) = oneshot::channel::<()>();
+
+    let jh = tokio::spawn(async move {
+        Server::builder()
+            .add_service(svc)
+            .serve_with_shutdown("127.0.0.1:1322".parse().unwrap(), rx.map(drop))
+            .await
+            .unwrap();
+    });
+
+    tokio::time::delay_for(Duration::from_millis(100)).await;
+
+    let channel = Endpoint::from_static("http://127.0.0.1:1322")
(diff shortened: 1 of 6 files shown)


tonic/src/transport/service/user_agent.rs
```diff
@@ -0,0 +1,47 @@
...
+    type Error = T::Error;
+    type Future = T::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, mut req: Request<ReqBody>) -> Self::Future {
+        let value = match self.user_agent.as_ref() {
+            Some(custom_ua) => {
```
LucioFranco, line 33: can we compute this at `fn new` time then just clone the headervalue everytime, instead of reallocating on each request?

LucioFranco (changes requested):

tonic/src/transport/service/user_agent.rs
alce: Good idea.

tonic/src/transport/channel/endpoint.rs
```diff
@@ -7,6 +7,7 @@ use crate::transport::service::TlsConnector;
 use crate::transport::Error;
 use bytes::Bytes;
 use http::uri::{InvalidUri, Uri};
+use http::HeaderValue;
```
LucioFranco, line 10: Can we combine these two imports since they come from the same crate?

LucioFranco (approved):
LGTM, just one style nit, should we consider adding tests for this?

tonic/src/transport/channel/endpoint.rs
alce: Missed this one. I'll push the change and merge.

alce:
Sure, I'll add a test.

tests/integration_tests/tests/user_agent.rs
```diff
@@ -0,0 +1,53 @@
+use futures_util::FutureExt;
+use integration_tests::pb::{test_client, test_server, Input, Output};
+use std::time::Duration;
+use tokio::sync::oneshot;
+use tonic::transport::Endpoint;
+use tonic::{transport::Server, Request, Response, Status};
```
LucioFranco, line 6: These can be merged

LucioFranco (changes requested):

alce:
ha, weird merging of imports, hang on...

LucioFranco (approved):

alertedsnake:
This is awesome, @alce - thanks for working on it!

alce:
@alertedsnake thanks! And thank you for reporting the issue.

alertedsnake:
Just confirmed this is working perfectly in my real-world situation, so thanks again :)

PR #458, opened 2020-09-19
fix(transport): reconnect lazy connections after first failure

### LINKED ISSUE #452: A connection opened with `connect_lazy` does not attempt to reconnect if the server is unavailable on first request
### Version

```
tonic = 0.3.1
tonic-build = 0.3.1
```

### Platform

Linux 4.9.104-microsoft-standard (WSL2), Linux 5.8.6-NixOS (NixOS)

### Crates

Likely `tonic` itself.

### Description

If a connection is created with `connect_lazy`, a first gRPC call is made, and the server is not available (e.g. not running or rejecting incoming connections), subsequent gRPC calls will not attempt to reconnect to the server, and will instead return an error.

Code and reproducible setup (Docker) here: https://github.com/K900/tonic-issue-repro

Following up from a Discord conversation with @LucioFranco.

alce:
Channels created with lazy connections never try to reconnect if the first connection attempt fails. This is because `Reconnect` returns `Poll::Ready(Err)` on `poll_ready` and the service is considered dead.

This change passes a flag to Reconnect to signal if the connection is intended to be lazy, in which case reconnect returns the error on the next call. A new `Connection::lazy` constructor is provided to avoid threading booleans around.

fixes #452

### DIFF
--- a/tests/integration_tests/tests/connection.rs
+++ b/tests/integration_tests/tests/connection.rs
@@ -3,7 +3,22 @@ use integration_tests::pb::{test_client::TestClient, test_server, Input, Output}
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tokio::sync::oneshot;
-use tonic::{transport::Server, Request, Response, Status};
+use tonic::{
+    transport::{Endpoint, Server},
+    Request, Response, Status,
+};
+
+struct Svc(Arc<Mutex<Option<oneshot::Sender<()>>>>);
+
+#[tonic::async_trait]
+impl test_server::Test for Svc {
+    async fn unary_call(&self, _: Request<Input>) -> Result<Response<Output>, Status> {
+        let mut l = self.0.lock().unwrap();
+        l.take().unwrap().send(()).unwrap();
+
+        Ok(Response::new(Output {}))
+    }
+}
 
 #[tokio::test]
 async fn connect_returns_err() {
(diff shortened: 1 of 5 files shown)


tests/integration_tests/tests/connection.rs
```diff
@@ -53,3 +56,38 @@ async fn connect_returns_err_via_call_after_connected() {
...
+
+    let channel = Endpoint::from_static("http://127.0.0.1:1339")
+        .connect_lazy()
+        .unwrap();
+
+    let mut client = TestClient::new(channel);
+
+    // First call should fail, the server is not running
+    let res = client.unary_call(Request::new(Input {})).await;
+    assert!(res.is_err());
```
LucioFranco, line 74: we should use unwrap_err here too

tonic/src/transport/service/reconnect.rs
```diff
@@ -89,12 +91,12 @@ where
 
                             state = State::Idle;
 
-                            if self.has_been_connected {
-                                self.error = Some(e.into());
-                                break;
-                            } else {
+                            if !self.has_been_connected && !self.is_lazy {
```
LucioFranco, line 94:
```suggestion
                            if !(self.has_been_connected || self.is_lazy) {
```

applying https://en.wikipedia.org/wiki/De_Morgan%27s_laws

tonic/src/transport/service/reconnect.rs
```diff
@@ -89,12 +91,12 @@ where
...
                                 return Poll::Ready(Err(e.into()));
                             }
+
+                            self.error = Some(e.into());
```
LucioFranco, line 98: should we put this in the else branch?

LucioFranco (approved):

dfreese:
It looks like this change, in combination with #392 leaves the `Result<Channel, Error>` return type, but no longer is able to fail.  Would there be interest in just having this function return `Channel` instead?

LucioFranco:
@dfreese yeah, I think we need to revisit some of the APIs but likely won't change anytime soon since its a breaking change and I would like to limit that until we have a proper design for larger changes.

Issue #403, opened 2020-07-16
0.3.0 Client::connect always returns Ok even if the sever URL does not exist.


DominicWrege:
### Platform 
tonic-client-bug v0.1.0 (/tmp/tonic-0.3.0-client-bug)
└── tonic v0.3.0
└── tonic-build v0.2.0

### Crates
```
tonic = { version = "0.3.0", features = ["tls"] }
prost = "0.6.1"
tokio = { version = "0.2.21", features = ["full"] }
futures = "0.3.5"
async-stream = "0.2.1"
```

### Description

I excepted that the client returns an ``Err`` if no connection to the sever could establish.
**Example Repository**: https://github.com/DominicWrege/tonic-0.3.0-client-bug

LucioFranco:
I wonder if it is related to https://github.com/hyperium/tonic/pull/392, will have to take a look later this week. Thanks for opening the issue :)

DominicWrege:
Thanks for your response. So the basically `connect` is the new `lazy_connect` ? Or did I misunderstood something?
I my opinion it should be very explicated if  the connection should be established lazy or not.
My problem is that I want to check if the server is available by simply trying if a can get a connection.

LucioFranco:
@DominicWrege no I mean't that that change may have introduced unexpected behavior.

Rulliam:
I experienced the same behavior, thank you for taking the time to fix this !

LucioFranco:
@DominicWrege this should fix it https://github.com/hyperium/tonic/pull/413, I wrote a small description in the PR for why it was happening.

DominicWrege:
thx

Rulliam:
Thank you ! works like a charm !

faern:
Will there be a `0.3.1` (or similar) release with this fix soon? We are currently holding off upgrading from `0.2` due to this issue.

LucioFranco:
@faern should be out in the next 10min https://github.com/hyperium/tonic/pull/436

LucioFranco:
Ok, `0.3.1` is out.

faern:
Awesome! Thank you for this incredible response time!

LucioFranco:
We like our low p99s 😉

PR #460, opened 2020-09-22
Endpoint::lazy_connect always tries to reconnect

boxdot:
When a connection could not be established on first connection
attempt, an error was reported and the connection never tried to  reconnect.
This behaviour is kept for `Endpoint::connect` and changed for
`Endpoint::lazy_connect`.

## Motivation

The behaviour to fail on the first attempt was introduced in #413.
This makes tonic clients behave two-fold:

1. If a client fails at the first connection attempt, the connection is broken
and it will never try to reconnect.
2. If a client connected successfully at the first attempt, but later the connection
breaks, the client will resiliently try to reconnect.

As explained in #413 and #403, the behaviour in 1 is sometimes needed, however
there is another use case where the server we are connecting to is not yet up. 
When connecting lazily, we expect the connection to
be always resilient, since we move the actual connection attempt to a later
point in time.

## Solution

* Keep the error reporting behaviour in `Endpoint::connect`.
* Keep the resilient behaviour in `Endpoint::connect_lazy`.

## Note

I would appreciate a proposal if there is a better way to achieve resilient initial reconnect. Also, I am not very happy to propagate the `always_reconnect` flag through several layers, but do not see a better way to do it.

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -215,7 +215,8 @@ impl Endpoint {
         #[cfg(not(feature = "tls"))]
         let connector = service::connector(http);
 
-        Channel::connect(connector, self.clone()).await
+        let always_reconnect = false;
+        Channel::connect(connector, self.clone(), always_reconnect).await
     }
 
     /// Create a channel from this config.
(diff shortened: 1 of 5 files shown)


boxdot:
After opening the PR saw #458. Sorry for not checking. Feel free to close in case you prefer the other solution.

LucioFranco:
Thanks for opening this, likely to go with #458

Issue #462, opened 2020-09-24
make codec::decode::Streaming a trait for easier testability


stefanhoelzl:
## Feature Request
make [codec::decode::Streaming](https://github.com/hyperium/tonic/blob/master/tonic/src/codec/decode.rs#L21) a trait.
This will allow easier mocking needed for testing of grpc calls using client streaming.

### Crates
`tonic`
`tonic-build`

### Motivation
currently when I have a grpc call with client streaming like [ByteStream::Write](https://github.com/googleapis/googleapis/blob/master/google/bytestream/bytestream.proto#L77) I get a rust function with this signature:
```rust
async fn write(&self, request: Request<Streaming<WriteRequest>>) -> Result<Response<WriteResponse>, Status>
```
If I want to test this function I would need to create a `Streaming` object which is quite complicated.
It would be much easier if the generated function `write` would be generic accepting a argument implementing the trait `Stream` and `Streaming` would implement this.

### Proposal
Adding a new trait
```rust
trait Stream {
    type Item;
    async fn message<'_>(&'_ mut self) -> Result<Option<Self::Item>, Status>;
    async fn trailers<'_>(&'_ mut self) -> Result<Option<MetadataMap>, Status>;
}
```

and generate function definitions  for `ByteStream::write` with such a signature
```rust
async fn write<S: Stream<Item = WriteRequest >>(&self, request: Request<S>);
```

This would allow to create simple Mocks which implement `Stream` for testing.
And otherwise `Streaming` is still used, which must then implement `Stream`.

### Alternatives
Please let me know if you know alternative how testing can be simplified.

alce:
Not exactly what you are after but maybe you could delegate complex logic to a function similar to the one you  propose, and write most of your tests against that:

```rust
impl BytestreamService for Foo {
    async fn write(
        &self,
        request: Request<Streaming<WriteRequest>>,
    ) -> Result<Response<WriteResponse>, Status> {
        do_write(request).await // if you need the request
        do_write2(request.into_inner()).await // or just the stream
    }
}

async fn do_write<S>(request: Request<S>) -> Result<Response<WriteResponse>, Status>
where
    S: Stream<Item = Result<WriteResponse, Status>> + Unpin,
{
    //..
}

async fn do_write2<S>(stream: S) -> Result<Response<WriteResponse>, Status>
where
    S: Stream<Item = Result<WriteResponse, Status>> + Unpin,
{
    //..
}

```

[...]

PR #464, opened 2020-09-25
Stream trait

stefanhoelzl:
see also #462 

## Motivation
currently when I have a grpc call with client streaming like `ByteStream::Write` I get a rust function with this signature:

`async fn write(&self, request: Request<Streaming<WriteRequest>>) -> Result<Response<WriteResponse>, Status>`
If I want to test this function I would need to create a Streaming object which is quite complicated.
It would be much easier if the generated function write would be generic accepting a argument implementing the trait Stream and Streaming would implement this.

## Solution
Adding a new trait

```rust
trait Stream {
    type Item;
    async fn message<'_>(&'_ mut self) -> Result<Option<Self::Item>, Status>;
    async fn trailers<'_>(&'_ mut self) -> Result<Option<MetadataMap>, Status>;
}
```
and generate function definitions for `ByteStream::write` with such a signature

`async fn write<S: Stream<Item = WriteRequest >>(&self, request: Request<S>);`
This allows to create simple Mocks which implement `Stream` for testing.
And otherwise `Streaming` is still used, which must then implement `Stream`.


examples/src/routeguide/server.rs
```diff
@@ -105,31 +103,30 @@ impl RouteGuide for RouteGuideService {
...
     ) -> Result<Response<Self::RouteChatStream>, Status> {
         println!("RouteChat");
 
-        let mut notes = HashMap::new();
-        let mut stream = request.into_inner();
-
-        let output = async_stream::try_stream! {
-            while let Some(note) = stream.next().await {
-                let note = note?;
+        todo!();
```
stefanhoelzl, line 112: I need some help here on how to get this working!

interop/src/server.rs
```diff
@@ -114,27 +115,27 @@ impl pb::test_service_server::TestService for TestService {
...
-                        tokio::time::delay_for(Duration::from_micros(param.interval_us as u64)).await;
-
-                        let payload = crate::server_payload(param.size as usize);
-                        yield StreamingOutputCallResponse { payload: Some(payload) };
-                    }
-                }
-            };
-
-            Ok(Response::new(Box::pin(stream) as Self::FullDuplexCallStream))
+            todo!();
```
stefanhoelzl, line 118: I think this is the same than [here](https://github.com/hyperium/tonic/pull/464/files#diff-b6467f7b50c468db706f5c04e0f3f73aR112)

alce:
While I agree that changing some trait functions to be generic can be useful (and I believe it is possible),  I'm not sure the change would fly. However, if you want to continue working on this, here are some pointers that may help. These are not gospel and I may be wrong but you might want to look into them.

* The `async_trait` dependency is feature gated. It is only available with the `codegen` feature.
* Some code in examples,  tests, and most likely other people's code, rely on the fact that `Streaming` implements `futures::Stream`.
* The future returned by `MessageStream::message` is not the same as the future returned by the current `Streaming::message` method. They can't be use interchangeably as is.
* Similarly, `MessageStream::message` can't always be used instead of `Stream::next`.
* Probably the easiest way to achieve what you want, is to use `futures::Stream` as a bound to the generic methods parameters. The additional `MessageStream` wouldn't be needed. This should work without much fuss, other than changing codegen. But, this will make function signatures a lot more involved. In the end, even if you can pull off introducing the new async trait I _think_ function signatures would still be a bit of an eye sore. There is also the object safety issue. Although I don't think it matters, I don't know for sure.

Issue #384, opened 2020-06-30
Remove `Code::__NonExhaustive` variant


LucioFranco:
cc @seanmonstar, looks like there are no more status codes being added, I think we can do this?

LucioFranco:
One issue with this will be when decoding a status code that doesn't exist. In this case we most likely want to error out. We want to consider this case and probably don't want to funnel it into the same error case as getting a correct status back from another remote peer.

seanmonstar:
There's a guarantee that no new codes would be added? I'm suspicious of that, unless adding a new code would blow up too many implementations. An alternative is to change to `non_exhaustive`.

LucioFranco:
I don't think they will change, there is basically zero movement right now with the spec. Though if it does change I also don't think there is any problem in creating a breaking change for tonic.

I think the idea is that if we don't use `non_exhaustive`  we will be forcing the user to statically handle the case when a new code is added. I don't think that is that bad of a consequence. Not sure though.

davidpdrsn:
I agree with @LucioFranco. I've made a PR that removes it https://github.com/hyperium/tonic/pull/625.

PR #467, opened 2020-09-28
chore: clippy lints

alce:
Fixes some clippy lints. 

There are some remaining, most of which are in the `metadata` module. I will open a separate PR for those. The rest can be fixed once there is a decision on #384

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -505,11 +505,12 @@ impl Status {
         Status {
             code,
             message: message.into(),
-            details: details,
-            metadata: metadata,
+            details,
+            metadata,
         }
     }
 
+    #[allow(clippy::wrong_self_convention)]
     /// Build an `http::Response` from the given `Status`.
     pub fn to_http(self) -> http::Response<BoxBody> {
         let (mut parts, _body) = http::Response::new(()).into_parts();
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -48,6 +48,7 @@ use tracing_futures::{Instrument, Instrumented};
 
 type BoxService = tower::util::BoxService<Request<Body>, Response<BoxBody>, crate::Error>;
 type TraceInterceptor = Arc<dyn Fn(&HeaderMap) -> tracing::Span + Send + Sync + 'static>;
+type BoxFuture<T, E> = Pin<Box<dyn Future<Output = Result<T, E>> + Send + 'static>>;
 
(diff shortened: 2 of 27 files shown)


tonic/src/status.rs
```diff
@@ -505,11 +505,12 @@ impl Status {
...
             code,
             message: message.into(),
-            details: details,
-            metadata: metadata,
+            details,
+            metadata,
         }
     }
 
+    #[allow(clippy::wrong_self_convention)]
```
LucioFranco, line 513: wut....

tonic/src/transport/server/mod.rs
```diff
@@ -98,11 +99,13 @@ where
     B::Error: Into<crate::Error> + Send,
 {
     type Response = Response<BoxBody>;
+    type Error = crate::Error;
+
+    #[allow(clippy::type_complexity)]
```
LucioFranco, line 104: This is like so simple, clippy why 😭

tonic/src/transport/service/discover.rs
```diff
@@ -12,12 +12,12 @@ use tokio::{stream::Stream, sync::mpsc::Receiver};
 
 use tower::discover::{Change, Discover};
 
+type BoxFuture<T, E> = Pin<Box<dyn Future<Output = Result<T, E>> + Send + 'static>>;
```
LucioFranco, line 15: we could move these to some `common.rs` file? I think hyper does this

LucioFranco (approved):

tonic/src/status.rs
alce: yeah, apparently it wants the method to be called `into_http`

LucioFranco: Yeah, I guess its right actually...

tonic/src/transport/server/mod.rs
alce: I tried to split them up into simpler type aliases but it makes them harder to read imo. We could add a `clippy.toml` increasing the threshold for this lint a bit and get rid of the allow annotations. I think there are 3 or so.

tonic/src/transport/service/discover.rs
alce: yup

alce: Done. I defined `BoxFuture` inline in `transport` for now since it's the only one I could find we can extract. It is also not used anywhere else other than codegen, but I guess that one can stay independent.

LucioFranco (approved):

PR #468, opened 2020-09-29
chore: fix warnings in metadata module

alce:
Fixes clippy lint warnings in metadata module.

Note that this PR manually implements `Hash` for `MatadataValue`.  This was necessary because there is a manual `PartialEq` implementation. This is particularly important for binary metadata values.

### DIFF
--- a/tonic/src/metadata/value.rs
+++ b/tonic/src/metadata/value.rs
@@ -6,6 +6,7 @@ use super::key::MetadataKey;
 use bytes::Bytes;
 use http::header::HeaderValue;
 use std::error::Error;
+use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
 use std::str::FromStr;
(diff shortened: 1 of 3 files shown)


alce:
On a side note, I don't know if it would be worth it to review `MatadataMap`'s implementation, particularly around constructors. It carried over some of `http::HeaderMap`s  details, that were written before `TryFrom` and possibly other traits were stable.

tonic/src/metadata/value.rs
```diff
@@ -134,6 +135,8 @@ impl<VE: ValueEncoding> MetadataValue<VE> {
     /// For MetadataValue<Binary> the provided parameter must be base64
     /// encoded without padding bytes at the end.
     ///
+    /// # Safety
```
LucioFranco, line 138: nice catch!

LucioFranco (approved):

PR #472, opened 2020-10-05
fix(build): Match namespace code with other generated packages

tylerhawkes:
## Motivation
I submitted a pr earlier that made it so that when prost supports protobufs without a package spec that tonic will already support it. Because prost doesn't actually support the feature I can't see a way to test the functionality. I just tried spinning up a server and had an issue with the implementation of the `NamedService` trait in hyper where it wasn't respecting the package spec. This is a fix for that.

## Solution
If the package is empty then don't add a preceding `.` to the namespace.


LucioFranco (approved):
Thanks!

PR #475, opened 2020-10-15
fix(transport): overwrite user-agent header for client requests

aheadmode:
x

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -80,14 +80,13 @@ impl Endpoint {
 
     /// Set a custom user-agent header.
     ///
-    /// `user_agent` will be prepended to Tonic's default user-agent string (`tonic/x.x.x`).
     /// It must be a value that can be converted into a valid  `http::HeaderValue` or building
     /// the endpoint will fail.
     /// ```
     /// # use tonic::transport::Endpoint;
     /// # let mut builder = Endpoint::from_static("https://example.com");
-    /// builder.user_agent("Greeter").expect("Greeter should be a valid header value");
-    /// // user-agent: "Greeter tonic/x.x.x"
+    /// builder.user_agent("Greeter 1.1").expect("Greeter should be a valid header value");
+    /// // user-agent: "Greeter 1.1"
     /// ```
     pub fn user_agent<T>(self, user_agent: T) -> Result<Self, Error>
     where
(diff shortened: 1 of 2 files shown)


alce:
The current implementation follows the spec's recommendation closely. It is also in line with grpc core and all implementations on top of it.

In what context is this a security risk?


https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#user-agents
https://github.com/grpc/grpc/issues/2661

aheadmode:
The protocol does not require a user-agent to function. It is _recommended_ to set a user-agent string and to follow a certain format.

The specification does not mandate to append Tonic's default user-agent string to all client requests. 

As a user, I can to completely overwrite the user-agent header with a custom string in Java and Go (Golang)

- github.com/grpc/grpc-java: https://github.com/grpc/grpc-java/blob/8c4088a9e95f9bc9af5632da13be19f816808c35/core/src/main/java/io/grpc/internal/ClientTransportFactory.java#L113

- github.com/grpc/grpc-go: https://github.com/grpc/grpc-go/blob/506b7730668b5a13465224b0d8133f974a3f843d/internal/transport/transport.go#L543

As a user, I can to completely overwrite the user-agent header with a custom string in browsers that support the [JavaScript API for WebExtensions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/onBeforeSendHeaders#Examples).

Browser fingerprinting is the capability of a site to identify or re-identify a visiting user, user agent or device via configuration settings or other observable characteristics.

W3C: Mitigating Browser Fingerprinting in Web Specifications

> ...
> 
> Data sources that may be used for browser fingerprinting include:
> - ...
> - browser characteristics
> 

https://www.w3.org/TR/fingerprinting-guidance/#identifying-fingerprinting-surface-and-evaluating-severity

alce:
> As a user, I can to completely overwrite the user-agent header with a custom string in Java and Go (Golang)

No you can't. The user agent provided in these configuration options will be prepended to a default, same as in tonic. In fact, I _think_ the java implementation even includes the specific transport in the user agent string (netty or okhttp).

You can't use a tonic client in the browser right now but if you could, it should probably behave as the js client: the user agent string is left for the browser to set and the library sets an `x-user-agent` metadata entry, currently `grpc-web-javascript/0.1`.

As far as I understand, the decision for user agent headers to behave this way is mostly an organizational one, made by the grpc org.  We _could_ go a different way, I just don't see a sufficiently solid reason to do so.

Issue #282, opened 2020-03-05
Compression Support


ericmcbride:
## Feature Request
Add compression to Tonic
### Crates
https://github.com/BurntSushi/rust-snappy
### Motivation
We use gzip (default) in our Go GRPC services.  Would like to normalize this across all services.  There are other compression options such as snappy.  I noticed burntsushi has a nice snappy implementation in rust.

### Proposal
Add some level of compression for larger payloads.

Id be more then happy to help with the implementation of this.  Been wanting to get my hands dirty on some more open source projects.  I don't know if any ideas have been floated or not about this, or if it belongs here or in tower.

LucioFranco:
@ericmcbride yeah, compression is something we should 100% add. I am curious what type of compression you are using with grpc-go? Is it at the connection level or is it per message?

I'd love to support this though I am not 100% sure on how to do this yet, so happy to chat about it! Feel free to ping me on discord as well and we can chat if this is something you are interested in contributing :)

ericmcbride:
Yeah we just use gzip and enable it as an option. I haven't really went deep into the rabbit hole.  At first glance 
```
// callInfo contains all related configuration and information about an RPC.
type callInfo struct {
	compressorType        string
	failFast              bool
	stream                ClientStream
	maxReceiveMessageSize *int
	maxSendMessageSize    *int
	creds                 credentials.PerRPCCredentials
	contentSubtype        string
	codec                 baseCodec
	maxRetryRPCBufferSize int
}


// CallOption configures a Call before it starts or extracts information from
// a Call after it completes.
type CallOption interface {
	// before is called before the call is sent to any server.  If before
	// returns a non-nil error, the RPC fails with that error.
	before(*callInfo) error

	// after is called after the call has completed.  after cannot return an
	// error, so any failures should be reported via output parameters.
	after(*callInfo)
}

```
Makes me thing its per message but this is a quick assumption.  Ill dive deeper into the code this weekend and look and id be more then happy to jump on!

Sherlock-Holo:
I think if compression is in connection-level, need to check peer support compression or not, but it makes developers can focus on the real job, instead of 'how to set the framework correctly'.

And if we could choose which rpc call enable compression, which one disable, it's very good, because when we just send some little bytes, compress them doesn't reduce time, sometimes even make data bigger. However, if we need to send a lot of data, like sending a photo, compress it may reduce a lot of transport time

LucioFranco:
[...]

PR #476, opened 2020-10-15
Gzip decompression

vbfox:
## Motivation

When a tonic server we deployed receive grpc messages that are compressed (as gzip by a Go client) it currently fail as decompression isn't currently supported (#282)

I wanted to support this limited case as simply as possible.

## Solution

I took the solution of only implementing this very specific case because:
* gzip is the most common compression for grpc and the one that is available by default in the go and c++ implementations
* It's an hard-failure, if a tonic server client doesn't send compressed requests the other party will understand it. But when a compressed message is received the only way to handle it is to reject it if compression isn't implemented.
* It doesn't require any additional API surface as a single hardcoded compressor is supported and only enabled via feature flags.

On the code decisions:
* `Compressor` contains compression and decompression method and aims to be/become the equivalent of https://godoc.org/google.golang.org/grpc/encoding#Compressor it could be the trait that users could implement to provide additional compression methods.
* `Decompression` hosts the decompression information provided to 'decode' (A single http header for now)
* `Compression` is the same on the other direction but is only used for responses in cases where the request was compressed in all other cases no compression take place. *It also avoid a problem currently existing in envoy grpc-web support and mimics what grpc-go does*
* I looked at enabling some of the interop tests on compression but go-grpc doesn't implement them. I'm not sure of how this can easily be integration-tested before compression is there and it can be tested via roundtrip. The C++ tests support compression but they seem to test advanced aspects too that aren't implemented yet in tonic with this PR.

Tests done:
* Added benches with gzip
* Greeter and routeguide were tested both way against go, enabling gzip from the go side clients
* Deploying an internal project using tonic and accessing it with Go requiring compression and via grpc-web translated by envoy

---

Fixes https://github.com/hyperium/tonic/issues/282

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -139,7 +139,7 @@ name = "hyper-warp-multiplex-server"
 path = "src/hyper_warp_multiplex/server.rs"
 
 [dependencies]
-tonic = { path = "../tonic", features = ["tls"] }
+tonic = { path = "../tonic", features = ["tls", "gzip"] }
 prost = "0.6"
 tokio = { version = "0.2", features = ["rt-threaded", "time", "stream", "fs", "macros", "uds"] }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -36,6 +36,7 @@ transport = [
 tls = ["transport", "tokio-rustls"]
 tls-roots = ["tls", "rustls-native-certs"]
 prost = ["prost1", "prost-derive"]
+gzip = ["flate2"]
 
 # [[bench]]
 # name = "bench_main"
@@ -48,6 +49,8 @@ futures-util = { version = "0.3", default-features = false }
 tracing = "0.1"
 http = "0.2"
 base64 = "0.12"
+flate2 = { version = "1.0", optional = true }
+once_cell = "1.0"
 
 percent-encoding = "2.0"
 tower-service = "0.3"
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -92,4 +92,3 @@ rustdoc-args = ["--cfg", "docsrs"]
 [[bench]]
 name = "decode"
 harness = false
-
--- a/tonic/benches/decode.rs
+++ b/tonic/benches/decode.rs
@@ -19,7 +19,7 @@ macro_rules! bench {
                 .build()
                 .expect("runtime");
 
-            let payload = make_payload($message_size, $message_count, &$encoding);
+            let payload = make_payload($message_size, $message_count, $encoding);
             let body = MockBody::new(payload, $chunk_size);
             b.bytes = body.len() as u64;
 
@@ -27,7 +27,7 @@ macro_rules! bench {
                 rt.block_on(async {
                     let decoder = MockDecoder::new($message_size);
 
-                    let decompression = Decompression::new($encoding);
+                    let decompression = Decompression::from_encoding($encoding);
                     let mut stream = Streaming::new_request(decoder, body.clone(), decompression);
 
                     let mut count = 0;
@@ -113,7 +113,7 @@ impl Decoder for MockDecoder {
     }
 }
 
-fn make_payload(message_length: usize, message_count: usize, encoding: &Option<String>) -> Bytes {
+fn make_payload(message_length: usize, message_count: usize, encoding: Option<&str>) -> Bytes {
     let mut buf = BytesMut::new();
 
     let raw_msg = vec![97u8; message_length];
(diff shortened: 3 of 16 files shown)


alce:
Thank you for this. It looks like a good approach at first glance.

I am going to go over the implementation in more depth in the next couple of days, primarily to see if I can spot details that may cause trouble (as in major breaking changes, for example) with further enhancements or other planned features. A bit later on, @LucioFranco can help us out sorting out the details.

alce:
@vbfox I looked at this in a little more detail and I think it's neat. However, we can't proceed yet  because we need to figure out how to best to integrate this with channels and clients. I understand this PR is not concerned with that (and it's fine) but wee need to have a clear path forward. We may want to make some changes to the transport before integrating this. It may take a little bit though.

These are some thoughts I had while going through this. Nothing actionable yet, just things we may want to consider.

* I know Go and Java use interfaces but maybe we could use enums instead of traits. We  could provide identity, gzip and deflate variants under a  single`compression` feature, maybe even throw in brotli. We loose the ability to easily add algorithms but I wonder how useful that is in practice and it could make the implementation much simpler.

* If we go with trait objects, maybe we could have an explicit registry instead of populating the compressors map based on feature flags.

* I think we should keep request and response compression asymmetric and not follow Go's lead on this one.

* Tiny tip: you can cast a boolean to an u8 `compression_enabled as u8`

We'll update with specific steps forward as soon as possible. Thanks again.

davidpdrsn:
@vbfox Sorry for the slow progress on this. Is this still something you're interested in working on? If so I would like to help reviewing. If not I also wouldn't mind taking over the implementation.

I also asked a general question about how the spec requires you to implement compression [here](https://github.com/hyperium/tonic/issues/282#issuecomment-830626404). Maybe you're able to help.

vbfox:
@davidpdrsn I'm interested in gzip compression being available 😄 

I can continue working on it but if someone wants to take over no problem, the remaining issues were more about how the project wants to approach integrating Compression and i'm not implicated enough in tonic design for that I admit :)

davidpdrsn:
Awesome! I'm gonna help out with reviewing then.

I guess the first thing to do is get it up to date with the current `master`. It
seems your branch is still using tokio 0.2 whereas `master` has updated to tokio
1.

Here are my comments about what @alce said.

> * I know Go and Java use interfaces but maybe we could use enums instead of
>   traits.

I agree that going with traits to support custom algorithms probably isn't
necessary. I'm wondering if something like this might work:

```rust
/// Which encodings are enabled?
///
/// Used to construct the `grpc-accept-encoding` header and determine if the
/// `grpc-encoding` of an incoming messsage is supported.
pub(crate) struct MessageEncodingConfig {
    gzip: bool,
    // more encodings added here later
}

/// Which encoding should be used for a message
///
/// Corresponds to the `grpc-encoding` header.
pub(crate) MessageEncoding {
    Gzip,
    // more encodings added here later
}
```

Where `Endpoint` and `Server` somehow stores a `MessageEncodingConfig` and have
methods allowing the user to customize which encodings to support.

This is similar to how compression/decompression works in
[tower-http](https://github.com/tower-rs/tower-http).

We could also have methods to enable/disable the compression on a single
`Response` but stream items are problematic since they aren't wrapped in a type
that tonic controls. So we have nowhere to add new fields. I would be fine with
making compression global for a given `Channel` or `Server` and not support
customizing it for individual messages. I wonder if this violates the spec or
otherwise breaks some use-case :thinking:

> We could provide identity, gzip and deflate variants under a
> single`compression` feature, maybe even throw in brotli.

I think only supporting gzip out of the box is fine but if supporting more
algorithms isn't too much additional work that would also be fine by me.

> I think we should keep request and response compression asymmetric and not
> follow Go's lead on this one.

I'm not quite sure what code is referenced here or what the implications would
be.

davidpdrsn:
> My main question is, could an adapted version of this PR lands by itself (no API change and add support for the other party sending compressed messages) with another PR later adding sending compressed messages and configuration or do you expect both at the same time ?

We're currently thinking about doing a 0.5 release which will include a few breaking changes and [grpc-web support](https://github.com/hyperium/tonic/pull/455). I would really like to have this feature included in that release as its very useful. However since this isn't a breaking change and therefore not something we _have_ to include in 0.5 I would prefer if we aimed to make a single PR that added a compelling compression feature, meaning we would have support for both sending and receiving compressed messages. If we break things into multiple PRs we risk getting into a situation where we want to ship 0.5 but can't because compression is only half done.

I'm still up for taking over the implementation if you don't feel you have the bandwidth for it.
 
> On the `plugability/trait` part the main interest is not only for new algorithms but also to allow users to provide their own dependencies for example to configure what gzip implementation to use (pure rust ? zlib-ng ? https://github.com/rust-lang/flate2-rs#backends ) as it can have performance/deployment complexity tradeofs that tonic might not want to impose on it's users.

Personally I don't think allowing users to plug in their own compression algorithms is an important feature. For example its not something supported by reqwest or warp. I think that is fine as it simplifies the implementation quite a bit.

vbfox:
> I'm still up for taking over the implementation if you don't feel you have the bandwidth for it.

I'll try to find time this weekend to advance on it but If I can't, I think i'll let you take over.

> Personally I don't think allowing users to plug in their own compression algorithms is an important feature. For example its not something supported by reqwest or warp. I think that is fine as it simplifies the implementation quite a bit.

👍🏼

davidpdrsn:
@vbfox thanks for your work on this! I've been taken in https://github.com/hyperium/tonic/pull/692.

PR #482, opened 2020-10-24
transport: impl Service for Channel instead of GrpcService

### LINKED ISSUE #481: Impl tower::Service for tonic::transport::Channel
### Motivation

A type that implements `tower::Service` in the proper way implements `GrpcService`, but not the other way around.

[`Channel`](https://docs.rs/tonic/0.3.1/tonic/transport/channel/struct.Channel.html) is a special implementation of `GrpcService`. It can potentially implement `Service` as well (making such a wrapper type is easy), but currently it doesn't. Due to this, we can't make use of various `tower`'s wrapper services for a channel.

```rust
#[tokio::main]
    let conn = tonic::transport::Endpoint::new("http://[::1]:50051")?
        .connect()
        .await?;
    // Wrap a channel with Timeout
    let conn = Timeout::new(conn, Duration::from_millis(1000));
    let mut client = GreeterClient::new(conn);  // Error!
    let request = tonic::Request::new(HelloRequest {
        name: "Tonic".into(),
    });

    let response = client.say_hello(request).await?;

    println!("RESPONSE={:?}", response);

    Ok(())
}
```

```
error[E0277]: the trait bound `tonic::transport::Channel: tower::Service<http::Request<tonic::body::BoxBody>>` is not satisfied
  --> src/client.rs:40:41
   |
40 |     let mut client = GreeterClient::new(conn);
   |                                         ^^^^ the trait `tower::Service<http::Request<tonic::body::BoxBody>>` is not implemented for `tonic::transport::Channel`
   |
   = note: required because of the requirements on the impl of `tower::Service<http::Request<tonic::body::BoxBody>>` for `tower::tower_timeout::Timeout<tonic::transport::Channel>`
   = note: required because of the requirements on the impl of `tonic::client::GrpcService<tonic::body::BoxBody>` for `tower::tower_timeout::Timeout<tonic::transport::Channel>`

```

### Proposal

Let's implement `tower::Service<http::Request<BoxBody>>` for `Channel`, so that we can make use of `tower` service wrappers more easily.

### Alternatives

To achieve the same goal, users may write a small wrapper for `Channel` which implements `Service`.

```rust
struct ChannelWrapper(Channel);

impl Service<http::Request<BoxBody>> for ChannelWrapper {
    type Response = http::Response<Body>;
    type Error = tonic::transport::Error;
    type Future = tonic::transport::channel::ResponseFuture;

    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
        self.0.poll_ready(cx)
    }

    fn call(&mut self, req: http::Request<BoxBody>) -> Self::Future {
        let mut channel = self.0.clone();

        channel.call(req)
    }
}
```

```rust
#[tokio::main]
    let conn = tonic::transport::Endpoint::new("http://[::1]:50051")?
        .connect()
        .await?;
    let conn = ChannelWrapper(conn);
    let conn = Timeout::new(conn, Duration::from_millis(1000));
    let mut client = GreeterClient::new(conn);
...
```

We can find [a similar one in `examples`](https://github.com/hyperium/tonic/blob/master/examples/src/tower/client.rs). But providing a canonical `Service` implementation for `Channel` would simplify user code?

ajalab:
## Motivation

Fixes #481. Adds `Service` impl for `Channel`.

## Solution

This PR replaces `GrpcService` impl for `tonic::transport::Channel` with `Service` impl. The two trait implementations cannot co-exist due to name conflict, but the former impl will be still alive thanks to the [general trait implementation for `Service` implementors](https://github.com/hyperium/tonic/blob/90858926b6a788f78f9c7bbdd7dda127f7f41e2c/tonic/src/client/service.rs#L33-L51).

**Notes**:

After applying this change, _existing codes that both imports `GrpcService` and `Service` traits can be broken_ due to multiple method definitions. This can be easily fixed by removing an import whichever or calling the preferable method explicitly. In this crate, only `examples/src/tower/client.rs` will be affected, so I fixed that in the second commit.

Finally, I added `timeout` example, which makes use of the unlocked feature by this change: wraps a channel with `tower` service.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -75,6 +75,14 @@ name = "tower-client"
 path = "src/tower/client.rs"
 
 [[bin]]
+name = "timeout-server"
+path = "src/timeout/server.rs"
+
+[[bin]]
+name = "timeout-client"
+path = "src/timeout/client.rs"
+
+[[bin]]
 name = "multiplex-server"
 path = "src/multiplex/server.rs"
 
--- a/examples/src/timeout/client.rs
+++ b/examples/src/timeout/client.rs
@@ -1,0 +1,28 @@
+use hello_world::greeter_client::GreeterClient;
+use hello_world::HelloRequest;
+use std::time::Duration;
+use tower::timeout::Timeout;
+
+use tonic::transport::Channel;
+
+pub mod hello_world {
+    tonic::include_proto!("helloworld");
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let channel = Channel::from_static("http://[::1]:50051").connect().await?;
+    let timeout_channel = Timeout::new(channel, Duration::from_millis(1000));
+
+    let mut client = GreeterClient::new(timeout_channel);
+
+    let request = tonic::Request::new(HelloRequest {
+        name: "Tonic".into(),
+    });
+
+    let response = client.say_hello(request).await?;
+
+    println!("RESPONSE={:?}", response);
+
+    Ok(())
+}
--- a/examples/src/timeout/server.rs
+++ b/examples/src/timeout/server.rs
@@ -1,0 +1,45 @@
+use std::time::Duration;
+use tokio::time::delay_for;
+use tonic::{transport::Server, Request, Response, Status};
+
+use hello_world::greeter_server::{Greeter, GreeterServer};
+use hello_world::{HelloReply, HelloRequest};
+
+pub mod hello_world {
+    tonic::include_proto!("helloworld");
+}
+
+#[derive(Default)]
+pub struct MyGreeter {}
+
+#[tonic::async_trait]
+impl Greeter for MyGreeter {
+    async fn say_hello(
+        &self,
+        request: Request<HelloRequest>,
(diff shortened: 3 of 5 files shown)


LucioFranco:
This looks good 👍 Its just a breaking change, so I need to think about how we will do all our breaking changes for the next release.

ajalab:
Definitely. Thank you for taking the time to review!

danburkert:
Just found this issue after being stymied when trying to wrap `Channel` in a custom `Service` wrapper.  I found the `GrpcService` machinery to be very difficult to work with, would it be possible to just remove it and inline the necessary bounds everywhere it's currently used?  Trait aliases would be a better solution, but AFAIK they are not stable.

LucioFranco:
@danburkert Yeah, I think there is benefit to the type but I think I made a few mistakes in actually using it correctly. That said, I am planning a large refactor of the transport system and want to simplify tonic a bit. So I am seriously taking this all into consideration but just a bit busy right now with other things. I am planning on breaking out a chunk of time in the next month or two to work on this. Sorry for the delay all.

LucioFranco (approved):

PR #489, opened 2020-11-07
Update bytes to v0.6

olix0r:
Currently uses patch dependencies for http, http-body, & prost.


olix0r:
See https://github.com/hyperium/http/pull/441 https://github.com/hyperium/http-body/pull/30 https://github.com/danburkert/prost/pull/381

PR #495, opened 2020-11-19
fix(transport): return Poll::ready until error is consumed

tl-helge-hoff:
When a lazy connection fails to connect the first time it returns Poll::ready and saves the error in the Reconnect service.
Subsequent calls to `Reconnect::poll_ready` return `Poll::Pending`  which makes tower_balance loop forever as it will never get a ready service.
This is due to `Reconnect::poll_ready` being called twice by `tower_balance` and returning first `Poll::Ready` and then
`Poll::Pending`.

## Motivation
We bumped into the infinite-looping issue while using the `Channel::balance_channel` together with tls,
when our tls config was incorrect and triggering an error during the handshake.
## Solution
`Reconnect::poll_ready` will consistently return `Poll::Ready` until the error is consumed in `Reconnect::call`.
After the error is consumed, and the error is surfaced through a client call, the `Reconnect` service will try to connect again.
The PR also adds trace logging in the `Reconnect` service.

### DIFF
--- a/tonic/src/transport/service/reconnect.rs
+++ b/tonic/src/transport/service/reconnect.rs
@@ -1,6 +1,7 @@
 use crate::Error;
 use pin_project::pin_project;
 use std::fmt;
+use std::fmt::Debug;
 use std::{
     future::Future,
     pin::Pin,
@@ -13,6 +14,7 @@ use tracing::trace;
 pub(crate) struct Reconnect<M, Target>
 where
     M: Service<Target>,
+    M::Error: Debug,
 {
     mk_service: M,
     state: State<M::Future, M::Response>,
@@ -32,6 +34,7 @@ enum State<F, S> {
 impl<M, Target> Reconnect<M, Target>
 where
     M: Service<Target>,
+    M::Error: Debug,
 {
     pub(crate) fn new(mk_service: M, target: Target, is_lazy: bool) -> Self {
         Reconnect {
@@ -52,6 +55,7 @@ where
     M::Future: Unpin,
     Error: From<M::Error> + From<S::Error>,
     Target: Clone,
+    <M as tower_service::Service<Target>>::Error: std::fmt::Debug,
 {
     type Response = S::Response;
     type Error = Error;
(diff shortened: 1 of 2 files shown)


tonic/src/transport/service/reconnect.rs
```diff
@@ -1,6 +1,7 @@
 use crate::Error;
 use pin_project::pin_project;
 use std::fmt;
+use std::fmt::Debug;
```
LucioFranco, line 4: We already import fmt above so you can just use `fmt::Debug`

tonic/src/transport/service/reconnect.rs
```diff
@@ -13,6 +14,7 @@ use tracing::trace;
 pub(crate) struct Reconnect<M, Target>
 where
     M: Service<Target>,
+    M::Error: Debug,
```
LucioFranco, line 17: I think we actually want to use `Into<Box<dyn Error + Send + Sync>>` here and then convert it. If it impls that trait and you call `into()` then it will automatically impl `Debug` because of the `Error` trait.

LucioFranco (changes requested):
Sorry for the delay on this! I'd like to see that one trait bound change, then we can merge it.

tonic/src/transport/service/reconnect.rs
tl-helge-hoff: I think I've done what you've suggested in the last comment.

tonic/src/transport/service/reconnect.rs
```diff
@@ -52,6 +54,7 @@ where
     M::Future: Unpin,
     Error: From<M::Error> + From<S::Error>,
     Target: Clone,
+    <M as tower_service::Service<Target>>::Error: fmt::Debug,
```
LucioFranco, line 57: This still needs to be `Into<Error>` or else it might break the stack down the line.

LucioFranco:
Closing in favor of #536

LucioFranco:
Thanks @tl-helge-hoff for getting this started!

PR #497, opened 2020-11-20
Load balancing example and test

tl-helge-hoff:
This PR adds, as an example, the load balancing scenario that provoked bug [1](https://github.com/hyperium/tonic/pull/493), and [2](https://github.com/hyperium/tonic/pull/495), as well as tests that cover the those fixes.

The load balancing example tries to solve a client-side load balancing scenario in which the set of server ips
are not static, and the way of retrieving the current set of ips at any time is done through resolving a hostname that maps
all server ips.
In the example, a detached task tries to resolve a hostname at a fixed interval and if any changes to the set of server ips
is detected, it will be reflected on the gRPC client by notifying the discovery logic through the channel retrieved with `Channel::balance_channel`.
While developing this load balancing scheme, the two aforementioned bugs were found to which we added tests.

The tests are added as an integration test that leverages the code from our load balancing example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,5 +16,6 @@ members = [
     "tests/extern_path/uuid",
     "tests/ambiguous_methods",
     "tests/extern_path/my_application",
-    "tests/integration_tests"
+    "tests/integration_tests",
+    "tests/load_balancing"
 ]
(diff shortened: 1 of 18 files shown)


LucioFranco:
While this is really good, I feel like this may be too complex for an example to live within this repo? Usually I try to keep examples simple so they are easy to follow and not too intimidating.

PR #500, opened 2020-11-29
feat(tonic): implement From<io::Error> for Status

jpopesculian:
Implements the conversion from `std::io::Error` to `tonic::Status`.

## Motivation

The `io::Error` conversion is currently left as
unimplemented. It either should be implemented or removed if it won't be
implemented.

## Solution

Implements the conversion from `std::io::Error` to
`tonic::Status`


LucioFranco (approved):
Thanks!

PR #501, opened 2020-11-30
Don't format generated code in tonic-health

akhilles:
## Motivation

Fix docs.rs build: https://docs.rs/crate/tonic-health/0.2.0/builds/274199

## Solution

Don't call `rustfmt`.


LucioFranco (approved):
Thanks!

PR #502, opened 2020-11-30
Remove unnecessary dependencies in tonic-health

akhilles:
## Motivation

fewer dependencies is better

## Solution

remove dependencies


edrevo (approved):

LucioFranco (approved):
Thanks!

PR #507, opened 2020-12-10
Update health.proto

smhmayboudi:
There is a new field. https://github.com/grpc/grpc/blob/master/doc/health-checking.md


LucioFranco (approved):
Thanks!

Issue #394, opened 2020-07-10
Enable mapping `h2` errors to correct `Status` codes.


LucioFranco:
We currently have this code https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L344 which maps `h2::Error` into proper `Code`'s. We should test and ensure that if the user enables the `transport` feature we can downcast the errors into `h2` variants and mapping them to correct status codes.

SkamDart:
Happy to take this over but after reading through the code, it looks as if maybe this has already been done with this [feature](https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L329-L334) and [test](https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L775-L791)?

Unless I'm misunderstanding this and you were looking to feature gate something like, `pub enum Http2Error` based on [this spec](https://github.com/grpc/grpc/blob/3977c30/doc/PROTOCOL-HTTP2.md#errors) under transport?

LucioFranco:
Last I remembered I don't think its hooked up would be good to test. We don't need to add a http2error we just need to ensure when we get an h2 error its translated into the correct status code.

davidpdrsn:
Is this related to https://github.com/hyperium/tonic/pull/612?

davidpdrsn:
It isn't. But it was fixed in https://github.com/hyperium/tonic/pull/509

PR #509, opened 2020-12-12
Enable `h2` errors

SkamDart:
## Motivation
Solves #394 

## Solution
Adds `h2` as a dependency and feature gate under `transport`


LucioFranco (approved):
Thanks!

PR #516, opened 2020-12-29
Support timeouts with "grpc-timeout" header

ParkMyCar:
## Motivation

This PR fixes issue #75, which is to implement support for the "grpc-timeout" header in tonic. Setting a timeout/deadline is an official part of the gRPC spec, and is a [recommended](https://grpc.io/blog/deadlines/) practice.

### Server

This PR creates a new `Timeout` struct that implements `tower::Service` and is added as a `layer_fn(...)` in a `Connection`'s `tower::ServiceBuilder`. The `Timeout` service then parses the `grpc-timeout` header, and wraps an inner future in a timeout.

I'm relying on `tokio::time::Delay` for the actual timeout logic (`Delay` seems to be 1:1 with `Sleep` in tokio 1.0, which should make migration easy). I realize this prevents the `Timeout` service from being runtime agnostic, but since the service lives within the `transport` module, which already depends on tokio, I thought this was okay. If we don't want to rely on `tokio::time::Delay`, we can hand write our own Delay-like future, with a downside being, we won't be able to use mock timers for testing.

A `tonic::transport::Server` already supports a default timeout, I handle the presence of a server timeout, and a `grpc-timeout` header, by using which ever duration is the shortest, essentially the server timeout becomes a ceiling. The reasoning here is I didn't want a client to be able to override a server's settings, but if a client wants to timeout faster, the server shouldn't care about that.

### Client

This PR doesn't (yet) make any changes to gRPC clients. It could, similar to what the [gRPC docs](https://grpc.io/blog/deadlines/#c) have as examples for other libraries, but I thought the server/client impls would be a good separation for different PRs.

### Note

I had previously opened #362 a few months ago which tries to add the same feature, supporting timeouts with the "grpc-timeout" header. I'm opening a new PR now because I had deleted my previous fork of tonic that was used to open that previous PR. I hope this is okay 🙂

### DIFF
--- a/tonic/src/metadata/map.rs
+++ b/tonic/src/metadata/map.rs
@@ -198,11 +198,10 @@ pub struct OccupiedEntry<'a, VE: ValueEncoding> {
 
 impl MetadataMap {
     // Headers reserved by the gRPC protocol.
-    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 8] = [
+    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 7] = [
         "te",
         "user-agent",
         "content-type",
-        "grpc-timeout",
         "grpc-message",
         "grpc-encoding",
         "grpc-message-type",
--- a/tonic/src/transport/service/timeout.rs
+++ b/tonic/src/transport/service/timeout.rs
@@ -1,0 +1,216 @@
+use http::Request;
+use std::{
+    task::{Context, Poll},
+    time::Duration,
+};
+use tower_service::Service;
+use tracing::{debug, warn};
+
+pub(crate) struct Timeout<S> {
+    inner: S,
+    server_timeout: Option<Duration>,
+}
+
+impl<S> Timeout<S> {
+    pub(crate) fn new(inner: S, server_timeout: Option<Duration>) -> Self {
+        Self {
+            inner,
+            server_timeout,
+        }
+    }
+}
+
+impl<S, ReqBody> Service<Request<ReqBody>> for Timeout<S>
+where
+    S: Service<Request<ReqBody>>,
+    S::Error: Into<crate::Error>,
+{
+    type Response = S::Response;
+    type Error = crate::Error;
+    type Future = future::TimeoutFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map_err(Into::into)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        // Try to parse the `grpc-timeout` header, if it is present
+        let header_timeout = headers::try_parse_grpc_timeout(req.headers()).unwrap_or_else(|e| {
+            warn!("Error parsing grpc-timeout header {:?}", e);
+            None
+        });
+
+        // Use the shorter of the two durations, if either are set
+        let timeout_duration = match (header_timeout, self.server_timeout) {
+            (None, None) => None,
+            (Some(dur), None) => Some(dur),
+            (None, Some(dur)) => Some(dur),
+            (Some(header), Some(server)) => {
+                let shorter_duration = std::cmp::min(header, server);
+                debug!(
+                    "both grpc-timeout header present: {:?},\
+                     and server timeout set: {:?}.\
+                     Using server timeout of: {:?}",
+                    header, server, shorter_duration,
+                );
+                Some(shorter_duration)
+            }
+        };
+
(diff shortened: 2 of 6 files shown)


tonic/src/metadata/map.rs
```diff
@@ -198,11 +198,10 @@ pub struct OccupiedEntry<'a, VE: ValueEncoding> {
 
 impl MetadataMap {
     // Headers reserved by the gRPC protocol.
-    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 8] = [
+    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 7] = [
         "te",
         "user-agent",
         "content-type",
-        "grpc-timeout",
```
ParkMyCar, line 205: removing this definitely isn't right, but otherwise the header was getting stripped when I tried to set it in the integration test. Would love some guidance on how to set this header from a test

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /// Tries to split the last character of the string, from the rest of the string,
+    /// returning (rest_of_string, last_char), if successful.
+    ///
+    /// `str.split_at(...)` panics if we're not on a UTF-8 code point boundary. This
+    /// should never happen in practice because the `grpc-timeout` header should be only
+    /// ASCII characters.
+    fn try_split_last(val: &str) -> Result<(&str, &str), &str> {
+        std::panic::catch_unwind(|| val.split_at(val.len() - 1)).map_err(|_| val)
```
ParkMyCar, line 124: I realized, after writing a unit test, that `HeaderValue` already prevents non-ASCII characters 🎉 so this extra `catch_unwind(|| ...)` probably isn't necessary, but will leave it in for now

Sushisource:
@LucioFranco Anything I could do to move along at least the client-side support for setting the timeout header? I'm planning on releasing a crate soon that requires it, and it would be great to not need to `[patch]` tonic.

I'm guessing you want to do something a little more formal than just removing it from reserved headers. If you let me know what that might be I would be happy to make a PR.

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,318 @@
...
+    type Future = future::TimeoutFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map_err(Into::into)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        // Try to parse the `grpc-timeout` header, if it is present
+        let header_timeout = headers::try_parse_grpc_timeout(req.headers()).unwrap_or_else(|e| {
+            warn!("Error parsing grpc-timeout header {:?}", e);
```
davidpdrsn, line 39: nit: not sure `warn` is the best fit here since clients might be sending garbage data. It isn't really tonics fault, modulo parsing bugs of course 😊 I kinda think we should use `trace` or drop the log entirely.

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,318 @@
...
+        });
+
+        // Use the shorter of the two durations, if either are set
+        let timeout_duration = match (header_timeout, self.server_timeout) {
+            (None, None) => None,
+            (Some(dur), None) => Some(dur),
+            (None, Some(dur)) => Some(dur),
+            (Some(header), Some(server)) => {
+                let shorter_duration = std::cmp::min(header, server);
+                debug!(
```
davidpdrsn, line 50: I'm thinking this should just be `trace`.

```suggestion
                trace!(
```

tonic/src/transport/service/timeout.rs
davidpdrsn: I guess `str.split_at(...)` should be fine since `HeaderValue::to_str` checks for ASCII as you mention.

davidpdrsn (commented):
@ParkMyCar Thanks for working on this! I've left a couple of comments but overall I think the approach looks good 👍 

Are you still up for driving this home? Getting this up to date with `master` might require some work as tonic has updated to tokio 1.x.

I do think having to create a new middleware not being able to use [`tower::timeout::Timeout`](https://docs.rs/tower/0.4.6/tower/timeout/struct.Timeout.html) is unfortunate but I don't see a nice way around it that doesn't involve creating new middlewares regardless.

davidpdrsn:
I'm going to be driving this home in a separate PR.

davidpdrsn:
Superseded by https://github.com/hyperium/tonic/pull/606

PR #520, opened 2021-01-01
Gracefully handle bad native certs

### LINKED ISSUE #519: fragile loading of native certs on MacOS with tls-roots
## Bug Report

When using the `tls-roots` feature, many MacOS users will run into [this issue](https://github.com/ctz/rustls-native-certs/issues/4). 

### Version

```
tonic v0.3.1
tonic-build v0.2.0
```

### Platform

Linux (Arch, btw)

### Crates

`tonic` (specifically [this line](https://github.com/hyperium/tonic/blob/master/tonic/src/transport/service/tls.rs#L61))

### Description

Failure to parse a cert can happen for a variety of reasons, some of which are outlined in that thread. But failure to parse a single cert from the MacOS Trust Record should not result in failing to load any cert at all if there are other trusted certs in the store.

 ### Prescription

`tonic` should use the best-effort strategy described in [this comment on the previous thread](https://github.com/ctz/rustls-native-certs/issues/4#issuecomment-647145085) and implemented in [`hyper-rustls`](https://github.com/ctz/hyper-rustls/blob/master/src/connector.rs#L34). Happy to make a PR for this if it would get merged!

NAlexPear:
## Description

Prevent problems reading individual certs from the cert store from interfering with the use of other valid certs when the `tls-roots` feature is enabled.

Fixes #519 

## Motivation

MacOS users will often have certs in their Trusted Settings Record that cannot be parsed by `rustls_native_roots` (used in conjunction with the `tls-roots` feature here). This change makes `tls-roots` usable for those users again.

## Solution

Similar to [this implementation in `hyper-rustls`](https://github.com/ctz/hyper-rustls/blob/master/src/connector.rs#L34), one-off errors are effectively ignored when loading the native cert store.

### DIFF
--- a/tonic/src/transport/service/tls.rs
+++ b/tonic/src/transport/service/tls.rs
@@ -58,7 +58,10 @@ impl TlsConnector {
 
         #[cfg(feature = "tls-roots")]
         {
-            config.root_store = rustls_native_certs::load_native_certs().map_err(|(_, e)| e)?;
+            config.root_store = match rustls_native_certs::load_native_certs() {
+                Ok(store) | Err((Some(store), _)) => store,
+                Err((None, error)) => Err(error)?,
+            };
         }
 
         if let Some(cert) = ca_cert {


LucioFranco (approved):
Thanks!

Issue #521, opened 2021-01-05
namespacing issues with services when not using prost-types for well-known protobuf types


tdyas:
### Version

`tonic` v0.3.1
`tonic-build` v0.3.1
`prost` at https://github.com/danburkert/prost/commit/a1cccbcee343e2c444e1cd2738c7fba2599fc391 (v0.6.x at the commit with `Bytes` support)

### Platform

macOS 10.15.7

### Crates

`tonic-build`

### Description

I am using Prost pinned to `master` at  https://github.com/danburkert/prost/commit/a1cccbcee343e2c444e1cd2738c7fba2599fc391 (which includes the use of `Bytes` for binary fields). 

I would like to have Prost compile the well-known types (instead of using `prost-types`) so that `Any` fields use `Bytes`. However, when I enable that mode via `config.compile_well_known_types()` in `build.rs`, generated grpc service stubs fail to compile because they are in a `_client` submodule and are missing an extra `super` reference necessary to get the identifier path out of that submodule.

For example, when compiling the [Google Operations API protos](https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto), errors like the following occur:

```
error[E0433]: failed to resolve: could not find `protobuf` in `super`
   --> XXX/target/debug/build/bazel_protos-4307470ac76a8035/out/google.longrunning.rs:183:40
    |
183 |     ) -> Result<tonic::Response<super::protobuf::Empty>, tonic::Status> {
    |                                        ^^^^^^^^ could not find `protobuf` in `super`

```

This code was within an `operations_client` submodule. The reference needs to be `super::super::protobuf::Empty` in this case, not `super::protobuf::Empty`. (The service in this example returns a `google.protobuf.Empty`.)

I assume the issue is the special casing of identifiers in `google.protobuf` performed in the service generator at https://github.com/hyperium/tonic/blob/3c2a4a28134515ac842a81ee14bfb49609b39d3c/tonic-build/src/prost.rs#L96 and https://github.com/hyperium/tonic/blob/3c2a4a28134515ac842a81ee14bfb49609b39d3c/tonic-build/src/prost.rs#L106. That special casing does not apply the `proto_path` to `google.protobuf` types but implicitly assumes that they will be found in `prost-types`, which is not the case if they are being compiled by Prost.

I am willing to contribute a fix, but would need some basic guidance on what the right fix is. I assume the fix be as simple as just taking the `else` branch when well-known types are being compiled?

tdyas:
[...]

PR #522, opened 2021-01-05
feat(build): Support compiling well-known protobuf types

tdyas:
## Motivation

Tonic does not currently handle the case when Prost is configured to compile well-known protobuf types instead of just referencing `prost_types`. See https://github.com/hyperium/tonic/issues/521 for a full description of the problem.

## Solution

Allow the user to configure Tonic to compile well-known protobuf types. That configuration is passed through to Prost.


tdyas:
Caveats:

- This still needs a proper test. I did not see code generation tests within the `tonic-build` crate. Are they somewhere else in the code base?
- Now that there are two options (`proto_path` and `compile_well_known_types`) as arguments everywhere, it would probably be useful to pass around some sort of a context struct instead.

tdyas:
Added a test of using `tonic-build` with compiling well-known types.

Note: I had to modify some of the well-known protos to add doctest `ignore` attributes on indented code blocks in the proto comments because Rust was treating them as Rust doctests. This is probably an issue that should be reported to Prost.

LucioFranco:
With prost releasing a new version supporting `Bytes` (not sure if `prost-types` got an update) can we just update `prost-types` instead of trying to side step it here? This seems like a pretty extreme change for something that should honestly be a pretty trivial update to prost?

tdyas:
> With prost releasing a new version supporting `Bytes` (not sure if `prost-types` got an update) can we just update `prost-types` instead of trying to side step it here? This seems like a pretty extreme change for something that should honestly be a pretty trivial update to prost?

No, because I do not believe prost-types will be compiled with the `Bytes` support because `Bytes` support is entirely optional, `Vec<u8>` is still the default. First, note that enabling Bytes is a non-default config option: https://docs.rs/prost-build/0.7.0/prost_build/struct.Config.html#method.bytes

Moreover, you can see the updated version of `google.protobuf.Any` in prost-types here: https://github.com/danburkert/prost/blob/79f0dfd8ed703fcce27f053c3b29c360c7ead5cb/prost-types/src/protobuf.rs#L1033

Note that the `value` field in `google.protobuf.Any` is still compiled as a `Vec<u8>`, not a `Bytes`.

Also, I submit that the fact that my ultimate purpose is to gain access to `Bytes` support is somewhat irrelevant to this PR. The goal of this PR is to allow users of tonic-build the option of compiling the well-known protobuf types with Prost, which is still a bug in tonic-build, regardless of my ultimate purpose.

LucioFranco:
Okay, that is fair, I need to think about this a bit more but I'd like to first ship a new version of tonic ASAP that works with tokio 1.0 then I plan on working on some enhancements. Since I don't consider this a major blocker to that I would like to look at this once we get that version out. Hopefully that version will go out this week. I really appreciate the time you took to look into this :)

tdyas:
> Since I don't consider this a major blocker to that I would like to look at this once we get that version out. 

I agree, this PR is not a major blocker.

tdyas:
Just wanted to provide a gentle nudge now that v0.4 is out. I'd love to get some feedback on this PR to keep it moving.

My main question is how to thread through the `compile_well_known_types` value. Should the generation methods take a context struct carrying `proto_path` and `compile_well_known_types` instead of those being passed as separate parameters?

tustvold:
FYI instead of copying the full proto definitions, it is sufficient to create a dummy proto file imports them

```
syntax = "proto3";

package google.protobuf;

import "google/protobuf/any.proto";
import "google/protobuf/api.proto";
import "google/protobuf/descriptor.proto";
import "google/protobuf/duration.proto";
import "google/protobuf/empty.proto";
import "google/protobuf/field_mask.proto";
import "google/protobuf/source_context.proto";
import "google/protobuf/struct.proto";
import "google/protobuf/timestamp.proto";
import "google/protobuf/type.proto";
import "google/protobuf/wrappers.proto";
```

tdyas:
I removed the vendored proto files as suggested, and then rebased (with some churn for some typos of mine).

Thoughts on whether a context structure makes sense? With three parameters now in some of the generation methods (e.g., `emit_package`, `proto_path`, and `compile_well_known_types`), that might be a good refactor to simplify the function signatures.

davidpdrsn (approved):
I'm not super familiar with the code but the changes look pretty simple. Sounds like a useful feature to have as well.

The final 👍 should probably come from @LucioFranco though 😊

LucioFranco (approved):

dfreese:
Apologies for the drive by comment, but I believe this functionality was already supported with [Builder::compile_with_config](https://docs.rs/tonic-build/0.4.0/tonic_build/struct.Builder.html#method.compile_with_config).  That's what we used to handle more specific configurations.  [Here is an example](https://github.com/bazelbuild/rules_rust/blob/3b6fc1e0349f4ec4a7c0002d364133ac2f3e757e/proto/prostgen/lib.rs#L63-L65) of how we used compile_with_config to specify this.

tdyas:
> Apologies for the drive by comment, but I believe this functionality was already supported with [Builder::compile_with_config](https://docs.rs/tonic-build/0.4.0/tonic_build/struct.Builder.html#method.compile_with_config). That's what we used to handle more specific configurations. [Here is an example](https://github.com/bazelbuild/rules_rust/blob/3b6fc1e0349f4ec4a7c0002d364133ac2f3e757e/proto/prostgen/lib.rs#L63-L65) of how we used compile_with_config to specify this.

So how did you work-around https://github.com/hyperium/tonic/issues/521 then? Or was there another solution?

dfreese:
> > Apologies for the drive by comment, but I believe this functionality was already supported with [Builder::compile_with_config](https://docs.rs/tonic-build/0.4.0/tonic_build/struct.Builder.html#method.compile_with_config). That's what we used to handle more specific configurations. [Here is an example](https://github.com/bazelbuild/rules_rust/blob/3b6fc1e0349f4ec4a7c0002d364133ac2f3e757e/proto/prostgen/lib.rs#L63-L65) of how we used compile_with_config to specify this.
> 
> So how did you work-around #521 then? Or was there another solution?

Thanks for pointing that out.  I needed to look more closely.  Where things differed is I was specifying the well known types in a different crate and then pulling them into the current crate using [extern_path](https://docs.rs/tonic-build/0.4.0/tonic_build/struct.Builder.html#method.extern_path).  That doesn't appear to run into the same issue.

PR #523, opened 2021-01-06
Remove async-stream dependency

akhilles:
## Motivation

Reduce dependencies, proc macro use.

## Solution

Use stream combinators instead of `async_stream` generators.


LucioFranco:
Whats the motivation around not using `async-stream`? To me this is almost always preferred as they are easier to use and much more intuitive than combinators.

ParkMyCar:
To add a bit of context on the compile times, the change here is relatively small. For procedural macros the largest dependency is `syn`, which continues to be a dependency because of other crates. We can see this by creating a graph of all the transitive dependencies using [`cargo deps`](https://crates.io/crates/cargo-deps)

[Before](https://www.dropbox.com/s/z7vge0isy7lx6lr/fe4d5b9d9%20deps.png?dl=0), [After](https://www.dropbox.com/s/d6vjm4gavxd9oog/fe4d5b9d9%20after-patch%20deps.png?dl=0)

Going a bit further into the actual compile times, we can use `cargo`s experimental [`timing`](https://internals.rust-lang.org/t/exploring-crate-graph-build-times-with-cargo-build-ztimings/10975) feature to see the impact it has. On a 2014 MacBook Pro with an i7, I get the following:

[Before](https://www.dropbox.com/s/t1akh3xqfslkjxc/fe4d5b9d9%20timings.html?dl=0), [After](https://www.dropbox.com/s/0l0tbc1b3kixcmg/fe4d5b9d9%20after-patch%20timings.html?dl=0)

`async-stream-impl` takes my machine 5.6 seconds to compile, but the overall compile time of `tonic` goes from 59.7 seconds to 57.5 seconds, the total change is only 2.2 seconds because of `cargo`s parallelization.

My two cents, `tonic` should continue to use `async-stream` because of the ergonomics it provides, particularly with the `yield` keyword. I agree it's frustrating that `rustfmt` doesn't work within the macro, and it breaks rust-analyzer, but ideally these problems can be solved, it looks like there is already an [issue filed for rust-analyzer](https://github.com/rust-analyzer/rust-analyzer/issues/4046), albeit there isn't much activity on it.

LucioFranco:
@ParkMyCar thanks for the detailed post!

@akhilles yeah, I think some of the TLS/runtime overhead we already get with tokio anyways. In reality all this will be dominated by network/io/syscall overhead.

I think the biggest thing is rustfmt but I don't think this is enough motivation to remove it. Hopefully, we can improve on all these points in the future.

Issue #264, opened 2020-02-14
Allow setting a message max size


L1AN0:
### Motivation

See https://github.com/tensorflow/serving/issues/1382, if the message size is large, tonic currently does not support changing the max message size.

### Proposal

Expose the API the change allowed message size.

akiradeveloper:
@L1AN0 @LucioFranco What is the maximum message size allowed now? As far as I understand, gRPC's default max size in spec is 4MB but tonic server seems to be capable of receive larger requests.

LucioFranco:
I believe there is a cap with h2, its a bug. related issue is here for that @akiradeveloper https://github.com/hyperium/tonic/issues/352

akiradeveloper:
@LucioFranco Thanks for you reply but my question is why tonic doesn't have an upper limit as official implementations do 

https://github.com/grpc/grpc/releases/tag/v1.29.0
> A significant change was made in #22575 where gRPC Core will check whether the decompressed message fits in the configured limit for the maximum message length that the channel can receive. (The default limit is 4MB.)

Is that your intention not to comply with the official behaviors?

LucioFranco:
> @LucioFranco Thanks for you reply but my question is why tonic doesn't have an upper limit as official implementations do

I don't believe this is specified in the protocol definition so it was never added initially. Would be happy to add it if people want this behavior.

 > Is that your intention not to comply with the official behaviors?

Just because gRPC core does it, doesn't mean its "official" per-say. Each implemenation of gRPC has its own niche things. I would love to add this but I don't have the time currently. I would totally accept a PR though.

davidpdrsn:
It seems this was fixed by https://github.com/hyperium/tonic/pull/529. Just re-open if that is not the case.

LucioFranco:
Yes, I believe the C++ implementation has a non-h2 limit as well. We could add this via `Grpc` types.

PR #529, opened 2021-01-08
tonic: add max http2 frame size to server.

SkamDart:
Exposes option to configure the max http2 frame size used by the
underlying hyper server via the `tonic::transport::Server` builder.

Refs: #264

## Motivation

Refs #264 

## Solution
Add as optional parameter to the `tonic::transport::Server` builder to be passed to the hyper server builder.


LucioFranco (approved):
Thanks!

PR #530, opened 2021-01-08
Upgrade to Tokio 1.0.0 ecosystem

QuentinPerez:
Hey 👋 ,

A quick PR to finish the work done by @marshaled here https://github.com/hyperium/tonic/pull/514.

Before merging, we have to wait for the release of `warp`, this PR depends on a fork.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -148,28 +148,29 @@ path = "src/hyper_warp_multiplex/server.rs"
 
 [dependencies]
 tonic = { path = "../tonic", features = ["tls"] }
-prost = "0.6"
-tokio = { version = "0.2", features = ["rt-threaded", "time", "stream", "fs", "macros", "uds"] }
+prost = "0.7"
+tokio = { version = "1.0", features = ["rt-multi-thread", "time", "fs", "macros", "net"] }
+tokio-stream = { version =  "0.1", features = ["net"] }
+async-stream = "0.3"
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
-async-stream = "0.2"
-tower = "0.3"
+tower = { version = "0.4" }
 # Required for routeguide
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-rand = "0.7"
+rand = "0.8"
 # Tracing
 tracing = "0.1.16"
 tracing-subscriber = { version = "0.2", features = ["tracing-log"] }
 tracing-attributes = "0.1"
 tracing-futures = "0.2"
 # Required for wellknown types
-prost-types = "0.6"
+prost-types = "0.7"
 # Hyper example
-hyper = "0.13"
-warp = { version = "0.2", default-features = false }
+hyper = "0.14"
+warp = { git = "https://github.com/aknuds1/warp", branch = "chore/upgrade-tokio",  default-features = false }
 http = "0.2"
-http-body = "0.3"
-pin-project = "0.4.17"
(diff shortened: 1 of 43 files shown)


coolreader18:
It looks like `tower-balance` and some other crates have been merged into `tower` submodules gated by features (which you've enabled). The current tower documentation says:

>Tower is made up of the following crates:
>
>-    tower (this crate)
>-    tower-service
>-    tower-layer
>-    tower-test

So I think it would be good to drop the other crates, which still depend on tokio 0.2 as well.

QuentinPerez:
Oh yes good catch, fixed thx

simonsan:
- [ ] `warp` https://github.com/seanmonstar/warp/pull/753

davidpdrsn:
Hm I've noticed something kinda annoying. It seems `Channel` is not longer `Sync` because it contains a `Semaphore` from tower. `Channel` being `Sync` is something I relied upon in my codebase. Not sure if I should find a solution that doesn't depend on `Channel` being `Sync` or if there is some way to rectifying it.

examples/Cargo.toml
```diff
@@ -148,28 +148,29 @@ path = "src/hyper_warp_multiplex/server.rs"
...
 tracing-attributes = "0.1"
 tracing-futures = "0.2"
 # Required for wellknown types
-prost-types = "0.6"
+prost-types = "0.7"
 # Hyper example
-hyper = "0.13"
-warp = { version = "0.2", default-features = false }
+hyper = "0.14"
+warp = { git = "https://github.com/aknuds1/warp", branch = "chore/upgrade-tokio",  default-features = false }
```
LucioFranco, line 170: This is just for examples, so not blocking a release!

tonic-health/src/server.rs
```diff
@@ -148,15 +140,21 @@ impl Health for HealthService {
...
         };
 
-        let output = status_rx.map(|status| {
-            let status = crate::proto::health_check_response::ServingStatus::from(status) as i32;
-            Ok(HealthCheckResponse { status })
-        });
+        let output = async_stream::try_stream! {
+            // yield the current value
+            let status = crate::proto::health_check_response::ServingStatus::from(*status_rx.borrow()) as i32;
+            yield HealthCheckResponse{ status };
```
LucioFranco, line 151:
```suggestion
            yield HealthCheckResponse { status };
```

tonic-health/src/server.rs
```diff
@@ -148,15 +140,21 @@ impl Health for HealthService {
...
+
+            while let Ok(_) = status_rx.changed().await {
+                let status = crate::proto::health_check_response::ServingStatus::from(*status_rx.borrow()) as i32;
+                yield HealthCheckResponse{ status };
```
LucioFranco, line 155:
```suggestion
                yield HealthCheckResponse { status };
```

coolreader18:
I'm not sure how that'd be useful for `Channel` to be `Sync`, since all the uses of it require `&mut` anyway. Maybe try putting it in a Mutex to make it Sync? Since I assume you just need to store it if it was working before with mutable access

LucioFranco (approved):
Thanks! Will get this merged so we can move forward.

davidpdrsn:
@coolreader18 its being fixed in tower as we speak 😊

edrevo:
Are there any other changes needed for a new release of tonic? Can't wait to get all the good bits that have been added since 0.3.1!

LucioFranco:
@edrevo going through that all right now :)

PR #535, opened 2021-01-14
feat(transport): Fix TLS accept w/ peer certs

LucioFranco:


dingelish:
i don't know where to leave this comment ... but this commit doubled my service's performance. it's really amazing!!!

dingelish:
if needed, i can help backport this commit to 0.1, 0.2 and 0.3 and let everyone enjoy it!!!

LucioFranco:
@dingelish were you using an older version like 0.2 before we attempted to fix this? If so it was doing the handshake slowly. I don't think we should backport but instead have people upgrade. Glad it worked :)

PR #536, opened 2021-01-15
fix(transport): return Poll::ready until error is consumed 

LucioFranco:
Fixup of #495

### DIFF
--- a/tonic/src/transport/service/reconnect.rs
+++ b/tonic/src/transport/service/reconnect.rs
@@ -18,7 +18,7 @@ where
     mk_service: M,
     state: State<M::Future, M::Response>,
     target: Target,
-    error: Option<M::Error>,
+    error: Option<crate::Error>,
     has_been_connected: bool,
     is_lazy: bool,
 }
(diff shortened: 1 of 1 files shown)


tonic/src/transport/service/reconnect.rs
```diff
@@ -130,8 +139,10 @@ where
     }
 
     fn call(&mut self, request: Request) -> Self::Future {
+        tracing::trace!("Reconnect::call");
         if let Some(error) = self.error.take() {
-            return ResponseFuture::error(error);
+            tracing::error!("error: {:?}", error);
```
olix0r, line 144: Why log messages at error when we're returning the error? As this is a library, it seems preferable to keep these at something like `debug` so that the application has more control over whether these are handled/logged, etc.

olix0r: Also, if I understand correctly, this would log the errors _twice_ -- once above in poll_ready and then again at call-time.

LucioFranco: Yeah, no one has complained about it yet, but I think thats not a bad idea. Let me do a pass of those in a follow up PR.

olix0r: > no one has complained about it yet

If I'm correct, this logging was introduced by this PR and not there before...

LucioFranco: We had two other places where we logged at `error!`, so I moved them all to `debug!`, thats what I meant.

Issue #545, opened 2021-01-27
example tower-client "buffer full; poll_ready must be called first"


smhmayboudi:
### Version

```bash
│   ├── tonic v0.4.0
├── tonic v0.4.0 (*)
├── tonic-health v0.3.0
│   └── tonic v0.4.0 (*)
│   └── tonic-build v0.4.0
└── tonic-build v0.4.0 (*)
```
### Platform

Darwin local.local 20.2.0 Darwin Kernel Version 20.2.0: Wed Dec  2 20:39:59 PST 2020; root:xnu-7195.60.75~1/RELEASE_X86_64 x86_64

### Description

I tried to run `cargo run --bin tower-client` and I got a panic.
```bash
thread 'main' panicked at 'buffer full; poll_ready must be called first', /Users/smhmayboudi/.cargo/registry/src/github.com-1ecc6299db9ec823/tower-0.4.3/src/buffer/service.rs:132:14
```

davidpdrsn:
This is probably the same issue as https://github.com/tower-rs/tower/issues/547. That issue includes a fix.

smhmayboudi:
Thanks. should I worry about "tower-server" as well. Since it is very similar to this case.
```rust
        let mut svc = self.inner.clone();

        Box::pin(async move {
            // Do async work here....

            svc.call(req).await
        })
```

davidpdrsn:
Yes. That kind of usage of tower might panic. You're very welcome to submit a PR that fixes the examples 😊

PR #547, opened 2021-01-27
fix(examples): Fix tower examples

smhmayboudi:
based on [tower issue 547] (https://github.com/tower-rs/tower/issues/547), "The original service is ready, but the clone is not necessarily ready."

Issue Number: #545

### DIFF
--- a/examples/src/tower/client.rs
+++ b/examples/src/tower/client.rs
@@ -1,5 +1,6 @@
-use hello_world::greeter_client::GreeterClient;
+use futures::TryFutureExt;
 use hello_world::HelloRequest;
+use hello_world::greeter_client::GreeterClient;
 use service::AuthSvc;
 
 use tonic::transport::Channel;
@@ -57,13 +58,9 @@ mod service {
         }
 
         fn call(&mut self, req: Request<BoxBody>) -> Self::Future {
-            let mut channel = self.inner.clone();
+            // Do extra async work here...
 
-            Box::pin(async move {
-                // Do extra async work here...
-
-                channel.call(req).await.map_err(Into::into)
-            })
+            Box::pin(self.inner.call(req).err_into::<Self::Error>())
         }
     }
 }
--- a/examples/src/tower/client.rs
+++ b/examples/src/tower/client.rs
@@ -1,6 +1,6 @@
 use futures::TryFutureExt;
-use hello_world::HelloRequest;
 use hello_world::greeter_client::GreeterClient;
+use hello_world::HelloRequest;
 use service::AuthSvc;
 
 use tonic::transport::Channel;
--- a/examples/src/tower/server.rs
+++ b/examples/src/tower/server.rs
@@ -1,3 +1,4 @@
+use futures::TryFutureExt;
 use hyper::{Body, Request as HyperRequest, Response as HyperResponse};
 use std::task::{Context, Poll};
 use tonic::{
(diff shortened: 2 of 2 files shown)


examples/src/tower/client.rs
```diff
@@ -57,13 +58,9 @@ mod service {
         }
 
         fn call(&mut self, req: Request<BoxBody>) -> Self::Future {
-            let mut channel = self.inner.clone();
```
LucioFranco, line 60: We actually want to keep it this way because it shows how you can embed async work around the future of the channel.

LucioFranco (commented):
Im not totally sure how this fixes that bug?, because once you produce the future you can call call at any time from that cloned service.

examples/src/tower/client.rs
smhmayboudi: There is another solution which we can apply.

```rust
let clone = self.inner.clone();
let mut inner = std::mem::replace(&mut self.inner, clone);
Box::pin(async move {
    // Do extra async work here...
    channel.call(req).await.map_err(Into::into)
})
```

smhmayboudi:
The solution is working in my project and your example. The clone is acting properly on the server-side, but not on client-side. It seems the clone one needs more time to be ready or somehow have to call pull_ready again.

examples/src/tower/client.rs
olix0r: @LucioFranco See my comment here https://github.com/tower-rs/tower/issues/547#issuecomment-767629149

LucioFranco: Oh interesting.....maybe the mem replace works? @olix0r what do you think?

moh-abk:
Any update on this @smhmayboudi ?

examples/src/tower/client.rs
smhmayboudi: Everything is fine?

smhmayboudi:
@LucioFranco Did you reach any conclusion?

LucioFranco:
@davidpdrsn since you've been working a lot on tower recently any thoughts on this?

examples/src/tower/server.rs
```diff
@@ -71,13 +72,9 @@ where
...
     fn call(&mut self, req: HyperRequest<Body>) -> Self::Future {
-        let mut svc = self.inner.clone();
+        // Do async work here....
 
-        Box::pin(async move {
-            // Do async work here....
-
-            svc.call(req).await
-        })
+        Box::pin(self.inner.call(req).err_into::<Self::Error>())
```
davidpdrsn, line 77:
```suggestion
        let clone = self.inner.clone();
        let mut inner = std::mem::replace(&mut self.inner, clone);
        Box::pin(async move {
            // Do extra async work here...
            inner.call(req).await.map_err::<Self::Error>()
        })
```

This was we still make it clear to users where to add additional async code.

examples/src/tower/client.rs
```diff
@@ -57,13 +58,9 @@ mod service {
...
+            // Do extra async work here...
 
-            Box::pin(async move {
-                // Do extra async work here...
-
-                channel.call(req).await.map_err(Into::into)
-            })
+            Box::pin(self.inner.call(req).err_into::<Self::Error>())
```
davidpdrsn, line 63:
```suggestion
        let clone = self.inner.clone();
        let mut inner = std::mem::replace(&mut self.inner, clone);
        Box::pin(async move {
            // Do extra async work here...
            inner.call(req).await.map_err::<Self::Error>()
        })
```

This was we still make it clear to users where to add additional async code.

davidpdrsn:
I agree with @olix0r. This fix is correct. However as mentioned I think we should use an explicit async block to make it clear to users who to add additional async. I submitted two suggestions.

davidpdrsn:
Superseded by https://github.com/hyperium/tonic/pull/624

PR #549, opened 2021-01-30
Update tokio dependencies in tutorials to be compatible with tonic 0.4

hasheddan:
## Motivation

The existing tutorials do not build due to incompatibility of tonic 0.4 and tokio 0.2.

## Solution

Update tokio dependencies to 1.0 and adjust required features accordinly.

Thanks for the awesome work y'all do on this library!


LucioFranco (approved):
There is one conflict, once that is fixed we can merge this. Thanks!

hasheddan:
@LucioFranco should be ready now :) I just dropped the helloworld update because it looks like another PR got in already that fixed it :+1:

PR #559, opened 2021-02-13
feat(tonic): implement `From<Code>` for `i32`

davidpdrsn:
`Code` is already trivial to convert into an `i32` however that isn't
immediately obvious by looking at the docs. This has tripped me up
before.

Implementing `From<Code>` for `i32` should hopefully make it a bit more
obvious.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -746,6 +746,13 @@ impl From<i32> for Code {
     }
 }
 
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):

LucioFranco:
This doesn't seem like a chore? maybe feat?

PR #566, opened 2021-02-21
feat(build): Use `RUSTFMT` to find `rustfmt` binary

UebelAndre:
## Motivation

The Rust Bazel rules ([rules_rust](https://github.com/bazelbuild/rules_rust)) is looking to add some tools for building targets with Tonic. In development of https://github.com/bazelbuild/rules_rust/pull/479 it was found that `tonic-build` calls `rustfmt` and expects this to be in the user's PATH. This is generally not going to be the case when Building in Bazel. We're looking for a way to specify this binary without needing to rely on or modify PATH.

## Solution
Cargo already has a [pattern for specifying environment variables](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads) to control what binaries are used during builds. The change here adopts the same functionality as Cargo and allows for an environment variable (`RUSTFMT`) to optionally specify the location of the `rustfmt` binary.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -161,13 +161,14 @@ pub fn fmt(out_dir: &str) {
         if !file.ends_with(".rs") {
             continue;
         }
-        let result = Command::new("rustfmt")
-            .arg("--emit")
-            .arg("files")
-            .arg("--edition")
-            .arg("2018")
-            .arg(format!("{}/{}", out_dir, file))
-            .output();
+        let result =
+            Command::new(std::env::var("RUSTFMT").unwrap_or_else(|_| "rustfmt".to_owned()))
+                .arg("--emit")
+                .arg("files")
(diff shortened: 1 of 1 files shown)


UebelAndre:
@LucioFranco Hey, sorry to ping but do you have a moment to take a look at this? 🙏

UebelAndre:
Hey @LucioFranco, just pining this PR one more time in hopes you have time for a quick review 😅 🙏

LucioFranco (approved):

UebelAndre:
> Hey @LucioFranco, just pining this PR one more time in hopes you have time for a quick review 😅 🙏

It's the same thing Cargo uses: https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads

That was my primary goal here, was to have parity with cargo.

Issue #365, opened 2020-06-01
Unix Domain Socket Credential Support with Connected Trait


dfreese:
We're using tonic, and want to be able to use it with unix domain sockets.  When using them, we want to be able to validate the [peer_cred](https://docs.rs/tokio/0.2.21/tokio/net/struct.UnixStream.html#method.peer_addr) of the tokio::net::UnixStream, for example.

It looks like Request gets it's information from the [Connected](https://docs.rs/tonic/0.2.1/tonic/transport/server/trait.Connected.html) trait, which is empty in the [UDS example](https://github.com/hyperium/tonic/blob/master/examples/src/uds/server.rs#L68)

Is peer_cred something that you'd be willing to expose in [Request](https://docs.rs/tonic/0.2.1/tonic/struct.Request.html)?  I understand that would be a little messy, in that it starts to open you up to supporting a number of different connection types through the Request interface.

(BTW: I'm happy to add this, or something else if it seems like there would be a better way of going about it)

LucioFranco:
If I am miss understanding something, shouldn't this work? https://docs.rs/tonic/0.2.1/tonic/struct.Request.html#method.remote_addr

dfreese:
So two of the functions UnixStream exposes are [`peer_cred(&self)`](https://docs.rs/tokio/0.2.21/tokio/net/struct.UnixStream.html#method.peer_cred) which returns [`tokio::net::unix::UCred`](https://docs.rs/tokio/0.2.21/tokio/net/unix/struct.UCred.html) and [`peer_addr(&self)`](https://docs.rs/tokio/0.2.21/tokio/net/struct.UnixStream.html#method.peer_addr) which returns a [` std::os::unix::net::SocketAddr`](https://doc.rust-lang.org/nightly/std/os/unix/net/struct.SocketAddr.html).

Neither of those are really compatible with [`std::net::SocketAddr`](https://doc.rust-lang.org/nightly/std/net/addr/enum.SocketAddr.html) because that expects an IP and a port, but the unix credentials have a user id and group id of the connecting process and the unix socket address is generally a file path.

I could, technically, shoehorn the group id and user id into the flowinfo and scope_id of [`std::net::SocketAddrV6`](https://doc.rust-lang.org/nightly/std/net/struct.SocketAddrV6.html), since they're u32, but that would be quite the hack.

LucioFranco:
@dfreese ah yeah, socketaddr from net didn't make sense but I forgot to check the module :)

[...]

PR #570, opened 2021-02-26
Update Unix Domain Socket example to provide concrete type

dfreese:
In the process of updating our code to tokio 1.x, I had a little more
trouble migrating the code I had on tokio 0.2 because the types that we
used disappeared, particularly tokio::next::unit::Incoming.  Drawing on
the example of hyper::server::conn::AddrIncoming, this creates a
UnixIncoming struct in the example, and then uses that within the server
code.

I generally found this cleaner, though I understand that is subjective.
I also thought it could be useful to inform any further unix api
discussions, similar in the vein of #365.

### DIFF
--- a/examples/src/uds/server.rs
+++ b/examples/src/uds/server.rs
@@ -1,9 +1,6 @@
 #![cfg_attr(not(unix), allow(unused_imports))]
 
-use futures::TryFutureExt;
 use std::path::Path;
-#[cfg(unix)]
-use tokio::net::UnixListener;
 use tonic::{transport::Server, Request, Response, Status};
 
 pub mod hello_world {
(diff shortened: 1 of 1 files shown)


LucioFranco:
Hmm I kinda liked the approach with `async-stream` is there a reason you removed that? It should continue to work with tokio 1.0

dfreese:
Main reason is that we also had a Drop impl to clean up the socket file when the struct was done being used.  Didn't bring that in from our implementation to keep things simple.

I also find having the concrete type a little easier to work with and test, and clearer from what was required of the type, but that's very much a personal preference.

It's a very subjective PR, so no offense taken if you'd rather not include it.

davidpdrsn:
I think I'll a call here and say we don't merge this. As its an example I think the use of `async_stream::stream!` is fine. Feel free to re-open if you want to discuss things further.

PR #574, opened 2021-03-05
add reporting infrastructure on server RPCs + time-series metrics

tdyas:
## Motivation

The goal of this PR is to allow applications to generate comprehensive time-series metrics for server-side RPCs. gRPC servers written in Go can already easily add such functionality through Go modules like [go-grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus). This PR provides a comparable solution for Tonic servers.

_Why is this PR necessary?_ The existing mechanisms for hooking into the request stack did not seem to support the time-series metrics use case well enough because:

1. Interceptors do receive the status code from calling the underlying service (nor any notification for that matter that an RPC had ended). Thus, an interceptor cannot be used to generate metrics relating to completion of RPCs.

2. The generated server stubs implement `Service<http::Request, Response = <http::Response>>`. Any "metrics service" wrapping those servers would have no access to some of the data it would need without additional hacky parsing. For example, the method name would need to be parsed again from the request path (while the service name could come from `NamedService`). And the status code would have to be parsed from the `grpc-status` header stored in the `http::Response`.  Both of those solutions seem very hacky.

## Solution

This PR defines a `tonic::reporter::Reporter` struct which wraps a function/closure that is called with the service name, method name, and the RPC type of each inbound RPC call. The function returns an instance of the `ReportingCallback` trait which receives all further events related to that specific RPC, specifically, stream message received, stream message sent, and RPC completion.

The generated server stubs gain support for:
- `with_reporter` and `with_interceptor_reporter` methods which add a reporter to a server.
- a `NamedMethod` trait impl to supply the parsed service name and method name to `tonic::server::Grpc` without any need for double parsing of the request path.

The new `tonic-metrics` crate contains a reporting callback that use the [`metrics` crate](https://crates.io/crates/metrics) to generate the following metrics:

- Counters:
  * `grpc_server_started_total` with labels `grpc_type`, `grpc_service`, and `grpc_method`
  * `grpc_server_handled_total` with labels `grpc_type`, `grpc_service`, `grpc_method`, and `grpc_code`
  * `grpc_server_msg_received_total` with labels `grpc_type`, `grpc_service`, and `grpc_method`
  * `grpc_server_msg_sent_total` with labels `grpc_type`, `grpc_service`, and `grpc_method`
- Histogram:
  * `grpc_server_handling_seconds` with labels `grpc_type`, `grpc_service`, and `grpc_method`

Open Questions:

- This change is somewhat behind the `transport` feature. Maybe it should be behind a separate `reporting` feature instead? And probably needs me to recheck it is all behind the `transport` feature.

## Future

- _Client Support_: The changes made to `tonic::server::Grpc` could also be made in a similar manner to `tonic::client::Grpc`. This would serve the same purpose of generating time-series metrics and tracing spans.

## Alternatives

- This use case could also probably be solved by reorganizing Tonic and generated stubs to better support middleware that want to use Tonic types in their signature like `tonic:Status` for errors instead of `http::Response`. That sort of solution seemed more complicated to me though than providing the reporter abstraction in this PR. Moreover, such a reorganization  would probably require a significant amount of design work by someone more familiar with Tonic than me.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "tonic",
     "tonic-build",
     "tonic-health",
+    "tonic-metrics",
     "tonic-types",
     "tonic-reflection",
 
--- a/tests/integration_tests/Cargo.toml
+++ b/tests/integration_tests/Cargo.toml
@@ -10,12 +10,14 @@ license = "MIT"
 
 [dependencies]
 tonic = { path = "../../tonic" }
+tonic-metrics = { path = "../../tonic-metrics" }
 prost = "0.7"
 futures-util = "0.3"
 bytes = "1.0"
 
 [dev-dependencies]
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "net"] }
+metrics = "0.14"
 
 [build-dependencies]
 tonic-build = { path = "../../tonic-build" }
--- a/tests/integration_tests/tests/reporter.rs
+++ b/tests/integration_tests/tests/reporter.rs
@@ -1,0 +1,176 @@
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use bytes::Bytes;
+use futures_util::FutureExt;
+use integration_tests::pb::{test_client, test_server, Input, Output};
+use metrics::{GaugeValue, Key, Recorder, Unit};
+use tokio::sync::oneshot;
+use tonic::{transport::Server, Code, Request, Response, Status};
+
+#[tokio::test]
+async fn generates_metrics() {
+    struct Svc;
+
+    #[tonic::async_trait]
+    impl test_server::Test for Svc {
+        async fn unary_call(&self, _: Request<Input>) -> Result<Response<Output>, Status> {
+            Err(Status::with_details(
+                Code::ResourceExhausted,
+                "Too many requests",
+                Bytes::from_static(&[1]),
+            ))
+        }
+    }
+
+    struct TestRecorderInner {
+        counters: Vec<(Key, u64)>,
+        histograms: Vec<(Key, f64)>,
+    }
+
+    struct TestRecorder {
+        inner: Arc<Mutex<TestRecorderInner>>,
+    }
+
+    impl Recorder for TestRecorder {
+        fn register_counter(
+            &self,
+            _key: Key,
+            _unit: Option<Unit>,
+            _description: Option<&'static str>,
+        ) {
+        }
+
+        fn register_gauge(
+            &self,
+            _key: Key,
+            _unit: Option<Unit>,
+            _description: Option<&'static str>,
+        ) {
+        }
+
+        fn register_histogram(
+            &self,
+            _key: Key,
+            _unit: Option<Unit>,
+            _description: Option<&'static str>,
+        ) {
+        }
+
+        fn increment_counter(&self, key: Key, value: u64) {
+            let mut inner = self.inner.lock().unwrap();
+            inner.counters.push((key, value));
+        }
+
+        fn update_gauge(&self, _key: Key, _value: GaugeValue) {}
+
+        fn record_histogram(&self, key: Key, value: f64) {
+            let mut inner = self.inner.lock().unwrap();
+            inner.histograms.push((key, value));
+        }
+    }
+
+    let recorder_inner = Arc::new(Mutex::new(TestRecorderInner {
+        counters: Vec::new(),
+        histograms: Vec::new(),
+    }));
+
+    let recorder = TestRecorder {
+        inner: recorder_inner.clone(),
+    };
+
+    metrics::set_boxed_recorder(Box::new(recorder)).unwrap();
+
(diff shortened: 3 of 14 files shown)


davidpdrsn:
> For example, the method name would need to be parsed again from the request path (while the service name could come from NamedService). And the status code would have to be parsed from the grpc-status header stored in the http::Response. Both of those solutions seem very hacky.

To me a generic solution based on tower doesn't sound hacky at all. In fact that is the exact solution we will be exploring in tower-http https://github.com/tower-rs/tower-http/issues/57. I have already built a similar middleware at work and it's not bad.

tdyas:
> > For example, the method name would need to be parsed again from the request path (while the service name could come from NamedService). And the status code would have to be parsed from the grpc-status header stored in the http::Response. Both of those solutions seem very hacky.
> 
> To me a generic solution based on tower doesn't sound hacky at all. In fact that is the exact solution we will be exploring in tower-http [tower-rs/tower-http#57](https://github.com/tower-rs/tower-http/issues/57). I have already built a similar middleware at work and it's not bad.

Is that middleware at work just for HTTP or does it include gRPC metrics? Generic middlewares with Tonic currently seem limited to just mapping `http::Request` to `hyper::Response` from which a few issues arise in my view:

- The gRPC status code is stored in the `grpc-status` header of the `http::Response` and not in the HTTP status. It seems duplicative to me for Tonic to encode `tonic::Status` into that header, only for a generic middleware to have to decode it again (or at least the integer code).

- This PR is able to instrument the number of message sent/received over streaming RPCs. It seemed much easier to just hook into `tonic::Streaming` to count those messages. Any ideas on how a generic Hyper middleware could do that? (I'm not that familiar with how `http::Request` is used by streaming RPCs.)

- The service name and method name are already parsed by the stubs. The generic Hyper middleware would need to parse the path and extract those separately, but would not be able from the path alone to know the type of RPC. (This is the least of the problems but is at least straightforward to do.)

I would also prefer a solution that leverages more of the `tower::Service` abstraction. However, as just mentioned, Tonic doesn't seem to have any "intermediate" way to insert middleware "closer" in the stack to the RPC call and use types that are more gRPC-specific than the generic HTTP types.

Using metrics as an example, I would like to be able to write a middleware that could intercept the `tonic::Status` for an RPC *before* it is encoded into a `http::Response` and also `map` over the streams used for streaming APIs before they are encoded. That seemed like it would require more than just me to think it over. Hence:

> Moreover, such a reorganization would probably require a significant amount of design work by someone more familiar with Tonic than me.

Are there any planned refactors that would help in the regard?

davidpdrsn:
> Is that middleware at work just for HTTP or does it include gRPC metrics?

gRPC is still HTTP. So yes it does include gRPC. Not quite at the level that something thats able to hook into the guts of tonic could do but still enough for measuring latency, traffic, errors, and throughput. That is all I've ever needed in terms of "framework level" metrics. Most other things goes in application code in my experience.

For measuring streams you should be able to write a `Body` wrapper that adds metrics to `poll_data` or `poll_trailers`. You'll only receive opaque `Bytes` chunks but thats probably fine but that should be fine for counting the number of messages. I do something similar for [the `Trace` middleware](https://github.com/tower-rs/tower-http/blob/tracing-middleware/tower-http/src/trace/body.rs) in tower-http thats currently being developed.

I'm assuming that since each message in a gRPC stream is provided to Tonic as one struct it'll comes through `Body::poll_data` as one encoded `Bytes` chunk but I could very well be wrong. If its not the case it might require new APIs in tonic to support metrics for streams.

> Using metrics as an example, I would like to be able to write a middleware that could intercept the tonic::Status for an RPC before it is encoded into a `http::Response`

Yeah that wouldn't be possible as the tower `Service` would sit outside of the tonic `Channel`. However I'm not sure what the use case for that is. What metrics does that enable?

> Are there any planned refactors that would help move toward a Service-based solution?

I don't think there are plans to change how tonic hooks into Tower.

tdyas:
> gRPC is still HTTP. So yes it does include gRPC. Not quite at the level that something thats able to hook into the guts of tonic could do but still enough for measuring latency, traffic, errors, and throughput. That is all I've ever needed in terms of "framework level" metrics. Most other things goes in application code in my experience.

gRPC at some level is still HTTP but they differ in error semantics. The following test passes:

```
 #[test]
    fn to_http_test() {
        let status = Status::internal("foo");
        let http = status.to_http();
        assert_eq!(http.status(), hyper::StatusCode::OK);
    }
```

I.e., gRPC status codes are encoded as 200 OK at the HTTP level. Thus inspecting HTTP status alone is insufficient to monitor error conditions with gRPC services.

davidpdrsn:
True but you can look at the "grpc-status" response header instead.

tdyas:
Closing as this approach seems to not be favored (and I solved locally in a different way in any event).

rohitjoshi:
I really liked the generic metrics support in tonic. Can this be reconsidered to avoid each one implementing their own?

davidpdrsn:
@rohitjoshi please file a separate issue if you have any feature requests :)

PR #578, opened 2021-03-10
fix(codec): Allocate inbound buffer once

athre0z:
During profiling for bottlenecks in my tonic powered message queue, I noticed that a large amount of compute was spent in reallocating and `memmov`ing over read buffers. 

This PR gets rid of the reallocs entirely by pre-allocating the buffer for the entire message body directly after the message size is known instead of only reserving space for the next chunk when they come in.

In my testing with 8MiB messages, this improved throughput from 370MiB/s to 560MiB/s, which is roughly a 50% improvement. I originally mistyped the numbers, thus the "30-40%" in the commit message -- didn't want to force-push just for that.

### DIFF
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -180,6 +180,7 @@ impl<T> Streaming<T> {
                 }
             };
             let len = self.buf.get_u32() as usize;
+            self.buf.reserve(len);
 
             self.state = State::ReadBody {
                 compression: is_compressed,
@@ -235,16 +236,6 @@ impl<T> Stream for Streaming<T> {
             };
 
             if let Some(data) = chunk {
-                if data.remaining() > self.buf.remaining_mut() {
-                    let amt = if data.remaining() > BUFFER_SIZE {
-                        data.remaining()
-                    } else {
-                        BUFFER_SIZE
-                    };
-
-                    self.buf.reserve(amt);
-                }
-
                 self.buf.put(data);
             } else {
                 // FIXME: improve buf usage.


davidpdrsn (commented):
I think this looks good and I assume some tests would fail if it was wrong.

But would like to hear what @LucioFranco thinks.

tonic/src/codec/decode.rs
```diff
@@ -180,6 +180,7 @@ impl<T> Streaming<T> {
                 }
             };
             let len = self.buf.get_u32() as usize;
+            self.buf.reserve(len);
```
LucioFranco, line 183: Would this be a vector for attack? Since a user could pass any number here and reserve as much memory as possible even if they can't send that on the wire?

athre0z: Yes and no. As far as I see, even before this PR, there was also no protection against the user accumulating an arbitrary large buffer (by, well, just actually sending it). So in this way, this PR doesn't make the situation worse, except that it lowers the attacker's effort. 

On macOS and Linux, allocating huge buffers doesn't actually fault & reserve the corresponding pages until they are actually written to ([overcommit](https://en.wikipedia.org/wiki/Memory_overcommitment)), so the situation doesn't change at all. On Windows, only the size of the page file can be overcommitted, so this would crash immediately on very large values.

Unless I'm wrong and this is actually verified but I just didn't see where (which is very much possible!), I'd argue: yes, this clearly is an issue, but one that should probably be discussed and fixed in a dedicated issue?

LucioFranco: Thanks for the great write up. Yeah, I think this makes sense then!

LucioFranco (approved):
Thanks again!

PR #579, opened 2021-03-16
feat: Expose status constructors

tdyas:
## Motivation

I have a `tower::Service` that generates time-series metrics for gRPC RPCs. That code currently has reimplementations of `Status::from_header_map` and `Code::from_bytes` to extract the gRPC status from a `hyper::http::Response`. I would rather not reimplement those functions.

## Solution

Export `Status::from_header_map` and `Code::from_bytes` from the crate.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -380,7 +380,8 @@ impl Status {
         Status::from_error(&*err.into())
     }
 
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
I think this is reasonable. Thanks!

PR #584, opened 2021-03-19
Fix Prost codec tests and cfg

grippy:
## Motivation

I was curious how one would go about writing a custom Tonic Codec. While looking at the MockEncoder and MockDecoder I came across this issue: https://github.com/hyperium/tonic/issues/583

## Solution

The configures the Prost tests module, fixes the imports, and reading the decode buffer.


davidpdrsn (approved):
Thanks!

PR #586, opened 2021-03-19
Fix ListFeaturesStream return type on RouteGuide Tutorial

JunichiSugiura:
## Motivation
I followed [routeguide-tutorial.md](https://github.com/hyperium/tonic/blob/master/examples/routeguide-tutorial.md) but there's a problem with ListFeaturesStream at compile time.

## Solution
Luckily I found a [solution](https://github.com/hyperium/tonic/issues/540#issuecomment-771739578) in one of the issues.
This PR is to adopt what is suggested in the comment above both in `routeguide-tutorial.md` and `server.rs`.

### DIFF
--- a/examples/routeguide-tutorial.md
+++ b/examples/routeguide-tutorial.md
@@ -185,6 +185,7 @@ prost = "0.7"
 futures-core = "0.3"
 futures-util = "0.3"
 tokio = { version = "1.0", features = ["rt-multi-thread", "macros", "sync", "time"] }
+tokio-stream = "0.1"
 
 async-stream = "0.2"
 serde = { version = "1.0", features = ["derive"] }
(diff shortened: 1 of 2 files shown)


examples/routeguide-tutorial.md
```diff
@@ -273,6 +273,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use tonic::{Request, Response, Status};
+use tokio_stream::wrappers::ReceiverStream;
```
davidpdrsn, line 276: I think this will require adding `tokio-stream` as a dependency [here](https://github.com/hyperium/tonic/blob/2e88100179f2197121d207cb55c3d105d695d938/examples/routeguide-tutorial.md#generating-client-and-server-code).

davidpdrsn (changes requested):
Thanks for noticing! I have also run into this myself actually.

I noticed some small additional thing.

davidpdrsn:
Thanks for fixing this!

PR #597, opened 2021-04-07
Improve comment on Channel::builder()

djc:
This seems to be confused about plural vs singular (and gets the plural form wrong). I think plural is more appropriate since `Endpoint::connect()` and `Endpoint::connect_lazy()` only take a `&self`.

### DIFF
--- a/tonic/src/transport/channel/mod.rs
+++ b/tonic/src/transport/channel/mod.rs
@@ -77,7 +77,7 @@ pub struct ResponseFuture {
 }
 
 impl Channel {
-    /// Create a [`Endpoint`] builder that can create [`Channel`]s.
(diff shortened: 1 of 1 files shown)


tonic/src/transport/channel/mod.rs
```diff
@@ -77,7 +77,7 @@ pub struct ResponseFuture {
 }
 
 impl Channel {
-    /// Create a [`Endpoint`] builder that can create a [`Channel`]'s.
+    /// Create a [`Endpoint`] builder that can create [`Channel`]s.
```
davidpdrsn, line 80:
```suggestion
    /// Create an [`Endpoint`] builder that can create [`Channel`]s.
```

davidpdrsn (approved):
Thanks!

PR #599, opened 2021-04-09
add healthcheck example

cortopy:
## Motivation

I didn't know how to run health checks for a tonic server. As per https://github.com/hyperium/tonic/issues/471, the client for the health service is not exposed.

This example is an easy way I found to run the checks. Other users may benefit from it 

## Solution

See the script added

### DIFF
--- a/examples/src/health/README.md
+++ b/examples/src/health/README.md
@@ -1,0 +1,16 @@
+# Health checks
+
+gRPC has a [health checking protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) that defines how health checks for services should be carried out. Tonic supports this protocol with the optional [tonic health crate](https://docs.rs/tonic-health).
+
+This example uses the crate to set up a HealthServer that will run alongside the application service. In order to test it, you may use community tools like [grpc_health_probe](https://github.com/grpc-ecosystem/grpc-health-probe).
+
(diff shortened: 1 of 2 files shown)


davidpdrsn (commented):
Thanks for working on this!

What do you think about instead adding a `README.md` in `examples/src/health` that explains how to probe the server? I could perhaps link to https://github.com/grpc-ecosystem/grpc-health-probe. Similarly to what we do [here](https://github.com/hyperium/tonic/tree/master/examples/src/gcp).

I'm personally not a huge fan of having a script that downloads some file and just executes it. Would probably be fine but still scares me. The version number would also have to be kept up to date.

cortopy:
that makes a lot of sense. To be honest, I wasn't sure about the script myself either. What if I added a README to the tonic-health package itself so that the rs docs page also gets it? At the moment is quite empty

Not sure if it would be appropriate to mention external tools in the main docs, but I wouldn't mind tackling documentation for both the example and `tonic-health` since I'm at it

davidpdrsn:
@cortopy sounds good 👍

cortopy:
@davidpdrsn just added my changes as content for README files

davidpdrsn (approved):
Love it! Thanks for adding this 😊

Issue #471, opened 2020-10-01
tonic-health: expose client


LukeMathWalker:
### Crates

`tonic-health`

### Motivation

Testing a server that exposes the health check service now requires:

- embedding the protos in your own project;
- adding a build script to generate the client.

### Proposal

Add a new feature flag, `client`, to `tonic-health` to expose a gRPC client for the health check service.
Would you prefer to expose the "raw" client as it comes out of `tonic-build` or would you prefer to wrap it in a nice Rust API that does not expose the underlying protobuf types?

LucioFranco:
iirc tonic-health was just for the server side, but I am not opposed to exposing a client. I don't have much time right now to get to this but would love a PR :)

cc @jen20

LukeMathWalker:
A PR you shall have :guitar: 
Shall we go for raw protobuf types or "wrapped" up nicely in a Rust-ier shell?

LucioFranco:
We could expose the `pb` since I doubt it will change. Just put it into its own module. I think we can then provide a transport feature that enables the `connect` ability? Not sure tbh

jen20:
I definitely planned to get to a client at some point! A PR would be great @LukeMathWalker! Personally I'd tend towards exposing a more idiomatic Rust API if possible, though the PB-generated code is not too bad in this case.

cortopy:
It would be great to have the client generated for health checks. Kubernetes [does not support grpc health checks natively](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/) and requires either a command line tool or a side container that acts as http/grpc proxy for health checks

LucioFranco:
Yeah, I agree, we should expose the client and also the proto module (which would include the client).

PR #602, opened 2021-04-12
feat(health): Expose proto and client (#471)

TransientError:
## Motivation
Resolves https://github.com/hyperium/tonic/issues/471
It can be hard to test the health check server without the client and the proto module. By exposing this, we make it easier for users to do so.

## Solution
Making the proto module public for `health.proto` and flipping build_client to true in `build.rs`

### DIFF
--- a/tonic-health/src/lib.rs
+++ b/tonic-health/src/lib.rs
@@ -22,8 +22,10 @@
 
 use std::fmt::{Display, Formatter};
 
-mod proto {
+/// These are the generated types from our health check proto file.
+pub mod proto {
     #![allow(unreachable_pub)]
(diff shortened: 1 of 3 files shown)


tonic-health/src/lib.rs
```diff
@@ -22,8 +22,10 @@
 
 use std::fmt::{Display, Formatter};
 
-mod proto {
+/// These are the generated types from our health check proto file.
```
LucioFranco, line 25:
```suggestion
/// Generated protobuf types from the `grpc.healthy.v1` package.
```

LucioFranco (approved):

PR #603, opened 2021-04-12
Remove grpc-timeout header from reserved headers

liufuyang:
## Motivation

We have some gRPC servers in Java which requires callers to it set deadline. Currently seems it is not possible to use tonic as a client to talk to those servers as we cannot set `grpc-timeout` on the header?

As it also mentioned and others would like to remove this as well in PR https://github.com/hyperium/tonic/pull/516, also related to this discussion https://github.com/hyperium/tonic/issues/75.

Not sure when can https://github.com/hyperium/tonic/pull/516 be shipped in but this change could perhaps go in first as it blocks not me (see https://github.com/hyperium/tonic/issues/75#issuecomment-772143514)

## Solution

Remove `grpc-timeout` from the reserved list.

Otherwise please let me know how to set `deadline` on the call from tonic as a client.

### DIFF
--- a/tonic/src/metadata/map.rs
+++ b/tonic/src/metadata/map.rs
@@ -198,11 +198,10 @@ pub struct OccupiedEntry<'a, VE: ValueEncoding> {
 
 impl MetadataMap {
     // Headers reserved by the gRPC protocol.
-    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 8] = [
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):

davidpdrsn:
Makes sense while we wait for https://github.com/hyperium/tonic/pull/516

liufuyang:
Thank you :)

PR #606, opened 2021-04-14
feat(transport): Support timeouts with "grpc-timeout" header

davidpdrsn:
This adds support for the `grpc-timeout` header as is [recommended](https://grpc.io/blog/deadlines/) by the gRPC spec.

It replaces use of `tower::timeout::Timeout` in the server with our own `Timeout` middleware. That tries to parse the timeout in `grpc-timeout` header and picks which ever of the server or client timeouts are the shortest.

---

Most of the implementation is based on https://github.com/hyperium/tonic/pull/516. cc @ParkMyCar
Fixes https://github.com/hyperium/tonic/issues/75


tests/integration_tests/tests/timeout.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    let res = client.unary_call(req).await;
+    dbg!(&res);
+
+    let err = res.unwrap_err();
+    assert!(err.message().contains("Timeout expired"));
+
+    // TODO(david): make this work. Will require mapping `TimeoutExpired` errors into
+    // `Code::Cancelled` but can't quite figure out how to do that.
+    assert_eq!(err.code(), Code::Cancelled);
```
davidpdrsn, line 49: @LucioFranco do you know how to fix this? Basically our new `TimeoutExpired` error should change the status code to `Code::Cancelled`. It gets mapped to `Unknown` currently. Did some digging but couldn't quite figure out where that happens.

tests/integration_tests/tests/timeout.rs
```diff
@@ -0,0 +1,94 @@
...
+    req.metadata_mut()
+        // 10 hours
+        .insert("grpc-timeout", "10H".parse().unwrap());
+
+    // TODO(david): for some reason this fails with "h2 protocol error: protocol error: unexpected
+    // internal error encountered". Seems to be happening on `master` as well. Bug?
+    let res = client.unary_call(req).await;
+    dbg!(&res);
+    let err = res.unwrap_err();
+    assert!(err.message().contains("Timeout expired"));
```
davidpdrsn, line 93: This is quite odd. I'm gonna test it on master to see if it behaves the same way.

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+    type Error = crate::Error;
+    type Future = ResponseFuture<S::Future>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx).map_err(Into::into)
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let client_timeout = try_parse_grpc_timeout(req.headers()).unwrap_or_else(|e| {
+            tracing::trace!("Error parsing `grpc-timeout` header {:?}", e);
```
LucioFranco, line 43:
```suggestion
            tracing::trace!("Error parsing `grpc-timeout` header {}", e);
```

Probably want to use `Display` here instead of `Debug`?

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+use http::{HeaderMap, HeaderValue};
+use pin_project::pin_project;
+use std::{fmt, pin::Pin};
+use std::{
+    future::Future,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tokio::time::Sleep;
+use tower_service::Service;
```
LucioFranco, line 11: We should merge all imports from the same crate

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+
+#[derive(Debug, Clone)]
+pub(crate) struct Timeout<S> {
```
LucioFranco, line 14: Might make sense to call this `GrpcTimeout` since it parses a header for that :D

```suggestion
pub(crate) struct GrpcTimeout<S> {
```

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    fn call(&mut self, req: Request<ReqBody>) -> Self::Future {
+        let client_timeout = try_parse_grpc_timeout(req.headers()).unwrap_or_else(|e| {
+            tracing::trace!("Error parsing `grpc-timeout` header {:?}", e);
+            None
+        });
+
+        // Use the shorter of the two durations, if either are set
+        let timeout_duration = match (client_timeout, self.server_timeout) {
```
LucioFranco, line 48: Beautiful use of a match :)

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+    }
+}
+
+#[pin_project(project = OptionPinProj)]
+enum OptionPin<T> {
+    Some(#[pin] T),
+    None,
+}
+
+const GRPC_TIMEOUT_HEADER: &str = "grpc-timeout";
```
LucioFranco, line 105: This should probably live in code that contains the list of all the other reserved headers?

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+                .map_err(|_| val)
+                .and_then(|s| if s.is_empty() { Err(val) } else { Ok(s) })?
+                // `HeaderValue::to_str` only returns `Ok` if the header contains ASCII so this
+                // `split_at` will never panic from trying to split in the middle of a character.
+                // See https://docs.rs/http/0.2.4/http/header/struct.HeaderValue.html#method.to_str
+                //
+                // `len - 1` also wont panic since we just checked `s.is_empty`.
+                .split_at(val.len() - 1);
+
+            // gRPC spec specifies `TimeoutValue` will be at most 8 digits
```
LucioFranco, line 130: Ah love trusting a spec xD

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+            Ok(Some(duration))
+        }
+        None => Ok(None),
+    }
+}
+
+// Note: The wrapped Duration should only be used for logging purposes. It is **not** the
+// actual duration that elapsed, resulting in a timeout, instead it is a close approximation
+#[derive(Debug)]
+struct TimeoutExpired;
```
LucioFranco, line 163:
```suggestion
struct TimeoutExpired(());
```

We might want to expose this so people can downcast to it?

LucioFranco: We also probably should actually enumerate all the errors.....lol

tonic/src/transport/service/timeout.rs
```diff
@@ -0,0 +1,251 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Timeout expired")
+    }
+}
+
+// std::error::Error only requires a type to impl Debug and Display
+impl std::error::Error for TimeoutExpired {}
+
+#[cfg(test)]
+mod tests {
```
LucioFranco, line 175: Might be interesting to add a quicktest fuzz test here?

tonic/src/metadata/map.rs
```diff
@@ -198,11 +198,10 @@ pub struct OccupiedEntry<'a, VE: ValueEncoding> {
 
 impl MetadataMap {
     // Headers reserved by the gRPC protocol.
-    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 8] = [
+    pub(crate) const GRPC_RESERVED_HEADERS: [&'static str; 7] = [
```
LucioFranco, line 201: So I think we should keep `grpc-timeout` here and instead expose a `Request::set_deadline` or something like that. What do you think?

LucioFranco (commented):
Ok just did a quick pass, will pull down the code to dig into those test failures shortly here. Overall looks great, left a few comments inline with some questions.

ParkMyCar:
Ah no worries! Thanks for driving this home 🙂

LucioFranco:
@ParkMyCar sorry for the long delay on all of this! Thanks for the patience 😄

tonic/src/transport/service/timeout.rs
ParkMyCar: @davidpdrsn if you don't have time for this, I've wanted to mess around with `quicktest` but haven't had a chance yet. Would be more than happy to add a follow up diff adding a fuzz test here 🙂

ParkMyCar:
@LucioFranco no problem! Just glad to see this get closer to landing

tonic/src/transport/service/timeout.rs
ParkMyCar: Out of curiosity, whats the difference between `struct TimeoutExpired` and struct `TimeoutExpired(())`?

LucioFranco: It allows someone to never construct the type and thus would allow us to add new stuff to it without breaking the type. Because the internal `0` indexed tuple item is private. So just good tactic to ensure you have flexibility in the future.

tonic/src/transport/service/timeout.rs
davidpdrsn: I've used quickcheck before so I'll look at it now 😊

tonic/src/transport/service/timeout.rs
davidpdrsn: Good idea!

tonic/src/metadata/map.rs
davidpdrsn: Afaik these headers are removed from all requests in `MetadataMap::into_sanitized_headers` so don't think that would work. In my testing that meant the `GrpcTimeout` middleware couldn't see the header value.

I do think adding `Request::set_deadline` is a good idea but given the odd formatting of the header value I don't think it'll be super easy. Think I'll look into that in a separate PR.

tests/integration_tests/tests/timeout.rs
davidpdrsn: Still not sure how to handle this. @LucioFranco do you have idea?

tonic/src/transport/service/timeout.rs
davidpdrsn: Implemented in 3452d3077653f.

davidpdrsn:
@LucioFranco ready for another round of review!

tonic/src/transport/service/timeout.rs
davidpdrsn: `HeaderValue` doesn't implement `Display` actually. From skimming the code it seem `Debug` does the appropriate thing for this case.

tonic/src/metadata/map.rs
LucioFranco: Alright that is acceptable :D lets just get this shipped and we can focus on these things later.

tests/integration_tests/tests/timeout.rs
LucioFranco: I pulled the code down locally its here

```
diff --git a/tonic/src/status.rs b/tonic/src/status.rs
index dcf381d..309fe0c 100644
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -331,6 +331,10 @@ impl Status {
                 if let Some(h2) = err.downcast_ref::<h2::Error>() {
                     return Some(Status::from_h2_error(h2));
                 }
+
+                if let Some(timeout) = err.downcast_ref::<crate::transport::TimeoutExpired>() {
+                    return Some(Status::cancelled(format!("{}", timeout)));
+                }
             }

             cause = err.source();
```

This was the fix for me.

tests/integration_tests/tests/timeout.rs
LucioFranco: Aha you're right its broken on master, the issue is that when we return an `Error` from the service handling the request on the server side h2 goes crazy about it and errors out. We actually need to convert the error into a http status and return it as an empty response rather than as an error. Does that make sense?

davidpdrsn: Yeah that makes sense. This should do it 214c948

tonic/src/metadata/mod.rs
```diff
@@ -29,6 +29,8 @@ pub use self::value::AsciiMetadataValue;
 pub use self::value::BinaryMetadataValue;
 pub use self::value::MetadataValue;
 
+pub(crate) use self::map::GRPC_TIMEOUT_HEADER;
```
LucioFranco, line 32: Do we need to add an allow here?

tonic/src/transport/server/recover_error.rs
```diff
@@ -0,0 +1,75 @@
...
+    type Output = Result<Response<BoxBody>, crate::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result: Result<Response<BoxBody>, crate::Error> =
+            ready!(self.project().inner.poll(cx)).map_err(Into::into);
+
+        match result {
+            Ok(res) => Poll::Ready(Ok(res)),
+            Err(err) => {
+                if let Some(status) = Status::try_from_error(&*err) {
```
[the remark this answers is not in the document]
LucioFranco: @seanmonstar what do you think here?

LucioFranco (approved):
LGTM! Thanks for pushing this through, I left a few inline questions but overall I think we can get this shipped. Feel free to prep a release (or tell me you don't have time and I can do it).

LucioFranco:
Oh and CI needs a fix

tonic/src/metadata/mod.rs
davidpdrsn: I gave it a `#[cfg(feature = "transport")]`

tonic/src/transport/server/recover_error.rs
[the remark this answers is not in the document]
davidpdrsn: Any error that `Status::try_from_error` can downcast will be converted to a response. Do you think we should mention that here or refer to the docs for `Status::try_from_error` (and write some docs for that method as well maybe)?

davidpdrsn:
I'll go ahead and merge this and work on `Request::set_deadline`. I think that would be cool to have in a release together with this as they kinda go together.

Will do follow up PRs if the docs could be made clearer 😊

PR #611, opened 2021-04-24
fix compile error when tower not updated

tdyas:
## Motivation

Fix the following compile error caused by the version of `tower` in `Cargo.lock` not being at least v0.4.6:

```
error[E0599]: no method named `layer_fn` found for struct `ServiceBuilder<tower::layer::util::Identity>` in the current scope
  --> tonic/src/transport/service/connection.rs:54:14
   |
54 |             .layer_fn(|s| AddOrigin::new(s, endpoint.uri.clone()))
   |              ^^^^^^^^ method not found in `ServiceBuilder<tower::layer::util::Identity>`
```

Since the existing version in tonic's `Cargo.toml` is v0.4.4, cargo did not know to update that dependency automatically.

## Solution

Specify v0.4.6 for `tower` to esure `layer_fn` is available.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -68,7 +68,7 @@ h2    = { version = "0.3", optional = true }
 hyper = { version = "0.14.2", features = ["full"], optional = true }
 tokio = { version = "1.0.1", features = ["net"], optional = true }
 tokio-stream = "0.1"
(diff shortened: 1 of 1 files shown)


davidpdrsn:
This was fixed as part of https://github.com/hyperium/tonic/pull/606

PR #612, opened 2021-04-24
add h2::Error as a `source` for tonic::Status when converting from h2::Error

tdyas:
## Motivation

A gRPC server may send a HTTP/2 GOAWAY frame with NO_ERROR status to gracefully shutdown a connection. This appears to Tonic users as a `tonic::Status` with `Code::Internal` and the message set to `h2 protocol error: protocol error: not a result of an error`.

The only way to currently detect this case and differentiate it from other internal errors (e.g., an application-level internal error) is to match on the message. A client may want to differentiate these cases because it may only want to alert on the application-level internal error and not on the transient transport-level issue. (Indeed, this is the use case for which I'm envisioning using this change.)

Matching on a message is not as robust, however, as matching on an `h2::Error` and its reason code. (The message could change for example if a future version of Tonic decided to vary the message. This would break any users that matched on the previous version of the message.)

## Solution

Store the `h2::Error` used when creating a `tonic::Status` from a `h2::Error` and provide it as the `source` for purposes of `std::error::Error`. This will allow users to downcast it and match on the original `h2::Reason`.

Note: `tonic::Status` must now manually implement `Clone` because`h2::Error` cannot implement `Clone` since one of the `Kind` variants stores a `std::io::Error` (which does not implement `Clone`). Since we know that `h2_error` will always have a reason since we only store it in that case, this manual implementation of `Clone` just extracts the reason and creates a new `h2::Error` from it to "clone" it.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -592,14 +592,6 @@ fn invalid_header_value_byte<Error: fmt::Display>(err: Error) -> Status {
     )
 }
 
-impl TryFrom<Box<dyn Error + Send + Sync + 'static>> for Status {
-    type Error = Box<dyn Error + Send + Sync + 'static>;
-
-    fn try_from(err: Box<dyn Error + Send + Sync + 'static>) -> Result<Self, Self::Error> {
-        Status::try_from_error(err)
-    }
-}
-
 #[cfg(feature = "transport")]
 impl From<h2::Error> for Status {
     fn from(err: h2::Error) -> Self {
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -3,6 +3,7 @@ use crate::metadata::MetadataMap;
 use bytes::Bytes;
 use http::header::{HeaderMap, HeaderValue};
 use percent_encoding::{percent_decode, percent_encode, AsciiSet, CONTROLS};
+use std::convert::TryFrom;
 use std::{borrow::Cow, error::Error, fmt};
 use tracing::{debug, trace, warn};
 
@@ -305,38 +306,31 @@ impl Status {
 
     #[cfg_attr(not(feature = "transport"), allow(dead_code))]
     pub(crate) fn from_error(err: Box<dyn Error + 'static>) -> Status {
-        Status::try_from_error(err).unwrap_or_else(|| Status::new(Code::Unknown, err.to_string()))
+        Status::try_from_error(err)
+            .unwrap_or_else(|err| Status::new(Code::Unknown, err.to_string()))
     }
 
-    pub(crate) fn try_from_error(err: &(dyn Error + 'static)) -> Option<Status> {
-        let mut cause = Some(err);
-
-        while let Some(err) = cause {
-            if let Some(status) = err.downcast_ref::<Status>() {
-                return Some(Status {
-                    code: status.code,
-                    message: status.message.clone(),
-                    details: status.details.clone(),
-                    metadata: status.metadata.clone(),
-                    source: Some(err),
-                });
+    pub(crate) fn try_from_error(err: Box<dyn Error + 'static>) -> Result<Status, Box<dyn Error + 'static>> {
+        let err = match err.downcast::<Status>() {
+            Ok(status) => {
+                return Ok(*status);
             }
+            Err(err) => err,
+        };
 
-            #[cfg(feature = "transport")]
-            {
-                if let Some(h2) = err.downcast_ref::<h2::Error>() {
-                    return Some(Status::from_h2_error(h2));
-                }
-
-                if let Some(timeout) = err.downcast_ref::<crate::transport::TimeoutExpired>() {
-                    return Some(Status::cancelled(timeout.to_string()));
-                }
+        #[cfg(feature = "transport")]
+        let err = match err.downcast::<h2::Error>() {
+            Ok(h2) => {
+                return Ok(Status::from_h2_error(&*h2));
             }
+            Err(err) => err,
+        };
 
-            cause = err.source();
+        if let Some(status) = find_status_in_cause_chain(&*err) {
+            return Ok(status);
         }
 
-        None
+        Err(err)
     }
 
     // FIXME: bubble this into `transport` and expose generic http2 reasons.
@@ -383,7 +377,8 @@ impl Status {
     where
         E: Into<Box<dyn Error + Send + Sync>>,
     {
-        Status::from_error(&*err.into())
+        let err: Box<dyn Error + Send + Sync> = err.into();
+        Status::from_error(err)
     }
 
     /// Extract a `Status` from a hyper `HeaderMap`.
@@ -536,6 +531,27 @@ impl Status {
     }
 }
 
+pub(crate) fn find_status_in_cause_chain(err: &(dyn Error + 'static)) -> Option<Status> {
+    let mut cause = Some(err);
+
+    while let Some(err) = cause {
+        if let Some(status) = err.downcast_ref::<Status>() {
+            return Some(Status {
+                code: status.code,
+                message: status.message.clone(),
+                details: status.details.clone(),
+                metadata: status.metadata.clone(),
+                // Since `Status` is not `Clone`, any `source` on the original Status
+                // cannot be cloned so must remain with the original `Status`.
+                source: None,
+            });
+        }
+
+        cause = err.source();
+    }
+
+    None
+}
 impl fmt::Debug for Status {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // A manual impl to reduce the noise of frequently empty fields.
@@ -569,9 +585,11 @@ fn invalid_header_value_byte<Error: fmt::Display>(err: Error) -> Status {
     )
 }
 
-impl From<Box<dyn Error + Send + Sync + 'static>> for Status {
-    fn from(err: Box<dyn Error + Send + Sync>) -> Self {
-        Status::from_error()
+impl TryFrom<Box<dyn Error + Send + Sync + 'static>> for Status {
+    type Error = Box<dyn Error + 'static>;
+
+    fn try_from(err: Box<dyn Error + Send + Sync>) -> Result<Self, Self::Error> {
+        Status::try_from_error(err)
     }
 }
 
@@ -799,25 +817,25 @@ mod tests {
     #[test]
     fn from_error_status() {
         let orig = Status::new(Code::OutOfRange, "weeaboo");
-        let found = Status::from_error(&orig);
+        let found = Status::from_error(Box::new(orig));
 
-        assert_eq!(orig.code(), found.code());
-        assert_eq!(orig.message(), found.message());
+        assert_eq!(found.code(), Code::OutOfRange);
+        assert_eq!(found.message(), "weeaboo");
     }
 
     #[test]
     fn from_error_unknown() {
         let orig: Error = "peek-a-boo".into();
-        let found = Status::from_error(&*orig);
+        let found = Status::from_error(orig);
 
         assert_eq!(found.code(), Code::Unknown);
-        assert_eq!(found.message(), orig.to_string());
+        assert_eq!(found.message(), "peek-a-boo".to_string());
     }
 
     #[test]
     fn from_error_nested() {
         let orig = Nested(Box::new(Status::new(Code::OutOfRange, "weeaboo")));
-        let found = Status::from_error(&orig);
+        let found = Status::from_error(Box::new(orig));
 
         assert_eq!(found.code(), Code::OutOfRange);
         assert_eq!(found.message(), "weeaboo");
@@ -829,7 +847,7 @@ mod tests {
         use std::error::Error as _;
 
         let orig = h2::Error::from(h2::Reason::CANCEL);
-        let found = Status::from_error(&orig);
+        let found = Status::from_error(Box::new(orig));
 
         assert_eq!(found.code(), Code::Cancelled);
 
@@ -837,7 +855,7 @@ mod tests {
             .source()
             .and_then(|err| err.downcast_ref::<h2::Error>())
             .unwrap();
-        assert_eq!(orig.reason(), source.reason());
+        assert_eq!(source.reason(), Some(h2::Reason::CANCEL));
     }
 
     #[test]
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -3,7 +3,7 @@ use crate::metadata::MetadataMap;
 use bytes::Bytes;
 use http::header::{HeaderMap, HeaderValue};
 use percent_encoding::{percent_decode, percent_encode, AsciiSet, CONTROLS};
-use std::{borrow::Cow, error::Error, fmt};
+use std::{borrow::Borrow, borrow::Cow, error::Error, fmt};
 use tracing::{debug, trace, warn};
 
 const ENCODING_SET: &AsciiSet = &CONTROLS
@@ -45,7 +45,7 @@ pub struct Status {
     /// or by `Status` fields above, they will be ignored.
     metadata: MetadataMap,
     /// Optional underlying error.
-    source: Option<Box<(dyn Error + Send + Sync + 'static)>>,
+    source: Option<Box<dyn Error + Send + Sync + 'static>>,
 }
 
 /// gRPC status codes used by [`Status`].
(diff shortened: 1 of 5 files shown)


tdyas:
This could also be simplified to just have `Status` carry a `h2_reason` field and not support any of the `std::error::Error::source` functionality. (There would be a `h2_reason` method returning an `Option<h2::Reason>`).

tonic/src/status.rs
```diff
@@ -546,6 +582,13 @@ impl fmt::Debug for Status {
             builder.field("metadata", &self.metadata);
         }
 
+        #[cfg(feature = "transport")]
+        {
+            if self.h2_error.is_some() {
+                builder.field("h2_error", &self.h2_error);
+            }
```
davidpdrsn, line 589:
```suggestion
            builder.field("h2_error", &self.h2_error);
```

I guess staying close to the actual struct definition is fine as well.

tonic/src/status.rs
```diff
@@ -45,6 +44,31 @@ pub struct Status {
...
+        Status {
+            code: self.code,
+            message: self.message.clone(),
+            details: self.details.clone(),
+            metadata: self.metadata.clone(),
+            // The following contortion to clone `h2_error` is required because `h2::Error`
+            // cannot implement `Clone` because one of the `Kind` variants  stores a
+            // `std::io::Error` (which does not implement `Clone`). Since we know that
+            // `h2_error` will always have a reason, just extract the reason and create a new
+            // `h2::Error` from it.
```
davidpdrsn, line 63:
```suggestion
            // The following contortion to clone `h2_error` is required because `h2::Error`
            // cannot implement `Clone` because one of the `Kind` variants  stores a
            // `std::io::Error` (which does not implement `Clone`).
            //
            // We know that if `h2_error` is a `Some` it will have a `reason` since it was
            // created from that `reason`. If its a `None` this clone wont change anything.
```

nit: I find this wording a bit more clearer.

davidpdrsn (commented):
Nice! I think this makes sense 😊 I had a couple of suggestions. Just minor things.

> This could also be simplified to just have Status carry a h2_reason field and not support any of the std::error::Error::source functionality. (There would be a h2_reason method returning an Option<h2::Reason>).

I actually prefer the implementation in this PR as that we wont need special methods for each new kind of error source we might add in the future.

tonic/src/status.rs
tdyas: Made this change plus fixed some grammatical nits (e.g., `wont`).

tonic/src/status.rs
tdyas: Made this change as well.

davidpdrsn (approved):
I think this looks good but should hear from @LucioFranco as well.

tonic/src/status.rs
```diff
@@ -45,6 +44,32 @@ pub struct Status {
     /// If the metadata contains any headers with names reserved either by the gRPC spec
     /// or by `Status` fields above, they will be ignored.
     metadata: MetadataMap,
+    /// Optional underlying `h2::Error` if this `Status` was converted from a `h2::Error`.
+    #[cfg(feature = "transport")]
+    h2_error: Option<h2::Error>,
```
LucioFranco, line 49: I would like to not make this the specific h2 error but actually a more general `Box<dyn Error + Send + sync>` and to expose that via the `Error::source` methods. Then a user can pull in `h2` and extract that error and do what they need with it.

LucioFranco (changes requested):

tonic/src/status.rs
LucioFranco: Yeah, I could see it down the line storing other errors which is why I wanted to go more generic.

The clone does put a wrench into this doesn't it...I think we could drop clone to simplify this. What do you think? We are already preparing a breaking release so we could include this. cc @davidpdrsn

davidpdrsn:
@tdyas do you wanna drive this home? We would like to include the change to `Status` in the upcoming 0.5 release.

tdyas:
I modified the PR to add a generic source field, although my Rust knowledge is not as advanced so not sure what the right invocation to use to fix this error: 

```
error[E0277]: the trait bound `Box<dyn std::error::Error + Sync + std::marker::Send>: Borrow<dyn std::error::Error>` is not satisfied
   --> tonic/src/status.rs:633:35
    |
633 |         self.source.map(|err| err.borrow())
    |                                   ^^^^^^ the trait `Borrow<dyn std::error::Error>` is not implemented for `Box<dyn std::error::Error + Sync + std::marker::Send>`
    |
    = help: the following implementations were found:
              <Box<T, A> as Borrow<T>>
```

davidpdrsn:
This seems to do the trick

```rust
impl Error for Status {
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        self.source.as_ref().map(|err| (&**err) as _)
    }
}
```

Removing Clone on Status causes a test in prost.rs to fail but that can be fixed with

```rust
let messages = std::iter::repeat_with(move || Ok::<_, Status>(msg.clone())).take(10000);
```

davidpdrsn (commented):
Changing my review "comment" since I wanna give it final look when things are compiling.

Edit: Which apparently github doesn't let you do... How do you go from "approved" to "not approved yet" without "request changes" 🤷

tdyas:
> Edit: Which apparently github doesn't let you do... How do you go from "approved" to "not approved yet" without "request changes" 🤷

Not sure but I can hit the re-request review button to reset that state (which I just did).

tonic/src/status.rs
```diff
@@ -359,7 +364,13 @@ impl Status {
             _ => Code::Unknown,
         };
 
-        Status::new(code, format!("h2 protocol error: {}", err))
+        let mut status = Self::new(code, format!("h2 protocol error: {}", err));
+        let error: Option<Box<dyn Error + Send + Sync + 'static>> = err
+            .reason()
+            .map(h2::Error::from)
+            .map(|err| Box::new(err) as Box<dyn Error + Send + Sync + 'static>);
```
davidpdrsn, line 371: I'm not sure if this could be problematic but if the `h2::Error` doesn't have a `reason` we don't store it as the source. I'm wondering if it would make sense to fix that.

It would require some further since this method would have to receive an owned `h2::Error` which cascades into a few more places in the code. `try_from_error` for example should probably get a `crate::Error` (alias for `Box<dyn Error + Send + Sync + 'static>`). I briefly looked around at the callsites and it seems to be doable.

But again I'm not sure if thats actually necessary. @LucioFranco what do you think?

tonic/src/status.rs
```diff
@@ -800,10 +820,18 @@ mod tests {
...
         let orig = h2::Error::from(h2::Reason::CANCEL);
         let found = Status::from_error(&orig);
 
         assert_eq!(found.code(), Code::Cancelled);
+
+        let source = found
+            .source()
+            .and_then(|err| err.downcast_ref::<h2::Error>())
+            .unwrap();
+        assert_eq!(orig.reason(), source.reason());
```
davidpdrsn, line 834: I'm thinking it would also be nice to have a high level integration test kinda thing that uses the APIs users would. Not sure how to write such a test though.

davidpdrsn (changes requested):
I think it looks good! I had one question and an integration test would be cool as well.

tonic/src/status.rs
tdyas: > It would require some further since this method would have to receive an owned h2::Error which cascades into a few more places in the code. try_from_error for example should probably get a crate::Error (alias for Box<dyn Error + Send + Sync + 'static>). I briefly looked around at the callsites and it seems to be doable.

Correct, that is why I didn't just store the `h2::Error` here since it would require `h2::Error` to be `Clone`.

Also, using an owned value for this conversion would probably be better done via a `TryFrom` impl?

LucioFranco: Yeah, I wonder if we can make it so we get an owned copy of h2::Error and then store it. That would then allow the user to get the h2 error -> source error?

tdyas: That would work when converting `h2::Error` directly by changing `Status::from_h2_error` to take an owned value.

It would not work for `Status::try_from_error` though because the chain of errors retrieved via `.source` are references to the errors. There does not seem to be an API on `Error` to unwrap a chain of errors (e.g., a form of `into_inner`).

tonic/src/status.rs
tdyas: The `h2` crate has test helpers for low level testing of HTTP/2 streams. Maybe this integration test would be similar to the tests in https://github.com/hyperium/h2/tree/master/tests/h2-tests?

tonic/src/status.rs
davidpdrsn: Hm yeah thats a bit unfortunate. Although I pulled down your branch and found a solution that might work. If we change `Status::try_from_error` to

```rust
impl Status {
    fn try_from_error(err: Box<dyn Error + 'static>) -> Result<Status, Box<dyn Error + 'static>> {
        let err = match err.downcast::<Status>() {
            Ok(status) => {
                return Ok(*status);
            }
            Err(err) => err,
        };

        #[cfg(feature = "transport")]
        let err = match err.downcast::<h2::Error>() {
            Ok(h2) => {
                return Ok(Status::from_h2_error(*h2));
            }
            Err(err) => err,
        };

        if let Some(status) = find_status_in_cause_chain(&*err) {
            return Ok(status);
        }

        Err(err)
    }
}

fn find_status_in_cause_chain(err: &(dyn Error + 'static)) -> Option<Status> {
    let mut cause = Some(err);

    while let Some(err) = cause {
        if let Some(status) = err.downcast_ref::<Status>() {
            return Some(Status {
                code: status.code,
                message: status.message.clone(),
                details: status.details.clone(),
                metadata: status.metadata.clone(),
                // Since `Status` is not `Clone`, any `source` on the original Status
                // cannot be cloned so must remain with the original `Status`.
                source: None,
            });
        }

        cause = err.source();
    }

    None
}
```

That captures almost all the previous behavior. The use case that doesn't work where someone has an `h2::Error` thats nested inside the chain of causes. Is that a common use case, probably not but I don't know. There isn't a test for. It does support `Status` being nested inside the chain, which there is a test for so I figure its something we want to keep working.

tdyas: I've incorporated this change into the PR. Please let me know how it looks. I had to convert call sites to box and move errors.

>  Is that a common use case, probably not but I don't know. 

I don't know either.

tonic/src/status.rs
tdyas: Any thoughts on whether an h2-level integration test would be the way to go?

LucioFranco: I don't think we need those low level helpers, we should be able to trigger most things from hyper/h2 directly, no?

LucioFranco:
@tdyas @davidpdrsn any updates on this?

tdyas:
I haven't had a chance to work on an integration test for this due to my work schedule. I see there is now a merge conflict, in trying to fix that, I am having a compile error in `tonic/src/transport/server/recover_error.rs` due to `err`'s ownership passing into `Status::try_from_error` and not being available in the else branch:

```
                if let Some(status) = Status::try_from_error(&*err) {
                    let mut res = Response::new(MaybeEmptyBody::empty());
                    status.add_header(res.headers_mut()).unwrap();
                    Poll::Ready(Ok(res))
                } else {
                    Poll::Ready(Err(err))
                }
```

I will need to adjust the `Status::try_from_error` signature to pass the `err` back out so the caller can recover it if the conversion fails.

tdyas:
I rebased the PR and now it fails with:

```
---- picks_client_timeout_if_thats_sorter stdout ----
thread 'picks_client_timeout_if_thats_sorter' panicked at 'assertion failed: err.message().contains(\"Timeout expired\")', tests/integration_tests/tests/timeout.rs:60:5

---- cancelation_on_timeout stdout ----
thread 'cancelation_on_timeout' panicked at 'assertion failed: err.message().contains(\"Timeout expired\")', tests/integration_tests/tests/timeout.rs:22:5
```

Are those tests relying on `TimeoutExpired` being preferred?

LucioFranco:
@davidpdrsn ^

tdyas:
Also, I had to add `Send + Sync` bounds to the `Error` passed into `Status::try_from_error` to get it compiling. Not sure how to have Rust "pass through" Send+Sync if present onto the result type of a method.

davidpdrsn:
> Are those tests relying on `TimeoutExpired` being preferred?

Yes we should still downcast that error correctly. I did some digging and this fixes it:

```diff
diff --git a/tonic/src/status.rs b/tonic/src/status.rs
index ef9731a..927d95b 100644
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -320,12 +320,6 @@ impl Status {
             Err(err) => err,
         };
 
-        #[cfg(feature = "transport")]
-        let err = match err.downcast::<crate::transport::TimeoutExpired>() {
-            Ok(timeout) => return Ok(Status::cancelled(timeout.to_string())),
-            Err(err) => err,
-        };
-
         #[cfg(feature = "transport")]
         let err = match err.downcast::<h2::Error>() {
             Ok(h2) => {
@@ -555,6 +549,11 @@ fn find_status_in_source_chain(err: &(dyn Error + 'static)) -> Option<Status> {
             });
         }
 
+        #[cfg(feature = "transport")]
+        if let Some(timeout) = err.downcast_ref::<crate::transport::TimeoutExpired>() {
+            return Some(Status::cancelled(timeout.to_string()));
+        }
+
         source = err.source();
     }
```

> Also, I had to add Send + Sync bounds to the Error passed into Status::try_from_error to get it compiling. Not sure how to have Rust "pass through" Send+Sync if present onto the result type of a method.

Nah I don't think there is another way. Its fine 👌

tdyas:
Pushed @davidpdrsn's fix for the timeout error decoding.

tonic/src/status.rs
```diff
@@ -555,6 +592,14 @@ fn invalid_header_value_byte<Error: fmt::Display>(err: Error) -> Status {
...
 }
 
+impl TryFrom<Box<dyn Error + Send + Sync + 'static>> for Status {
+    type Error = Box<dyn Error + Send + Sync + 'static>;
+
+    fn try_from(err: Box<dyn Error + Send + Sync + 'static>) -> Result<Self, Self::Error> {
+        Status::try_from_error(err)
+    }
+}
+
```
davidpdrsn, line 602: Is there a specific use-case this impl enables, or is mainly for convenience? I dunno, feels a bit out of place for me.

davidpdrsn (commented):
@tdyas Thank you for your excellent work on this! I only had one minor comment/question. So close! 😃

tonic/src/status.rs
tdyas: I don't recall, and I can't find a relevant use of this in the main code bases in which I work. Shall I remove it?

davidpdrsn: Yeah I think we should just remove it.

davidpdrsn (approved):
LGTM!

tonic/src/status.rs
```diff
@@ -356,7 +355,13 @@ impl Status {
             _ => Code::Unknown,
         };
 
-        Status::new(code, format!("h2 protocol error: {}", err))
+        let mut status = Self::new(code, format!("h2 protocol error: {}", err));
+        let error: Option<Box<dyn Error + Send + Sync + 'static>> = err
```
LucioFranco, line 359: why do you need this extra type hint?

tonic/src/status.rs
```diff
@@ -3,6 +3,7 @@ use crate::metadata::MetadataMap;
 use bytes::Bytes;
 use http::header::{HeaderMap, HeaderValue};
 use percent_encoding::{percent_decode, percent_encode, AsciiSet, CONTROLS};
+use std::convert::TryFrom;
```
LucioFranco, line 6: Looks like this is failing CI?

LucioFranco (approved):
LGTM, just need to get CI working then we can merge! Thanks for pushing this across the finish line!

tonic/src/status.rs
tdyas: Looking.

tonic/src/status.rs
tdyas: I don't recall. I removed it and it still compiles so pushed the removal of it.

tonic/src/status.rs
tdyas: Fixed.

tdyas: It seems like `deny(warnings)` is only set in CI? When I run `cargo clippy`, it only warned for this and some other issues. It would probably be useful to make the crates enable `#![deny(warnings)]` (but in a separate PR obviously).

riking:
> GOAWAY frame with NO_ERROR status to gracefully shutdown a connection. This appears to Tonic users as a tonic::Status with Code::Internal and the message set to h2 protocol error: protocol error: not a result of an error.

I believe the official GRPC-on-h2 spec says that this should be mapped by the library to a UNAVAILABLE code; the user of the library shouldn't have to deal with recognizing it. https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#goaway-frame

PR #615, opened 2021-04-29
feat(tonic): Add `Request::set_timeout`

davidpdrsn:
This will set the `grpc-timeout` to a value formatted according to [the spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).


LucioFranco (approved):
So slick!

liufuyang:
Aha, this is very nice to have, thank you very much :)

PR #621, opened 2021-05-01
build: show all rustfmt output on errors

davidpdrsn:
Sometimes error output from rustfmt might be printed to stdout instead
of stderr. So to help users debug rustfmt issues this makes stdout get
printed as well.

Fixes https://github.com/hyperium/tonic/issues/600


LucioFranco (approved):
good idea!

PR #622, opened 2021-05-01
feat(tonic): Use `BoxBody` from `http-body` crate

davidpdrsn:
As of `http-body` 0.4.1 its has had a `BoxBody` type similar to
`tonic::body::BoxBody`. It also has `Empty` and `Body::map_{data,err}`.

That means all the custom body things we had in tonic can basically be
replaced with that.

Note that this is a breaking change so we should merge this next time we
decide to ship a breaking release.

The breaking changes are:

- `tonic::body::Body` has been removed. I think its fine for users to
depend directly on `http-body` if they need this trait.
- `tonic::body::BoxBody` is now just a type alias for
`http_body::combinators::BoxBody<Bytes, Status>`. So the methods it
previously had are gone. The replacements are
  - `tonic::body::Body::new` -> `http_body::Body::boxed`
  - `tonic::body::Body::map_from` -> `http_body::Body::map_data` and
  `http_body::Body::map_err` depending on which part you want to map.
  - `tonic::body::Body::empty` -> `http_body::Empty`

Additionally a `Sync` bound has been added to a few methods. I actually
don't think this is a breaking change because the old
`tonic::body::Body` trait had `Sync` as a supertrait meaning the `Sync`
requirement was already there.

Fixes https://github.com/hyperium/tonic/issues/557


davidpdrsn:
Not quite sure why the interop tests are failing. They work for me locally. I'm on macOS.

LucioFranco:
@davidpdrsn are you sure you are running the right tests locally? There is a custom body impl to merge the trailers in the interop tests.

tonic/src/codegen.rs
```diff
@@ -31,3 +30,7 @@ impl std::fmt::Display for Never {
 }
 
 impl std::error::Error for Never {}
+
+pub fn empty_body() -> crate::body::BoxBody {
+    http_body::Empty::new().map_err(|err| match err {}).boxed()
+}
```
LucioFranco, line 36: What is different here than the one in `body.rs`?

LucioFranco (commented):
Seeing all that red in the diff makes me extremely happy!!!

davidpdrsn:
I can reproduce the interop failure locally as well now. I'll look into it.

tonic-reflection/tests/proto/grpc.reflection.v1alpha.rs
```diff
@@ -166,9 +166,9 @@ pub mod server_reflection_client {
     impl<T> ServerReflectionClient<T>
     where
         T: tonic::client::GrpcService<tonic::body::BoxBody>,
-        T::ResponseBody: Body + HttpBody + Send + 'static,
+        T::ResponseBody: Body + Body + Send + 'static,
```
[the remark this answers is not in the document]
LucioFranco: delete them?

davidpdrsn:
The bug caught by the interop tests was in `http_body::Empty`. Fix is https://github.com/hyperium/http-body/pull/42

tonic-reflection/tests/proto/grpc.reflection.v1alpha.rs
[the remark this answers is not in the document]
davidpdrsn: Removed in fa29bae80945b.

tonic-reflection/tests/proto/mod.rs
```diff
@@ -1,337 +0,0 @@
-/// The message sent by the client when calling ServerReflectionInfo method.
```
LucioFranco, line 1: why were these deleted?

tonic/src/client/grpc.rs
```diff
@@ -71,8 +71,8 @@ impl<T> Grpc<T> {
     ) -> Result<Response<M2>, Status>
     where
         T: GrpcService<BoxBody>,
-        T::ResponseBody: Body + HttpBody + Send + 'static,
-        <T::ResponseBody as HttpBody>::Error: Into<crate::Error>,
+        T::ResponseBody: Body + Send + Sync + 'static,
```
LucioFranco, line 74: Sync is new here?

tonic-reflection/tests/proto/mod.rs
davidpdrsn: Thought thats what you mean by https://github.com/hyperium/tonic/pull/622#discussion_r628253343.

Not sure what those files were for anyway. Removing them didn't break anything 🤔

tonic/src/client/grpc.rs
davidpdrsn: Yes and no. The old `tonic::body::Body` [had `Sync` as a super trait](https://docs.rs/tonic/0.4.3/src/tonic/body.rs.html#16-42) so previously `Sync` was implicitly required.

LucioFranco: Make sense!

LucioFranco (approved):

PR #629, opened 2021-05-04
Map tonic::transport::Error in tonic::Status::Unavailable

### LINKED ISSUE #628: Tonic Transport Errors not mapped to Unavailable
### Version
For completeness, our the all of the potentially relevant deps in our Cargo.toml are:
```toml
[dependencies]
async-stream = "=0.3.1"
futures = "=0.3.14"
futures-util = "=0.3.14"
http = "=0.2.4"
hyper = "=0.14.5"
prost = "=0.7.0"
prost-derive = "=0.7.0"
prost-types = "=0.7.0"
prost-build = "=0.7.0"
ring = "=0.16.20"
rustls = { version = "=0.19.0", features = ["default", "dangerous_configuration"] }
rustls-native-certs = "=0.5.0"
tokio = { version = "=1.5.0", features = ["full"] }
tokio-stream = "=0.1.5"
tokio-util = "=0.6.6"
tonic = { version = "=0.4.2", features = ["default", "tls", "tls-roots", "transport"] }
tonic-build = "=0.4.2"
tower = "=0.4.6"
```

### Description
We were doing robustness testing of our connections.  We start with a basic `tonic::transport::let channel = tonic::transport::Channel::from_static("http://some-endpoint.com").connect_lazy()`.  We will either have the network disconnected at the start, or disconnect it during the use of the client.  We've seen a copy different errors related to the underlying connection, but they all show up as transport errors.

> Failed to upload file "4": Tonic gRPC Error: status: Unknown, message: "transport error: connection error: broken pipe", details: [], metadata: MetadataMap { headers: {} }

> Failed to upload file "4": Tonic gRPC Error: status: Unknown, message: "transport error: error trying to connect: dns error: failed to lookup address information: Name or service not known", details: [], metadata: MetadataMap { headers: {} }

I believe Unavailable is the more correct code here, so that the client knows it's valid to retry.  I'm relatively sure that is consistent with gRPC's implementation.  It's been a little bit, so I'll need to put together a specific example.

dfreese:
This came up when testing retry logic when using the
tonic::transport::Error.  Specifically, we would run into Unknown errors
during the following sequence:

* unplug ethernet
* create client with connect_lazy
* unary call -> Unknown error, representing a tonic::transport::Error

Fixes #628

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -332,6 +332,10 @@ impl Status {
                     return Some(Status::from_h2_error(h2));
                 }
 
+                if let Some(transport) = err.downcast_ref::<crate::transport::Error>() {
+                    return Some(Status::unavailable(transport.to_string()));
+                }
+
                 if let Some(timeout) = err.downcast_ref::<crate::transport::TimeoutExpired>() {
                     return Some(Status::cancelled(timeout.to_string()));
                 }
@@ -817,6 +821,25 @@ mod tests {
         let err = orig.to_h2_error();
 
         assert_eq!(err.reason(), Some(h2::Reason::CANCEL));
+    }
+
+    #[test]
+    #[cfg(feature = "transport")]
+    fn from_error_transport() {
+        #[derive(Debug)]
+        struct TestErr {}
+        impl fmt::Display for TestErr {
+            fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                write!(f, "hello, it's me")
+            }
+        }
+        impl std::error::Error for TestErr {}
+
+        let orig = crate::transport::Error::from_source(TestErr {});
+        let found = Status::from_error(&orig);
+
+        assert_eq!(found.code(), Code::Unavailable);
+        assert_eq!(found.message(), "hello, it's me");
     }
 
     #[test]


davidpdrsn:
I'm curious what the underlying error is actually causing what you're seeing. Its most likely `h2::Error` but would be nice to know what its `reason` is. Perhaps you try to add some more. Maybe some `println`s [here](https://github.com/hyperium/tonic/blob/master/tonic/src/transport/error.rs#L34-L41) would show the exact error.

We already map `h2::Reason::REFUSED_STREAM` to `Code::Unavailable` [here](https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs#L358) according to https://github.com/grpc/grpc/blob/3977c30/doc/PROTOCOL-HTTP2.md#errors.

`transport::Error` can also happen for a bunch of different reasons not all of which feel correct to map to unavailable. [An invalid user agent for example.](https://github.com/hyperium/tonic/blob/master/tonic/src/transport/error.rs#L24)

dfreese:
That's fair. Let me see if I can come up with a clear test case to demonstrate what I was seeing.

riking:
#612 potentially related where GOAWAY would cause new streams to fail with a `NO_ERROR` error

tdyas:
> Its most likely h2::Error but would be nice to know what its reason is.

FYI https://github.com/hyperium/tonic/pull/612 was merged recently and makes the underlying `h2::Error` available as the `source` of the `Status` via the `std::error::Error::source` method.

dfreese:
I've updated this to handle `hyper::Error`, which I think is more representative of the underlying problem.  Since there isn't a way to create hyper errors, this is covered by integration, not unit, tests.  I updated the connection integration test to check the updated error.

I haven't created a good integration test for the keep_alive behavior yet (which is the `is_timeout()` branch of the check added here).  I wanted to check about the overall approach first before diving into that.

dfreese:
@tdyas, it looks like #612 changed how the h2 error was being mapped to an error.  tonic is no longer checking for an h2 error on every level, and then mapping it to the appropriate gRPC error code.  It looks like it now only check in the top-level case.  I think I agree from @riking's perspective that the user should only be required to check the gRPC status in order to know how to appropriately handle the gRPC error.

dfreese:
@davidpdrsn It looks like the connection.rs integration test case was reasonable enough to validate what I was looking at.  In that case, what I'm seeing in terms of an error chain is:
1. tonic transport error
2. hyper error (`is_connect() == true`)
3. std::io::Error (Connection Refused [os error 111])

There's no underlying h2::Error

riking:
> I think I agree from @riking's perspective that the user should only be required to check the gRPC status in order to know how to appropriately handle the gRPC error.

And in particular, correctly returning UNAVAILABLE is necessary for automatic retry machinery to function.

PR #640, opened 2021-05-12
feat(metadata): remove manual `Send + Sync` impls for metadata types

davidpdrsn:
Removes manual `Send` and `Sync` impls for these types as they weren't
necessary:

- `metadata::Iter`
- `metadata::IterMut`
- `metadata::ValueDrain`
- `metadata::ValueIterMut`

Since all the types they contained were `Send + Sync` we can rely on the
compiler automatically adding the impls.

Also added some tests to ensure we don't accidentally make them `!Send`
or `!Sync` in the future.

Fixes https://github.com/hyperium/tonic/issues/636

### DIFF
--- a/tonic/src/metadata/map.rs
+++ b/tonic/src/metadata/map.rs
@@ -1244,9 +1244,6 @@ impl<'a> Iterator for Iter<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
perfect, thanks!

PR #642, opened 2021-05-13
feat(tonic): add `Request` and `Response` extensions

davidpdrsn:
Adds `tonic::Extensions` which is a newtype around `http::Extensions`.

Request extensions can be set by interceptors with
`Request::extensions_mut` and retrieved from RPCs with
`Request::extensions`. Extensions can also be set in tower middleware
and will be carried through to the RPC.

Since response extensions cannot be set by interceptors the main use
case is to set them in RPCs and retrieve them in tower middlewares.
Figured that might be useful.

Fixes https://github.com/hyperium/tonic/issues/255

### DIFF
--- a/tests/integration_tests/tests/extensions.rs
+++ b/tests/integration_tests/tests/extensions.rs
@@ -1,0 +1,150 @@
+use futures_util::FutureExt;
+use hyper::{Body, Request as HyperRequest, Response as HyperResponse};
+use integration_tests::pb::{test_client, test_server, Input, Output};
+use std::{
+    task::{Context, Poll},
+    time::Duration,
+};
+use tokio::sync::oneshot;
+use tonic::{
+    body::BoxBody,
+    transport::{Endpoint, NamedService, Server},
+    Request, Response, Status,
+};
+use tower_service::Service;
+
+struct ExtensionValue(i32);
+
+#[tokio::test]
+async fn setting_extension_from_interceptor() {
+    struct Svc;
+
+    #[tonic::async_trait]
+    impl test_server::Test for Svc {
+        async fn unary_call(&self, req: Request<Input>) -> Result<Response<Output>, Status> {
+            let value = req.extensions().get::<ExtensionValue>().unwrap();
+            assert_eq!(value.0, 42);
+
+            Ok(Response::new(Output {}))
+        }
+    }
+
+    let svc = test_server::TestServer::with_interceptor(Svc, |mut req: Request<()>| {
+        req.extensions_mut().insert(ExtensionValue(42));
+        Ok(req)
+    });
+
+    let (tx, rx) = oneshot::channel::<()>();
+
+    let jh = tokio::spawn(async move {
+        Server::builder()
+            .add_service(svc)
+            .serve_with_shutdown("127.0.0.1:1323".parse().unwrap(), rx.map(drop))
+            .await
+            .unwrap();
+    });
+
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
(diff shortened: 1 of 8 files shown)


tests/integration_tests/tests/extensions.rs
```diff
@@ -0,0 +1,150 @@
...
+    let channel = Endpoint::from_static("http://127.0.0.1:1323")
+        .connect()
+        .await
+        .unwrap();
+
+    let mut client = test_client::TestClient::new(channel);
+
+    match client.unary_call(Input {}).await {
+        Ok(_) => {}
+        Err(status) => panic!("{}", status.message()),
```
LucioFranco, line 58: wouldn't an unwrap here work?

tests/integration_tests/tests/extensions.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    jh.await.unwrap();
+}
+
+#[derive(Debug, Clone)]
+struct InterceptedService<S> {
+    inner: S,
+}
+
+impl<S> Service<HyperRequest<Body>> for InterceptedService<S>
```
LucioFranco, line 118: Its crazy how complex this is :(

tonic/src/extensions.rs
```diff
@@ -0,0 +1,67 @@
+use std::fmt;
+
+/// A type map of protocol extensions.
+///
+/// `Extensions` can be used by [`Interceptor`] and [`Request`] to store extra data derived from
+/// the underlying protocol.
+///
+/// [`Interceptor`]: crate::Interceptor
+/// [`Request`]: crate::Request
+pub struct Extensions(http::Extensions);
```
LucioFranco, line 10: nit (not blocking): Should we make this a proper struct with an `inner`? Idk what style I like better honestly but I tend to do that over a tuple struct 🤷 up to you doesn't matter too much.

tonic/src/lib.rs
```diff
@@ -62,6 +62,7 @@
 //! [`transport`]: transport/index.html
 
 #![recursion_limit = "256"]
+#![allow(clippy::inconsistent_struct_constructor)]
```
LucioFranco, line 65: the heck is this lint lmao

LucioFranco (approved):

tests/integration_tests/tests/extensions.rs
davidpdrsn: Yeah true. I copied some other test that didn't have it. Will fix it.

tests/integration_tests/tests/extensions.rs
davidpdrsn: Yeah I really wanna try and find a way to make this easier. Having to implement `NamedService` also complicates things a bit because you cannot take a middleware from tower and use it with tonics router.

tonic/src/lib.rs
davidpdrsn: The most pedantic lint ever https://rust-lang.github.io/rust-clippy/master/#inconsistent_struct_constructor. rust-analyzer was complaining about it.

tonic/src/extensions.rs
davidpdrsn: I don't really have a preference. I'll change it.

tests/integration_tests/tests/extensions.rs
LucioFranco: Yeah, though I don't have a proper solution either :(

tonic/src/lib.rs
LucioFranco: lmaooooo

PR #647, opened 2021-05-14
feat(transport): provide generic access to connect info

davidpdrsn:
Provides generic access to "connection info" provided by the stream of incoming connections through request extensions.

The `Connected` trait is changed to:

```rust
/// Trait that connected IO resources implement and use to produce info about the connection.
///
/// [...]
pub trait Connected {
    /// The connection info type the IO resources generates.
    type ConnectInfo: Clone + Send + Sync + 'static;

    /// Create type holding information about the connection.
    fn connect_info(&self) -> Self::ConnectInfo;
}
```

`TcpStream` and `AddrStream` implement this and return `TcpConnectInfo` which has `remote_addr: Option<SocketAddr>`. `TlsStream` implements it by returning `TlsConnectInfo<T>` where `T` is some other connect info type. That has `get_ref` to access the inner type and `peer_certs` to get the certificates.

`ServerIo` is then changed to no longer box the IO type but instead be generic such that no type information is erased:

```rust
pub(crate) enum ServerIo<IO> {
    Io(IO),
    #[cfg(feature = "tls")]
    TlsIo(TlsStream<IO>),
}
```

Where `IO` and `TlsStream<IO>` implement `Connected`.

In `MakeSvc` we access the connect info from the IO object and set it as a request extension so its accessible later.

Should be fairly flexible since we don't have to keep adding methods to the `Connected` trait but can instead add those on the types produced.

It is slightly less discoverable but I've updated the uds example to demo it and the `remote_addr` and `peer_certs` methods still exist on `Request`.

Fixes https://github.com/hyperium/tonic/issues/365


LucioFranco (approved):
This is slick, love it!

PR #651, opened 2021-05-16
feat(tonic): make it easier to add tower middleware to servers

davidpdrsn:
This is an attempt at making tower easier to use with tonic. It only touches the server parts because I think [adding tower middleware to a client is already easy](https://github.com/hyperium/tonic/blob/master/examples/src/tower/client.rs#L13-L16).

The main limitations of the current setup I think are:

1. Tonic's router requires that all services implement `NamedService`. This is big downside because it means middleware defined in tower cannot be used since tower doesn't know anything about tonic.
2. Users who want to do middleware kind of things, that interceptors don't support, have to know that tower exists and works with tonic.
3. Tower middleware must use `http::Request` and `http::Response` to work with tonic. Thats quite surprising since tonic already has its own request and response types.

What I have so far solves 1 and 2. I'm not quite sure 3 has a good solution but might be possible to write a middleware that converts things back and forth behind the scenes.

## Using tower now

With this patch adding tower middleware to a tonic server looks like:

```rust
let svc = GreeterServer::new(greeter);

// some large layer who's output service _doesn't_
// implement `NamedService`
let layer = tower::ServiceBuilder::new()
    // good old tower middlewares
    .timeout(Duration::from_secs(30))
    .concurrency_limit(42)
    // some custom middleware we wrote
    .layer(MyMiddlewareLayer::default())
    .into_inner();

Server::builder()
    // wrap all services in the layer
    .layer(layer)
    .add_service(svc.clone())
    // also supports optional services
    .add_optional_service(Some(svc))
    .serve(addr)
    .await?;
```

So `Server` and `Router` becomes generic over a tower layer. The layer starts out as `Identity` but can be changed with the `Server::layer` method. The layer will be applied before handing the service off to hyper via `MakeSvc`.

`Server::layer` will show up in the docs and gives us a place to mention and demo using tower to add middleware. That should hopefully increase awareness a bit.

## TODO

Things to fix should be decide to move forward with this.

- [x] Write docs for eveything
- [x] Update interceptor client and server examples
- [x] Remove old intercept API?
- [x] Don't derive Debug for InterceptorFn and InterceptedService
- [x] Update generated `with_interceptor` constructors to use InterceptedService
- [x] Don't use `BoxBody` as bounds for services in `add_service` and `add_optional_service` as that requires `Error = Status`.

Fixes https://github.com/hyperium/tonic/issues/372 https://github.com/hyperium/tonic/issues/499 https://github.com/hyperium/tonic/issues/70 
Ref https://github.com/hyperium/tonic/issues/69


davidpdrsn:
Update: Interceptors can now be added with `tonic::service::interceptor_fn`:

```rust
Server::builder()
    .layer(tonic::service::interceptor_fn(intercept))
    .add_service(svc)
    .serve(addr)
    .await?;
```

And since `interceptor_fn` just returns a `tower::Layer` they can be composed just like any other tower service.

davidpdrsn:
Converting this back to a draft. Wanna see if I can find a solution to

> Don't use BoxBody as bounds for services in add_service and add_optional_service as that requires Error = Status.

That means you currently cannot use middleware that change the body type which quite a few in tower-http does, namely `Trace`.

tonic/src/transport/server/mod.rs
```diff
@@ -360,20 +364,44 @@ impl Server {
         Router::new(self.clone(), svc)
     }
 
-    pub(crate) async fn serve_with_shutdown<S, I, F, IO, IE>(
+    /// TODO(david): docs
```
davidpdrsn, line 367: Reminder to self to write docs for this.

tonic/src/transport/server/mod.rs
```diff
@@ -25,26 +25,33 @@ use crate::transport::Error;
...
 use tokio::io::{AsyncRead, AsyncWrite};
-use tower::{limit::concurrency::ConcurrencyLimitLayer, util::Either, Service, ServiceBuilder};
-use tracing_futures::{Instrument, Instrumented};
+use tower::{
+    layer::util::Identity, layer::Layer, limit::concurrency::ConcurrencyLimitLayer, util::Either,
+    Service, ServiceBuilder,
+};
 
-type BoxService = tower::util::BoxService<Request<Body>, Response<BoxBody>, crate::Error>;
+type BoxHttpBody = http_body::combinators::BoxBody<Bytes, crate::Error>;
```
davidpdrsn, line 53: `tonic::body::BoxBody` uses `Status` as the error type which break tower middleware that wraps the response body, as is common in tower-http. Have to use this different error type for response bodies instead since it supports any error. Its only. used in the `serve_*` methods and not in `add_service` and `add_optional_service`.

davidpdrsn: Requiring the `Data` type to be `Bytes` should be fine and simplifies parts of the implementation.

tonic/src/transport/service/router.rs
```diff
@@ -10,7 +10,7 @@ use std::{
 use tower_service::Service;
 
 #[derive(Debug)]
-pub(crate) struct Routes<A, B, Request> {
+pub struct Routes<A, B, Request> {
```
davidpdrsn, line 13: Note to self: Check if this really needs to be `pub`.

examples/src/tower/client.rs
```diff
@@ -1,8 +1,9 @@
 use hello_world::greeter_client::GreeterClient;
 use hello_world::HelloRequest;
 use service::AuthSvc;
+use tower::ServiceBuilder;
```
LucioFranco, line 4: should tonic re-export? Or even better could the channel build expose a way to get the servicebuilder?

tests/integration_tests/tests/complex_tower_middleware.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn new() -> Self {
+        todo!()
+    }
+}
+
+impl<S> Layer<S> for MyServiceLayer {
+    type Service = MyService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        todo!()
```
LucioFranco, line 47: if these are not actual todo we should use `unimplemented!`.

tonic/src/service/interceptor.rs
```diff
@@ -0,0 +1,183 @@
...
+/// request, going through the [tower] abstraction is recommended.
+///
+/// Interceptors is not recommend should not be used to add logging to your service. For that a
+/// [tower] middleware is more appropriate since it can also act on the response.
+///
+/// See the [interceptor example][example] for more details.
+///
+/// [tower]: https://crates.io/crates/tower
+/// [example]: https://github.com/hyperium/tonic/tree/master/examples/src/interceptor
+// TODO: when tower-http is shipped update the docs to mention its `Trace` middleware which has
```
LucioFranco, line 34: gh issue for this?

tonic/src/transport/server/mod.rs
LucioFranco: Yeah I think this sounds right

tonic/src/transport/server/mod.rs
LucioFranco: I assume you plan to include this in this PR?

LucioFranco (approved):
Everything looks good to me, I see you still have a few things to check but approving since I trust you to complete those before merging! I love this PR!

tonic/src/transport/server/mod.rs
davidpdrsn: Yes totally!

tonic/src/transport/service/router.rs
davidpdrsn: It does unfortunately since it needs to appear in trait bounds. I've made it `#[doc(hidden)]`.

tests/integration_tests/tests/complex_tower_middleware.rs
davidpdrsn: Good point 👍

examples/src/tower/client.rs
davidpdrsn: I considered that as well but opted not to. Since `ServiceBuilder` isn't very useful on its own its not clear which other things from tower we would also have to re-export.

I might consider adding some convenience stuff to channel builders but I think I'll do that as a separate PR.

tonic/src/service/interceptor.rs
davidpdrsn: yes!

drauschenbach:
Doesn't this refactor break the `interceptor/client.rs` and `authentication/client.rs` examples?

davidpdrsn:
@drauschenbach indeed it does. That's why it was released as part of tonic 0.5.

PR #662, opened 2021-05-23
feat(transport): add connect timeout to `Endpoint`

davidpdrsn:
Pulls in [hyper-timeout] which has a connector that can have a timeout
applied. I did consider vendoring hyper-timeout since its a fairly small
crate. I guess we can always do that later since its not exposed
publicly.

I would also like to add a test but I'm not sure about the best way of
testing this.

Fixes https://github.com/hyperium/tonic/issues/498

[hyper-timeout]: https://github.com/hjr3/hyper-timeout


LucioFranco:
hmmm I'd rather vendor this personally, what do you think @davidpdrsn probably not worth an extra dependency etc

LucioFranco (approved):

davidpdrsn:
I will get some time later this week to focus on OSS. Will vendor hyper-timeout then.

PR #670, opened 2021-05-28
fix: cargo-web compilation

gmorer:
## Motivation

Since the changement with the BoxBody #622 tonic-web wasn't compiling, the CI didn't detect it (this will be fixed with #648).

## Solution

Make `tonic::body::empty_body` public and use it instead of the old and non existant anymore `BoxBody::empty()`


davidpdrsn:
Thanks!

PR #673, opened 2021-06-01
modify visibility

### LINKED ISSUE #672: Leaked private struct in public function
### Version

0.4.3

### Platform

N/A

### Description

In the `add_service` function on `Router`, you are retuning a new `Router` which uses `Or` as a type parameter. While `Or` itself is a public struct, its only public inside its module, since the module it lives in is private.

On the Rust Discord, they suggested that this really is a bug in Rust itself, but it would be a breaking change to fix so it's probably expected in Rust 2024... Regardless, it should be fixed. I'm not sure if we just want to make the module public or re-export the struct

conradludgate:
Fixes #672

### DIFF
--- a/tonic/src/transport/mod.rs
+++ b/tonic/src/transport/mod.rs
@@ -90,7 +90,7 @@ pub mod channel;
 pub mod server;
 
 mod error;
-mod service;
+pub mod service;
 mod tls;
 
 #[doc(inline)]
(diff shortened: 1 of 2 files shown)


davidpdrsn:
As discussed in https://github.com/hyperium/tonic/issues/672 this isn't quite the way we want to go. So I'm going to close this for now.

Issue #504, opened 2020-12-08
Support for adding attributes to generated *_client/*_server modules


tarcieri:
### Crates

`tonic_build`

### Motivation

We're using `tonic_build` to compile both proto and service definitions. However, we'd like to gate the service definitions on a cargo feature (e.g. `grpc`), so it's possible to use the protos independently of Tonic for use in e.g. WASM environments.

### Proposal

Add a new method to `tonic_builder::Builder` similar to the existing `field_attribute` and `type_attribute` methods, but instead: `module_attribute`.

This would make it possible to add an attribute like `#[cfg(feature = "grpc")]` to the generated modules to gate usages of Tonic.

### Alternatives

We managed to "solve" our immediate problem by [munging the output of `tonic_build`](https://github.com/cosmos/cosmos-rust/pull/8/files#diff-58406f56aa9701745b57c8af4adb067401ed9cb97ee7cad24938a905f8314f36R271-R281), however it's something of a brittle hack. A first-class feature for this would be much more robust.

davidpdrsn:
I might be wrong but wouldn't this feature require support for `module_attribute` in prost build?

davidpdrsn:
Fixed by https://github.com/hyperium/tonic/pull/684

PR #684, opened 2021-06-17
feat(build): support adding attributes to clients and servers

cab:
## Motivation

Implements #504.

Having the ability to add attributes to generated servers and clients allows things like feature gating or deriving additional service functionality without the need for a fork.

## Solution

This PR only changes `tonic-build`. Four new methods are added to `Builder`:

| method | description |
| -------- | ----------- |
| `server_mod_attribute` | Adds a new attribute to the `mod` wrapping the server implementation. |
| `server_attribute` | Adds a new attribute to the service server `struct`. |
| `client_mod_attribute` | Adds a new attribute to the `mod` wrapping the client implementation. |
| `client_attribute` | Adds a new attribute to the service client `struct`. |

These are all optional values. Calling them more than once is additive. 

Each method accepts a `pattern` and an `attribute`. `pattern` follows similar rules as `tonic`'s attribute matching rules.

If `pattern` matches for a given service, the respective `attribute` values are parsed as `syn::Attribute`s and added to the generated code in `server::generate` and `client::generate`.

### DIFF
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -1,4 +1,4 @@
-use super::{Method, Service};
+use super::{Attributes, Method, Service};
 use crate::{generate_doc_comments, naive_snake_case};
 use proc_macro2::TokenStream;
 use quote::{format_ident, quote};
@@ -12,6 +12,7 @@ pub fn generate<T: Service>(
     emit_package: bool,
     proto_path: &str,
     compile_well_known_types: bool,
+    attributes: &Attributes,
 ) -> TokenStream {
     let service_ident = quote::format_ident!("{}Client", service.name());
     let client_mod = quote::format_ident!("{}_client", naive_snake_case(&service.name()));
@@ -20,13 +21,26 @@ pub fn generate<T: Service>(
     let connect = generate_connect(&service_ident);
     let service_doc = generate_doc_comments(service.comment());
 
+    let package = if emit_package { service.package() } else { "" };
+    let path = format!(
+        "{}{}{}",
+        package,
+        if package.is_empty() { "" } else { "." },
+        service.identifier()
+    );
+
+    let mod_attributes = attributes.for_mod(package);
+    let struct_attributes = attributes.for_struct(&path);
+
     quote! {
         /// Generated client implementations.
+        #(#mod_attributes)*
         pub mod #client_mod {
             #![allow(unused_variables, dead_code, missing_docs)]
             use tonic::codegen::*;
 
             #service_doc
+            #(#struct_attributes)*
             #[derive(Debug, Clone)]
             pub struct #service_ident<T> {
                 inner: tonic::client::Grpc<T>,
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -155,6 +155,68 @@ pub trait Method {
     ) -> (TokenStream, TokenStream);
 }
 
+/// Attributes that will be added to `mod` and `struct` items.
+#[derive(Debug, Default, Clone)]
+pub struct Attributes {
+    /// `mod` attributes.
+    module: Vec<(String, String)>,
+    /// `struct` attributes.
+    structure: Vec<(String, String)>,
+}
+
+impl Attributes {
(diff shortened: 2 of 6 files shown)


tonic-build/src/lib.rs
```diff
@@ -218,6 +218,51 @@ fn generate_doc_comments<T: AsRef<str>>(comments: &[T]) -> TokenStream {
...
+// Checks whether a path pattern matches a given path.
+pub(crate) fn match_name(pattern: &str, path: &str) -> bool {
+    if pattern.is_empty() {
+        false
+    } else if pattern == "." {
+        true
+    } else if pattern == path {
+        true
+    } else {
+        let pattern_segments = pattern.split('.').collect::<Vec<_>>();
```
cab, line 247: this is based on  https://github.com/tokio-rs/prost/blob/5eddc35e3152bc9e1afb426481cd340d5919a239/prost-build/src/ident.rs#L62

tonic-build/src/client.rs
```diff
@@ -22,19 +23,34 @@ pub fn generate<T: Service>(
...
+    let path = format!(
+        "{}{}{}",
+        package,
+        if package.is_empty() { "" } else { "." },
+        service.identifier()
+    );
+
+    let mod_attributes = attributes.for_mod(package);
+    let struct_attributes = attributes.for_struct(&path);
+    let impl_attributes = attributes.for_impl(&path);
```
LucioFranco, line 36: is impl attributes actually useful?

tonic-build/src/prost.rs
```diff
@@ -388,3 +447,53 @@ impl Builder {
         Ok(())
     }
 }
+
+#[derive(Debug, Default, Clone)]
+pub struct Attributes {
```
LucioFranco, line 452: does this need to live in the `prost` module?

LucioFranco (approved):
LGTM, left a few questions.

tonic-build/src/prost.rs
cab: nope! do you have a preferred module? otherwise i'll move it to the crate root

tonic-build/src/client.rs
cab: the more i think about it, probably not -- any use case would likely be better served by writing a `prost_build::ServiceGenerator`

tonic-build/src/prost.rs
LucioFranco: lib rs is probably fine

tonic-build/src/client.rs
LucioFranco: I think we should just keep mod and struct ones and then I think i'm ok to merge this!

tonic-build/src/lib.rs
```diff
@@ -155,6 +155,50 @@ pub trait Method {
...
 }
 
+/// Attributes that will be added to `mod` and `struct` items. Should be pairs of (`pattern`, `attribute`).
+#[derive(Debug, Default, Clone)]
+pub struct Attributes {
+    /// `mod` attributes.
+    pub module: Vec<(String, String)>,
+    /// `struct` attributes.
+    pub structure: Vec<(String, String)>,
+}
```
davidpdrsn, line 165: Should we perhaps add `#[non_exhaustive]` so we can add new fields without breakage?

cab: good idea! adding now

davidpdrsn (approved):

tonic-build/src/lib.rs
```diff
@@ -155,6 +155,53 @@ pub trait Method {
...
 }
 
+/// Attributes that will be added to `mod` and `struct` items. Should be pairs of (`pattern`, `attribute`).
+#[non_exhaustive]
+#[derive(Debug, Default, Clone)]
+pub struct Attributes {
+    /// `mod` attributes.
+    pub module: Vec<(String, String)>,
+    /// `struct` attributes.
+    pub structure: Vec<(String, String)>,
```
LucioFranco, line 165: why do these need to be public?

LucioFranco (approved):
Great! Thank you for getting this over the finish line!

Issue #678, opened 2021-06-10
Authentication example is doesn't work


Dentosal:
### Version

Clean Tonic pulled from github master branch, at commit 4e0d2fc.

### Platform

WSL 1 on WIndows 10. (`4.4.0-19041-Microsoft #488-Microsoft x86_64 GNU/Linux`)

### Crates

Tonic itself, although I think it might be because Tower has changed.

### Description

I tried

```bash
cd examples
 cargo run --bin authentication-client
```

And it crashes:

```
thread 'main' panicked at 'buffer full; poll_ready must be called first', /home/user/.cargo/registry/src/github.com-1ecc6299db9ec823/tower-0.4.8/src/buffer/service.rs:146:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

davidpdrsn:
Thanks for noticing!

I did some digging and this is most likely because we don't propagate backpressure [here](https://github.com/hyperium/tonic/blob/master/tonic/src/service/interceptor.rs#L106), which leads to the client pancking since it contains a `tower::buffer::Buffer`. The fix is the replace that line with `self.inner.poll_ready(cx)`. I'll do that this week unless someone snatches it.

PR #685, opened 2021-06-19
fix(tonic): propagate call to `poll_ready` in `InterceptedService`

Szymongib:
Fixes: #678

## Motivation

As mentioned in the issue authentication example is broken as `poll_ready` is not called on [`tower::buffer::Buffer`](https://github.com/tower-rs/tower/blob/tower-0.4.8/tower/src/buffer/service.rs#L146).

## Solution

In `InterceptedService` propagate call to `poll_ready` to inner service.


davidpdrsn (approved):
Thanks!

PR #692, opened 2021-06-25
feat(tonic): compression support

davidpdrsn:
Don't be scared by the large number of lines added. Most of it is tests 😊 

This adds compression/decompression support to `tonic::server::Grpc` and `tonic::client::Grpc`.

Features:

- Supports transparently compressing/decompressing requests, responses, and streams. Only encoding supported so far is gzip but adding new ones is straight forward.
- Uses [flate2](https://crates.io/crates/flate2) but the compression backend can be changed without breaking changes.
- Generated clients and servers have some new methods for configuring which encodings are accepted/enabled. All these methods forward for `tonic::server::Grpc` or `tonic::client::Grpc` respectively:
    - `accept_gzip` (client): It enables receiving gzip compressed responses (sets `grpc-accept-encoding` accordingly).
    - `accept_gzip` (server): It enables receiving gzip compressed requests. If a server receives a request with an unsupported encoding it responds with `Unimplemented` as per the gRPC spec.
    - `send_gzip` (client): This will compress requests with gzip. Again server will respond with `Unimplemented` if that isn't supported/enabled.
    - `send_gzip` (server): This will compress responses if the client supports it.

## Example usage

```rust
// build server the supports receiving and sending gzip compressed content
let svc = test_server::TestServer::new(Svc).accept_gzip().send_gzip();

// spawn the server
let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
let addr = listener.local_addr().unwrap();
tokio::spawn(async move {
    Server::builder()
        .add_service(svc)
        .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(listener))
        .await
        .unwrap();
});

// create channel connected to the server
let channel = Channel::builder(format!("http://{}", addr).parse().unwrap())
    .connect()
    .await
    .unwrap();

// create client that also supports receiving and sending gzip compressed content
let mut client = test_client::TestClient::new(channel)
    .accept_gzip()
    .send_gzip();

// make a request as per usual
let response = client
    .compress_input_unary(SomeData {
        data: vec![1, 2, 3, 4],
    })
    .await
    .unwrap();
```

## Why not compress in the transport layer?

My original thought was to bake compression/decompression into the transport layer. So to enable compression on a client you'd call `accept_gzip` on the `Endpoint` (`Channel` builder), and for the server you'd call it on `Server`. However it turns out to not work great. The issue being that compression settings need to be funneled into `tonic::server::Grpc` and `tonic::client::Grpc` because thats where the actual protocol stuff happens.

The downside to that is that it requires new generated methods for client and server structs, which hurts discoverability. Hopefully some decent docs and an example will help with that.

## Outstanding issues

Compression is currently enabled or disabled globally for a client. It is not possible to configure for individual requests/responses/stream messages. Its easy enough for requests/responses, can be done with a new field on `Request`/`Response`, or with an extension. However since stream messages aren't wrapped in anything there is no place for tonic to add additional configuring hooks. I've considered a breaking change where streams are required to yield something like `tonic::StreamMessage<ActualMessage>` but not sure. What do you think @LucioFranco?

Edit: Compression on unary/client stream responses can now be disabled with `Response::disable_compression`. It doesn't do anything for server streams as that would require a config on each item in the stream.

## TODO

Besides the comments I've left for myself below:

- [x] Add example
- [x] When server receives request with encoding that isn't supported, include `grpc-accept-encoding` with which encodings are supported.
- [x] Test `bidirectional_stream::client_enabled_server_enabled` is flaky. It sometimes fails. Have to fix that.

---

Encoding/decoding parts based on https://github.com/hyperium/tonic/pull/476

Fixes https://github.com/hyperium/tonic/issues/282


tonic-build/src/client.rs
```diff
@@ -59,21 +58,19 @@ pub fn generate<T: Service>(
...
+                // TODO(david): docs
+                pub fn send_gzip(mut self) -> Self {
+                    self.inner = self.inner.send_gzip();
+                    self
                 }
-            }
 
-            impl<T> std::fmt::Debug for #service_ident<T> {
-                fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-                   write!(f, #struct_debug)
```
davidpdrsn, line 75: `tonic::client::Grpc` actually contains useful data now. So figured it makes sense to change this `Clone` and `Debug` impl to derives. Made sure the `Derive` doesn't require `T: Derive`.

tonic/src/client/grpc.rs
```diff
@@ -28,12 +31,30 @@ use std::fmt;
...
+    }
+
+    pub fn send_gzip(mut self) -> Self {
+        self.send_compression_encodings = Some(CompressionEncoding::Gzip);
+        self
+    }
+
+    pub fn accept_gzip(mut self) -> Self {
+        self.accept_compression_encodings.enable_gzip();
+        self
```
davidpdrsn, line 57: Note to self: Write some docs here.

tonic-build/src/client.rs
```diff
@@ -59,21 +58,19 @@ pub fn generate<T: Service>(
...
+                // TODO(david): docs
+                pub fn accept_gzip(mut self) -> Self {
+                    self.inner = self.inner.accept_gzip();
```
davidpdrsn, line 69: Note to self: Docs

tonic-build/src/server.rs
```diff
@@ -65,6 +71,18 @@ pub fn generate<T: Service>(
...
+                pub fn accept_gzip(mut self) -> Self {
+                    self.accept_compression_encodings.enable_gzip();
+                    self
+                }
+
+                // TODO(david): docs
+                pub fn send_gzip(mut self) -> Self {
+                    self.send_compression_encodings.enable_gzip();
+                    self
+                }
```
davidpdrsn, line 85: Note to self: Docs

tonic/src/codec/compression.rs
```diff
@@ -0,0 +1,161 @@
...
+) -> Result<(), std::io::Error> {
+    let capacity = ((len / BUFFER_SIZE) + 1) * BUFFER_SIZE;
+    out_buf.reserve(capacity);
+
+    match encoding {
+        CompressionEncoding::Gzip => {
+            let mut gzip_encoder = GzEncoder::new(
+                &decompressed_buf[0..len],
+                // FIXME: support customizing the compression level
+                flate2::Compression::new(6),
```
davidpdrsn, line 125: It would be nice to support customizing this. I'll look into that when this has been merged while I also look into adding support for more compression encodings.

tonic/src/codec/compression.rs
```diff
@@ -0,0 +1,161 @@
...
+            );
+            let mut out_writer = out_buf.writer();
+
+            tokio::task::block_in_place(|| std::io::copy(&mut gzip_encoder, &mut out_writer))?;
```
davidpdrsn, line 129: Have to use `block_in_place` here to avoid blocking the executor. Making `compress` and `decompress` async is be quite hard because of how `Streaming` works.

tonic/src/codec/decode.rs
```diff
@@ -156,19 +173,13 @@ impl<T> Streaming<T> {
 
     fn decode_chunk(&mut self) -> Result<Option<T>, Status> {
         if let State::ReadHeader = self.state {
-            if self.buf.remaining() < 5 {
+            if self.buf.remaining() < HEADER_SIZE {
```
davidpdrsn, line 176: Look a while to figure out what this hardcoded `5` was. Figured it made sense to extract into a `const`.

tonic/src/codec/decode.rs
```diff
@@ -156,19 +173,13 @@ impl<T> Streaming<T> {
...
             let is_compressed = match self.buf.get_u8() {
                 0 => false,
-                1 => {
-                    trace!("message compressed, compression not supported yet");
-                    return Err(Status::new(
-                        Code::Unimplemented,
-                        "Message compressed, compression not supported yet.".to_string(),
-                    ));
-                }
+                1 => true,
```
davidpdrsn, line 182: 🎊

tonic/src/codec/decode.rs
```diff
@@ -191,17 +202,45 @@ impl<T> Streaming<T> {
...
-            return match self
-                .decoder
-                .decode(&mut DecodeBuf::new(&mut self.buf, *len))
-            {
+            let decoding_result = if *compression {
+                self.decompress_buf.clear();
+
+                if let Err(err) = decompress(
+                    self.encoding.unwrap_or_else(|| {
+                        unreachable!("message was compressed but `Streaming.encoding` was `None`. This is a bug in Tonic. Please file an issue")
```
davidpdrsn, line 217: This should be caught by higher levels so this case should be unreachable 🤞

tonic/src/server/grpc.rs
```diff
@@ -29,7 +45,42 @@ where
...
+
+    pub fn accept_gzip(mut self) -> Self {
+        self.accept_compression_encodings.enable_gzip();
+        self
+    }
+
+    pub fn send_gzip(mut self) -> Self {
+        self.send_compression_encodings.enable_gzip();
+        self
+    }
```
davidpdrsn, line 63: Note to self: Docs

tests/compression/src/compressing_request.rs
```diff
@@ -0,0 +1,97 @@
...
+    assert_eq!(status.code(), tonic::Code::Unimplemented);
+    assert_eq!(
+        status.message(),
+        "Content is compressed with `gzip` which isn't supported"
+    );
+
+    // TODO(david): include header with which encodings are supported as per the spec:
+    //
+    // > The server will then include a grpc-accept-encoding response header which specifies the
+    //   algorithms that the server accepts.
```
davidpdrsn, line 96: Note to self: Fix this.

davidpdrsn:
@roblabla supporting custom compression algorithms was discussed a bit in https://github.com/hyperium/tonic/pull/476. Personally I don't feel its worth the compressing it brings. Other crates like reqwest and warp don't support custom algorithms either. If its something there is more demand for we can of course reconsider.

@LucioFranco What do you think?

mat-gas:
@davidpdrsn  I think there's a slight misunderstanding about what @roblabla said :
I think he actually wants to use AES **encryption** (and not compression, typo in his message) so it's not a custom compression algorithm he wants, but the hability to chain compression than encryption (because encrypting data then compressing is totally useless)

tonic/src/client/grpc.rs
```diff
@@ -28,12 +30,102 @@ use std::fmt;
...
+    pub fn send_gzip(mut self) -> Self {
+        self.send_compression_encodings = Some(CompressionEncoding::Gzip);
+        self
+    }
+
+    #[doc(hidden)]
+    #[cfg(not(feature = "compression"))]
+    pub fn send_gzip(self) -> Self {
+        panic!(
+            "`send_gzip` called on a server but the `compression` feature is not enabled on tonic"
```
LucioFranco, line 89: This mentions server but we are in the client? Also, wouldn't it be better to just have this not compile, aka just don't include a `not` version.

tonic/src/codec/compression.rs
```diff
@@ -0,0 +1,189 @@
...
+        } else {
+            return Ok(None);
+        };
+
+        let EnabledCompressionEncodings { gzip } = enabled_encodings;
+
+        match header_value_str {
+            "gzip" if gzip => Ok(Some(CompressionEncoding::Gzip)),
+            other => {
+                let mut status = Status::unimplemented(format!(
```
LucioFranco, line 84: is this the correct status other implementations return?

tonic/src/codec/compression.rs
LucioFranco: ugh I really don't like adding this as a dep basically doens't allow you to use compression when using `tokio::test`. Its not really doing any io could we get away without the block in place?

tonic/src/server/grpc.rs
```diff
@@ -95,13 +267,31 @@ where
         S: ClientStreamingService<T::Decode, Response = T::Encode>,
         B: Body + Send + Sync + 'static,
         B::Error: Into<crate::Error> + Send + 'static,
+        T: std::fmt::Debug,
```
LucioFranco, line 270: hmmmm this is new, why?

LucioFranco (commented):
LGTM overall just a few questions before I approve

tonic/src/client/grpc.rs
davidpdrsn: Yes not compiling would be better, and its also what I did initially, but it doesn't work with the ["Check features"](https://github.com/hyperium/tonic/blob/master/.github/workflows/CI.yml#L32) step in CI. It seems to test all combinations of features. It doesn't seem to be possible to tell it to exclude a set of features.

tonic/src/codec/compression.rs
davidpdrsn: Yes:

> If a client message is compressed by an algorithm that is not supported by a server, the message WILL result in an UNIMPLEMENTED error status on the server. The server will then include a grpc-accept-encoding response header which specifies the algorithms that the server accepts. If the client message is compressed using one of the algorithms from the grpc-accept-encoding header and an UNIMPLEMENTED error status is returned from the server, the cause of the error MUST NOT be related to compression. If a server sent data which is compressed by an algorithm that is not supported by the client, an INTERNAL error status will occur on the client side.

From https://github.com/grpc/grpc/blob/master/doc/compression.md

tonic/src/server/grpc.rs
davidpdrsn: Good catch! Was leftover from some debugging.

tonic/src/codec/compression.rs
davidpdrsn: I've removed it in 5458db1806b12.

LucioFranco (approved):

PR #710, opened 2021-07-09
feat(examples): add grpc-web example

davidpdrsn:


examples/src/grpc-web/client.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    println!("REPLY={:?}", reply);
+
+    Ok(())
+}
+
+// one byte for the compression flag plus four bytes for the length
+const GRPC_HEADER_SIZE: usize = 5;
+
+fn encode_body<T>(msg: T) -> Bytes
```
LucioFranco, line 45: what is this code for? Mocking grpc-web?

LucioFranco: I see we are working directly with h1 and hyper, nice.

LucioFranco (approved):

PR #714, opened 2021-07-11
feat(tonic): derive Clone for RouterService

cab:
## Motivation

Implementing `Clone` for `RouterService` makes it easier to use with `hyper::service::make_service_fn` and `tower::service_fn`. (For example, when using the pattern in [this example](https://github.com/hyperium/tonic/blob/master/examples/src/hyper_warp/server.rs#L54)).

## Solution

Derive `Clone` for `RouterService`.


davidpdrsn (approved):
LGTM! Thanks!

PR #721, opened 2021-07-12
Split path types in compile

aquarhead:
Same idea as https://github.com/tokio-rs/prost/pull/496 (it should explain both motivation and solution)

### DIFF
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -368,24 +368,22 @@ impl Builder {
     }
 
     /// Compile the .proto files and execute code generation.
-    pub fn compile<P>(self, protos: &[P], includes: &[P]) -> io::Result<()>
-    where
-        P: AsRef<Path>,
-    {
+    pub fn compile(
+        self,
+        protos: &[impl AsRef<Path>],
+        includes: &[impl AsRef<Path>],
+    ) -> io::Result<()> {
         self.compile_with_config(Config::new(), protos, includes)
     }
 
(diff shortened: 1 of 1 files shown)


davidpdrsn:
Could this be a breaking change?

aquarhead:
Yes, I believe it will break if `includes` was empty, like `&[]`. The workaround I used in `prost` is change to `&[""]`, but I'd like to know if there's a better solution.

davidpdrsn:
Ah right yeah. Can't think of a way around that 🤔 [`std::env::join_paths`](https://doc.rust-lang.org/stable/std/env/fn.join_paths.html) has the same issue, although that does allow you to use turbofish. I see thats also how you wrote it initially [here](https://github.com/tokio-rs/prost/pull/496#discussion_r664748885).

I personally not sure using `impl AsRef<Path>` is strictly better than regular generic bounds but I suppose an empty `includes` is quite a niche use-case, though I'm not sure. Never done it myself 🤷

aquarhead:
The problem is that both `protos` and `includes` are using the same generic parameter, so they need to be of same type to make the compiler happy, which can be unnecessary and manual type conversion code.

I guess _a_ solution is to provide another `compile_simple` function that does not take `includes`? Or just document this case?

davidpdrsn:
> Or just document this case?

I would be okay with that I think. @LucioFranco what do you think?

LucioFranco:
Yes, this is a breaking change so lets hold off on it until we go for 0.6. We can keep the PR open though since it will still work down the line.

PR #734, opened 2021-07-30
fix converting `google.protobuf.Empty` to `()` when compiling well-known protobuf types

tdyas:
## Motivation

https://github.com/hyperium/tonic/issues/521 covers the work for Tonic to compile the well-known protobuf types instead of relying on the `prost-types` crate. (One reason to do that is for projects that want `google.protobuf.Any` to use `Bytes` instead of `Vec<u8>` for binary data.) 

https://github.com/hyperium/tonic/pull/522 added initial support for Tonic to configure Prost and itself to compile the well-known protobuf types.

That support is incomplete, however: `tonic-build` fails to compile when configured to convert `google.protobuf.Empty` to `()`. This configuration occurs if the user calls `config.extern_path(".google.protobuf.Empty", "()")` where `config` is an instance of `prost_build::Config`. The error occurs because the existing code tries to interpret `()` as a Rust path and parsing, for example, `super::()` as a Rust path will always fail.

## Solution

This PR fixes the conversion of `google.protobuf.Empty` to `()` by specifically allowing the use of `()` as a valid Rust conversion without attempting to prepend any Rust path.

There are several stacked commits in this PR:
1. Update the `wellknown-compiled` integration test to configure the `google.protobuf.Empty` -> `()` conversion. This first commit fails to compile without the fix.
2. Refactor `tonic_build::prost::Service::request_response_name` to remove the duplication of logic between converting request and response types.
3. Fix the conversion by introducing an allow list of types (consisting only of `()` for now) and just generate a token stream in that case without trying to interpret as a Rust path.


tdyas:
Rebased cleanly and pushed.

Issue #755, opened 2021-09-02
Panic when enabling gzip compression and decompression on both sides


vbfox:
### Version

```
tonic v0.5.2
tonic-build v0.5.2
```

### Platform

MacOS client, Ubuntu 18.04.1 server

### Crates

`tonic`

### Description

When i'm connecting to a tonic server running with `.accept_gzip().send_gzip()` from a tonic client (Configured with `.accept_gzip().send_gzip()`) the following panic happens:

```
thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code: message was compressed but `Streaming.encoding` was `None`. This is a bug in Tonic.
```

vbfox:
I think there are 2 problems here:
* Setting `send_gzip()` doesn't send the expected header on clients
* Receiving a compressed message without any header shouldn't panic but respond as specified in https://grpc.github.io/grpc/core/md_doc_compression.html
  > An ill-constructed message with its Compressed-Flag bit set but lacking a grpc-encoding entry different from identity in its metadata MUST fail with INTERNAL status, its associated description indicating the invalid Compressed-Flag condition.

PR #763, opened 2021-09-12
compression: handle compression flag but no header

vbfox:
## Motivation

Due to some yet unknown reason I ended up with the crash described in #755 where the client set the compressed bit but doesn't send the correct header for the server for compression and the server panics as it reaches an `unreachable!` statement with:

```
thread 'tokio-runtime-worker' panicked at 'internal error: entered unreachable code: message was compressed but `Streaming.encoding` was `None`. This is a bug in Tonic.
```

Note: the client issue might be a problem with one of the configured interceptors in my codebase, to be investigated separately.

## Solution

The case where a compression flag is set but no compression header is present is checked while transitioning between the header and body phases and an internal error is returned in this case.

### DIFF
--- a/tests/compression/src/compressing_request.rs
+++ b/tests/compression/src/compressing_request.rs
@@ -87,3 +87,40 @@ async fn client_enabled_server_disabled() {
         "identity"
     );
 }
+
+#[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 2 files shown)


QuentinPerez:
hey @davidpdrsn 👋 

by chance, could you take a look on this patch ? we would like to move away from our patches/forks 🙂

PR #764, opened 2021-09-15
Updated server_streaming_echo example

utx0:
## Motivation

Just updated the example to include a basic stream as I found this hard to get my head around when I started trying to work with tonic.

### DIFF
--- a/examples/src/authentication/server.rs
+++ b/examples/src/authentication/server.rs
@@ -6,9 +6,9 @@ use futures::Stream;
 use pb::{EchoRequest, EchoResponse};
 use std::pin::Pin;
 use tonic::{metadata::MetadataValue, transport::Server, Request, Response, Status, Streaming};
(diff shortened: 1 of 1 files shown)


g0hl1n:
Just scrolled through the issues and came across this one.

@LucioFranco, I guess it can be closed as there is a correct streaming example in `examples/src/streaming`?

@utx0 is the streaming example somehow misleading/unclear for you?

utx0:
> Just scrolled through the issues and came across this one.
> 
> @LucioFranco, I guess it can be closed as there is a correct streaming example in `examples/src/streaming`?
> 
> @utx0 is the streaming example somehow misleading/unclear for you?

yes when I first came to tonic it was unclear how to do streaming. Mind you this is almost a year old so feel free to close if there is a better example now. I just found too much was implied for the user to already know.

PR #785, opened 2021-10-11
health, reflection: make rustfmt dependency optional

DavDave1:
## Motivation

Rust Docker images for automated builds don't include rustfmt, so building tonic-health or tonic-reflection crates requires adding a rustfmt installation step to build scripts.

## Solution

Make rustfmt requirement optional for tonic-health and tonic-reflection. 
This feature is kept enabled as default so that existing code won't break.


LucioFranco:
lol sorry for the noise totally botched the merge xD, thanks for the contribution!

PR #838, opened 2021-11-15
fix: Return error on non https uri instead of panic

djc:
## Motivation

In getting started with tonic, I ran into an issue where issuing a first request with the default features and no explicit TLS configuration set would just result in a H2 `GoAway` frame without further context. Given that I was providing a `https` URL to the `Endpoint`, it seems reasonable that tonic should be able to give me a more meaningful error in this case.

## Solution

If the URI scheme is `https` and the `tls` feature is disabled, return an error from the connector. If the feature is enabled but there is no TLS config present, also return an error.

### DIFF
--- a/tonic/src/transport/service/connector.rs
+++ b/tonic/src/transport/service/connector.rs
@@ -5,6 +5,7 @@ use super::tls::TlsConnector;
 use http::Uri;
 #[cfg(feature = "tls-roots-common")]
 use std::convert::TryInto;
+use std::fmt;
 use std::task::{Context, Poll};
 use tower::make::MakeConnection;
 use tower_service::Service;
@@ -78,9 +79,17 @@ where
         #[cfg(feature = "tls-roots-common")]
         let tls = self.tls_or_default(uri.scheme_str(), uri.host());
 
+        let is_https = uri.scheme_str() == Some("https");
         let connect = self.inner.make_connection(uri);
 
         Box::pin(async move {
+            #[cfg(not(feature = "tls"))]
+            {
+                if is_https {
+                    return Err(HttpsUriWithoutTlsSupport(()).into());
+                }
+            }
+
             let io = connect.await?;
 
             #[cfg(feature = "tls")]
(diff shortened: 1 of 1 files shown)


tonic/src/transport/service/connector.rs
```diff
@@ -80,16 +80,26 @@ where
...
         let tls = self.tls_or_default(uri.scheme_str(), uri.host());
 
+        let is_https = uri.scheme_str() == Some("https");
         let connect = self.inner.make_connection(uri);
 
         Box::pin(async move {
+            #[cfg(not(feature = "tls"))]
+            {
+                if is_https {
+                    return Err("connecting to HTTPS without TLS enabled".to_owned().into());
```
LucioFranco, line 90: Should we make this an actual error type?

djc: Seemed unlikely to me that code would be handling this, but I'm happy to change it if you prefer it the other way.

LucioFranco: Yeah lets make it a concrete error or error variant?

djc: Like this?

tonic/src/transport/service/connector.rs
```diff
@@ -78,20 +79,43 @@ where
...
             }
 
             Ok(BoxedIo::new(io))
         })
     }
 }
+
+/// Error returned when trying to connect to an HTTPS endpoint without TLS enabled.
+#[derive(Debug)]
+pub(crate) struct NoTlsSupport(());
```
LucioFranco, line 112: Probably should use a name like this https://github.com/hyperium/hyper-tls/blob/master/src/client.rs#L125 bit more informative than NoTlsSupport imo

LucioFranco (changes requested):
LGTM one nit

tonic/src/transport/service/connector.rs
djc: Changed the name to `HttpsUriWithoutTlsSupport`.

LucioFranco (approved):

guswynn:
This never panicked for me? The `GoAway` was definitely really hard to debug when I first was debugging, but this pr prevents me from using `tonic` 0.7 with a custom connector (`connect_with_connector` and friends) that uses tls through some other means (in my case, `tokio-native-tls`). I have shared how I have set this up in the discord, and I would be curious if @LucioFranco's openssl setup still works on 0.7

Issue #840, opened 2021-11-18
Content-Type missing in responses triggered by interceptors returning Status


NAlexPear:
## Bug Report

Services guarded by interceptors that return `Err(Status::*)` for any reason are missing `Content-Type` headers in the responses when queried with `grpcurl`.

### Version

Tested on the main `v0.6.1` release and the latest `master` branch

### Platform

`Linux plopear 5.14.16-arch1-1 x86_64 GNU/Linux`

### Crates

`tonic`

### Description

TL;DR: `Err(Status::*)` responses are handled differently when they occur in `Interceptor`s versus methods, leading `grpcurl` to complain that the `content-type` header is missing in `Interceptor`-generated responses.

Similar to #431, #700, and #759, but creating a new issue since this appears to be a broader case than a specific example not working (or represents a regression to the fix applied in #701).

I've created a repro of the issue [here](https://github.com/NAlexPear/tonic-interceptor-response-repro). This repro lets you configure different `FAILURE_MODE`s through environment variables: `None` means that there is no failure, `Method` means that the configured methods return a `Status`, and `Interceptor` means that the interceptor returns a `Status`. I'd expect `Method` and `Interceptor` failure modes to generate the exact same response, but they do not. Listed below are the tests I've run on each relevant failure mode.

**Failure Mode**: `Method`
**Run command** (in repro repo): `RUST_LOG=info FAILURE_MODE=method cargo run`
**Query command**: `grpcurl -plaintext -import-path ./proto -proto health.proto [::]:50051 grpc.health.v1.Health/Check`
**Query response**:
```
ERROR:
  Code: Internal
  Message: Failure within the Check method
``` 

**Failure Mode**: `Interceptor`
**Run command** (in repro repo): `RUST_LOG=info FAILURE_MODE=interceptor cargo run`
**Query command**: `grpcurl -plaintext -import-path ./proto -proto health.proto [::]:50051 grpc.health.v1.Health/Check`
**Query response**:
```
ERROR:
  Code: Unknown
  Message: OK: HTTP status code 200; transport: missing content-type field
``` 

The first is behaving as expected, the second is not.

### Next Steps:

I'll do a bit more digging to see how these requests are different, and I suspect that a fix like #701 will be in order. But want to make sure that this is floated as a possible regression on _all_ interceptors, rather than those used in a particular example.

NAlexPear:
[...]

Issue #759, opened 2021-09-06
authentication example incompatible with grpcurl and grpcui (when token doesn't match)


ebiggs:
### Version

fresh clone of master, running authentication example's server

### Platform

Darwin MacBook-Pro.local 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:31 PDT 2021; root:xnu-7195.141.2~5/RELEASE_X86_64 x86_64

### Description

When executing 
```
 grpcurl -v -plaintext -import-path . -proto echo.proto -H 'authorization: Bearer some-secret-token' -d '{ "message": "Hello World" }' localhost:50051 grpc.examples.echo.Echo.UnaryEcho
```

I get a happy result:
```
Resolved method descriptor:
// UnaryEcho is unary echo.
rpc UnaryEcho ( .grpc.examples.echo.EchoRequest ) returns ( .grpc.examples.echo.EchoResponse );

Request metadata to send:
authorization: Bearer some-secret-token

Response headers received:
content-type: application/grpc
date: Mon, 06 Sep 2021 00:15:30 GMT

Response contents:
{
  "message": "Hello World"
}

Response trailers received:
(empty)
Sent 1 request and received 1 response
```

However when I execute this, changing just the token
```
grpcurl -v -plaintext -import-path . -proto echo.proto -H 'authorization: Bearer not-a-secret' -d '{ "message": "Hello World" }' localhost:50051 grpc.examples.echo.Echo.UnaryEcho
```

I get this unexpected result:
```
Resolved method descriptor:
// UnaryEcho is unary echo.
rpc UnaryEcho ( .grpc.examples.echo.EchoRequest ) returns ( .grpc.examples.echo.EchoResponse );

Request metadata to send:
authorization: Bearer not-a-secret

Response headers received:
(empty)

Response trailers received:
(empty)
Sent 1 request and received 0 responses
ERROR:
  Code: Unknown
  Message: OK: HTTP status code 200; transport: missing content-type field
```

My expectations are to get an "unauthencated" status code and the message: "No valid auth token" [per line 66 of tonic/examples/src/authentication/server.rs](https://github.com/hyperium/tonic/blob/b30b1a1e2589dc609ed9dd20cfadefd4e82018b2/examples/src/authentication/server.rs#L66).

PR #842, opened 2021-11-19
fix: Handle interceptor errors as responses (#840)

NAlexPear:
## Summary

This PR forwards `Interceptor` errors/`Status`es as valid HTTP responses, just like any other gRPC service generated by `tonic`.

## Motivation

In the very specific case of missing Content-Type headers when requests are intercepted, this PR fixes the bugs reported in #840 and #759.

In the more general case, this class of bug arose because interceptors and their statuses are treated differently from the rest of the gRPC service stack: their errors/`Status` are bubbled up as errors to be handled at the tower service layer, but not as a gRPC response. That means that a basic HTTP response is returned, sans trailers or headers of any kind (including the missing `Content-Type` header). [See this repository](https://github.com/NAlexPear/tonic-interceptor-response-repro) and [this comment](https://github.com/hyperium/tonic/issues/840#issue-1057846258) to compare the responses generated by returning a `Status` from a method versus an interceptor.

This behavior means that clients like `grpcurl` that encounter the responses from interceptors don't know how to handle them, and it's likely that this affects any consumer that depends on the `Content-Type` header (among others).

## Solution

The solution in this PR is to handle `Status` errors the same way in both interceptors and regular ol' gRPC services: by converting those statuses to valid responses in the Interceptor service itself, matching the behavior in the existing `grpc` module. 

The downside to this approach is [a few additional constraints on the response body type](https://github.com/NAlexPear/tonic/blob/b8ae4daf23770ea829e4b861060a92ee3486f7a1/tonic/src/service/interceptor.rs#L220). These constraints seem reasonable given that they match up with the current uses of interceptors in `tonic`, but they do make wrapping non-`tonic` services trickier.

## Next Steps

EDIT: looks like those constraints are too restrictive for `tonic-health` :facepalm:. I'll take a look at either relaxing those constraints or updating `tonic-health` accordingly.

EDIT 2: I've both relaxed the constraints _and_ added a bit of restriction to clients in `tonic-build`. I'm not sure what the right mix between the two should be for a final implementation, but this does solve the issue for my use-case.

### DIFF
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -15,7 +15,7 @@ pub fn generate<T: Service>(
     attributes: &Attributes,
 ) -> TokenStream {
     let service_ident = quote::format_ident!("{}Client", service.name());
-    let client_mod = quote::format_ident!("{}_client", naive_snake_case(&service.name()));
+    let client_mod = quote::format_ident!("{}_client", naive_snake_case(service.name()));
     let methods = generate_methods(service, emit_package, proto_path, compile_well_known_types);
 
     let connect = generate_connect(&service_ident);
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):
LGTM!

okkero:
Looking forward to seeing this merged

olliebatch:
Looking forward to seeing this merged, Prevents the use of grpcCurl as it errors so will be useful!

davidpdrsn (approved):

PR #847, opened 2021-11-25
chore: revert to 2018 edition

davidpdrsn:
Fixes https://github.com/hyperium/tonic/issues/843

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ the generated code.
 
 ### Rust Version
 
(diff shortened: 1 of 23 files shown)


NateLing:
why？

davidpdrsn:
Because tonic is used in libraries that don’t want to require the latest version of Rust. Such as opentelemetry.

olix0r (approved):

PR #853, opened 2021-12-01
Add json-codec for application/grpc+json

zbysir:
## Motivation

A [`Codec`] that implements `application/grpc+json` via the serde library.

### DIFF
--- a/tonic/src/codec/json.rs
+++ b/tonic/src/codec/json.rs
@@ -60,7 +60,7 @@ impl<T, U> Default for JsonCodec<T, U> {
 impl<T, U> Codec for JsonCodec<T, U>
     where
         T: serde::Serialize + Send + 'static,
-        U: serde::de::DeserializeOwned + Send + Default + 'static,
+        U: serde::de::DeserializeOwned + Send + 'static,
 {
     type Encode = T;
     type Decode = U;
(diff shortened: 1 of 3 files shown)


olix0r:
This introduces dependencies that are only needed when this extra functionality is desired. Can it be feature-gated?

zbysir:
> This introduces dependencies that are only needed when this extra functionality is desired. Can it be feature-gated?

Good idea, I will try to add ‘feature-gated’

zbysir:
> Make json-codec be feature-gated


> This introduces dependencies that are only needed when this extra functionality is desired. Can it be feature-gated?

I did it, please review it when you have time, thank you for your help.

tonic/src/codec/json.rs
```diff
@@ -0,0 +1,76 @@
...
+        let s = serde_json::to_string(&item).expect("Message only errors if not enough space");
+        buf.put(s.as_bytes());
+
+        Ok(())
+    }
+}
+
+pub struct JsonDecoder<U>(PhantomData<U>);
+
+impl<U: for<'a> serde::Deserialize<'a>> Decoder for JsonDecoder<U> {
```
LucioFranco, line 26: Can we use this https://docs.rs/serde/latest/serde/de/trait.DeserializeOwned.html should be the same just a bit cleaner

LucioFranco (changes requested):
I would be happy to include this. But I think first I would like this PR to have an example and some tests. Then we can merge it.

zbysir:
> I would be happy to include this. But I think first I would like this PR to have an example and some tests. Then we can merge it.

Okay, I will add some tests and example later.

zbysir:
Hello?

LucioFranco (changes requested):
Sorry for the delay. I think I would like to merge this on the condition that we add the change for build and add an example + tests (I think you added tests so that is good!).

PR #858, opened 2021-12-06
chore: fix warnings around unused TLS code

djc:
## Motivation

Fix warnings about unused fields.

## Solution

Remove a type that seems to be unused.


davidpdrsn:
Thanks!

PR #859, opened 2021-12-06
feat(tls): upgrade to tokio-rustls 0.23 (rustls 0.20)

djc:


LucioFranco:
FYI this is will require a breaking release because we accept the client/server config from rustls in an api.

LucioFranco (approved):
Thanks for doing this!

djc:
What's the planning for 0.7? Would be nice not to carry around duplicate deps for this.

djc:
As discussed a bit on Discord, I added a commit that removes API surface referencing specific rustls types so that hopefully a future rustls upgrade won't need to be a semver-incompatible version bump.

Would still appreciate some context on planning for 0.7, since I'd like to make some changes to opentelemetry-stackdriver that would pull in rustls 0.20 -- but that crate also depends on tonic, and it would be nice not to depend on two versions of rustls/webpki.

LucioFranco (approved):

tiziano88:
Is it still possible to pass in a custom rustls ClientConfig now?

tiziano88:
Thanks, that would be great! We would definitely need this, we use a custom cert validator, and it seems after this commit there is no way of using it in tonic.

https://github.com/project-oak/oak/blob/a34217684954466983636ea09bb59c99759591e1/oak_client/src/attestation.rs#L122-L153

cc @ipetr0v @conradgrobler

djc:
Can you file a separate issue so we can be sure to track your use case? I have some ideas but will need to investigate in more detail.

exFalso:
Hi, we're in the process of upgrading from tonic 0.6 to 0.8, and the direct rustls configuration seems to have been removed in this PR, and TLS configuration is now done completely differently. We'll try to update to the new way of configuring, but I think it'd be nice if there was still a way to configure rustls directly with a `rustls::ClientConfig`, as the new configuration options are quite limited and may not fit all use cases. Cheers!

PR #861, opened 2021-12-08
feat(tonic): add unix socket support in server

agreen17:
Addresses https://github.com/hyperium/tonic/issues/856

Adds `tonic::transport::server::UdsConnectInfo` to facilitate consumers using a unix domain socket on the server side. Updates the existing `uds` example to use this `UdsConnectInfo` and tokio's `UnixListenerStream` as the latter cuts down on consumer boilerplate. Adds an integration test for `UdsConnectInfo`.

## Motivation

Currently using a unix socket requires consumers to provide an implementation of the `tonic::transport::server::Connected` trait (`UdsConnectInfo`).

When I started using `tonic`, I copy+pasted this stuff from the `uds` example into my application. I bet other users have done this as well, and while this is not the end of the world, it would be nice if unix sockets worked out of the box with a `tonic` server.

This PR aims to allow consumers to more easily use a `tokio::net::UnixListener` with a tonic server.

## Solution

The `UdsConnectInfo` implementation here is pulled directly from the existing `uds` example. I pulled `UnixListenerStream` into the `uds` example to remove some boilerplate code there.


tests/integration_tests/tests/connect_info.rs
```diff
@@ -48,3 +48,80 @@ async fn getting_connect_info() {
...
+            Ok(Response::new(Output {}))
+        }
+    }
+
+    #[tokio::test]
+    async fn getting_connect_info() {
+        let unix_socket_path = "/tmp/tonic/uds-integration-test";
+        tokio::fs::create_dir_all(Path::new(unix_socket_path).parent().unwrap())
+            .await
+            .unwrap();
```
LucioFranco, line 92: Would it make sense to use https://doc.rust-lang.org/stable/std/env/fn.temp_dir.html and not have to create a dir then we can remove the need to add the `fs` stuff.

tonic/src/transport/server/conn.rs
```diff
@@ -98,6 +98,36 @@ impl Connected for TcpStream {
...
 
+/// Connection info for Unix domain socket streams.
+///
+/// This type will be accessible through [request extensions][ext] if you're using
+/// a unix stream.
+///
+/// See [`Connected`] for more details.
+///
+/// [ext]: crate::Request::extensions
+#[cfg(unix)]
```
LucioFranco, line 109: Maybe we can just move this to a `unix.rs` file and `cfg` that whole file?

LucioFranco (commented):
Left a few comments but I like this idea, we should have probably added this a while ago.

tests/integration_tests/tests/connect_info.rs
agreen17: Sounds reasonable. Only small issue there is that tokio `UnixListener` needs the path to the unix socket to exist on instantiation, but _does not clean it up_ when dropped. So, we need to remove the file at the end of the test or else we won't be able to run the integration tests multiple times (subsequent runs will fail with `AddrInUse`).

Seems like you want to remove the added `tokio::fs` dep, so I should be able to use `std::env::temp_dir` and just do a blocking rm at the end of the test `std::fs::remove_file` if that sounds good to you?

tonic/src/transport/server/conn.rs
agreen17: Sounds good to me 👍🏽

agreen17:
> Left a few comments but I like this idea, we should have probably added this a while ago.

Great to hear - his simplifies our server usage a bit and hopefully it will do the same for others! I've addressed your comments, let me know what you think

agreen17:
Oops, looks like I missed a `#[cfg(unix)]` in the UDS example and that broke non-Unix builds. Just pushed a fixup to address that

tests/integration_tests/tests/connect_info.rs
steven-joruk: An option would be use [pair ](https://docs.rs/tokio/latest/tokio/net/struct.UnixStream.html#method.pair) to avoid the filesystem altogether. You'd have to convert one of them to a `UnixListener` though, maybe through a raw fd.

ShaneEverittM:
Is this gonna land in the next release?

ShaneEverittM:
Also, does this support abstract sockets?

examples/src/uds/server.rs
```diff
@@ -26,8 +29,13 @@ impl Greeter for MyGreeter {
...
         #[cfg(unix)]
         {
-            let conn_info = request.extensions().get::<unix::UdsConnectInfo>().unwrap();
+            let conn_info = request.extensions().get::<UdsConnectInfo>().unwrap();
             println!("Got a request {:?} with info {:?}", request, conn_info);
+
+            // Client-side unix sockets are unnamed.
+            assert!(conn_info.peer_addr.as_ref().unwrap().is_unnamed());
+            // This should contain process credentials for the client socket.
+            assert!(conn_info.peer_cred.as_ref().is_some());
```
LucioFranco, line 38: Probably makes sense to have these as doc items and not have asserts in an example, what do you think?

LucioFranco (commented):
Overall LGTM just would like to see the asserts removed.

examples/src/uds/server.rs
agreen17: Sounds good to me 👍🏽 there's already sufficient documentation on these fields in the `UdsConnectInfo` struct, so I'll just go ahead and remove these assertions - the info will be printed by the statement just above anyways.

LucioFranco: Yup that works for me then we can merge! Thank you for the patience on this!

agreen17: @LucioFranco For sure! Thanks for taking the time to review.

Fixup commits have been squashed - hopefully we're good to merge now 👀

LucioFranco (approved):
LGTM! Thanks!

PR #871, opened 2021-12-24
Fix typo in the streaming example.

jimmykiang:
## Motivation
Typo in one of the comments from the streaming example.

https://github.com/hyperium/tonic/blob/master/examples/src/streaming/server.rs#L47

## Solution
Fixed typo.


davidpdrsn:
Thanks!

PR #873, opened 2021-12-28
chore: Update tracing-subscriber to 0.3

tottoto:
## Motivation

Update tracing-subscriber to version 0.3.

## Solution

Update tracing-subscriber version to 0.3 and add `env-filter` that is removed from default features in this version. `interop` and `tests/integration_tests` use [`tracing_subscriber::fmt::SubscriberBuilder::with_env_filter`](https://docs.rs/tracing-subscriber/0.3.4/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_env_filter) and [`tracing_subscriber::EnvFilter`](https://docs.rs/tracing-subscriber/0.3.4/tracing_subscriber/struct.EnvFilter.html), but `examples` does not use `env-filter` feature because `tracing_subscriber::fmt()` returns `tracing_subscriber::fmt::SubscriberBuilder` and its `init` method does not use `env-filter` functionality.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -194,7 +194,7 @@ serde_json = "1.0"
 tracing = "0.1.16"
 tracing-attributes = "0.1"
 tracing-futures = "0.2"
-tracing-subscriber = {version = "0.2", features = ["tracing-log"]}
+tracing-subscriber = {version = "0.3", features = ["tracing-log"]}
 # Required for wellknown types
 prost-types = "0.9"
 # Hyper example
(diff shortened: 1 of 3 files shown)


olix0r:
@hawkw is this backwards-compatible? i.e. if an application is using tracing-subscriber 0.2, can tonic be used with t-s 0.3?

davidpdrsn:
I think this only touches tests. iirc tonic doesn’t use tracing-subscriber outside of dev builds.

hawkw:
> @hawkw is this backwards-compatible? i.e. if an application is using tracing-subscriber 0.2, can tonic be used with t-s 0.3?

Yeah, I believe `tonic` only uses `tracing-subscriber` in tests and examples; it's not a dependency of the `tonic` crate itself. So this shouldn't effect crates that depend on `tonic` at all.

hawkw (approved):

Issue #879, opened 2022-01-02
Tonic does not handle `grpc-encoding: identity` when compression is enabled


tzilist:
### Version
0.6.2

### Platform
Unix (Darwin and Debian)

### Description
When receiving a `grpc-encoding` header that is i.e. `grpc-encoding: identity`, Tonic fails to uncompressed the response and returns an 
```rust
Err(
    Status {
        code: Unimplemented,
        message: "Content is compressed with `identity` which isn't supported",
        metadata: MetadataMap {
            headers: {
                "grpc-accept-encoding": "gzip,identity",
            },
        },
        source: None,
    },
)
```

I believe this is coming from here: https://github.com/hyperium/tonic/blob/0583cff80f57ba071295416ee8828c3430851d0d/tonic/src/codec/compression.rs#L63-L100

Ideally, this code should ignore compression if it is enabled and the `identity` value is set for `grpc-encoding`

As a note for testing this: we are currently using the [GCP PubSub emulator ](https://cloud.google.com/pubsub/docs/emulator#prereq) to do some testing. The `grpc-encoding` header sent back is `identity` and this is causing an issue for our code if we have compression enabled.

matze:
Is there anything blocking from reviewing the change in the PR? We have a similar issue with a Java server who's responses are encoded with `identity` and fail to be delivered.

davidpdrsn:
@matze its merged 😊

matze:
Much appreciated, thanks David!

PR #880, opened 2022-01-02
Add `identity` to `from_encoding_header`

tzilist:
This change is for #879 . It allows for `identity` to be sent back as a `grpc-encoding` header as it requires no decompression

## Motivation
Some servers return `identity` as a `grpc-encoding` header

## Solution
Allow the `identity` value of `grpc-encoding` to fall back to no compression


davidpdrsn (approved):

davidpdrsn:
Thanks!

PR #882, opened 2022-01-03
chore(interop): Change structopt to clap

tottoto:
## Motivation

Change structopt to clap version 3.

## Solution

Change structopt to clap version 3.

### DIFF
--- a/interop/Cargo.toml
+++ b/interop/Cargo.toml
@@ -17,6 +17,7 @@ path = "src/bin/server.rs"
 [dependencies]
 async-stream = "0.3"
 bytes = "1.0"
+clap = {version = "3", features = ["derive"]}
 console = "0.14"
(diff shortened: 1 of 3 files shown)


djc (approved):
(Unofficial review:) LGTM, thanks!

LucioFranco:
Did something happen to structopt? Whats the motivation here?

djc:
clap 3 came out which brings structopt (which used clap under the hood) into the clap crate. So this is basically a massively modernized version of structopt, which works the same but has a bunch of refinements and also relies on clap 3 under the hood instead of clap 2. AFAICT structopt/clap 2 are a dead end.

davidpdrsn:
Yep! At embark we've moved all our internal stuff to clap as well.

LucioFranco (approved):

PR #883, opened 2022-01-04
fix(tonic): Expose h2 error instead of reason

roblabla:
## Motivation

Currently, when an error occurs at the H2 layer, it is transformed into a `tonic::Status` with a `source()` containing only the H2 error reason, instead of the full error. This loses a ton of context, as it is now impossible to know if the failure was caused by a protocol error, a general IO error (and if so, what what IO error), or something else entirely. It makes it much harder to debug problems.

It is unclear why this choice was made in #612. There's talk about `Clone` implementations causing this to be necessary, but I can't find that implementation.

## Solution

This MR simply bubbles up the underlying `h2::Error`. `tonic::Status::source` will contain an `h2::Error` instead of an `h2::Reason`.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -325,7 +325,7 @@ impl Status {
         #[cfg(feature = "transport")]
         let err = match err.downcast::<h2::Error>() {
             Ok(h2) => {
-                return Ok(Status::from_h2_error(&*h2));
+                return Ok(Status::from_h2_error(h2));
             }
             Err(err) => err,
         };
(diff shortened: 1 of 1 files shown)


davidpdrsn:
iirc `Status` only contained the h2 error reason, since `Status` needed to be `Clone`. However when we removed that requirement in https://github.com/hyperium/tonic/pull/612 I think we just forgot to store the whole error instead of only the reason.

I think this change makes sense but its probably a breaking change since source downcasts will have to change.

LucioFranco (approved):

PR #884, opened 2022-01-09
Enhancement of streaming examples

super-horst:
I worked through gRPC streaming with tonic and discovered that the respective examples are incomplete.

Changes:
- server implementation for 3 streaming scenarios (server-, client-, bidirectional-streaming)
- corresponding client implementations in different modules

Open questions:
- Is there a better way to structure the clients?
- Is [this](https://github.com/super-horst/tonic/blob/54f9df5221d2a2e18fded7601ca47fe4779e4b46/examples/src/streaming/server.rs#L123) use of Pin correct? I haven't needed to pin anything before so I'm not sure.


LucioFranco:
Hi thanks for making this change. I would love to improve the examples but I would rather keep them all in one file rather than multiple. what do you think?

super-horst:
I combined all clients into a single file, similar to the client example in _routeguide_

examples/src/streaming/client.rs
```diff
@@ -2,28 +2,120 @@ pub mod pb {
...
+        if self.counter >= self.count_to {
+            return Poll::Pending;
+        }
+
+        let message = format!("foo {}", self.counter);
+        self.get_mut().counter += 1;
+
+        return Poll::Ready(Some(EchoRequest { message }));
+    }
+}
```
LucioFranco, line 40: Why this over using `async-stream`?

rohitjoshi:
Can you please also include an example for an interceptor with streaming?

2escapereality:
An example with long lived client and server bidirectional streams would be great.

PR #894, opened 2022-01-18
fix(codec): Remove `Default` bound on `Codec`

andrewhickman:
## Motivation

I implemented a custom codec to use with a dynamic method where the types are not known at compile time. However this bound is mildly annoying since there's no sensible way to implement default
https://github.com/andrewhickman/grpc-client/blob/487c5409e78afef200a064c81ae1ade1b37c3c28/src/grpc/codec.rs#L63

## Solution

The bound is never used so we can just remove it. I think this is technically a breaking change for consumers of the `Codec` trait, however anything using `ProstCodec`, like generated code, still works.

### DIFF
--- a/tonic/src/codec/mod.rs
+++ b/tonic/src/codec/mod.rs
@@ -33,7 +33,7 @@ const HEADER_SIZE: usize =
     std::mem::size_of::<u32>();
 
 /// Trait that knows how to encode and decode gRPC messages.
-pub trait Codec: Default {
+pub trait Codec {
     /// The encodable message.
     type Encode: Send + 'static;
     /// The decodable message.


LucioFranco:
@davidpdrsn I forget did you remove the need for default when you changed some of the compression stuff?

davidpdrsn:
Hm no I don't think so.

I tried bisecting it but had to go so far back that I hit

```
~/de/ma/tonic [HEAD @ 6dbe88d445e37 * bisect]
❯ cargo c --all-features --tests -p tonic
    Updating crates.io index
    Updating git repository `https://github.com/tower-rs/tower`
error: no matching package named `tower-balance` found
location searched: https://github.com/tower-rs/tower
required by package `tonic v0.1.0-alpha.6`
```

Seems fine to me to the `Default` requirement.

PR #895, opened 2022-01-18
Expose `Or` and `Routes`

Miaxos:
## Motivation

Linked to https://github.com/hyperium/tonic/issues/886
https://github.com/hyperium/tonic/issues/886#issuecomment-1012221009

When you manipulate the `Router` struct to create tooling around in another crate, you'll need these two privates types. 

https://github.com/hyperium/tonic/blob/8cc65a696ca95b8795256839dc1e95dddbe3cde8/tonic/src/transport/server/mod.rs#L92 

For instance when you want to implement a trait like this

```rust
impl<C, A, B, L> SomeTrait<C> for GRPCDriver<C, Router<A, B, L>> {
  ...
}
```

## Solution

Just expose them.

### DIFF
--- a/tonic/src/transport/mod.rs
+++ b/tonic/src/transport/mod.rs
@@ -98,6 +98,10 @@ pub use self::channel::{Channel, Endpoint};
 pub use self::error::Error;
 #[doc(inline)]
 pub use self::server::{NamedService, Server};
+#[doc(hidden)]
+pub use self::service::Or;
+#[doc(hidden)]
+pub use self::service::Routes;
 #[doc(inline)]
 pub use self::service::TimeoutExpired;
 pub use self::tls::Certificate;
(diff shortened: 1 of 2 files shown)


davidpdrsn:
I believe we've discussed this before but I cannot find that discussion 🤷 

However these types are intentionally private because they're likely to be removed in the future, namely as part of https://github.com/hyperium/tonic/pull/830. I would suggest you use generics (if possible) or [`tower::util::BoxCloneService`](https://docs.rs/tower/latest/tower/util/struct.BoxCloneService.html) to erase the types.

Miaxos:
All right! Good to know for https://github.com/hyperium/tonic/pull/830, it's going to break some of my code! Yeah I tried with generics but it's doesn't work for that case, I didn't know about BoxCloneService I'll try.

PR #897, opened 2022-01-23
tonic-health: Implement overall server health as described in protocol standard

snowpoke:
This PR adds a service of empty name (`""`) to each Health Reporter, which will always return a `SERVING` status.

## Motivation
The description of the GRPC Health Checking Protocol includes the following line:
> The server should use an empty string as the key for server's overall health status, so that a client not interested in a specific service can query the server's status with an empty request. 
(https://github.com/grpc/grpc/blob/master/doc/health-checking.md)

At the moment, this behavior is not implemented in the `tonic-health` crate. In particular, this can cause problems when running health checkers in their default settings.
For example, if you try to run Google's GRPC Health Probe script (https://github.com/grpc-ecosystem/grpc-health-probe) like described in the example, it will cause an unexpected error:
```
error: health rpc failed: rpc error: code = NotFound desc = service not registered
```
The expected behavior would be to return `status: SERVING`. 

## Solution
My PR changes `HealthReporter::new` such that on creation, we register a service `""` and associate it with `ServingStatus::Serving`. Instead of `HashMap::new()`, a HashMap based on this service is used when the HealthReporter is initiated.
I also expanded the tests to check for this particular behavior.

### DIFF
--- a/tonic-health/src/server.rs
+++ b/tonic-health/src/server.rs
@@ -250,6 +250,21 @@ mod tests {
     async fn test_service_watch() {
         let (mut reporter, service) = make_test_service().await;
 
+        // Overall server health
+        let resp = service
+            .watch(Request::new(HealthCheckRequest {
+                service: "".to_string(),
+            }))
+            .await;
+        assert!(resp.is_ok());
+        let mut resp = resp.unwrap().into_inner();
+        let item = resp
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Good catch! Thanks for the PR!

Issue #900, opened 2022-01-30
Inconsistent Server Builder api - timeout func returns &mut Self instead of Self


jdahlq:
### Crates

tonic

### Motivation

Similar functions in the Builder return `Self` rather than `&mut Self` enabling move semantics, but `timeout` doesn't, leading to quirky behavior.

```rust
// This compiles because `add_service` takes a `&mut Self` and clones it.
let router = Server::builder()
  .accept_http1(true)
  .timeout(Duration::from_secs(10))
  .add_service(my_svc);

// This does NOT compile because `accept_http1` takes a `Self` but cannot move or copy it.
let router = Server::builder()
  .timeout(Duration::from_secs(10))
  .accept_http1(true)
  .add_service(my_svc);
```

Issue #115 changed most of these functions to return `Self`, but `timeout` was (re-)introduced in PR #199 and (presumably unintentionally) uses the old behavior of returning `&mut Self`.

### Proposal

Make `timeout` consistent with other funcs like `accept_http1`, `concurrency_limit_per_connection`, etc.

jdahlq:
TODO: Update this PR to avoid the CORS issue described in Issue #911 once PR #912 is submitted.

LucioFranco:
Nice catch! Yeah, we should get this fixed :)

PR #901, opened 2022-01-30
fix(transport): Make server builder more consitient

jdahlq:
## Motivation

See Issue #900 

## Solution

See Issue #900

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -191,12 +191,14 @@ impl<L> Server<L> {
     /// # use tonic::transport::Server;
     /// # use tower_service::Service;
     /// # use std::time::Duration;
-    /// # let mut builder = Server::builder();
+    /// # let builder = Server::builder();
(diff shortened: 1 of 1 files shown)


jdahlq:
I guess this might cause a silent breakage for code like this:

```rust
let router = Server::builder();
router.timeout(Duration::from_secs(10));
```

`router.timeout` would no longer do anything if not chained or reassigned. We could add a `#[must_use]` annotation (and probably should add it for all these functions which return Self), which presumably would cause a compiler error for users with such code. Not sure what the best course of action is, here.

LucioFranco:
Yeah I think adding must use makes sense here then. If you want to go ahead and add that change here as well we can merge this.

jdahlq:
Done.

LucioFranco:
Can you rebase against master? I fixed the ci failure https://github.com/hyperium/tonic/pull/914

jdahlq:
Thanks, rebased.

LucioFranco (approved):

Issue #890, opened 2022-01-14
build: Switch to prettyprinter for formating


LucioFranco:
https://github.com/dtolnay/prettyplease

PR #904, opened 2022-02-02
feat(build): use prettyplease to format output (#890)

matze:
Use prettyplease to format the output unconditionally and drop the optional rustfmt feature.

## Motivation

Having to ask people to install `rustfmt` to avoid spurious build errors was annoying. prettyplease can be installed as a regular dependency and format the code without having to shell out. Moreover this feature was asked for in #890.

## Solution

Add prettyplease 0.1.x and use it to format the token stream unconditionally. The few protos I tested were indistinguishable from having them formatted with rustfmt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,6 @@ the generated code.
 
 ```bash
 $ rustup update
-$ rustup component add rustfmt
 $ cargo build
 ```
 
--- a/examples/helloworld-tutorial.md
+++ b/examples/helloworld-tutorial.md
@@ -26,7 +26,6 @@ feature.
 
 ```bash
 $ rustup update
-$ rustup component add rustfmt
 ```
 
 ## Defining the HelloWorld service
--- a/examples/routeguide-tutorial.md
+++ b/examples/routeguide-tutorial.md
@@ -47,12 +47,6 @@ Change your current directory to Tonic's repository root:
 $ cd tonic
 ```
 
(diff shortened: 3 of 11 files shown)


davidpdrsn:
Hot take: Why does tonic have to format the generated code at all? Can't users do that, if necessary?

matze:
Certainly possible but I see myself freaked out, first time I jump into unformatted, computer-generated code. And I definitely do sometimes browse the code.

davidpdrsn:
You mean after opening it from within the target directory?

I normally format the code in the rust playground if I'm just interested in peeking at it. Otherwise I browse the generated docs.

matze:
> You mean after opening it from within the target directory?

Kind of. rust-analyzer let's me jump to any piece of code no matter if the current crate, some dependency or generated code like the tonic's.

davidpdrsn:
Ah I see. That doesn't work for me for some reason and it just jumps to the place where I include the generated code 🤷‍♂️

I think formatting the code is still worth it especially given how easy with prettyplease. I was just wondering about your use case.

LucioFranco:
@davidpdrsn originally when I was developing tonic I ran into compiler errors and what the compiler does is selects the line to print. When all the code is not formatted it fits onto one line and you can probably guess what the errors looked like xD

LucioFranco:
@matze could you post a copy of what the formatted code looks like?

LucioFranco (approved):
Thanks for adding this!

Issue #870, opened 2021-12-23
Asynchronous interceptor


thekeenant:
I'm working with Tonic (+tokio) trying to introduce authentication/authorization for my gRPC server. I can see a simple example of this [here](https://github.com/hyperium/tonic/blob/master/examples/src/authentication/server.rs). However, I think auth would typically involve dispatching some additional request (e.g. http, rpc, or some database request) to make the decision to return Ok or Err (to check db for presence of some token, or username/password, etc.)

Is it possible to call such an async function from a Tonic interceptor? Or maybe it's necessary to use Tower to accomplish this and if that's the case, any plans to make regular Tonic interceptors async capable?

davidpdrsn:
Async interceptors currently requires writing a tower service middleware.

Several people have asked about it so probably does make sense to have a convenient API for 😊

jdahlq:
FYI I have published v0.1.0 of [tonic-async-interceptor](https://crates.io/crates/tonic-async-interceptor) which uses tonic v0.8.1. Hope it is useful!

liufuyang:
@jdahlq Do you know if it is possible to also use your tonic-async-interceptor for a client/Channel?  For now we have an [Interceptor](https://github.com/liufuyang/bigtable_rs/blob/154caed2a12a7deba282732bcb22d9d027d0f509/bigtable_rs/src/bigtable.rs#L426) and used it on a client with the generated [.with_interceptor() method](https://github.com/liufuyang/bigtable_rs/blob/154caed2a12a7deba282732bcb22d9d027d0f509/bigtable_rs/src/google/google.bigtable.v2.rs#L1072)

Later we might need to switch the interceptor with [some auth manager](https://github.com/hrvolapeter/gcp_auth/blob/a3edad5f7c2f0cc2c3e8d2b8c0cbcbbfa43cdc95/examples/simple.rs) that seems only providing an async API. But now I am not sure how we can switch it to an async interceptor. (Or perhaps there are other ways to do this?) cc @LucioFranco

talksik:
is this not supported? I feel like this is a fundamental feature for allowing production rust grpc

rthomas:
[...]

PR #910, opened 2022-02-11
feat(tonic): introduce async_interceptor

jdahlq:
## Motivation

See Issue #870.

Tonic server provides `interceptor` for easy creation of a middleware layer from a closure. This `interceptor` can inspect and transform the headers and metadata of a request, as well as rejecting it by returning a status.

This PR introduces `async_interceptor`, a version of `interceptor` that accepts async functions and closures, allowing for the creation of more powerful middleware without the usual Tower boilerplate.

## Solution

The code for `interceptor` was duplicated and adapted for use with an async interceptor. I was initially hoping to reuse most of the functions which support `interceptor`, but it would have required heavy modification of those functions and possibly less performant code, so ultimately I decided it was better to leave that code mostly untouched.

I found it challenging to write clean code without using async/await or FutureExt functions like `then` which chain futures together. As far as I can tell, those features would have required boxing the returned futures, and I'm assuming that we want to keep the futures in the core library on the stack, which this PR manages to do.

I also added some tests and updated the Tower example.

I'm still rather new to Rust (coming from a mostly Java and C++ background), so I'd really appreciate advice on code style. Thanks!

### DIFF
--- a/tonic/src/service/interceptor.rs
+++ b/tonic/src/service/interceptor.rs
@@ -51,8 +51,8 @@ pub trait Interceptor {
 }
 
 impl<F> Interceptor for F
-    where
-        F: FnMut(crate::Request<()>) -> Result<crate::Request<()>, Status>,
+where
+    F: FnMut(crate::Request<()>) -> Result<crate::Request<()>, Status>,
 {
     fn call(&mut self, request: crate::Request<()>) -> Result<crate::Request<()>, Status> {
         self(request)
@@ -68,9 +68,9 @@ pub trait AsyncInterceptor {
 }
 
 impl<F, U> AsyncInterceptor for F
-    where
-        F: FnMut(crate::Request<()>) -> U,
-        U: Future<Output = Result<crate::Request<()>, Status>>,
+where
+    F: FnMut(crate::Request<()>) -> U,
+    U: Future<Output = Result<crate::Request<()>, Status>>,
 {
     type Future = U;
 
@@ -83,22 +83,22 @@ impl<F, U> AsyncInterceptor for F
 ///
 /// See [`Interceptor`] for more details.
 pub fn interceptor<F>(f: F) -> InterceptorLayer<F>
-    where
-        F: Interceptor,
+where
+    F: Interceptor,
 {
     InterceptorLayer { f }
 }
 
 #[deprecated(
-since = "0.5.1",
-note = "Please use the `interceptor` function instead"
+    since = "0.5.1",
+    note = "Please use the `interceptor` function instead"
 )]
 /// Create a new interceptor layer.
 ///
 /// See [`Interceptor`] for more details.
 pub fn interceptor_fn<F>(f: F) -> InterceptorLayer<F>
-    where
-        F: Interceptor,
+where
+    F: Interceptor,
 {
     interceptor(f)
 }
@@ -107,8 +107,8 @@ pub fn interceptor_fn<F>(f: F) -> InterceptorLayer<F>
 ///
 /// See [`AsyncInterceptor`] and [`Interceptor`] for more details.
 pub fn async_interceptor<F>(f: F) -> AsyncInterceptorLayer<F>
-    where
-        F: AsyncInterceptor,
+where
+    F: AsyncInterceptor,
 {
     AsyncInterceptorLayer { f }
 }
@@ -123,8 +123,8 @@ pub struct InterceptorLayer<F> {
 }
 
 impl<S, F> Layer<S> for InterceptorLayer<F>
-    where
-        F: Interceptor + Clone,
+where
+    F: Interceptor + Clone,
 {
     type Service = InterceptedService<S, F>;
 
@@ -143,9 +143,9 @@ pub struct AsyncInterceptorLayer<F> {
 }
 
 impl<S, F> Layer<S> for AsyncInterceptorLayer<F>
-    where
-        S: Clone,
-        F: AsyncInterceptor + Clone,
+where
+    S: Clone,
+    F: AsyncInterceptor + Clone,
 {
     type Service = AsyncInterceptedService<S, F>;
 
@@ -155,8 +155,8 @@ impl<S, F> Layer<S> for AsyncInterceptorLayer<F>
 }
 
 #[deprecated(
-since = "0.5.1",
-note = "Please use the `InterceptorLayer` type instead"
+    since = "0.5.1",
+    note = "Please use the `InterceptorLayer` type instead"
 )]
 /// A gRPC interceptor that can be used as a [`Layer`],
 /// created by calling [`interceptor`].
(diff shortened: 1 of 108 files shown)


tonic/src/service/interceptor.rs
```diff
@@ -52,6 +54,97 @@ where
     }
 }
 
+/// Async version of `Interceptor`.
+pub trait AsyncInterceptor {
```
LucioFranco, line 58: This feels almost exactly like the tower service trait, what is the advantage and why don't we just use the tower traits?

jdahlq: Hi @LucioFranco , sorry for letting this PR sit for so long. I've merged the latest master into my branch, updated the code to work with those changes, and refactored a bit to reduce code duplication.

Note that I had to constrain ReqBody to implement Default. Seems like that should be ok (and ResBody already had that constraint).

In response to your question...

I would love to use something in Tower rather than create all of this gnarly generics+futures code, but I couldn't find anything that will do the trick. You could ask the same thing about the non-async interceptor, too (written by @davidpdrsn   ?). I think the main reason is that Tower doesn't have anything that both (a) modifies a request and (b) allows an error response to be returned immediately.

Tower has [`service_fn`](https://tower-rs.github.io/tower/tower/fn.service_fn.html) which makes an async function into a Tower Service, but it doesn't wrap an inner service, so I don't think it can be used as a middleware layer.

Looking at the [`tower::util::ServiceExt`](https://docs.rs/tower/latest/tower/util/trait.ServiceExt.html) trait, there are are some promising methods that don't quite fit the async_interceptor case:

 - [`map_request`](https://docs.rs/tower/latest/tower/util/trait.ServiceExt.html#method.map_request) transforms the request and wraps the service, but it doesn't accept async functions, and it cannot skip calling the wrapped service.
 - [`filter_async`](https://docs.rs/tower/latest/tower/util/trait.ServiceExt.html#method.filter_async) is doesn't transform the request, but is otherwise close to what we want.

I also spent quite a bit of time trying to simplify this code with async functions, but since their return types are generated at compile time, it's hard to make them play nicely with the `Service` trait since it requires explicitly specifying the `Future` associated type, and we can't set `Future = impl Future<Output = ...>` until [this rust RFC](https://github.com/rust-lang/rust/issues/63063) is accepted and shipped.

davidpdrsn: The main advantage is that this uses tonic's `Request` and `Status` types. Whereas tower middleware have to use the types from `http`.

jdahlq:
The interop tests for windows-latest failed, looks like the test is just hanging? @LucioFranco maybe just needs to be re-run?

jdahlq:
It's a prost-build error:

>  CMake Error at CMakeLists.txt:24 (project):
>     The CMAKE_C_COMPILER:
> 
>       C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.31.31103/bin/Hostx64/x64/cl.exe
> 
>     is not a full path to an existing compiler tool.

Probably isn't caused by this PR. I'm assuming it's related to: https://github.com/tokio-rs/prost/pull/620

jdahlq:
It passes now, I think it was just flaky or a dependency issue. I merged the latest (unrelated) commit just to re-trigger the tests.

UnsolvedCypher:
Hi @LucioFranco, this looks like it is complete and checks are passing, can it be merged now?

LucioFranco:
I will try and give this a review in the coming weeks I have a decent amount of backlog right now.

jdahlq:
Hi @LucioFranco , it's been a few months, but you seem to have a lot on your plate, and I understand that this merge request is relatively unimportant. Please let me know if realistically speaking this won't be merged so I can put it in its own crate or something and don't have to rely on a forked version of tonic. Thanks!

LucioFranco:
At this point I don't have much time for this and in general tonic will be moving to be lighter (will eventually post the roadmap when I finish it lol). So if you can move this to your own crate for now that would probably be the easiest path forward. Sorry for the delay.

jdahlq:
> At this point I don't have much time for this and in general tonic will be moving to be lighter (will eventually post the roadmap when I finish it lol). So if you can move this to your own crate for now that would probably be the easiest path forward. Sorry for the delay.

Got it, I'll give that a try and post a link here for people who are interested.

jdahlq:
Hey Lucio, would you consider a few small function visibility changes if I sent you a new merge request?

Currently async_interceptor relies on a number of non-public functions. I can work around that except for one place: Neither [`Request::into_http`](https://github.com/hyperium/tonic/blob/master/tonic/src/request.rs#L169) nor [`Extensions::into_http`](https://github.com/hyperium/tonic/blob/master/tonic/src/extensions.rs#L62) are public. Either one being public is sufficient. Note that `Request::from_http` is already public, so it's not too much of a stretch to do the same for `into_http`.

For proper error handling, we also need  [`body::boxed`](https://github.com/hyperium/tonic/blob/master/tonic/src/body.rs#L9) to be public because it calls a long chain of non-public functions.

Finally, it would be nice if `Request::from_parts`, and `Request::into_parts` were public, but it's not strictly necessary.

Thanks for your time!

jdahlq:
Here is the minimal set of changes (3 lines) that I need to make an external async_interceptor crate work:
https://github.com/hyperium/tonic/commit/6af8bde4bb41d3bdd265c8c6a56ad7fdd9302cac

If that is acceptable, I'll open a merge request.

LucioFranco:
@jdahlq the `into_http` I think we can merge but the boxed one I believe we shouldn't expose that method maybe but we can expose stuff on `Status` to let you compose your own body type. That to me feels more flexible and less exposing internals about the body. What do you think?

jdahlq:
@LucioFranco yes, I agree it would be a little weird to exposed `boxed`, especially since it is doing more than just boxing. Are you thinking we could expose `Status::from_error`? That seems like a reasonable part of the API to me, and it would be sufficient for my purposes.

kwiesmueller:
@jdahlq do you already have a repo with the standalone interceptor?
I'd be interested in helping test it for an auth interceptor.

jdahlq:
Hi @kwiesmueller , you can grab the code here: https://github.com/arcanyx-pub/tonic-async-interceptor

It is published as [`tonic-async-interceptor`](https://crates.io/crates/tonic-async-interceptor) and currently points to a forked version of tonic, so you can use it in your Cargo.toml like this:

```toml
[dependencies]
# ...
tonic = { package = "tonic-arcanyx-fork", version = "0.8.1-alpha.0" }
tonic-async-interceptor = "0.1.0-alpha.0"
# If you use tonic-web, you'll also have to use a forked version:
tonic-web = { package = "tonic-web-arcanyx-fork", version = "0.8.1-alpha.1" }
```

The forked versions are taken from the [most recent commit](https://github.com/hyperium/tonic/commit/2a83f8ba1ad20dc2322eb16520694dcbb2dc795b) on hyperium/tonic:master.

jdahlq:
@kwiesmueller and all, I have published v0.1.0 of [tonic-async-interceptor](https://crates.io/crates/tonic-async-interceptor) which uses tonic v0.8.1. Hope it is useful!

Issue #911, opened 2022-02-12
Interceptor doesn't preserve HTTP method, causing CORS failures


jdahlq:
Before the request is sent to an interceptor, the HTTP request is converted into a Tonic request and its body is removed. After the interceptor has transformed the Tonic request, the body is added back in, and it is converted back into an HTTP request. However, this last conversion assumes that the request will always be an HTTP2 POST.

Tonic-web handles preflight CORS requests (which have the HTTP OPTIONS method), but if there is an interceptor before the Tonic service, the request is changed from OPTIONS to POST before tonic-web sees it, and an incorrect response is sent, causing a CORS failure in the browser.

jdahlq:
Tonic-web does handle it, but only when the grpc service itself is called. Consider the following server definition.

```rust
let router = Server::builder()
    .layer(interceptor(my_interceptor_fn))
    .add_service(tonic_web::enable(my_svc));
```

`tonic_web::enable` wraps `my_svc` with a `GrpcWeb` service which handles CORS before calling `my_svc`. `GrpcWeb` relies on the headers, specifically the HTTP Method, to identify preflight CORS requests vs normal gRPC POST requests. So, the problem is that `interceptor` changes the HTTP method to POST (from OPTIONS, the one used by preflight CORS) before tonic-web even sees it.

An alternative to modifying `interceptor` to preserve HTTP method is to refactor `Server` and tonic-web to handle CORS before all layers, but I don't think that makes much sense. It would be a big refactor, and you wouldn't be able to apply different tonic-web settings for different services.

The input and output of `interceptor` is an HTTP request, not a gRPC request, so I think it makes sense to preserve all the HTTP attributes of the input request there.

LucioFranco:
I believe with #912 we can close this.

PR #912, opened 2022-02-12
fix(tonic): Preserve HTTP method in interceptor

jdahlq:
## Motivation

See Issue #911 

## Solution

The function for converting a Tonic request into an HTTP request, `Request::into_http`, is public only to the crate and only used in a few places. Currently, HTTP2 and POST are hardcoded for HTTP method and version. Thus, we can simply add `method` and `version` params and update the callers, and then fix the interceptor to preserve them.

### DIFF
--- a/tonic/src/client/grpc.rs
+++ b/tonic/src/client/grpc.rs
@@ -248,7 +248,12 @@ impl<T> Grpc<T> {
             })
             .map(BoxBody::new);
 
-        let mut request = request.into_http(uri, SanitizeHeaders::Yes);
+        let mut request = request.into_http(
+            uri,
+            http::Method::POST,
+            http::Version::HTTP_2,
+            SanitizeHeaders::Yes,
+        );
 
         // Add the gRPC related HTTP headers
         request
--- a/tonic/src/request.rs
+++ b/tonic/src/request.rs
@@ -169,12 +169,14 @@ impl<T> Request<T> {
     pub(crate) fn into_http(
         self,
         uri: http::Uri,
+        method: http::Method,
+        version: http::Version,
         sanitize_headers: SanitizeHeaders,
     ) -> http::Request<T> {
         let mut request = http::Request::new(self.message);
 
-        *request.version_mut() = http::Version::HTTP_2;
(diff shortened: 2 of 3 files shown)


tonic/src/client/grpc.rs
```diff
@@ -248,7 +248,12 @@ impl<T> Grpc<T> {
             })
             .map(BoxBody::new);
 
-        let mut request = request.into_http(uri, SanitizeHeaders::Yes);
+        let mut request = request.into_http(
+            uri,
+            http::Method::POST,
+            http::Version::HTTP_2,
+            SanitizeHeaders::Yes,
+        );
```
LucioFranco, line 256: Im not following why these values are not just hard coded into`into_http` since they should always be POST and HTTP2 and can be overridden outside (aka for grpc-web).

jdahlq: Thanks for taking the time to review! Firstly, see my response on issue #911 to make sure we are on the same page.

If Rust supported default parameters, then I would propose making POST and HTTP2 defaults for `Request::into_http`.

But one issue here is that `Request::from_http` will happily accept HTTP requests of any version and method, silently dropping that information and creating subtle bugs like this one. By adding method and version to `Request::into_http`, we are making it more explicit that method and version are not preserved.

In any case, `Request::into_http` only has two callers currently, and it is restricted to the crate, so the stakes aren't very high here. I could leave `grpc.rs` alone and instead write a new method `Request::from_http_with_custom_method` if you think that would be cleaner.

LucioFranco:
looks like a fmt failure is causing ci to not pass here

jdahlq:
Whoops, not sure what happened there; during rebasing, the fmt check was fixed in the initial commit, the next commit added a fat-fingered "g" breaking the compile, and I omitted a `tokio::test` annotation. Sorry about the noise, all fixed now, CI passing.

tonic/src/client/grpc.rs
LucioFranco: Yup, thanks for writing that up it all makes sense!

LucioFranco (approved):
LGTM! Thanks!

PR #928, opened 2022-02-25
fix(web): Fix error tonic-web doc url

attila-lin:
## Motivation

Currently, `tonic-web` use the tonic-reflection doc url.

## Solution

Fixed it.


LucioFranco (approved):
Good catch! Thanks!

PR #930, opened 2022-02-26
feat(health): Expose `HealthService` publically

### LINKED ISSUE #929: Support retrieve name of NamedService by instance
### Crates

`tonic`

### Motivation

When I'm trying to [retrieve name of health service](https://github.com/engula/engula/blob/2ba773d2a32781f0b04a863873636a04cd9722de/src/engula/src/server.rs#L87-L89), it's impossible since the exact type consist an inner struct. Although I can still copy the name string literally and fallback to `set_service_status` interface, it would be reasonable to retrieve the name of `NamedService` by instance.

### Proposal

I'll submit a patch and it should simply add a getter.

### Alternatives

N/A

tisonkun:
Signed-off-by: tison <wander4096@gmail.com>

This closes #929.

I'm glad to add some tests, but the existing `NAME` doesn't have tests and I don't have idea where to add some.

### DIFF
--- a/tonic-health/src/server.rs
+++ b/tonic-health/src/server.rs
@@ -103,7 +103,7 @@ impl HealthReporter {
     }
 }
 
-/// A service providing implementations of gRPC health checking protocol. 
+/// A service providing implementations of gRPC health checking protocol.
 #[derive(Debug)]
 pub struct HealthService {
     statuses: Arc<RwLock<HashMap<String, StatusPair>>>,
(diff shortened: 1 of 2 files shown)


tisonkun:
cc @LucioFranco

LucioFranco:
@tisonkun looks like there is a CI failure?

tisonkun:
@LucioFranco 2c0da18 is the fix..

tisonkun:
@LucioFranco I think this PR is ready to merge now.

PR #931, opened 2022-02-27
Add example of detecting client drops in bidirectional streams on server side

### LINKED ISSUE #781: Detect client drops in bidirectional streams
Hi, 

Im following the routeguide tutorial and am wondering if there is a way for detecting when a client drops when working with bidirectional streams.

Here is the code snippet:
```
in the bidirectional stream, how can know if a client was not able to receive the response?

    async fn route_chat(
        &self,
        request: Request<tonic::Streaming<RouteNote>>,
    ) -> Result<Response<Self::RouteChatStream>, Status> {
        println!("RouteChat");

        let mut notes = HashMap::new();
        let mut stream = request.into_inner();

        let output = async_stream::try_stream! {
            while let Some(note) = stream.next().await {
                let note = note?;

                let location = note.location.clone().unwrap();

                let location_notes = notes.entry(location).or_insert(vec![]);
                location_notes.push(note);

                for note in location_notes {
                    yield note.clone();
                }
            }
        };

        Ok(Response::new(Box::pin(output)))
    }
```
Any help would be appreciated.

xoac:
closes #781 

I think that this example shows also a weakness of tonic. For example in `async fn server_streaming_echo()` function is impossible (or I do not know how) to get reason why `output_stream` wasn't able to send message to client. 

## Motivation

Provide example how to detect client disconnection on server side

## Solution

Use `tokio::spawn` to separate await on input stream from output stream.

### DIFF
--- a/examples/src/streaming/client.rs
+++ b/examples/src/streaming/client.rs
@@ -2,28 +2,84 @@ pub mod pb {
     tonic::include_proto!("grpc.examples.echo");
 }
 
+use futures::stream::Stream;
 use pb::{echo_client::EchoClient, EchoRequest};
+use std::time::Duration;
+use tokio_stream::StreamExt;
+use tonic::transport::Channel;
 
-#[tokio::main]
-async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut client = EchoClient::connect("http://[::1]:50051").await.unwrap();
+fn echo_requests_iter() -> impl Stream<Item = EchoRequest> {
+    tokio_stream::iter(1..usize::MAX).map(|i| EchoRequest {
+        message: format!("msg {:02}", i),
+    })
+}
 
+async fn streaming_echo(client: &mut EchoClient<Channel>, num: usize) {
     let stream = client
         .server_streaming_echo(EchoRequest {
             message: "foo".into(),
         })
         .await
+        .unwrap()
+        .into_inner();
+
+    let rev_task = tokio::spawn(async move {
+        // stream is infinite - take just 5 elements and then disconnect
+        let mut stream = stream.take(num);
+        while let Some(item) = stream.next().await {
+            println!("\trecived: {}", item.unwrap().message);
+        }
+        // stream is droped here and the disconnect info is send to server
+    });
+
+    rev_task.await.unwrap();
+}
+
+async fn bidirectional_streaming_echo(client: &mut EchoClient<Channel>, num: usize) {
+    let in_stream = echo_requests_iter().take(num);
+
+    let response = client
+        .bidirectional_streaming_echo(in_stream)
+        .await
         .unwrap();
 
-    println!("Connected...now sleeping for 2 seconds...");
+    let mut resp_stream = response.into_inner();
 
-    tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+    while let Some(recived) = resp_stream.next().await {
+        let recived = recived.unwrap();
+        println!("\trecived message: `{}`", recived.message);
+    }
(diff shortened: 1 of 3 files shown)


examples/src/streaming/client.rs
```diff
@@ -2,28 +2,84 @@ pub mod pb {
...
-    println!("Connected...now sleeping for 2 seconds...");
+    let mut resp_stream = response.into_inner();
+
+    while let Some(recived) = resp_stream.next().await {
+        let recived = recived.unwrap();
+        println!("\trecived message: `{}`", recived.message);
+    }
+}
+
+async fn bidirectional_streaming_echo_throttle(client: &mut EchoClient<Channel>, dur: Duration) {
```
LucioFranco, line 54: Not following why we want this example as well?

examples/src/streaming/server.rs
```diff
@@ -29,28 +56,37 @@ impl pb::echo_server::Echo for EchoServer {
...
+                        // the item was send to channel but there is no guarntee it was send to client!
+                        // If you need such guarntee the solution can be "rendezvous channel"
+                        // (channel with 0 capacity), tokio::sync::mpsc do not support this special case
+                        // but you can use `flume::bounded` https://docs.rs/flume/latest/flume/fn.bounded.html
+                    }
+                    Err(_item) => {
+                        println!("\tclient disconnected"); //out_stream (build on rx) was droped
+                        break;
+                    }
+                }
```
LucioFranco, line 83: So I kinda disagree with this method of tracking client disconnects. I would 1) I don't want to suggest flume for a few reason I don't want to dig into right now but at a high level we shouldn't be recommending it to users. 2) If you want to know if your client got a message this should be handled at a high level at the application level via acks. This is because no matter what with tcp/udp there is no way to ensure a user actually got the message. So the best way is to get an ack on the user level code that the users client has gotten the message. 

If a client disconnects the `while let` loop should exit.

examples/src/streaming/server.rs
```diff
@@ -65,9 +101,50 @@ impl pb::echo_server::Echo for EchoServer {
...
+        // to `in_stream`.
+        tokio::spawn(async move {
+            while let Some(result) = in_stream.next().await {
+                match result {
+                    Ok(v) => tx
+                        .send(Ok(EchoResponse { message: v.message }))
+                        .await
+                        .expect("working rx"),
+                    Err(err) => {
+                        if let Some(io_err) = match_for_io_error(&err) {
```
LucioFranco, line 123: I like this nice 👍

LucioFranco (changes requested):
Looks like a good start! Left a few comments.

examples/src/streaming/client.rs
xoac: I wanted to show difference between graceful stream end the one when client disconnected for example by broken connection or just crush. This 2 functions could be unified to one. 

Here is the output from server part.
```
EchoServer::server_streaming_echo
	client connected from: Some([::1]:36750)
	client disconnected
EchoServer::Bidirectional_streaming_echo
	stream ended
EchoServer::Bidirectional_streaming_echo
	client disconnected: broken pipe
	stream ended
```

xoac: In other words the difference is on the server side. On the client side I just needed a function that sends request slowly enough that you can write what is printed. I added additional comments to clarify that.

Should I unify this to `bidirectional_streaming_echo(client &mut EchoClient<Channel>, num_reqs: usize, throttle: Duration)`?

LucioFranco (approved):
Thanks!

PR #932, opened 2022-02-28
fix(transport): Make `Server::layer()` support more than one layer

Arcterus:
The current behavior is to replace the previous layer with the newly specified layer.  The new behavior is to chain the layers together.  The current behavior might actually be intentional, but IMO it's not very intuitive given how services are added by chaining method calls.

I can adjust the docs and/or tests as necessary, so let me know if you want me to do so.

## Motivation

We were using `Server::layer()` and assumed it supported more than one layer (without using `Stack` from `tower`) given the example in the docs (which has an interceptor for authentication that, with the current behavior, gets replaced with effectively an identity interceptor afaict).

## Solution

This change basically just uses `Stack` internally rather than requiring users to use it themselves manually.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -75,7 +75,7 @@ const DEFAULT_HTTP2_KEEPALIVE_TIMEOUT_SECS: u64 = 20;
 /// a very good out of the box http2 server for use with tonic but is also a
 /// reference implementation that should be a good starting point for anyone
 /// wanting to create a more complex and/or specific implementation.
-#[derive(Default, Clone)]
+#[derive(Clone)]
 pub struct Server<L = Identity> {
     trace_interceptor: Option<TraceInterceptor>,
     concurrency_limit: Option<usize>,
@@ -91,7 +91,29 @@ pub struct Server<L = Identity> {
     http2_keepalive_timeout: Option<Duration>,
     max_frame_size: Option<u32>,
     accept_http1: bool,
-    layer: L,
+    service_builder: ServiceBuilder<L>,
+}
+
+impl Default for Server<Identity> {
+    fn default() -> Self {
+        Self {
+            trace_interceptor: None,
+            concurrency_limit: None,
+            timeout: None,
+            #[cfg(feature = "tls")]
+            tls: None,
+            init_stream_window_size: None,
+            init_connection_window_size: None,
+            max_concurrent_streams: None,
+            tcp_keepalive: None,
+            tcp_nodelay: false,
+            http2_keepalive_interval: None,
+            http2_keepalive_timeout: None,
+            max_frame_size: None,
+            accept_http1: false,
+            service_builder: Default::default(),
+        }
+    }
 }
 
 /// A stack based `Service` router.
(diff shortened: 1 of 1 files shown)


davidpdrsn (approved):
I've actually been thinking that we should consider changing `Server::layer` so it can be called after services have been added, and applies the middleware to the services already added. This is how axum does it and I personally find it easier to understand. Adding the middleware first, then the services, seems a bit backwards to me.

In general we're not really making much use of the fact that tonic's uses axum internally for its routing. We could add more flexible APIs such as merging routers. That is probably outside the scope of this PR but thought I would mention it.

I think this changes makes sense on its own though.

tonic/src/transport/server/mod.rs
```diff
@@ -408,9 +411,9 @@ impl<L> Server<L> {
     /// [eco]: https://github.com/tower-rs
     /// [`ServiceBuilder`]: tower::ServiceBuilder
     /// [interceptors]: crate::service::Interceptor
-    pub fn layer<NewLayer>(self, new_layer: NewLayer) -> Server<NewLayer> {
+    pub fn layer<NewLayer>(self, new_layer: NewLayer) -> Server<Stack<L, NewLayer>> {
         Server {
-            layer: new_layer,
+            layer: Stack::new(self.layer, new_layer),
```
davidpdrsn, line 416: Actually now that I think about it. We should use consider using `tower::ServiceBuilder`. Then we ensure the layers are added in the right order. For example I believe the new layer should be added as the left layer in the stack as that is what [`ServiceBuilder::layer`](https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#method.layer) does.

davidpdrsn (changes requested):

tonic/src/transport/server/mod.rs
Arcterus: This is a `tower` issue, but unless I'm misunderstanding the code, those docs are wrong.  It seems like it wraps the given layer with the inner layer rather than wrapping the inner layer with the given one.

Anyway, I'll look into using `ServiceBuilder` later tonight or tomorrow.

Arcterus:
I haven't used `axum` much, so ¯\_(ツ)_/¯.

When you say the middleware applies to the services you specified previously, does that mean you can layer the middleware so some only apply to a subset of services?  Like, you could add more services after adding a middleware and those new services would not be affected by the middleware?  If so, I'd definitely prefer that approach.  We currently have an ad hoc mechanism to do something like this, which I'd prefer to remove if possible.

LucioFranco:
It should be possible to achieve this right now by just passing in your own service builder correct? I honestly want to just nix the entire server code and just use axum at this point 😆

davidpdrsn:
> It should be possible to achieve this right now by just passing in your own service builder correct?

Yeah this would just be a convenience.

> I honestly want to just nix the entire server code and just use axum at this point 😆

👌

tonic/src/transport/server/mod.rs
```diff
@@ -88,7 +91,29 @@ pub struct Server<L = Identity> {
     http2_keepalive_timeout: Option<Duration>,
     max_frame_size: Option<u32>,
     accept_http1: bool,
-    layer: L,
+    service_builder: ServiceBuilder<L>,
+}
+
+impl Default for Server<Identity> {
```
LucioFranco, line 97: I feel like we shouldn't have to implement this manually, is there something missing in tower?

LucioFranco (approved):
LGTM just one question

davidpdrsn (approved):

tonic/src/transport/server/mod.rs
LucioFranco: Ah okay huh I guess derive doesn't take into account default bounds

PR #935, opened 2022-03-04
feat(transport): support customizing `Channel`'s async executor

davidpdrsn:
## Motivation

At Embark we've been working on integrating tokio-console into our project and have used `hyper::client::Builder::executor` to pass our own executor that spawns named tasks. However its not possible to do that tonic, so tasks spawned by tonic are left unnamed.

## Solution

- Add tonic's own `Executor` trait. Its identical to hyper's but added here not to have a public dependency on hyper.
- Added `Endpoint::executor` builder method to support setting another executor.
- Spawn `tower::buffer::Buffer` tasks on the given executor.
- Still uses `tokio::spawn` by default.

I intentionally didn't touch the server side because we don't need that at Embark but could do that if we think its relevant.

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -4,10 +4,9 @@ use super::Channel;
 use super::ClientTlsConfig;
 #[cfg(feature = "tls")]
 use crate::transport::service::TlsConnector;
-use crate::transport::{service::SharedExec, Error};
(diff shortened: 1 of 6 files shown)


LucioFranco (approved):
Not opposed to this at all, though I am happy to just use hyper's since there is a plan for 1.0 there.

davidpdrsn:
I've moved to hyper's `Executor` trait.

LucioFranco (approved):

PR #939, opened 2022-03-14
Fix inner balance channel

belltoy:
## Motivation

There are a few issues with the `Channel::balance_list` API.

First, the inner mpsc capacity was hardcoded to `1024` which is not necessary, I think. For a gRPC service cluster, the endpoints of it won't be changed so frequently to consume a 1024 length channel. Or maybe it would, it will be back pressured via the Future. Maybe 1024 is wasteful. Here I use `8` as default.

On the other hand, the original `Channel::balance_list` API only returns the created channel with the sender dropped. That means users can't change the service endpoints after the initialization. And also the inner mpsc channel will never get `Change`s.

## Solution

- Reduce the inner mpsc channel capacity
- Use async `send` function and awaiting instead of `try_send` in case of the shrunk inner channel capacity has been reached (breaking)
- Return the inner sender for later updates (breaking)

### DIFF
--- a/examples/src/load_balance/client.rs
+++ b/examples/src/load_balance/client.rs
@@ -11,7 +11,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .iter()
         .map(|a| Channel::from_static(a));
 
-    let channel = Channel::balance_list(endpoints);
+    let (channel, _) = Channel::balance_list(endpoints).await;
 
     let mut client = EchoClient::new(channel);
 
--- a/tonic/src/transport/channel/mod.rs
+++ b/tonic/src/transport/channel/mod.rs
@@ -40,6 +40,7 @@ use tower::{
 type Svc = Either<Connection, BoxService<Request<BoxBody>, Response<hyper::Body>, crate::Error>>;
 
 const DEFAULT_BUFFER_SIZE: usize = 1024;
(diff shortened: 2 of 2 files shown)


belltoy:
Or is the original `balance_list` function just a convenient helper function for a common use? If that's the purpose, maybe its doc should be more detailed.

LucioFranco:
So I am not totally clear on why 1024 is wasteful, tokio's mpsc channel doesn't allocate that memory upfront so its not wasting that memory. As for the change, I don't see why we need to have a second method also return the TX? if you would like the sender you can use `balance_channel` that gives you the sender where balance_list is to just load balance over some provided non-dynamic list. 

As for the future of this code I actually plan on removing it from tonic soon so I am also not sure if I want to iterate on the API right now.

belltoy:
> So I am not totally clear on why 1024 is wasteful, tokio's mpsc channel doesn't allocate that memory upfront so its not wasting that memory. As for the change, I don't see why we need to have a second method also return the TX? if you would like the sender you can use balance_channel that gives you the sender where balance_list is to just load balance over some provided non-dynamic list.

My mistake. mpsc channel is using semaphore inner.

> As for the future of this code I actually plan on removing it from tonic soon so I am also not sure if I want to iterate on the API right now.

Since that, `balance_channel` is enough.

PR #941, opened 2022-03-18
Server signals ephemeral port when ready

robs-zeynet:
Creating client/server unit tests has two classical problems:

1) How do you find a port that's unused that the server can
    bind but also communicate it to the client.  Typically
    hard coding one in a test means the test will fail anytime
    that port is bound (1338 in current code)
2) How do you start the server and the client as fast as possible
    (think 100s of tests running) without using a sleep() which
    either is too small (causing race conditions) or too large
    (causing testing delays)

This patch proposes a way for the client to register a signal
and the server to send the ephemeral port back on that signal
when it's ready, solving both problems.

I created an example unit test to show how this could be used
but I can fix up the rest and add the right documentation if
folks agree with the general direction.

Feedback welcome.

### DIFF
--- a/tests/integration_tests/tests/connection.rs
+++ b/tests/integration_tests/tests/connection.rs
@@ -28,6 +28,43 @@ async fn connect_returns_err() {
 }
 
 #[tokio::test]
+async fn connect_ephemeral_no_race() {
+    let (tx, rx) = oneshot::channel();
+    let sender = Arc::new(Mutex::new(Some(tx)));
+    let svc = test_server::TestServer::new(Svc(sender));
+
+    let (sock_tx, sock_rx) = tokio::sync::oneshot::channel();
+
+    let jh = tokio::spawn(async move {
+
+        Server::builder()
+            .add_service(svc)
+            .add_signal_when_ready(sock_tx)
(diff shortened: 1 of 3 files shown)


davidpdrsn:
The recommend way of doing this is with [`Router::serve_with_incoming`](https://docs.rs/tonic/latest/tonic/transport/server/struct.Router.html#method.serve_with_incoming):

```rust
#[tokio::test]
async fn some_test() {
    let service = GreeterServer::new(MyGreeter);

    // create the listener up front so the server is immediately ready
    // bind to port `0` so the OS finds a free port
    let listener = tokio::net::TcpListener::bind("0.0.0.0:0").await.unwrap();
    let addr = listener.local_addr().unwrap();

    tokio::spawn(async move {
        tonic::transport::Server::builder()
            .add_service(service)
            .serve_with_incoming(tokio_stream::wrappers::TcpListenerStream::new(listener))
            .await
            .unwrap();
    });

    // make the channel bound to `addr`
}
```

Or mock all the IO if thats more your thing. See https://github.com/hyperium/tonic/blob/master/examples/src/mock/mock.rs.

PR #949, opened 2022-03-29
chore: Update tokio-util to 0.7

adamchalmers:
## Motivation

The tokio ecosystem has been moving to tokio-util 0.7 -- e.g. h2 has just upgraded to it. In my project, tonic is the only holdout still on 0.6, so if we upgrade, my project will have fewer crates to compile.

## Solution

Bump tokio-util from 0.6 to 0.7

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -59,7 +59,7 @@ async-stream = "0.3"
 http-body = "0.4.4"
 percent-encoding = "2.1"
 pin-project = "1.0"
-tokio-util = {version = "0.6", features = ["codec"]}
+tokio-util = {version = "0.7", features = ["codec"]}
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Thanks!

adamchalmers:
Great, thanks! Looks like this can be merged.

adamchalmers:
Actually, once H2 cuts their new release, it will be 0.3.13, which is still compatible with tonic's bound on `h2 = {version = "0.3", optional = true}`. 

So, this PR causes tonic to use both h2 0.6 and 0.7, but only for the short time until h2 cuts a new patch release. I spoke to the Hyper folks in Discord and they said they can cut a release tomorrow. So I feel OK merging this.

PR #960, opened 2022-04-04
feat(transport): Add `channel` feature flag

### LINKED ISSUE #959: tonic provides no way to disable server-only dependencies
If you pull in tonic 0.7 as a dependency, the transport feature (which is enabled by default) pulls in the axum dependency. It appears only servers need the axum dependency, so for downstream users who only need client code, this is unnecessary.

LucioFranco:
cc @djc 

Closes #959


djc (approved):
Nice, LGTM!

PR #962, opened 2022-04-04
chore: Prepare `0.7.1` release

LucioFranco:


davidpdrsn (commented):
Hm I this a breaking change? Does it change what `default_features = false` pulls in?

LucioFranco:
@davidpdrsn no this only adds a new feature that is a subset of the `transport` feature which is included automatically. So this only _adds_ a feature with one less dep than `transport`. I don't think its a breaking change.

davidpdrsn:
Yeah I think you're right.

If someone did `default_features = false, features = ["transport"]` then this doesn't change, since like you say `channel` is a subset of `transport`.

🚢

davidpdrsn (approved):

PR #969, opened 2022-04-09
Decouple `NamedService` from the `transport` feature

olix0r:
The `tonic::transport::NamedService` trait is gated by the `transport`
feature. This means that generated server implementations can only
include this implementation when the `transport` feature is enabled on
`tonic-build`; and if the `transport` feature is not used at
generation-time, then a server cannot be used with
`tonic::transport::Server` at runtime.

We would like to publish a crate including generated protobuf/gRPC
bindings that does not _require_ a dependency on the `transport` feature
(and its fairly heavyweight dependency tree). But it would be nice if
its servers could still be used by applications that want to use
`tonic::transport::Server`.

By decoupling the `NamedService` trait from the `transport` feature,
`tonic-build` can be changed to always provide an implementation for
this trait without requiring all of the `transport` dependencies. This
enables generated servers to be used with `tonic::transport::Server`.

### DIFF
--- a/tonic/src/server/mod.rs
+++ b/tonic/src/server/mod.rs
@@ -15,3 +15,12 @@ pub use self::grpc::Grpc;
 pub use self::service::{
     ClientStreamingService, ServerStreamingService, StreamingService, UnaryService,
 };
+
+/// A trait to provide a static reference to the service's
+/// name. This is used for routing service's within the router.
+pub trait NamedService {
+    /// The `Service-Name` as described [here].
+    ///
+    /// [here]: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+    const NAME: &'static str;
+}
--- a/tonic-build/src/server.rs
+++ b/tonic-build/src/server.rs
@@ -35,7 +35,7 @@ pub fn generate<T: Service>(
         if package.is_empty() { "" } else { "." },
         service.identifier()
     );
-    let transport = generate_transport(&server_service, &server_trait, &path);
+    let named = generate_named(&server_service, &server_trait, &path);
     let mod_attributes = attributes.for_mod(package);
     let struct_attributes = attributes.for_struct(&path);
 
@@ -172,7 +172,7 @@ pub fn generate<T: Service>(
                 }
             }
 
-            #transport
+            #named
         }
     }
 }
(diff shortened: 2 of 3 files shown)


LucioFranco:
I wonder if this counts as a breaking change? I don't think so...so maybe we can get this in a patch release for yall. What do you think?

LucioFranco:
Yeah, I think its worth it. I suspect this will be a feature that will live beyond the removal of the `transport` module.

LucioFranco:
@olix0r do you still want this feature?

LucioFranco:
We could probably get this into the next release

olix0r:
I think I'd still like this... I'm juggling a few things at the moment but I'll try to get back to this; or, if someone else has a chance to take it over that would be great, too. Cc @hawkw

LucioFranco:
@hawkw 👍 I wanna try and get some sort of tonic release this week and we could get this merged.

tonic/src/server/mod.rs
```diff
@@ -15,3 +15,12 @@ pub use self::grpc::Grpc;
...
     ClientStreamingService, ServerStreamingService, StreamingService, UnaryService,
 };
+
+/// A trait to provide a static reference to the service's
+/// name. This is used for routing service's within the router.
+pub trait NamedService {
+    /// The `Service-Name` as described [here].
+    ///
+    /// [here]: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
+    const NAME: &'static str;
```
LucioFranco, line 25: Do we want to change this to a method that returns a &'static str? OR keep it as a const?

hawkw:
@olix0r https://github.com/olix0r/tonic/pull/1 updates your branch from the latest master and fixes the compilation error, if you want to merge that into this PR? otherwise, i'm happy to open a new PR that supplants this one.

tonic/src/server/mod.rs
LucioFranco: Isn't this also a breaking change since its moving modules?

LucioFranco: Ah its not :)

hawkw: yeah, it's re-exported in `transport`.

PR #974, opened 2022-04-14
fix(build): Reduce `Default` bound requirement

conradludgate:
## Motivation

https://github.com/hyperium/tonic/issues/973

It seems the `Default` bound was overly restrictive

## Solution

Move the `Default` bound to the `with_interceptor` function instead of the whole impl

### DIFF
--- a/tests/integration_tests/Cargo.toml
+++ b/tests/integration_tests/Cargo.toml
@@ -23,6 +23,7 @@ http-body = "0.4"
 hyper = "0.14"
 tokio-stream = {version = "0.1.5", features = ["net"]}
 tower = {version = "0.4", features = []}
+tower-http = { version = "0.2", features = ["full"] }
 tower-service = "0.3"
 tracing-subscriber = {version = "0.3", features = ["env-filter"]}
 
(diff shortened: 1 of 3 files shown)


LucioFranco:
I think in something like this an overly expressive bound is good to help guide the compiler to better errors. Is there something that this is blocking for you?

conradludgate:
Yeah, as mentioned in the other issue, it blocks a lot of service layers that would otherwise be valid. I see your point however.

There was a fix that I can make to box the responses or something on my end, but it seemed better to me if it wasn't necessary in the first place, as this is more churn for our company wide jump to tonic 0.7

If you don't want this, then I'd advise this limitation and ways to get around it be documented somewhere that's easy to find

LucioFranco:
Oh I see now in https://github.com/hyperium/tonic/issues/973 :)

davidpdrsn:
Outside of the one comment I had I think this looks good!

LucioFranco:
Ok looks like we need to fix the one CI failure and then we can merge. I don't think this is a breaking change so we can get this out in a patch release.

davidpdrsn (approved):

PR #990, opened 2022-04-29
feat: Add TryFrom implementations for MetadataValue

adamchalmers:
## Motivation

As described [in the issue](https://github.com/hyperium/tonic/issues/986) there are several adhoc methods that say "This function is intended to be replaced in the future by a `TryFrom` implementation once the trait is stabilized in std." Well, TryFrom is stable in std now, so let's use it :)

## Solution

I looked at which TryFrom impls [http::HeaderValue](https://docs.rs/http/latest/http/header/struct.HeaderValue.html) has, and copied those where it made sense. Seems like MetadataValue is _mostly_ a copy of http::HeaderValue, so I thought it was a good guide.

I tried to avoid repeating code, so often various TryFrom implementations either call into each other, or into the underlying `http::HeaderValue` TryFrom implementations. I kept the #[inline] attribute when the original call had it, and added a deprecation warning to the original adhoc methods.

Apparently MetadataValue already implements `From` for all numeric types >= 16 bits. They just don't show up in rustdoc, which is surprising, but I didn't dive in to figure out why. So I only implemented TryFrom around different string and byte representations.

### DIFF
--- a/examples/src/authentication/client.rs
+++ b/examples/src/authentication/client.rs
@@ -9,7 +9,7 @@ use tonic::{metadata::MetadataValue, transport::Channel, Request};
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let channel = Channel::from_static("http://[::1]:50051").connect().await?;
 
-    let token = MetadataValue::from_str("Bearer some-auth-token")?;
+    let token: MetadataValue<_> = "Bearer some-auth-token".parse()?;
 
     let mut client = EchoClient::with_interceptor(channel, move |mut req: Request<()>| {
         req.metadata_mut().insert("authorization", token.clone());
--- a/tonic/src/metadata/value.rs
+++ b/tonic/src/metadata/value.rs
@@ -286,7 +286,8 @@ impl<VE: ValueEncoding> MetadataValue<VE> {
 ///
 /// ```
 /// # use tonic::metadata::*;
-/// let val = AsciiMetadataValue::try_from_bytes(b"hello\xfa").unwrap();
+/// # use std::convert::TryFrom;
+/// let val = AsciiMetadataValue::try_from(b"hello\xfa").unwrap();
 /// assert_eq!(val, &b"hello\xfa"[..]);
 /// ```
 ///
@@ -294,7 +295,8 @@ impl<VE: ValueEncoding> MetadataValue<VE> {
 ///
 /// ```
 /// # use tonic::metadata::*;
-/// let val = AsciiMetadataValue::try_from_bytes(b"\n");
+/// # use std::convert::TryFrom;
+/// let val = AsciiMetadataValue::try_from(b"\n");
 /// assert!(val.is_err());
 /// ```
 impl<'a, VE: ValueEncoding> TryFrom<&'a [u8]> for MetadataValue<VE> {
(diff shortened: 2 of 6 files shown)


adamchalmers:
As requested, tagging @LucioFranco for a review.

examples/src/authentication/client.rs
```diff
@@ -9,7 +9,7 @@ use tonic::{metadata::MetadataValue, transport::Channel, Request};
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let channel = Channel::from_static("http://[::1]:50051").connect().await?;
 
-    let token = MetadataValue::from_str("Bearer some-auth-token")?;
+    let token: MetadataValue<_> = "Bearer some-auth-token".parse()?;
```
LucioFranco, line 12: hmm I actually kinda prefer the look of `from_str` vs using `parse`. I wonder if we should _not_ deprecate it and keep it. But just have the TryFrom's use it or the other way around?

tonic/src/metadata/value.rs
```diff
@@ -282,6 +273,161 @@ impl<VE: ValueEncoding> MetadataValue<VE> {
...
+/// ```
+///
+/// An invalid value
+///
+/// ```
+/// # use tonic::metadata::*;
+/// let val = AsciiMetadataValue::try_from_bytes(b"\n");
+/// assert!(val.is_err());
+/// ```
+impl<'a, VE: ValueEncoding> TryFrom<&'a [u8]> for MetadataValue<VE> {
```
LucioFranco, line 300: Can this lifetime be elided?

tonic/src/metadata/value.rs
```diff
@@ -282,6 +273,161 @@ impl<VE: ValueEncoding> MetadataValue<VE> {
...
+/// (inclusive) are permitted, excluding byte 127 (DEL).
+///
+/// For Binary metadata values this method cannot fail. See also the Binary
+/// only version of this method `from_bytes`.
+///
+/// # Examples
+///
+/// ```
+/// # use tonic::metadata::*;
+/// let val = AsciiMetadataValue::try_from_bytes(b"hello\xfa").unwrap();
```
LucioFranco, line 289: Shouldn't this use the `TryFrom` impl rather than the `try_from_bytes`?

tonic/src/metadata/value.rs
adamchalmers: It can be, but then Rustdoc generates docs with a signature like `TryFrom<&'0 str>`. If I name the lifetime, then it outputs the signature `TryFrom<&'a str>` which I think is more readable, more what Rust programmers are used to seeing.

I copied this style of doing it [from the `hyper` docs](https://docs.rs/hyper/latest/hyper/header/struct.HeaderValue.html#impl-TryFrom%3C%26%27a%20str%3E) -- happy to change it to the elided lifetime if you prefer, but I thought matching Hyper's docs is a good tiebreaker when I wasn't sure which was better.

examples/src/authentication/client.rs
adamchalmers: Hmm. So, MetadataValue impls `FromStr`, so it will already have a `from_str` method from stdlib, which is equivalent to `parse`. Unfortunately it's not in the stdlib prelude, so if I delete the ad-hoc `from_str` method then old code will fail to compile (although the compiler will suggest the easy fix of "use std::str::FromStr").

It does _kinda_ annoy me to have an adhoc method named `from_str` when the stdlib one exists too, just because it seems to make the API a bit larger than it needs to be. But this is a _very_ minor gripe :) So overall I'm happy with the solution of not deprecating `from_str` if that's what you'd like.

(I'm pretty used to `parse()` because it's used very often in the [stdlib IP types](https://doc.rust-lang.org/stable/std/net/struct.Ipv4Addr.html#examples)).

So yeah, happy to do whichever, just let me know what you think now that you've ready this silly long reply :)

tonic/src/metadata/value.rs
LucioFranco: Oh cool didn't know that, seems fine to me!

LucioFranco (approved):

LucioFranco:
Thanks!

PR #993, opened 2022-05-02
tonic: Unhide docs for Ascii and Binary

adamchalmers:
`MetadataValue::from<i16>` and other numeric types aren't shown in Rustdoc, because they're implemented for `Metadata<Ascii>`, and Ascii is hidden.

## Motivation

The [docs for MetadataValue](https://docs.rs/tonic/0.7.1/tonic/metadata/struct.MetadataValue.html) don't show any of its `From<>` impls for numerics like i16/i32/usize/etc. 

## Solution

Unhide the docs. I understand this makes the docs a bit noisier, so if there's a better solution I'd be open to that instead.

### DIFF
--- a/tonic/src/metadata/encoding.rs
+++ b/tonic/src/metadata/encoding.rs
@@ -51,11 +51,27 @@ pub trait ValueEncoding: Clone + Eq + PartialEq + Hash + self::value_encoding::S
     fn is_valid_key(key: &str) -> bool;
 }
 
+/// gRPC metadata values can be either ASCII strings or binary. Note that only
+/// visible ASCII characters (32-127) are permitted.
+/// This type should never be instantiated -- in fact, it's impossible
+/// to, because there's no variants to instantiate. Instead, it's just used as
+/// a type parameter for [`MetadataKey`] and [`MetadataValue`].
+///
+/// [`MetadataKey`]: struct.MetadataKey.html
+/// [`MetadataValue`]: struct.MetadataValue.html
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
-#[doc(hidden)]
+#[non_exhaustive]
 pub enum Ascii {}
+
+/// gRPC metadata values can be either ASCII strings or binary.
+/// This type should never be instantiated -- in fact, it's impossible
+/// to, because there's no variants to instantiate. Instead, it's just used as
+/// a type parameter for [`MetadataKey`] and [`MetadataValue`].
+///
+/// [`MetadataKey`]: struct.MetadataKey.html
+/// [`MetadataValue`]: struct.MetadataValue.html
 #[derive(Clone, Debug, Eq, PartialEq, Hash)]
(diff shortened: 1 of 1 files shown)


LucioFranco:
I wonder if we want to add `#[non_exhaustive]` to the enums?

adamchalmers:
> I wonder if we want to add `#[non_exhaustive]` to the enums?

"The non_exhaustive attribute indicates that a type or variant may have more fields or variants added in the future" [(doc)](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute), which I guess is true. 

I think the main advantage of adding `#[non_exhaustive]` is to prevent people from matching on the enum, so that if we add more variants in the future we don't cause backwards-compatibility problems. But it seems _really_ unlikely that anyone is matching on it right now, because neither `Binary` nor `Ascii` have any variants, so they're impossible to instantiate. Nobody can even _have_ a value to match on!

So I don't think it would make any practical difference. It's simple enough to add it. But I think I rule _slightly_ in favor of not adding this annotation until it actually serves a purpose... if you decide to ever add a variant to one of these, you could always add `#[non_exhaustive]` then, and you'd know it wasn't a breaking change because nobody is using these types directly. I think it just makes the code a little noisier without solving a real problem? 

Instead, I'm adding a comment to explain that this type can't be instantiated, and is only used as a type parameter in `MetadataValue`.

LucioFranco:
Yeah, that sounds fair to me. I wonder if we want to convert them to be 

```rust
struct Ascii {
   _p:  ()
}
```

So they actually can never be constructed. I don't actually know why they are enums in the first place lol.

Anyways, I appreciate the thorough responses :)

davidpdrsn:
You can construct

```rust
struct Ascii {
   _p:  ()
}
```

because the field is public within your module (the root module). It wont be constructable in other modules or crates.

Generally speaking I've moved to using `#[non_exhaustive]` for all my crates. It works the same as a private field but is more modern I'd say.

Re breaking changes: Technically one could write code like this

```rust
// defined in tonic
pub enum Ascii {}

// defined in the user's crate
trait IntoInfallible {
    fn into_infallible(self) -> Infallible;
}

impl IntoInfallible for Ascii {
    fn into_infallible(self) -> Infallible {
        match self {}
    }
}
```

which would break if `Ascii` gained private fields.

I think we should keep as they are now (but make the docs public) and in the next major convert them to

```rust
#[non_exhaustive]
pub struct Ascii;

#[non_exhaustive]
pub struct Binary;
```

LucioFranco:
Well we can make the breaking change now since they were doc hidden and thus there are not technically apart of our public API, right?

davidpdrsn:
I guess kind of gray area but yeah I'd be fine with that.

adamchalmers:
OK, done

LucioFranco (approved):
Thanks! I appreciate you following up on this!

PR #995, opened 2022-05-02
Small wording change for newbie-friendly clarity

josephleblanc:
## Motivation
I was a bit confused in the tutorial about where to put build.rs, and think this small change might help other people learning rust who use the tutorial helloworld-tonic.

## Solution
Slightly change wording to emphasize location of build.rs is in root and not /src.


josephleblanc:
@LucioFranco

LucioFranco (approved):

PR #999, opened 2022-05-04
feat(build): Better support for custom codecs

bmwill:
@LucioFranco after we chatted briefly on discord yesterday I did a bit more thinking about how improve accessibility of the code generation functionality in tonic-build. This is one potential approach. Its in a WIP state but I wanted to get your thoughts and see if this is something you'd be interested in as a contribution to tonic. If so I can spend some time actually making it usable.

## Motivation

gRPC is a general purpose RPC framework that, while generally is paired with protobuf, is agnostic to the type of serialization format used for messages and in some cases it may be desired to use a different serialization format other than protobuf. 

## Solution

This PR takes a look at one possible way to enable using a custom, non-protobuf codec for serializing messages to the wire. This is done via:
- changing the `CODEC_PATH` constant to a method which returns a `&str` for better programability
- introduces a set of builders for Services and Methods so that you can build up a service with various methods and invoke the code generation without interacting with prost or proto files. This could eventually enable developing a proc-macro that does the generation at compile time instead of needing a build script (for services which don't leverage proto files and protobuf)

### DIFF
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -1,5 +1,4 @@
-use std::env;
-use std::path::PathBuf;
+use std::{env, path::PathBuf};
 
 fn main() {
     tonic_build::configure()
@@ -30,4 +29,30 @@ fn main() {
             &["proto/googleapis"],
         )
         .unwrap();
+
+    build_json_codec_service();
+}
+
+// Manually define the json.helloworld.Greeter service which used a custom JsonCodec to use json
+// serialization instead of protobuf for sending messages on the wire.
+// This will result in generated client and server code which relies on its request, response and
+// codec types being defined in a module `crate::common`.
+//
+// See the client/server examples defined in `src/json-codec` for more information.
+fn build_json_codec_service() {
+    let greeter_service = tonic_build::manual::Service::builder()
+        .name("Greeter")
+        .package("json.helloworld")
+        .method(
+            tonic_build::manual::Method::builder()
+                .name("say_hello")
+                .route_name("SayHello")
+                .input_type("crate::common::HelloRequest")
+                .output_type("crate::common::HelloResponse")
+                .codec_path("crate::common::JsonCodec")
+                .build(),
+        )
+        .build();
+
+    tonic_build::manual::Builder::new().compile(&[greeter_service]);
 }
--- a/examples/src/json-codec/client.rs
+++ b/examples/src/json-codec/client.rs
@@ -1,0 +1,28 @@
+//! A HelloWorld example that uses JSON instead of protobuf as the message serialization format.
+//!
+//! Generated code is the output of codegen as defined in the `build_json_codec_service` function
+//! in the `examples/build.rs` file. As defined there, the generated code assumes that a module
+//! `crate::common` exists which defines `HelloRequest`, `HelloResponse`, and `JsonCodec`.
+
+pub mod common;
+use common::HelloRequest;
+
+pub mod hello_world {
+    include!(concat!(env!("OUT_DIR"), "/json.helloworld.Greeter.rs"));
+}
+use hello_world::greeter_client::GreeterClient;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let mut client = GreeterClient::connect("http://[::1]:50051").await?;
+
+    let request = tonic::Request::new(HelloRequest {
+        name: "Tonic".into(),
+    });
+
+    let response = client.say_hello(request).await?;
+
+    println!("RESPONSE={:?}", response);
+
+    Ok(())
+}
--- a/examples/src/json-codec/common.rs
+++ b/examples/src/json-codec/common.rs
@@ -1,0 +1,80 @@
+//! This module defines common request/response types as well as the JsonCodec that is used by the
+//! json.helloworld.Greeter service which is defined manually (instead of via proto files) by the
+//! `build_json_codec_service` function in the `examples/build.rs` file.
+
+use bytes::{Buf, BufMut};
+use serde::{Deserialize, Serialize};
+use std::marker::PhantomData;
+use tonic::{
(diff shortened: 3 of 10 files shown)


LucioFranco (commented):
I like this a lot! I think just need to add some docs and we can merge this. Thanks!

tonic-build/src/rust.rs
```diff
@@ -0,0 +1,365 @@
...
+        let mut output = String::new();
+        generator.finalize(&mut output);
+
+        //TODO group services into packages and put all packages into separate files
+        let out_file = out_dir.join("tonic.rs");
+        fs::write(out_file, output).unwrap();
+    }
+}
+
+#[test]
```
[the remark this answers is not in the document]
bmwill: Yeah i just whipped this together to see what you thought so now that I know you're interested i'll clean it up with docs and the like

bmwill:
Ok this should be ready for a serious review now. My main question is how you want to handle the module naming, right now its all just in a public module `rust` and nothing is reexported at the top level.

LucioFranco:
@bmwill how about `manual` instead of `rust` for the module name?

examples/build.rs
```diff
@@ -30,4 +29,24 @@ fn main() {
...
+    let greeter_service = tonic_build::rust::Service::builder()
+        .name("Greeter")
+        .package("json.helloworld")
+        .method(
+            tonic_build::rust::Method::builder()
+                .name("say_hello")
+                .route_name("SayHello")
+                .input_type("crate::HelloRequest")
+                .output_type("crate::HelloResponse")
+                .codec_path("crate::JsonCodec")
```
LucioFranco, line 46: should we make these `crate::common` to be more clear this is tying in the types from that module file? My concern is that this build.rs is quite far from that code and crate seems a bit less obvious.

examples/src/json-codec/client.rs
```diff
@@ -0,0 +1,23 @@
+use common::{HelloRequest, HelloResponse, JsonCodec};
```
LucioFranco, line 1: We should add a module comment explaining/pointing out that the codegen comes from `build_json_codec_service` in the `build.rs` file.

examples/src/json-codec/common.rs
```diff
@@ -0,0 +1,76 @@
+use bytes::{Buf, BufMut};
```
LucioFranco, line 1: Could also use some module docs here maybe explaining what is going on. Nothing crazy is needed imo

tonic-build/src/rust.rs
```diff
@@ -0,0 +1,444 @@
...
+
+use super::{client, server, Attributes};
+use proc_macro2::TokenStream;
+use quote::ToTokens;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
+
+/// Service builder.
```
LucioFranco: Could use a bit more docs here

tonic-build/src/manual.rs
```diff
@@ -0,0 +1,444 @@
...
+
+use super::{client, server, Attributes};
+use proc_macro2::TokenStream;
+use quote::ToTokens;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
+
+/// Service builder.
```
LucioFranco, line 39: I think we should expand this doc comment to explain high level what these builders do (I don't think we need to add more to the other builders only this one).

tonic-build/src/manual.rs
```diff
@@ -0,0 +1,444 @@
...
+    methods: Vec<Method>,
+}
+
+impl ServiceBuilder {
+    /// Set the name for this Service.
+    ///
+    /// This value will be used both as the base for the generated rust types and service trait as
+    /// well as part of the route for calling this service. Routes have the form:
+    /// `/<package_name>.<service_name>/<method_route_name>`
+    pub fn name<T: AsRef<str>>(mut self, name: T) -> Self {
```
LucioFranco, line 58: small nit ive been using impl trait here for these types of conversions, wonder if we should switch these?

tonic-build/src/manual.rs
```diff
@@ -0,0 +1,444 @@
...
+            input_type: self.input_type.unwrap(),
+            output_type: self.output_type.unwrap(),
+            client_streaming: self.client_streaming,
+            server_streaming: self.server_streaming,
+            codec_path: self.codec_path.unwrap(),
+        }
+    }
+}
+
+struct ServiceGenerator {
```
LucioFranco, line 311: Do we actually need this extra struct? I don't mind keeping it just seems like we could just embed the code in the `compile` method?

tonic-build/src/manual.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+}
+
+impl Builder {
+    /// Create a new Builder
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Enable or disable gRPC client code generation.
```
LucioFranco, line 398: prob wanna mention the default here? (maybe we missed this also in the prost side).

examples/build.rs
```diff
@@ -30,4 +29,24 @@ fn main() {
             &["proto/googleapis"],
         )
         .unwrap();
+
+    build_json_codec_service();
+}
+
+fn build_json_codec_service() {
```
LucioFranco, line 36: I think we should add a comment to point to the example here

LucioFranco (changes requested):
Overall looks great left a few comments/questions.

tonic-build/src/manual.rs
LucioFranco: we can keep it nbd

bmwill:
@LucioFranco Ok I've updated the PR, hopefully addressing all of your comments

LucioFranco (approved):
Thanks! LGTM

zbysir:
@bmwill Take a look at this: https://github.com/hyperium/tonic/issues/851, can you give me some advice? First of all thank you.

What I want most is to support multiple codecs, not a specific one.

PR #1004, opened 2022-05-20
feat!: remove codegen dependency on compression feature

neoeinstein:
## Motivation

An issue was raised in neoeinstein/protoc-gen-prost#6 and tokio-rs/console#349. When compiled, the `compression` feature flag used to generate code from `tonic-build` _must_ match the feature flag used for `tonic`. This can result in problems if code is generated and then checked in rather than being generated as part of each build.

## Solution

This PR removes the `compression` feature flag from `tonic-build`, making it agnostic to the compression feature flag that may later be used for `tonic`. In support of this, I've removed the `compression` feature flag from `tonic` as well, replacing it with a more granular `gzip` feature flag.

Through the changes in the PR, when compiled without `gzip`, `EnabledCompressionEncodings` and `CompressionEncodings` both become zero-sized types. In addition, because `CompressionEncodings` ends up becoming an uninhabited enum, dead code elimination can remove most of the compression machinery at compile time, maintaining most of the benefits of having the more general `compression` feature flag.

One area that does have minor detriment is the `tonic::codec::decode::Streaming<T>`, which grows by 2 pointers, 2 sizes, and 1 byte (33 bytes on x86-64). This type is already several pointers and lengths (183 bytes), so this is an increase of about 18% to 216 bytes.

In order to better support removing the `compression` flag, `accept_gzip()` and `send_gzip()`, where used, were replaced with `accept_compressed(encoding)` and `send_compressed(encoding)`. This allows for better forward-compatibility in the event support for a future compression encoding is added, but does have the expense of being a little more verbose:

```rust
let client = TestClient::new(channel).accept_gzip();
```

becomes

```rust
let client = TestClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
```

⚠️ Note: The change to the feature flags and the change to the interface for compression do constitute breaking changes.

## `transport` feature flag

This same issue may also affect the `transport` feature flag. This PR does not make any attempt to reduce out that issue.


tonic/src/codec/compression.rs
```diff
@@ -10,35 +11,53 @@ pub(crate) const ACCEPT_ENCODING_HEADER: &str = "grpc-accept-encoding";
...
 impl EnabledCompressionEncodings {
-    /// Check if `gzip` compression is enabled.
-    pub fn gzip(self) -> bool {
-        self.gzip
+    /// Check if a [`CompressionEncoding`] is enabled.
+    pub fn is_enabled(&self, encoding: CompressionEncoding) -> bool {
+        match encoding {
+            #[cfg(feature = "gzip")]
+            CompressionEncoding::Gzip => self.gzip,
+        }
```
LucioFranco, line 24: how does this compile without the gzip feature?

LucioFranco (commented):
Overall this looks good! I'd like @davidpdrsn to take a look at this as well.

tonic/src/codec/compression.rs
neoeinstein: Without the `gzip` feature, `CompressionEncoding` becomes an enum with no variants, so it can't be constructed, similar to `core::convert::Infallible`. Thus, this match, even with no arms, is exhaustive. The Rust compiler assigns the `!` type to these types of `match` expressions which are exhaustive, but have no arms.

I use a similar strategy for `aliri`, which _can_ be build without support for any signing algorithm.

CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
...
+### Breaking Changes
+
+* **build:** Remove `compression` feature
+* **build:** Generated `accept_gzip` and `send_gzip` replaced with `accept_compressed` and `send_compressed`
+* **tonic:** Remove `compression` feature, add `gzip`
+* **tonic:** `accept_gzip` and `send_gzip` replaced with `accept_compressed` and `send_compressed`
+
+### Features
+
+* Code generated with `tonic-build` will now work regardless of whether `tonic` compression support is enabled
```
davidpdrsn, line 12: I believe we usually fills in the changelog when doing releases so perhaps this change should be removed. @LucioFranco knows :)

davidpdrsn (approved):
Sorry for the delay in reviewing this!

It looks good to me :)

CHANGELOG.md
LucioFranco: Oh Ya good point if you back this out we will add this during release.

neoeinstein: Dropped that commit off the end.

LucioFranco:
ci failures :( ping me on discrod and ill get the ci to run again

PR #1005, opened 2022-05-21
tonic: add methods to retrieve the uri and the method from a Request

QuentinPerez:
## Motivation

Fixes https://github.com/hyperium/tonic/issues/772

## Solution

Two new methods have been added to `Request`:
 - `uri()` to get a `&Uri` which is now owned by `Request` 
 - `method()` to get the method name of the RPC

## Misc

I've also reduced the clones done by the interceptor service, let me if it's useful and if yes if you want a dedicated PR.

### DIFF
--- a/tonic/src/client/grpc.rs
+++ b/tonic/src/client/grpc.rs
@@ -248,8 +248,13 @@ impl<T> Grpc<T> {
             })
             .map(BoxBody::new);
 
+        let (headers, extensions, _, msg) = request.into_parts();
+
+        // inject uri into the request
+        let request =
+            Request::from_http_parts(headers.into_headers(), extensions.into_http(), uri, msg);
+
         let mut request = request.into_http(
-            uri,
             http::Method::POST,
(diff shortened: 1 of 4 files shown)


QuentinPerez:
feel free to close it, if there is no need :)

LucioFranco:
Hey! Sorry for the extreme delay on this, I was out on vacation. So I've been thinking about this for a bit (been open as a tab for ever now haha). The original issue wanted this feature to use it within the `Interceptors` but the direction tonic will eventually go in will be to remove the interceptors in-order for users to just use tower layers. Since that is the case users will get access to the `http::Request` type instead of the tonic version.

That said, I am curious what your motivation for this is?

QuentinPerez:
No problem, I do understand the infinite tabs :)
The motivation was to close some opened issues, you are right, I questioned myself during the implementation about why we could not use tower layers directly, but now it's clearer, closing it then!

PR #1013, opened 2022-06-12
tonic: Introduce a new method on Endpoint to override the origin

QuentinPerez:
## Motivation

At @znly, we use Envoy as a service mesh to distribute the traffic across the services.
Meaning that:
- we have an auto-scaled farm of Envoy which proxy the entire traffic.
- we have to refresh the IPs periodically (due to auto-scaling).
- we have to talk to the IPs directly.

That said, the requests end with an unavailable error, why? because Envoy is called with the URI (the IP in our case) and can't determine which service we are trying to reach.

## Solution

A new method named `origin` has been added to override the origin sent.
We tried it in production and it's working fine.
Let me know if you have a better solution.

### DIFF
--- a/tests/integration_tests/tests/origin.rs
+++ b/tests/integration_tests/tests/origin.rs
@@ -1,0 +1,100 @@
+use futures::future::BoxFuture;
+use futures_util::FutureExt;
+use integration_tests::pb::test_client;
+use integration_tests::pb::{test_server, Input, Output};
+use std::task::Context;
+use std::task::Poll;
+use std::time::Duration;
+use tokio::sync::oneshot;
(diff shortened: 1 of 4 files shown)


QuentinPerez:
@LucioFranco 👋🏻
Let me know if you need more explanation or time to review this PR

LucioFranco (changes requested):
LGTM only thing missing I think is a simple unit test.

LucioFranco (approved):
Thanks!

QuentinPerez:
@LucioFranco thx ! do you think that we could have a minor soon ? do not want to stay on my fork too long :)

PR #1015, opened 2022-06-13
chore(reflection): Use write macro instead of write_fmt

tottoto:
## Motivation

Refactoring.

## Solution

Follows to the document of [`write_fmt`](https://doc.rust-lang.org/stable/std/fmt/trait.Write.html#method.write_fmt).


LucioFranco (approved):

LucioFranco:
Thanks!

PR #1017, opened 2022-06-15
feat: Add `Grpc::with_origin` for clients

LucioFranco:


tonic/src/client/grpc.rs
```diff
@@ -296,6 +316,7 @@ impl<T: Clone> Clone for Grpc<T> {
     fn clone(&self) -> Self {
         Self {
             inner: self.inner.clone(),
+            origin: self.origin.clone(),
```
kvc0, line 319: While I cannot judge whether this clone or the `ServiceBuilder`’s clone would be more efficient, this tradeoff with the scheme and authority being owned by the `Grpc` seems right to optimize for the expected use cases, making it clear how to replace Channels going forward without needing to learn Tower as well just to get started.

kvc0 (approved):
Thanks for this; I’m looking forward to using it soon!

LucioFranco:
@kvc0 thanks for the feedback!

Issue #1020, opened 2022-06-17
Automatically emitting rerun-if-changed instructions for cargo


mickvangelderen:
### Discussed in https://github.com/hyperium/tonic/discussions/1019

<div type='discussions-op-text'>

<sup>Originally posted by **mickvangelderen** June 17, 2022</sup>
I find myself writing a build script similar to the following for every project:

```rust
fn main() {
    let protos = ["proto/rocsys/canonical.proto"];

    let includes = ["proto"];

    tonic_build::configure()
        .server_mod_attribute("rocsys.canonical", "#[cfg(feature = \"server\")]")
        .client_mod_attribute("rocsys.canonical", "#[cfg(feature = \"client\")]")
        .compile(&protos, &includes)
        .unwrap();

    for path in protos.iter().copied().chain(includes.iter().copied()) {
        println!("cargo:rerun-if-changed={}", path)
    }
}
```

I was wondering if we could let tonic emit those rerun-if-changed statements for us. It would be even better if we could narrow down the rerun-if-changed instructions only to the included files rather than asking cargo to monitor the entire include directories.

It is possible that this feature already exists but I couldn't find any signs of tonic_build writing to `stdout` nor does my build script print anything to `stdout` when removing my own `println` statements.
</div>

Is this indeed a missing feature? If so, should it maybe be added to `tonic_build` (perhaps under a configuration flag since not everyone using `tonic_build` may be using `cargo`).

LucioFranco:
Oh we should for sure add this! I would totally accept a PR for this.

PR #1021, opened 2022-06-17
feat(build): Add option to emit rerun-if-changed instructions

mickvangelderen:
Fixes: #1020
Refs: #1019

## Motivation

Currently, tonic-build does not emit cargo:rerun-if-changed
instructions. This will cause the build script to be rerun if any file
changes in the package.

## Solution

This PR implements an option on the Builder called
emit_rerun_if_changed that, when enabled, will emit the
cargo:rerun-if-changed instructions for all of the proto file paths
and include directories.

### DIFF
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -373,7 +373,7 @@ impl Builder {
     /// Enable or disable emitting
     /// [`cargo:rerun-if-changed=PATH`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed)
     /// instructions for Cargo.
-    /// 
+    ///
     /// If set, writes instructions to `stdout` for Cargo so that it understands
     /// when to rerun the build script. By default, this setting is enabled if
     /// the `CARGO` environment variable is set. The `CARGO` environment
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -27,6 +27,7 @@ pub fn configure() -> Builder {
         emit_package: true,
         protoc_args: Vec::new(),
         include_file: None,
+        emit_rerun_if_changed: std::env::var_os("CARGO").is_some(),
     }
 }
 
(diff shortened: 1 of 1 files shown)


tonic-build/src/prost.rs
```diff
@@ -27,6 +27,7 @@ pub fn configure() -> Builder {
         emit_package: true,
         protoc_args: Vec::new(),
         include_file: None,
+        emit_rerun_if_changed: std::env::var_os("CARGO").is_some(),
```
mickvangelderen, line 30: Not sure if run-time detection of whether the code is run as a build script is
1. right, and
2. necessary

We could also just set it to `true` by default.

tonic-build/src/prost.rs
```diff
@@ -368,6 +370,14 @@ impl Builder {
         self
     }
 
+    /// Enable or disable emitting the rerun-if-changed cargo instruction
+    /// 
+    /// If set, emits rerun-if-changed instructions so that Cargo understands when to rerun the build script.
```
mickvangelderen, line 375: Is this the only place I need to document this option?

mickvangelderen (commented):
Do I need to edit a changelog or any of the examples?

tonic-build/src/prost.rs
LucioFranco: Yeah I think this is fine, I would also add info about the default setting that if you're in a build script this will be enabled and outside of a build script it won't

tonic-build/src/prost.rs
mickvangelderen: I skimmed through the [cargo reference](https://doc.rust-lang.org/cargo/reference/build-scripts.html) but did not find anything else. I also failed to find results for how to detect Rust code is being run from a build script through Google. I hope this does a good enough job.

LucioFranco:
@mickvangelderen just need you to run a fmt pass

mickvangelderen:
> @mickvangelderen just need you to run a fmt pass

That was sloppy, sorry.

LucioFranco:
and no problem! its easy to miss the formatting, I usually just have it auto fmt on save.

PR #1022, opened 2022-06-18
add IntoResponse impl for Status

jvdwrf:
## Motivation

I'm currently working on a crate that allows you to nest tonic_services directly in axum routers.

While working on this, I've had to create a wrapper for `tonic::Status`, which implements into-response: https://github.com/jvdwrf/axum-tonic/blob/d86f102225ed0a22e864fe388fb78d151078e384/src/status.rs#L27-L32
It would be much more ergonomic if this would not be necessary.

## Solution

Implement it directly on `tonic::Status`

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -107,6 +107,13 @@ pub enum Code {
     Unauthenticated = 16,
 }
 
+impl axum::response::IntoResponse for Status {
+    fn into_response(self) -> axum::response::Response {
+        let (parts, tonic_body) = self.to_http().into_parts();
+        axum::response::Response::from_parts(parts, axum::body::boxed(tonic_body))
+    }
+}
+
 impl Code {
     /// Get description of this `Code`.
     /// ```


tonic/src/status.rs
```diff
@@ -107,6 +107,13 @@ pub enum Code {
     Unauthenticated = 16,
 }
 
+impl axum::response::IntoResponse for Status {
+    fn into_response(self) -> axum::response::Response {
+        let (parts, tonic_body) = self.to_http().into_parts();
+        axum::response::Response::from_parts(parts, axum::body::boxed(tonic_body))
+    }
+}
```
davidpdrsn, line 115: This makes axum a public dependency on tonic, which I don't think we should have.

Then a breaking release of axum requires tonic to ship a breaking release.

I suggest instead you make your own helper function in your crate. Though it's not quite as elegant.

davidpdrsn (changes requested):

tonic/src/status.rs
jvdwrf: Mhm yea that makes sense, didn't think about that.

PR #1043, opened 2022-07-28
reflection: export server types

zombiezen:
## Motivation

Makes it easier to store a reflection server in a variable or get the gRPC service's name.


LucioFranco (approved):
Seems fine to me, thanks!

PR #1046, opened 2022-07-31
Update docs to reflect changes to compression feature

nashley:
The [current docs](https://docs.rs/tonic/0.8.0/tonic/index.html#feature-flags) still reference the outdated feature flag `compression`, which has been superseded by `gzip` in `tonic` and removed from `tonic-build` entirely.

See #1004 (commit a585a72) for more information on the changes to the `compression` feature in `tonic` `0.8`.

## Motivation

Have the docs reflect the current state of the crate(s).

## Solution

Fix the `tonic` docs to reference `gzip` and to no longer state that `tonic-build` also needs a flag to be specified.

### DIFF
--- a/tonic/src/lib.rs
+++ b/tonic/src/lib.rs
@@ -32,6 +32,7 @@
 //! - `prost`: Enables the [`prost`] based gRPC [`Codec`] implementation.
 //! - `gzip`: Enables compressing requests, responses, and streams.
 //! Depends on [flate2]. Not enabled by default.
+//! Replaces the `compression` flag from earlier versions of `tonic`.
 //!
 //! # Structure
 //!
--- a/tonic/src/lib.rs
+++ b/tonic/src/lib.rs
@@ -30,9 +30,8 @@
 //! - `tls-webpki-roots`: Add the standard trust roots from the `webpki-roots` crate to
 //! `rustls`-based gRPC clients. Not enabled by default.
 //! - `prost`: Enables the [`prost`] based gRPC [`Codec`] implementation.
-//! - `compression`: Enables compressing requests, responses, and streams. Note
-//! that you must enable the `compression` feature on both `tonic` and
-//! `tonic-build` to use it. Depends on [flate2]. Not enabled by default.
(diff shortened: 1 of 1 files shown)


nashley:
Should a note be included in the docs indicating that `compression` has been superseded by `gzip` to assist developers in upgrading? In my mind, that's better suited for the changelog, which is why I didn't add it here.

LucioFranco:
@nashley yeah I think we should add that upgrade note

LucioFranco (approved):
Good catch, I think if we add your last suggestion we can merge this and I can get a patch release out.

nashley:
Does that seem like enough explanation, or should I link/recreate the comments in https://github.com/hyperium/tonic/pull/1004?

tonic/src/lib.rs
```diff
@@ -30,9 +30,9 @@
 //! - `tls-webpki-roots`: Add the standard trust roots from the `webpki-roots` crate to
 //! `rustls`-based gRPC clients. Not enabled by default.
 //! - `prost`: Enables the [`prost`] based gRPC [`Codec`] implementation.
-//! - `compression`: Enables compressing requests, responses, and streams. Note
-//! that you must enable the `compression` feature on both `tonic` and
-//! `tonic-build` to use it. Depends on [flate2]. Not enabled by default.
+//! - `gzip`: Enables compressing requests, responses, and streams.
+//! Depends on [flate2]. Not enabled by default.
+//! Replaces the `compression` flag from earlier versions of `tonic`.
```
LucioFranco, line 35: Probably should include <=0.7 to specify which version?
```suggestion
//! Replaces the `compression` flag from earlier versions of `tonic` (<= 0.7).
```

LucioFranco (approved):
Thanks!

matze:
> In my mind, that's better suited for the changelog, which is why I didn't add it here.

I would have appreciated an entry in the CHANGELOG but I figured it out on my own :laughing:

LucioFranco:
Oh yes, I made a mistake when I merged this, sorry about that! Let me get that fixed here https://github.com/hyperium/tonic/pull/1055

PR #1051, opened 2022-08-04
Make errors in Routes Infallible.

BratSinot:
Greetings!

## Motivation

So, why there `boxed error` instead of `Infallible`? There are no errors in these part, but because there `boxed error` it makes problem to use like this:
P.S. `MultiplexService` taken from there: https://github.com/tokio-rs/axum/blob/main/examples/rest-grpc-multiplex/src/multiplex_service.rs
```rust
    let grpc = tonic::transport::Server::builder()
        .add_service(WorkspaceServiceServer::new(workspace))
        .add_service(health_service)
        .into_service();

    let rest = axum::Router::new().route("/", get(|| async { "Hello".to_owned() }));

    let service = MultiplexService::new(management, grpc);

    axum::Server::bind(&address)
        .serve(tower::make::Shared::new(service))
        .await
        .unwrap();
```


LucioFranco:
Hmm this seems fine, only issue is that tonic just released a breaking change, so this will have to wait a bit. In addition, eventually the whole transport module will go away so I suspect it we will fix it via this or just remove all this code (still figuring out those timelines).

Issue #1034, opened 2022-07-16
Improve gRPC richer error model support


flemosr:
### Motivation

The [gRPC richer error model] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages. Implementing it using the [standard error messages] directly (as shown in [#401]) is not very convenient. Specially when sending multiple additional error detail messages.

### Proposal

I recently created a crate ([tonic-richer-error]) that implements a trait to `Status` (and provides some structs) aiming to make the implementation of the richer error model more straightforward. It seems to me that incorporating this crate to tonic as an auxiliary crate would be fairly simple. I could transfer the ownership of the released crate or it could be published after revision with another name.

### Alternatives

A trait similar to [WithErrorDetails] could be implemented on the tonic crate itself.

[gRPC richer error model]: https://www.grpc.io/docs/guides/error/
[standard error messages]: https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
[#401]: https://github.com/hyperium/tonic/issues/401
[tonic-richer-error]: https://github.com/flemosr/tonic-richer-error
[WithErrorDetails]: https://docs.rs/tonic-richer-error/0.2.1/tonic_richer_error/trait.WithErrorDetails.html

LucioFranco:
Hey! Sorry for the delay on this but I finally got some time to fully understand what you're proposing. I think this fits very well into `tonic-types` in similar vain to `prost-types` but for gRPC. I would totally accept a PR that adds this support there. Happy to mentor and help out.

flemosr:
Hi! Yes, I believe it would fit nicely into `tonic-types` too. I will start working on this PR.

flemosr:
Solved by #1317 .

PR #1060, opened 2022-08-18
feat(types): Add gRPC Richer Error Model support

flemosr:
## Motivation

The [gRPC richer error model] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages.

## Solution

This PR adds richer error model support to `tonic-types`, as per [#1034]. Example servers and clients were also added.

[gRPC richer error model]: https://www.grpc.io/docs/guides/error/
[#1034]: https://github.com/hyperium/tonic/issues/1034


flemosr:
Note: there are links at the bottom of the added `README.md` that are supposed to point to the crate docs. They would have to be updated to point to the correct released version of `tonic-types`.

LucioFranco:
Thank you for this! This is quite a large PR and will likely take me a while to review. Do you think you could reduce this into multiple PRs that we can merge independently?

flemosr:
Hi! Happy to contribute. I think the new code can be more or less evenly splitted between the functionallity corresponding to each standard error message type (`BadRequest`, `RetryInfo`, etc). I could add only the functionallity corresponding to `BadRequest` on the first PR, and add support for the other error types in subsequent PR's. Would this approach work better?

LucioFranco:
@flemosr yeah that would be great! Thanks

flemosr:
Hi! The first PR is still quite big, but most of it is the new proto file, docs and tests. The following ones can be smaller

PR #1064, opened 2022-08-19
Add From<T> for Response<T>

amrhassan:
Hello,

## Motivation

I've noticed that there's no `From<T>` implementation for `Response<T>` and thought it was worth having for the times when writing `t.into()` is snappier than `Response::new(t)`.

## Solution

I've added the trait impl.


LucioFranco:
I am trying to remember why we didn't include this. I know for requests we have `IntoRequest`.  But I can't come up with why we shouldn't include this. Would love to hear from others @neoeinstein @davidpdrsn

davidpdrsn (approved):
I think ergonomics-wise it doesn't make much difference but don't think it could hurt to add.

LucioFranco:
This is going to have to wait for the next breaking change release.

tottoto (approved):

PR #1071, opened 2022-09-06
feat(transport): Expose hyper's H2 adaptive window on server

adamchalmers:
## Motivation

HTTP/2 window sizes determine how many bytes the client or server is willing to allow in-flight at one time. Clients and servers maintain their own independent windows with their own independently-chosen sizes. Hyper's h2 client and servers can use "adaptive flow control" to dynamically adjust this size during the life of a connection. You can read about their adaptive scaling in the relevant Hyper [issue](https://github.com/hyperium/hyper/issues/1960) or [PR](https://github.com/hyperium/hyper/pull/2138).

Tonic lets users enable this [for clients](https://docs.rs/tonic/latest/tonic/transport/channel/struct.Endpoint.html#method.http2_adaptive_window) but not for servers.

## Solution

Add a new method to `transport::Server` which lets users set the underlying HTTP/2 server's adaptive window mode.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -90,6 +90,7 @@ pub struct Server<L = Identity> {
     tcp_nodelay: bool,
     http2_keepalive_interval: Option<Duration>,
     http2_keepalive_timeout: Option<Duration>,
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
thanks!

Issue #1084, opened 2022-09-23
Add `tonic::Result`


LucioFranco:
We should add a `tonic::Result<T>` which maps to `Result<T, Status>`.

punkeel:
https://github.com/hyperium/tonic/pull/1085 was merged, I believe this issue can be closed

PR #1085, opened 2022-09-23
add `Result<T>` type alias for `std::result::Result<T, tonic::Status>`

stphnsmpsn:
## Motivation

#1084 

## Solution

Added a simple alias in `lib.rs`: `pub type Result<T, E = Status> = std::result::Result<T, E>;`


tonic/src/lib.rs
```diff
@@ -122,3 +122,5 @@ pub(crate) type Error = Box<dyn std::error::Error + Send + Sync>;
 #[cfg(feature = "codegen")]
 #[cfg_attr(docsrs, doc(cfg(feature = "codegen")))]
 pub mod codegen;
+
+pub type Result<T> = std::result::Result<T, Status>;
```
davidpdrsn, line 126: This way users can still override the error type if they want.

```suggestion
pub type Result<T, E = Status> = std::result::Result<T, E>;
```

davidpdrsn: Also missing docs.

LucioFranco: TIL

LucioFranco (approved):
LGTM thanks!

LucioFranco:
One cI failure though that is probably cuz of formatting https://github.com/hyperium/tonic/actions/runs/3116115959/jobs/5054123932

stphnsmpsn:
@LucioFranco , it was a formatting issue my bad. Passes `cargo fmt` now.

PR #1068, opened 2022-08-26
feat(types): Add gRPC Richer Error Model support (BadRequest)

flemosr:
## Motivation

The [gRPC Richer Error Model][error-handling] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages.

## Solution

This PR is the first step to add richer error model support to `tonic-types`, as per [#1060]. It introduces the `WithErrorDetails` trait and adds support to the `BadRequest` standard error message type.

[error-handling]: https://www.grpc.io/docs/guides/error
[#1060]: https://github.com/hyperium/tonic/pull/1060

PR #1095, opened 2022-09-29
feat(types): Add gRPC Richer Error Model support (RetryInfo)

flemosr:
## Motivation

The [gRPC Richer Error Model][error-handling] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages.

## Solution

This PR continues the work initiated at [#1068]. It improves the organization of `tonic-types` modules, and adds support for the `RetryInfo` standard error message type to that crate.

[error-handling]: https://www.grpc.io/docs/guides/error
[#1068]: https://github.com/hyperium/tonic/pull/1068


LucioFranco:
Oh somehow this totally fell off my radar let me review this on monday! sorry about the delay.

tonic-types/src/richer_error/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+}
+
+fn gen_details_bytes(code: Code, message: &String, details: Vec<Any>) -> Bytes {
+    let status = pb::Status {
+        code: code as i32,
+        message: message.clone(),
+        details,
+    };
+
+    Bytes::from(status.encode_to_vec())
```
LucioFranco, line 30: Why encode to vec instead of using BytesMut + BufMut and regular `encode`? So that we avoid the vec round trip?

tonic-types/src/richer_error/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+}
+
+/// Used to implement associated functions and methods on `tonic::Status`, that
+/// allow the addition and extraction of standard error details.
+pub trait StatusExt {
```
LucioFranco, line 35: We probably want to seal this trait similar to https://docs.rs/tower-http/latest/src/tower_http/builder.rs.html#45 that way outside users can't implement it and we can add more required fn as we please without making a breaking change.

tonic-types/src/richer_error/std_messages/retry_info.rs
```diff
@@ -0,0 +1,151 @@
+use std::{convert::TryFrom, time};
+
+use prost::{DecodeError, Message};
+use prost_types::Any;
+
+use super::super::pb;
+use super::super::{FromAny, IntoAny};
```
LucioFranco, line 7: Nit: I like combining these since they have a similar path.

LucioFranco (changes requested):
LGTM overall have a few comments once those are addressed we can merge. Sorry for the delay again!

tonic-types/src/richer_error/mod.rs
flemosr: Using `BytesMut` indeed makes much more sense, thanks! It seems to me that using `BufMut` is not necessary in this case but I may be missing something.

tonic-types/src/richer_error/mod.rs
flemosr: You are right. I did not know this pattern, quite useful. Thanks!

tonic-types/src/richer_error/std_messages/retry_info.rs
flemosr: I missed that, thanks!

flemosr:
No problem! Hopefully I addressed all comments but please let me know if you think we should make any further changes.

LucioFranco (approved):
LGTM!

PR #1102, opened 2022-10-06
feat: support the SSLKEYLOGFILE environment variable for clients

bgeron:
Nothing changes unless SSLKEYLOGFILE is set in the environment.

If SSLKEYLOGFILE is set, then every time the client connnects, it will append a line to the file designated in SSLKEYLOGFILE with the pre-master secret for that connection.

Point a tool like Wireshark to that file, and suddenly it can decode the TLS conversations.

https://wiki.wireshark.org/TLS#using-the-pre-master-secret

https://docs.rs/rustls/0.20.6/rustls/struct.KeyLogFile.html

https://docs.rs/rustls/0.20.6/rustls/trait.KeyLog.html

### DIFF
--- a/tonic/src/transport/service/tls.rs
+++ b/tonic/src/transport/service/tls.rs
@@ -81,6 +81,7 @@ impl TlsConnector {
             None => builder.with_no_client_auth(),
         };
 
+        config.key_log = Arc::new(tokio_rustls::rustls::KeyLogFile::new());
         config.alpn_protocols.push(ALPN_H2.as_bytes().to_vec());
         Ok(Self {
             config: Arc::new(config),


LucioFranco:
I am leaning to not including this in tonic since its possible to configure this manually via rustls. The support for tls within tonic is quite primitive and will actually be removed in the future. So I recommend moving away from using the `transport` module in tonic.

bgeron:
Short of implementing a TLS connector myself, I think it's not possible to configure this manually. I can specify the `tonic::transport::ClientTlsConfig` but not the `rustls::ClientConfig`.

Alternatively I could make this a boolean flag to go into `ClientTlsConfig`. I guess that would be better. It could be off by default, like it is in `rustls`. 

But if you want to take TLS out of Tonic, I respect that of course. Thank you for making/maintaining Tonic.

LucioFranco:
Yes, I would use the `hyper-rustls` and make sure to only enable http2 then I think it should work. You can either use a customized incoming impl or you can use hyper directly.

esemeniuc:
I would love to have this for local debugging

PR #1110, opened 2022-10-13
chore: bumping dependency `axum` to version 0.5.16

Will-Low:
## Motivation
There is a security vulnerability in the `axum-core` crate ([RUSTSEC-2022-0055](https://rustsec.org/advisories/RUSTSEC-2022-0055)), which is used in `axum` and subsequently `tonic`.

I'm not sure if directly impacts `tonic` or not, but it may be related to https://github.com/hyperium/tonic/issues/1097.

## Solution
Proposing to upgrade the version of `axum` to 0.5.16, which contains the patched version of `axum-core`.


davidpdrsn:
It doesn't impact tonic. See https://github.com/hyperium/tonic/issues/1087

But @LucioFranco looks like https://github.com/hyperium/tonic/pull/1088 went to 0.5.15 when it should have been 0.5.16 😅

davidpdrsn (approved):

LucioFranco:
oh oops

PR #1113, opened 2022-10-14
test(tonic-web): fix grpc_web_content_types() test not covering all cases

nguiard:
## Motivation

The grpc_web_content_types() test function had a duplicate `GRPC_WEB_PROTO` in the array, where one of these should have been `GRPC_WEB_TEXT_PROTO`.


LucioFranco (approved):
Nice catch! Thanks!

PR #1123, opened 2022-10-25
Remove tonic-web custom cors implementation and replace it with tower-http `CorsLayer`

### LINKED ISSUE #1122: Cleanup tonic-web cors support and replace it with tower-http
## Motivation

The CORS code in tonic-web is not 100% compliant with cors spec and cannot be detached from the rest of the tonic-web protocol implementation. This can be problematic for some use cases where the tonic-web layer is just a part of the server, and some other cors layer is used before (as far as i understand in particular for https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests).

Some other grpc-web implementations where cors is separate from the protocol implementation:

* https://armeria.dev/docs/server-grpc/#grpc-web
* https://github.com/grpc/grpc-web/blob/35284bfe156fc41bbcdd554ac423a587d93ff8da/net/grpc/gateway/examples/helloworld/envoy.yaml#L30

## How

As described here: https://github.com/hyperium/tonic/pull/1119#issuecomment-1285123214. the idea is to:

* Remove all the tonic-web cors handling logic
* `tonic_web::enable` composes `CorsLayer` from tonic-http and `GrpcWebLayer` introduced by #1119. If a user wants more control over the `CorsLayer`, they can just instantiate manually `GrpcWebLayer` and compose it with whatever configured cors layer they want (not necessarily the one provided by tower-http)

## Breaking changes

* `Config` gets removed
* Behavioral changes for the cors layer, more conform to the spec

## Notes

* The current cors implementation is "strict", returning an error status code when headers, origin and whatnot are not compatible with the cors configuration, e.g. https://github.com/hyperium/tonic/blob/b409ddd478959e239aeef3cb8715cd3ace470a8f/tonic-web/src/service.rs#L300
* This seems to be the opposite of what should be, see https://github.com/tower-rs/tower-http/pull/224 and in particular the fetch spec https://fetch.spec.whatwg.org/#http-responses

> A successful HTTP response to a CORS-preflight request is similar, except it is restricted to an ok status, e.g., 200 or 204.
> Any other kind of HTTP response is not successful and will either end up not being shared or fail the CORS-preflight request. 

This because in the preflight request the server just replies back to the browser with its cors configuration, then it's up to the browser to figure out whether it should proceed or not with the actual request.

slinkydeveloper:
Fix #1122, see the issue for more details.

This PR depends on #1119


slinkydeveloper:
I tested this branch on a personal project, connecting with connect-web client with [grpc-web transport](https://connect.build/docs/web/choosing-a-protocol#grpc-web), and it works fine.

tonic-web/src/lib.rs
```diff
@@ -87,49 +87,58 @@
...
+use tower_http::cors::{AllowOrigin, Cors, CorsLayer};
+use tower_layer::Layer;
 use tower_service::Service;
 
-/// enable a tonic service to handle grpc-web requests with the default configuration.
+const DEFAULT_MAX_AGE: Duration = Duration::from_secs(24 * 60 * 60);
+const DEFAULT_EXPOSED_HEADERS: [&str; 3] =
+    ["grpc-status", "grpc-message", "grpc-status-details-bin"];
+const DEFAULT_ALLOW_HEADERS: [&str; 4] =
+    ["x-grpc-web", "content-type", "x-user-agent", "grpc-timeout"];
```
slinkydeveloper, line 110: Perhaps we could make these `pub` for people instantiating the `CorsLayer` manually?

LucioFranco: Is there a big benefit here? I personally like to stay on the conservative side of making things public just incase you need to change them in the future.

tonic-web/src/service.rs
```diff
@@ -170,7 +133,57 @@ where
...
+    E: Into<BoxError> + Send,
+{
+    type Output = Result<Response<BoxBody>, E>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
+
+        match this.case.as_mut().project() {
+            CaseProj::GrpcWeb { future, accept } => {
+                let res = match ready!(future.poll(cx)) {
```
LucioFranco, line 173: could this just be `ready!(...)?`?

LucioFranco (commented):
Looks pretty good to me left a few comments

slinkydeveloper:
For me this is ready now, it would be nice if someone else can test it though :)

LucioFranco (approved):

PR #1136, opened 2022-11-06
maybe a typo in the tutorial?

ichn-hu:
## Motivation

the example code makes build_client to be false, but claims it will build the client only.

## Solution

I guess this is a typo and it meant to build the server only.


LucioFranco (approved):
Yup seems like it! thanks!

PR #1142, opened 2022-11-08
fix: do not panic while encoding oversized bodies

### LINKED ISSUE #1141: Returning body w/ >4GB data panics
### Version
`master` (`e1885211495e63d962bc1d00f9be6eeaab2bb901`)

### Description

This test panics with `thread 'codec::prost::tests::encode_big' panicked at 'assertion failed: len <= std::u32::MAX as usize', tonic/src/codec/encode.rs:127:5`, but instead the stream should probably return an error:

```rust

    #[tokio::test]
    async fn encode_big() {
        let encoder = MockEncoder::default();

        let msg = vec![0u8; u32::MAX as usize + 1];

        let messages = std::iter::once(Ok::<_, Status>(msg));
        let source = futures_util::stream::iter(messages);

        let body = encode_server(
            encoder,
            source,
            None,
            SingleMessageCompressionOverride::default(),
        );

        futures_util::pin_mut!(body);

        while let Some(r) = body.data().await {
            r.unwrap();
        }
    }
```

crepererum:
## Motivation
Fixes #1141.

## Solution
Replace assertion w/ an ordinary error path.

### DIFF
--- a/tonic/src/codec/prost.rs
+++ b/tonic/src/codec/prost.rs
@@ -136,6 +136,37 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn encode_too_big() {
+        let encoder = MockEncoder::default();
+
+        let msg = vec![0u8; u32::MAX as usize + 1];
+
+        let messages = std::iter::once(Ok::<_, Status>(msg));
+        let source = futures_util::stream::iter(messages);
+
+        let body = encode_server(
+            encoder,
+            source,
+            None,
+            SingleMessageCompressionOverride::default(),
+        );
+
+        futures_util::pin_mut!(body);
+
+        assert!(body.data().await.is_none());
+        assert_eq!(
+            body.trailers()
+                .await
+                .expect("no error polling trailers")
+                .expect("some trailers")
+                .get("grpc-status")
+                .expect("grpc-status header"),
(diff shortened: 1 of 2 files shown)


LucioFranco (approved):
LGTM this seems extremely logical and a good change: :D

tonic/src/codec/prost.rs
```diff
@@ -136,6 +136,37 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn encode_too_big() {
+        let encoder = MockEncoder::default();
+
+        let msg = vec![0u8; u32::MAX as usize + 1];
```
LucioFranco, line 143: uh this seems like its failing on windows and I think that kinda makes sense when you try to allocate a vec of u32::MAX? Prob need to lower it for windows or something like that

crepererum: W/o that much data, we're not hitting the limit that triggers the bug. I can exclude the test for windows though.

LucioFranco: Yeah, lets see if that works, may just be some setting on windows that is making this fail.

crepererum: done

PR #1144, opened 2022-11-09
Follow up #1127

bouzuya:
See: https://github.com/hyperium/tonic/pull/1127#discussion_r1016758800

---

@LucioFranco 
I didn't understand how to centralize the impls, so I would like you to comment on the draft PRs.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -229,7 +229,7 @@ fn generate_attributes<'a>(
 fn generate_doc_comment<S: AsRef<str>>(comment: S) -> TokenStream {
     let comment = comment.as_ref();
 
-    let comment = if !comment.starts_with(" ") {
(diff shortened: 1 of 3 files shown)


tonic-build/src/lib.rs
```diff
@@ -211,11 +213,46 @@ fn generate_attributes<'a>(
...
+    service: &T,
+    method: &<T as Service>::Method,
+) -> String {
+    format!(
+        "{}{}{}.{}",
+        package,
+        if package.is_empty() { "" } else { "." },
+        service.identifier(),
+        method.identifier()
+    )
```
LucioFranco, line 236: I was thinking just this code could be put into a fn the rest I think we could leave

bouzuya: Is it like this? Should I not move format_method_name to the lib?

LucioFranco (approved):

Issue #687, opened 2021-06-21
Call credentials or per-call authentication


matze:
### Motivation

In order to implement a self-contained CA, I would like to issue a client certificate via a gRPC endpoint (and verify the cert via a side channel). However, once client authentication is enabled, all endpoints have to be called with a client certificate. Excluding certain calls from authentication or having [call credentials](https://grpc.io/docs/guides/auth/#credential-types) would allow me to implement that.

### Proposal

I am not in a position to say if it's these proposals are possible or not but I'd see it like this:

* Either annotate calls with a macro to opt in or out of client authentication or
* add an option that allows me to do the cert validation inside the call (and thus opt out if I deem so)

### Alternatives

I could start a separate non-authenticated server that just issues certificates but that is annoying because of the second port, sharing the cert information with the actual server etc.

LucioFranco:
I believe this would require configuring how we apply rustls and I am not an expert there. I would recommend looking through how you might approach this without tonic with a tcp/http stream. That solution should then be applicable to tonic.

dufkan:
I believe this is solved by https://github.com/hyperium/tonic/pull/1163.

matze:
Thanks! I will check once a release is made.

PR #1163, opened 2022-11-26
feat(tls): add an option for optional TLS client authentication

dufkan:
Previously there were only two options for client authentication – either no authentication or mandatory authentication. With this change, a server can allow for optional authentication with a given root CA certificate and enforce client authentication on a per-request basis.

Refs: #687

## Motivation

Currently, there is no easy way of performing client authentication on a per-request basis with the API provided by `ServerTlsConfig`. This behavior is useful in scenarios like the one described in #687, i.e., when some endpoints should be publicly accessible while others not.

## Solution

Since `rustls` provides `ClientCertVerifier` with the desired behavior, it only needs to be made accessible from `ServerTlsConfig`. Currently, when the method `client_ca_cert` is called on a `ServerTlsConfig`, it forces mandatory client authentication with a given root CA certificate.

I propose to rename this method and change its input to a three-valued enum – no authentication (default), optional authentication with a provided root CA certificate, and mandatory authentication with a provided root CA certificate. I understand this is a breaking change, but it seems like a much cleaner solution than introducing a new method along `client_ca_certificate` that would set some `optional` flag in `ServerTlsConfig`, since the flag would have no meaning if the `client_ca_certificate` was not used.

### DIFF
--- a/tonic/src/transport/server/tls.rs
+++ b/tonic/src/transport/server/tls.rs
@@ -51,6 +51,9 @@ impl ServerTlsConfig {
     /// Sets whether client certificate verification is optional.
     ///
     /// This option has effect only if CA certificate is set.
+    ///
+    /// # Default
+    /// By default, this option is set to `false`.
     pub fn client_auth_optional(self, optional: bool) -> Self {
         ServerTlsConfig {
             client_auth_optional: optional,
--- a/tonic/src/transport/server/tls.rs
+++ b/tonic/src/transport/server/tls.rs
@@ -11,6 +11,7 @@ use std::fmt;
 pub struct ServerTlsConfig {
     identity: Option<Identity>,
     client_ca_root: Option<Certificate>,
+    client_auth_optional: bool,
 }
 
 #[cfg(feature = "tls")]
@@ -27,6 +28,7 @@ impl ServerTlsConfig {
         ServerTlsConfig {
             identity: None,
             client_ca_root: None,
+            client_auth_optional: false,
         }
     }
 
@@ -46,7 +48,21 @@ impl ServerTlsConfig {
         }
     }
 
+    /// Sets whether client certificate verification is optional.
+    ///
+    /// This option has effect only if CA certificate is set.
+    pub fn client_auth_optional(self, optional: bool) -> Self {
+        ServerTlsConfig {
+            client_auth_optional: optional,
+            ..self
+        }
+    }
+
     pub(crate) fn tls_acceptor(&self) -> Result<TlsAcceptor, crate::Error> {
-        TlsAcceptor::new(self.identity.clone().unwrap(), self.client_ca_root.clone())
+        TlsAcceptor::new(
+            self.identity.clone().unwrap(),
+            self.client_ca_root.clone(),
+            self.client_auth_optional,
+        )
     }
 }
--- a/tonic/src/transport/service/tls.rs
+++ b/tonic/src/transport/service/tls.rs
@@ -118,12 +118,20 @@ impl TlsAcceptor {
     pub(crate) fn new(
         identity: Identity,
         client_ca_root: Option<Certificate>,
+        client_auth_optional: bool,
     ) -> Result<Self, crate::Error> {
         let builder = ServerConfig::builder().with_safe_defaults();
 
-        let builder = match client_ca_root {
-            None => builder.with_no_client_auth(),
-            Some(cert) => {
+        let builder = match (client_ca_root, client_auth_optional) {
+            (None, _) => builder.with_no_client_auth(),
+            (Some(cert), true) => {
+                use tokio_rustls::rustls::server::AllowAnyAnonymousOrAuthenticatedClient;
+                let mut roots = RootCertStore::empty();
+                rustls_keys::add_certs_from_pem(std::io::Cursor::new(&cert.pem[..]), &mut roots)?;
+                builder
+                    .with_client_cert_verifier(AllowAnyAnonymousOrAuthenticatedClient::new(roots))
+            }
+            (Some(cert), false) => {
                 use tokio_rustls::rustls::server::AllowAnyAuthenticatedClient;
                 let mut roots = RootCertStore::empty();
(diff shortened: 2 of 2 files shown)


tonic/src/transport/server/tls.rs
```diff
@@ -38,15 +38,29 @@ impl ServerTlsConfig {
...
+/// Client TLS authentication options.
+#[cfg(feature = "tls")]
+#[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
+#[derive(Clone, Debug, Default)]
+pub enum ClientTlsAuth {
+    /// No client authentication.
+    #[default]
+    None,
+    /// Optional client authentication with a provided root CA certificate.
+    Optional(Certificate),
```
LucioFranco, line 63: Could you explain more what the optional use case is? Do you know if any of the other gRPC implementations provide this sort of feature? As a side note, if this is a more niche use case you may want to consider using hyper directly and using `hyper-rustls` which will give you full access to all the rustls settings.

dufkan: This option is useful, for example, when you have two types of clients accessing your service. One type may be only viewers that can access some data you provide through your endpoints, and the other may have write access but have to authenticate first. Having to issue certificates to all viewers who may access your endpoints only once may not be ideal. Furthermore, you cannot issue such certificates over the gRPC with mandatory authentication (because the clients would first need to have the certificate).

The `Optional(Certificate)` option guarantees that if a client included a certificate, it was validated by standard procedures, but it may choose not to include any. This way, it is much safer than avoiding certificate verification altogether and doing it by yourself later on for endpoints that should require authentication.

I have not used any other gRPC libraries prior to `tonic`, but a quick search for a gRPC library led me to the grpc.io C++ library and to this page https://grpc.github.io/grpc/cpp/grpc__security__constants_8h.html#a29ffe63a8bb3b4945ecab42d82758f09 where are listed five different client authentication options. The enum values that I have included in the pull request correspond to the constants this way:
```
None <=> GRPC_SSL_DONT_REQUEST_CLIENT_CERTIFICATE 	
Optional(Certificate) <=> GRPC_SSL_REQUEST_CLIENT_CERTIFICATE_AND_VERIFY
Mandatory(Certificate) <=> GRPC_SSL_REQUEST_AND_REQUIRE_CLIENT_CERTIFICATE_AND_VERIFY
```

The remaining two options may be useful when you intend to perform certificate verification by yourself, but that can (and should) be considered dangerous. I do not currently see any good use cases for these options, so they are not included in  this pull request. They can be added by extending the enum that I proposed by additional fields that would take a trait object implementing `rustls::server::ClientCertVerifier`, but I would consider this quite a controversial addition that would be better left to using `hyper` directly if needed, as you mentioned.

examples/src/tls_client_auth/server.rs
```diff
@@ -67,7 +67,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let tls = ServerTlsConfig::new()
         .identity(server_identity)
-        .client_ca_root(client_ca_cert);
+        .client_auth(ClientTlsAuth::Mandatory(client_ca_cert));
```
LucioFranco, line 70: Can we update this to be the old behavior and then added `client_auth_mandatory` as additional methods that way we don't break current users?

tonic/src/transport/server/tls.rs
```diff
@@ -38,15 +38,29 @@ impl ServerTlsConfig {
...
-        TlsAcceptor::new(self.identity.clone().unwrap(), self.client_ca_root.clone())
+        TlsAcceptor::new(self.identity.clone().unwrap(), self.client_auth.clone())
     }
 }
+
+/// Client TLS authentication options.
+#[cfg(feature = "tls")]
+#[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
+#[derive(Clone, Debug, Default)]
+pub enum ClientTlsAuth {
```
LucioFranco, line 58: This should be a private enum so that we don't add additional api surface area.

tonic/src/transport/server/tls.rs
LucioFranco: Yeah, that seems fair!

LucioFranco (changes requested):
I think we can move forward with this change, I left a few comments on the api but otherwise I think this could be a good addition, thanks!

examples/src/tls_client_auth/server.rs
dufkan: Ok, that is a good suggestion. I would only propose adding `client_auth_optional(bool)` instead, as calling `client_ca_root` causes mandatory authentication by default.

However, this approach brings up the following question: How should it behave when `client_ca_root` is not set and `client_auth_optional(bool)` is called?

One option is to store the `bool` and use it only if `client_ca_root` call is eventually performed, otherwise it would be ignored, but that may be a bit counterintuitive.

The most idiomatic approach, I think, would be to have a single method taking in enum with acceptable configurations (like it is in this PR), but I understand that it would be a breaking change, so I can rework it once we decide on the desired behavior.

LucioFranco: > One option is to store the bool and use it only if client_ca_root call is eventually performed, otherwise it would be ignored, but that may be a bit counterintuitive.

I think since its a pretty niche use case we can just document it and that should be fine. I don't think its worth it to break users for the small improvement in clarity for a decently uncommon path.

dufkan: Ok, that's a fair point. I reworked it this way.

tonic/src/transport/server/tls.rs
dufkan: This is not used anymore.

tonic/src/transport/server/tls.rs
```diff
@@ -46,7 +48,21 @@ impl ServerTlsConfig {
         }
     }
 
+    /// Sets whether client certificate verification is optional.
+    ///
+    /// This option has effect only if CA certificate is set.
```
LucioFranco, line 53:
```suggestion
    /// This option has effect only if CA certificate is set.
    ///
    /// # Default
    /// By default, this option is set to `false`.
```

LucioFranco (approved):

PR #1169, opened 2022-11-29
chore: Update cargo-deny config

tottoto:
Updates it to support current project status.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -38,12 +38,12 @@ jobs:
     - name: Check all targets
       run: cargo check --all --all-targets --all-features
 
-  # deny-check:
-  #   name: cargo-deny check
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #   - uses: actions/checkout@v1
-  #   - uses: EmbarkStudios/cargo-deny-action@v1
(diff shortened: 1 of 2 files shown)


LucioFranco:
I believe the deny check isn't running in CI, we should probably fix that if we want to keep this up to date.

tottoto:
Fixed it. However I noticed that tonic depends on MPL-2.0 because of webpki-roots. It is a copyleft license. Should us allow to use it in the deny config?

LucioFranco:
Yeah, we can allow it.

tottoto:
The copyleft licenses seem to be warned for default. I added `copyleft = "deny"` config to make the cargo-deny job fail due to MPL-2.0 in https://github.com/hyperium/tonic/pull/1169/commits/8e7a949907f7d0c3b9df049b7c31db2bee510d2f ([the failed ci](https://github.com/hyperium/tonic/actions/runs/3626168774/jobs/6114916935)), and allowed it explicitly in https://github.com/hyperium/tonic/pull/1169/commits/ea3a7141dc1d2f5b7cc5ed80cabd6a57d3044a9a.

LucioFranco:
Thank you!

PR #1173, opened 2022-11-30
chore(examples): Fix blocking-server

tottoto:
- Moves to the blocking directory.
- Registers the blocking server example as bin.
- Fixes lint.
- Applies rustfmt.


LucioFranco:
Thanks!

PR #1178, opened 2022-12-03
chore(examples): Remove dummy echo implementation

tottoto:
Removes dummy echo implementations from examples as we might not need to use full echo proto in some examples.

These changes may improve its readability.


LucioFranco (approved):
Thanks

PR #1181, opened 2022-12-06
Fix clippy lint

tottoto:
Fixes clippy lints.

### DIFF
--- a/tests/root-crate-path/src/main.rs
+++ b/tests/root-crate-path/src/main.rs
@@ -1,4 +1,5 @@
-#[derive(Clone, PartialEq, Eq, ::prost::Message)]
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
 pub struct Animal {
     #[prost(string, optional, tag = "1")]
     pub name: ::core::option::Option<::prost::alloc::string::String>,
--- a/tests/root-crate-path/src/main.rs
+++ b/tests/root-crate-path/src/main.rs
@@ -9,6 +9,7 @@ pub struct Animal {
 pub mod foo {
     pub mod bar {
         pub mod baz {
+            #![allow(clippy::derive_partial_eq_without_eq)]
             tonic::include_proto!("foo.bar.baz");
         }
     }
--- a/tests/root-crate-path/src/main.rs
+++ b/tests/root-crate-path/src/main.rs
@@ -1,4 +1,4 @@
-#[derive(Clone, PartialEq, ::prost::Message)]
+#[derive(Clone, PartialEq, Eq, ::prost::Message)]
 pub struct Animal {
     #[prost(string, optional, tag = "1")]
     pub name: ::core::option::Option<::prost::alloc::string::String>,
--- a/tests/root-crate-path/src/main.rs
+++ b/tests/root-crate-path/src/main.rs
@@ -9,7 +9,6 @@ pub struct Animal {
 pub mod foo {
     pub mod bar {
         pub mod baz {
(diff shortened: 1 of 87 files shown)


LucioFranco (commented):
LGTM just want to see the eq stuff backed out into codegen and have one qusetion

tonic/src/transport/service/io.rs
```diff
@@ -70,7 +70,7 @@ impl AsyncWrite for BoxedIo {
 pub(crate) enum ServerIo<IO> {
     Io(IO),
     #[cfg(feature = "tls")]
-    TlsIo(TlsStream<IO>),
+    TlsIo(Box<TlsStream<IO>>),
```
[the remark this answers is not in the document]
tottoto: This fixes [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/#large_enum_variant).

tonic-build/src/prost.rs
```diff
@@ -15,6 +15,11 @@ use std::{
 ///
 /// Use [`compile_protos`] instead if you don't need to tweak anything.
 pub fn configure() -> Builder {
+    let type_attributes = vec![(
+        ".".to_string(),
+        "#[allow(clippy::derive_partial_eq_without_eq)]".to_string(),
+    )];
```
tottoto, line 21: As adding attributes might be only be configured via prost api, it seems like we need to set this as the default. However I'm not sure that this is an appropriate approach.

tottoto:
I have to remove clippy attributes from each fille which includes proto, and update generated codes.

tonic-build/src/prost.rs
LucioFranco: Ugh I wonder if we should just fix this in prost then? Somewhere in this macro impl https://github.com/tokio-rs/prost/blob/master/prost-derive/src/lib.rs#L21 since really the problem is I don't want code generated stuff to impl Eq or we can put this in prost-build?

tottoto: I published https://github.com/tokio-rs/prost/pull/775 as to try solving this.

LucioFranco:
The ci failure is interesting I think maybe you didn't run it with the clean up feature enabled?

LucioFranco:
Working on that now https://github.com/tokio-rs/prost/pull/777

LucioFranco:
ok its been published as v0.11.4

tottoto:
I updated prost-build to 0.11.4 in https://github.com/hyperium/tonic/pull/1181/commits/e7efd338322b97addc95d1b7d9d4330a17a38837 and fixed some warnings in https://github.com/hyperium/tonic/pull/1181/commits/7c455f70a731d23a5f2063bb4e228859ef183b4f.

tottoto:
I reverted https://github.com/hyperium/tonic/pull/1181/commits/7fe1d32111c76bd51bc7ce6bd0657d481c27d9a2 in https://github.com/hyperium/tonic/pull/1181/commits/728b040d584fcbb08a31d2075c91f072ff4953b5 as it was took over by prost-build 0.11.4.

tests/root-crate-path/src/main.rs
```diff
@@ -1,4 +1,4 @@
-#[derive(Clone, PartialEq, ::prost::Message)]
+#[derive(Clone, PartialEq, Eq, ::prost::Message)]
```
LucioFranco, line 1: This should get an allow with the new version of prost right?

tottoto: This looks like my mistake. I'll fix this.

tottoto: Fixed.

LucioFranco (approved):
Thanks!

PR #1182, opened 2022-12-09
tonic-build: use unified syntax to clone `Arc`s

teskje:
This PR changes the generated code to use the unified function syntax for cloning `Arc` values (`Arc::clone(foo)` instead of `foo.clone()`).

## Motivation

In https://github.com/MaterializeInc/materialize, we use tonic, and we also enforce the clippy lint [`clone_on_ref_ptr`](https://rust-lang.github.io/rust-clippy/stable/index.html#clone_on_ref_ptr). This lint currently triggers on code generated by tonic, which forces us to disable it for modules that import tonic-generated code.

The main motivation for this PR is that we don't have to disable the lint anymore. Using unified function syntax when cloning `Arc`s also makes the code more obvious, which is nice.

## Solution

Adjust the generated code to conform with the style clippy proposes.

### DIFF
--- a/tonic-build/src/server.rs
+++ b/tonic-build/src/server.rs
@@ -179,7 +179,7 @@ pub(crate) fn generate_internal<T: Service>(
 
             impl<T: #server_trait> Clone for _Inner<T> {
                 fn clone(&self) -> Self {
-                    Self(self.0.clone())
+                    Self(Arc::clone(&self.0))
                 }
             }
 
@@ -404,7 +404,7 @@ fn generate_unary<T: Method>(
             type Future = BoxFuture<tonic::Response<Self::Response>, tonic::Status>;
 
             fn call(&mut self, request: tonic::Request<#request>) -> Self::Future {
-                let inner = self.0.clone();
+                let inner = Arc::clone(&self.0);
                 let fut = async move {
                     (*inner).#method_ident(request).await
                 };
@@ -456,7 +456,7 @@ fn generate_server_streaming<T: Method>(
             type Future = BoxFuture<tonic::Response<Self::ResponseStream>, tonic::Status>;
 
             fn call(&mut self, request: tonic::Request<#request>) -> Self::Future {
-                let inner = self.0.clone();
+                let inner = Arc::clone(&self.0);
                 let fut = async move {
                     (*inner).#method_ident(request).await
                 };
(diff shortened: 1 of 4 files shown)


teskje:
This changes generated code, so I'm expecting some tests to break. I'm relying on CI to run the tests and tell me which break, so I can fix them. LMK if I should figure out how to run them myself instead!

LucioFranco (approved):
LGTM thanks!

LucioFranco:
Youll need to run the bootstrap test and commit the generated code to get CI to pass. then we can merge thanks!

teskje:
Thanks @LucioFranco! I've done that now, but I'm not entirely sure my changes are correct. They are different from what the CI output showed, and they are not related to the changes I made either.

LucioFranco:
how are you running the tests?

davidpdrsn:
nitpick: Its called "Fully Qualified Syntax" rather than "Unified Syntax". See https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name

teskje:
> how are you running the tests?

On a clean checkout, I simply run `cargo test`. I'm on macOs, in case that matters, with protobuf (3.21.11) installed via homebrew.

> nitpick: Its called "Fully Qualified Syntax" rather than "Unified Syntax".

I suppose you are right. I took the term from the clippy lint, no idea where they got it from. It was called ["universal function call syntax"](https://doc.rust-lang.org/stable/book/ufcs.html) at one point.

davidpdrsn:
> I suppose you are right. I took the term from the clippy lint, no idea where they got it from. It was called ["universal function call syntax"](https://doc.rust-lang.org/stable/book/ufcs.html) at one point.

Ah I see. I guess it has several names :P

tottoto:
The diffs are caused by new version of prost-build `0.11.4`. I think it will be fixed in https://github.com/hyperium/tonic/pull/1181.

tottoto:
And some docs changes in https://github.com/hyperium/tonic/pull/1182/commits/a684e510f21e144c368e92a2b1ee02dfbb553d68 might be caused by lack of cleanup-markdown feature of tonic-build in reflection manifest. It will be fixed in https://github.com/hyperium/tonic/pull/1181 too.

teskje:
Thanks, removing the second commit, `cargo update`ing, and rebasing on the latest master made the checks pass for me locally.

PR #1187, opened 2022-12-13
chore: Use namespaced dependency feature

tottoto:
We could use namespaced dependency feature introduced in Rust 1.60.


LucioFranco (approved):
LGTM our msrv is already 1.60 so this should be fine 👍

PR #1191, opened 2022-12-14
Move examples to separate crates

tottoto:
Moves examples to separate crates. This clarifies the dependencies of each example.


LucioFranco:
What is the advantage of having them all in separate crates? From what I can tell it just adds a lot more cargo files that will be a PITA to update when things like prost get updated (though we can use the new cargo feature to pass through workspace versions)?

tottoto:
The advantage is to clarify dependencies of each example feature. When I try to use each functionality in testing projects, it is difficult for me to find out what crates and their features are required. This change might solve the problem.

As you indicated, it is very troublesome to update each example. I think your opinion completely makes sense. To reduce the pain we could use workspace inheritance introduced in Rust 1.64. However it will break tonic's msrv 1.60. I think we could use this feature by moving examples to separate workspace keeping the root workspace's msrv 1.60.

LucioFranco:
Can users not look at the imports at the top of the file. What I tend to do is keep all the code you need to run the example in a single file such that you can look at the imports and see what crates they come from. If you need the version you can check the cargo.toml. Does this fall down somewhere that I am missing?

LucioFranco:
One improvement I would love to see tho is to have all the examples at the root of the examples folder and not under a src dir and to also make them work via --example instead of --bin.

tottoto:
I overlooked what you pointed out. I understand your opinion and I agree with you. I'll try to use the example target as advised.

davidpdrsn:
We do this in axum (https://github.com/tokio-rs/axum/tree/main/examples) and my impression is that it does help users a lot. I think its a bit more annoying for maintainers but overall worth it.

tottoto:
I noticed that we can not use optional dev-dependencies. So we can't specify the dependencies of each example. And the example target can not be used in a virtual manifest. These restriction would be resolved by defining actual manifest in root `Cargo.toml` and using dependencies, however I think it seems not like a good solution. From the above I'll try to use `required-features` in each binary target in `examples/Cargo.toml` to clarify the dependencies.

LucioFranco:
@davidpdrsn imho that just makes it harder to find the rust file with what you want. Personally I prefer to have a single examples folder with rs files/folders if client/server pairs (which tonic uses) and a single cargo.toml at the root of the examples folder. I don't really see an advantage of having separate actual crates for each example?

davidpdrsn:
If someone wants to know exactly which dependencies are used for axums jwt example they just go to https://github.com/tokio-rs/axum/blob/main/examples/jwt/Cargo.toml and they can see it right there. So the examples don't mix their dependencies. We still keep all the examples in a cargo workspace https://github.com/tokio-rs/axum/blob/main/examples/Cargo.toml

LucioFranco:
Right, but do people actually struggle to find the dependencies at the root cargo.toml if its not reduced to the min set needed for that example? I haven't seen someone come and ask which dep we used for an example, though maybe that has happened with axum more than tonic 🤷

davidpdrsn:
Yeah it was a frequent question for axum early on. Perhaps because there are more new rust devs using axum than tonic.

LucioFranco (changes requested):
So I think I am still in the boat that I want to not move every example into their own crate. I would actually rather also move them to be examples and not binaries. Mostly my reasoning is I don't think using the workspace version stuff that was recently introduced is easy to understand for beginners but updating all the examples is a lot of work.

So in that case I prefer the style it currently is though I think we could make them examples.

tottoto:
I'll close this pull request as I published another approach in https://github.com/hyperium/tonic/pull/1205. Thanks for all reviews.

PR #1194, opened 2022-12-15
chore: Use setup-rust-action v1 to setup rust in ci

tottoto:
- Uses hecrj/setup-rust-action which has been used before in tonic project as it's been updated more frequently than actions-rs/toolchain recently.
- Locks hecrj/setup-rust-action version to v1 to be safe against possible breaking changes.
- Uses hecrj/setup-rust-action's option to install components.


LucioFranco (approved):

LucioFranco:
Thanks!

PR #1198, opened 2022-12-16
chore: Use baptiste0928/cargo-install@v1 to install cargo-hack

tottoto:
In every check ci jobs, it takes about 3 (in ubuntu-latest), 4 (in windows-latest), 5 (in macOS-latest) minutes to install cargo-hack. This pull request tries to speed up the jobs by introducing [baptiste0928/cargo-install](https://github.com/baptiste0928/cargo-install) action using cache.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -24,7 +24,9 @@ jobs:
         components: rustfmt
     - uses: actions/checkout@v3
     - name: Install cargo-hack
-      run: cargo install cargo-hack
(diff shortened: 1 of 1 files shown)


LucioFranco (commented):
Just curious what is the benefit of using this action over what we had before?

tottoto:
Once we installed cargo-hack, in the next job, we can use cached cargo-hack. So we'll be able to spare time to install cargo-hack taking up to about 5 minutes every time.

LucioFranco:
makes sense to me!

Issue #1156, opened 2022-11-19
An error will occur if you name your protobuf service "Result"


stevenhansel:
### Version

```
tonic 0.8
tonic-build 0.8
```

### Platform

```
Darwin june.local 21.6.0 Darwin Kernel Version 21.6.0: Wed Aug 10 14:28:35 PDT 2022; root:xnu-8020.141.5~2/RELEASE_ARM64_T8101 arm64
```

### Description

I found out an unique bug, I happened to name my proto service as "Result" because it's supposed to stream the result of a service. Here is my proto definition:

`result.proto`
```
syntax = "proto3";

import "google/protobuf/any.proto";
import "google/protobuf/empty.proto";

option go_package = "github.com/stevenhansel/binus-logbook/scraper/internal/grpc/proto/result";

package result;

service Result {
  rpc Listen(google.protobuf.Empty) returns (stream Reply) {}
}

message Reply {
  int32 step = 1;
  string name = 2;
  google.protobuf.Any data = 3;
}
```

Then when I run `cargo build` where inside in the build.rs looks like:

`build.rs`
```
fn main() {
    let protos: &'static [&str] = &[
        "../proto/helloworld.proto",
        "../proto/scraper.proto",
        "../proto/result.proto",
    ];

    for proto in protos {
        tonic_build::compile_protos(proto)
            .unwrap_or_else(|e| panic!("Failed to compile protos {:?}", e));
    }
    tauri_build::build()
}
```

It will show this error:
```
error[E0782]: trait objects must include the `dyn` keyword
   --> /Users/june/binus-logbook/src-tauri/target/debug/build/binus-logbook-69ad9dfa394bbbc0/out/result.rs:110:24
    |
110 |                 Item = Result<super::Reply, tonic::Status>,
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: add `dyn` keyword before this trait
    |
110 -                 Item = Result<super::Reply, tonic::Status>,
110 +                 Item = dyn Result<super::Reply, tonic::Status>,
    |

error[E0107]: this trait takes 0 generic arguments but 2 generic arguments were supplied
   --> /Users/june/binus-logbook/src-tauri/target/debug/build/binus-logbook-69ad9dfa394bbbc0/out/result.rs:110:24
    |
110 |                 Item = Result<super::Reply, tonic::Status>,
    |                        ^^^^^^----------------------------- help: remove these generics
    |                        |
[...]

PR #1203, opened 2022-12-17
Fixes: #1156 Prevent error when "Result" is used as name of service in proto 

eskawl:
## Motivation

Fixes a name conflict when the service itself is named as `Result`. The conflict occurs with the `std::result::Result` included in the prelude. Issue described in #1156.

## Solution

To prevent the conflict, the generated code will now refer to the `Result` in a qualified manner viz, `std::result::Result`.

### Test Sample

**result.proto**

```protobuf
syntax = "proto3";

import "google/protobuf/any.proto";
import "google/protobuf/empty.proto";

option go_package = "github.com/stevenhansel/binus-logbook/scraper/internal/grpc/proto/result";

package result;

service Result {
  rpc Listen(google.protobuf.Empty) returns (Reply) {}
}

message Reply {
  int32 step = 1;
  string name = 2;
}
```

**server.rs**

```rust
use tonic::{transport::Server, Request, Response, Status};

pub mod pb {
    tonic::include_proto!("result");
}

use pb::{Reply, result_server};


#[derive(Debug, Default)]
pub struct ResultAPI {}

#[tonic::async_trait]
impl result_server::Result for ResultAPI {
    async fn listen(&self, _request: Request<()>) -> Result<Response<Reply>, Status> {
        return Ok(Response::new(Reply { step: 1, name: "23".to_string() }))
    }
}

#[tokio::main]
    let addr = "[::1]:50051".parse()?;
    let api = ResultAPI::default();

    Server::builder()
        .add_service(result_server::ResultServer::new(api))
        .serve(addr)
        .await?;

    Ok(())
}
```

**client.rs**

```rust
pub mod pb {
    tonic::include_proto!("result");
}

use pb::{result_client};

#[tokio::main]
    let mut client = result_client::ResultClient::connect("http://[::1]:50051").await?;

    let request = tonic::Request::new(());

    let response = client.listen(request).await?;

    println!("RESPONSE={:?}", response);

    Ok(())
}
```

### DIFF
--- a/tests/service_named_result/Cargo.toml
+++ b/tests/service_named_result/Cargo.toml
@@ -2,6 +2,7 @@
 name = "service_named_result"
 version = "0.1.0"
 edition = "2021"
+publish = false
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
--- a/tests/service_named_result/Cargo.toml
+++ b/tests/service_named_result/Cargo.toml
@@ -1,0 +1,13 @@
+[package]
+name = "service_named_result"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+prost = "0.11"
+tonic = {path = "../../tonic"}
+
+[build-dependencies]
+tonic-build = {path = "../../tonic-build"}
--- a/tests/service_named_result/LICENSE
+++ b/tests/service_named_result/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2020 Lucio Franco
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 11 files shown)


tests/service_named_result/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2020 Lucio Franco
```
LucioFranco, line 1: I don't think we need a license here?

LucioFranco (changes requested):
LGTM I think we just need to remove that license file

tests/service_named_result/LICENSE
eskawl: I've continued the existing pattern of including license files as followed by other examples. Should we stop including this file?

LucioFranco: ah yeah this is fine, we should be soft symlinking them but its ok since you're right the other crates in the test folder has the same thing.

tests/service_named_result/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "service_named_result"
+version = "0.1.0"
+edition = "2021"
```
LucioFranco, line 4:
```suggestion
edition = "2021"
publish = false
```

LucioFranco (approved):

PR #1179, opened 2022-12-04
feat(types): Add gRPC Richer Error Model support (DebugInfo)

flemosr:
## Motivation

The [gRPC Richer Error Model][error-handling] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages.

## Solution

This PR continues the work initiated in [#1068], building on the changes made in [#1095]. It adds support for the `DebugInfo` standard error message type to `tonic-types`.

[error-handling]: https://www.grpc.io/docs/guides/error
[#1068]: https://github.com/hyperium/tonic/pull/1068
[#1095]: https://github.com/hyperium/tonic/pull/1095

PR #1204, opened 2022-12-19
feat(types): Add gRPC Richer Error Model support (QuotaFailure)

flemosr:
## Motivation

The [gRPC Richer Error Model][error-handling] is quite useful to send additional feedback to clients, and is supported by many gRPC libraries in other languages.

## Solution

This PR continues the work initiated in [#1068], building on the changes made in [#1179]. It adds support for the `QuotaFailure` standard error message type to `tonic-types`.

[error-handling]: https://www.grpc.io/docs/guides/error
[#1068]: https://github.com/hyperium/tonic/pull/1068
[#1179]: https://github.com/hyperium/tonic/pull/1179

### DIFF
--- a/tonic-types/src/richer_error/error_details/mod.rs
+++ b/tonic-types/src/richer_error/error_details/mod.rs
@@ -166,23 +166,23 @@ impl ErrorDetails {
     }
 
     /// Get [`RetryInfo`] details, if any
-    pub fn retry_info(&self) -> &Option<RetryInfo> {
-        &self.retry_info
+    pub fn retry_info(&self) -> Option<RetryInfo> {
+        self.retry_info.clone()
     }
 
     /// Get [`DebugInfo`] details, if any
-    pub fn debug_info(&self) -> &Option<DebugInfo> {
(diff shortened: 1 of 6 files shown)


tonic-types/src/richer_error/error_details/mod.rs
```diff
@@ -120,18 +166,23 @@ impl ErrorDetails {
     }
 
     /// Get [`RetryInfo`] details, if any
-    pub fn retry_info(&self) -> Option<RetryInfo> {
-        self.retry_info.clone()
+    pub fn retry_info(&self) -> &Option<RetryInfo> {
```
LucioFranco, line 169: Should these be `Option<&Foo>`? This is also a breaking change so maybe we should do this later and save it?

flemosr: I did not see that v0.6.1 was released :sweat_smile: . Sorry about that. Yes, better to do this later, I pushed a commit to revert this change. Thanks!

LucioFranco (approved):

PR #1240, opened 2023-01-17
chore(interop): Lock clap version to 4.0

tottoto:
In order to fix msrv 1.60 test, locks clap version to 4.0 as clap 4.1 increased its msrv to 1.64.


LucioFranco (approved):
Thanks! We could also just consider not using clap and doing that stuff manually but we can do that later.

PR #1242, opened 2023-01-17
chore(interop): Use pico-args instead of clap

tottoto:
## Motivation

We have used clap to parse arguments in interop. This crate provides us with a lot of great functionalities. Since clap evolves quickly, its increase of msrv is relatively rapid. As we uses clap only in test crate, we might not need most of the clap's functionality. In order to reduce effects from dependencies in our test crates, we could use the other arguments parser which is aimed to be simple.

## Solution

Uses a more lightweight arguments parser library.

### DIFF
--- a/interop/Cargo.toml
+++ b/interop/Cargo.toml
@@ -16,8 +16,8 @@ path = "src/bin/server.rs"
 
 [dependencies]
 async-stream = "0.3"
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):
lgtm! Thanks

PR #1246, opened 2023-01-20
Fix panic in MetadataKey::from_bytes

andrewhickman:
## Motivation

I tried using `BinaryMetadataKey::from_bytes` to parse a header value, but noticed it inconveniently panics on values that don't end in `-bin`.

## Solution

I assume this is a bug since the signature of this function suggests it returns an error on failure, and the documentation makes no mention of panicking.

This PR fixes it to return an error instead.

### DIFF
--- a/tonic/src/metadata/key.rs
+++ b/tonic/src/metadata/key.rs
@@ -42,7 +42,7 @@ impl<VE: ValueEncoding> MetadataKey<VE> {
         match HeaderName::from_bytes(src) {
             Ok(name) => {
                 if !VE::is_valid_key(name.as_str()) {
-                    panic!("invalid metadata key")
+                    return Err(InvalidMetadataKey::new());
                 }
 
                 Ok(MetadataKey {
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Nice catch! Thanks

tonic/src/metadata/key.rs
```diff
@@ -282,3 +282,25 @@ impl Default for InvalidMetadataKey {
...
+#[test]
+fn test_from_bytes_ascii() {
+    assert!(AsciiMetadataKey::from_bytes(b"").is_err());
+    assert!(AsciiMetadataKey::from_bytes(b"\xFF").is_err());
+    assert_eq!(
+        AsciiMetadataKey::from_bytes(b"abc").unwrap().as_str(),
+        "abc"
+    );
+    assert!(AsciiMetadataKey::from_bytes(b"abc-bin").is_err());
+}
```
LucioFranco, line 306: shouild these live under #[cfg(test))

LucioFranco (changes requested):

LucioFranco (approved):

PR #1256, opened 2023-02-01
Make some functionality of Status public

nrc:
## Motivation

I'm writing a grpc transport layer based on the transport module of tonic. Currently a few bits of `Status` are `pub(crate)` so they can be used by transport, but not if the code is pulled into a separate crate.

## Solution

Change some `pub(crate)` to `pub` and add a public setter method.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -325,7 +325,10 @@ impl Status {
         })
     }
 
-    pub(crate) fn try_from_error(
+    /// Create a `Status` from various types of `Error`.
+    ///
+    /// Returns the error if a status could not be created.
+    pub fn try_from_error(
         err: Box<dyn Error + Send + Sync + 'static>,
     ) -> Result<Status, Box<dyn Error + Send + Sync + 'static>> {
         let err = match err.downcast::<Status>() {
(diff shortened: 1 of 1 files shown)


tonic/src/status.rs
```diff
@@ -325,7 +325,10 @@ impl Status {
         })
     }
 
-    pub(crate) fn try_from_error(
+    /// Create a `Status` from various types of `Error`.
+    ///
+    /// Returns the error if a status could not be created.
```
LucioFranco, line 330: I am a bit cautious about exposing these error conversions but I would be ok if we added a note that said, the internal downcast methods/types may change and that you can't depend on it to be stable internally.

nrc: Stable as in it might create a different status? Or as in might not work at all (i.e., returns the error)?

LucioFranco: Return a different status based on the input type.

LucioFranco, line 330:
```suggestion
    /// Create a `Status` from various types of `Error`.
    ///
    /// Returns the error if a status could not be created.
    ///
    /// # Downcast stability
    /// This function does not provide any stability guarantees around how it will downcast errors into
    /// status codes.
```

LucioFranco (approved):

PR #1257, opened 2023-02-04
chore: Add windows-sys and hermit-abi to cargo-deny skip-tree

tottoto:
## Motivation

Fixes cargo-deny check.

## Solution

Adds windows-sys to cargo-deny's skip-tree config. Since this crate is a fundamental crate to support windows, a lot of crates depend on it. As each of them defines their own update policies, it seems to be difficult to use one version of it entirely in dependency tree always.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -16,6 +16,7 @@ deny = [
 ]
 skip-tree = [
     { name = "rustls-pemfile" },
+    { name = "windows-sys" },
 ]
 
 [licenses]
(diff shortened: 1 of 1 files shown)


tottoto:
Added hermit-abi too.

tottoto:
Previously I proposed to remove these two crates from skip-tree in #1223. Now it seems that we should put them to skip-tree. As they are fundamental crates, we frequently have to depend on multiple versions of them.

LucioFranco (approved):

LucioFranco:
sounds good to me, thanks for following up!

PR #1263, opened 2023-02-10
Expose `Response#into_parts` and `Response#from_parts`

tonyvelichko:
## Motivation

In cases when you have to build for example an L7 gRPC proxy with data injection into responses from downstream API it's required to perform copying the whole responses to shift ownership this can be easily avoided by implementing the approach suggested for Request in PR: https://github.com/hyperium/tonic/pull/1118

## Solution

This PR contains two changes:

One is to expose `Response#into_parts` to easily control the ownership without making unnecessary copying. 
Another one is to expose `Response#from_parts`, in order to have a symmetric API, consistent with `Request` as well

### DIFF
--- a/tonic/src/response.rs
+++ b/tonic/src/response.rs
@@ -54,10 +54,12 @@ impl<T> Response<T> {
         self.message
     }
 
+    /// Consumes `self` returning the parts of the response.
     pub fn into_parts(self) -> (MetadataMap, T, Extensions) {
         (self.metadata, self.message, self.extensions)
     }
 
+    /// Create a new gRPC response from metadata, message and extensions.
     pub fn from_parts(metadata: MetadataMap, message: T, extensions: Extensions) -> Self {
         Self {
             metadata,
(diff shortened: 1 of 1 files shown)


tonic/src/response.rs
```diff
@@ -54,11 +54,11 @@ impl<T> Response<T> {
         self.message
     }
 
-    pub(crate) fn into_parts(self) -> (MetadataMap, T, Extensions) {
+    pub fn into_parts(self) -> (MetadataMap, T, Extensions) {
         (self.metadata, self.message, self.extensions)
     }
 
-    pub(crate) fn from_parts(metadata: MetadataMap, message: T, extensions: Extensions) -> Self {
+    pub fn from_parts(metadata: MetadataMap, message: T, extensions: Extensions) -> Self {
```
LucioFranco, line 61: Both needs docs

LucioFranco (changes requested):

tonic/src/response.rs
tonyvelichko: done

LucioFranco (approved):

PR #1274, opened 2023-02-15
feat(codec): Configure max request message size

### LINKED ISSUE #1097: Config max request message size to fend off potential malicious attacks
## Feature Request
If a message size is maliciously large, blindly accept it may not be a good idea.  

A quick test shows that with the current tonic server (0.8.1), a simple hello-world unary service was able to take a 1GB size request message and return a response; even though the service handler always responds "hello world" without processing the request at all, just accepting such a message significantly slow down the server.  It would be more reasonable to drop the connection instead of blindly consuming system resources to decompress/parse such a large message. 

This is the same request as in [#264](https://github.com/hyperium/tonic/issues/264).

aoudiamoncef:
## Motivation
In gRPC specs, by default there is a limit for the length of messages(4MB). Actually, Tonic Codec doesn't  set a limit which could cause security concerns. If an attacker decides to send us heavy messages, It'll be decoded and could fill all available RAM.

See #1097

## Solution
Support changing the max message size via a configuration. Set the limit to 4MB by default. 

resolves #1097


LucioFranco (changes requested):
This looks awesome! Really great work, I think I would like to see one integration test as well that shows that we can produce this error and then continue to make normal requests etc. I think once we have that we should be good, I would also try to think about where we could document this that would be helpful.

aoudiamoncef:
Hi @aoudiamoncef,

I have an issue with tonic-build client, I added the two parameters to the client config, but I got an error when I tried to add it in the generator side. 

image

tonic-build/src/client.rs
```diff
@@ -137,6 +137,20 @@ pub(crate) fn generate_internal<T: Service>(
                     self
                 }
 
+                /// Limits the maximum size of a decoded message.
+                #[must_use]
+                pub fn max_decoding_message_size(mut self, limit: usize) -> Self {
+                    self.inner = self.inner.max_decoding_message_size = Some(limit);
```
LucioFranco, line 143: I don't think this is valid rust xD

LucioFranco:
@aoudiamoncef commented I think those two lines in the client.rs of tonic-build need to just remove the `self.inner =` part and it should work?

tonic-build/src/client.rs
aoudiamoncef: I just copy and past from the code above 

```rust
                /// Compress requests with the given encoding.
                ///
                /// This requires the server to support it otherwise it might respond with an
                /// error.
                #[must_use]
                pub fn send_compressed(mut self, encoding: CompressionEncoding) -> Self {
                    self.inner = self.inner.send_compressed(encoding);
                    self
                }

                /// Enable decompressing responses.
                #[must_use]
                pub fn accept_compressed(mut self, encoding: CompressionEncoding) -> Self {
                    self.inner = self.inner.accept_compressed(encoding);
                    self
                }
```

aoudiamoncef:
> @aoudiamoncef commented I think those two lines in the client.rs of tonic-build need to just remove the `self.inner =` part and it should work?

I tired a couple of things without success. It's strange, as we have similar code above

LucioFranco:
@aoudiamoncef I think you need to call the fn on `Grpc`?

Like this 
image

aoudiamoncef:
> @aoudiamoncef I think you need to call the fn on `Grpc`?
> 
> Like this image

Ohhhhh yeah, copy and past is evil, my brain is fucked

aoudiamoncef:
Are you interested by `rust-toolchain.toml` which align rust toolchain with the Tonic CI ?

aoudiamoncef:
> > @aoudiamoncef I think you need to call the fn on `Grpc`?
> > Like this image
> 
> Ohhhhh yeah, copy and past is evil, my brain is fucked

My I was fucked by `inner` which has a nested `inner`. In my head, it's was the same integration as the `server.rs`

tonic/src/server/grpc.rs
```diff
@@ -114,6 +120,71 @@ where
         self
     }
 
+    /// Limits the maximum size of a decoded message.
+    ///
+    ///
```
LucioFranco, line 125:
```suggestion
```

tonic/src/server/grpc.rs
```diff
@@ -114,6 +120,71 @@ where
...
+    /// # };
+    /// ```
+    /// ```
+    pub fn max_decoding_message_size(mut self, limit: usize) -> Self {
+        self.max_decoding_message_size = Some(limit);
+        self
+    }
+
+    /// Limits the maximum size of a encoded message.
+    ///
```
LucioFranco, line 157:
```suggestion
```

LucioFranco (changes requested):
All the docs should state the default value beyond that everything looks pretty good. Left a few comments.

LucioFranco:
Also you will need to rebase your branch and run `cargo test --all` and commit the generated code for the sub libraries.

LucioFranco:
I wrote some instructions here https://github.com/hyperium/tonic/blob/master/CONTRIBUTING.md#generated-code

aoudiamoncef:
@LucioFranco, I corrected all the typos, tests passes locally and I tested the new feature in my projet with success 💯

aoudiamoncef:
The CI is green https://github.com/hyperium/tonic/actions/runs/4198616103

I just rebased from master to include the changes in the contributing file.

LucioFranco (approved):

BoynChan:
hi @LucioFranco, sorry for bother you in a merged PR, here's I want to ask a short but essential question.  I want to know does this change have any document or explain about how I can configure when I create an tonic grpc client?

aoudiamoncef:
I'll try to write an example to show the usage of this feature

PR #1275, opened 2023-02-16
feat: add GrpcMethod extension into request for client

linw1995:
## Motivation

The request was added with the **GrpcMethod** extension in the gRPC client code generated on **tonic-build**, for this reason, we can inspect the gRPC method info within the client interceptor function. In the future, this feature could be used for supporting Prometheus middleware metrics on a common basis.

```rust
    fn client_intercept(req: Request<()>) -> Result<Request<()>, Status> {
        println!("Intercepting client request: {:?}", req);

        let gm = req.extensions().get::<GrpcMethod>().unwrap();
        assert_eq!(gm.service(), "test.Test");
        assert_eq!(gm.method(), "UnaryCall");

        Ok(req)
    }
    let mut client = TestClient::with_interceptor(channel, client_intercept);
    client.unary_call(Request::new(Input {})).await.unwrap();
```

## Solution

Update **tonic-build** to generate client-side codes that support the **GrpcMethod** extension. Why not review the code changes in this file, tonic-health/src/generated/grpc.health.v1.rs, to see the differences? 😄


tonic/src/extensions.rs
```diff
@@ -71,3 +71,12 @@ impl fmt::Debug for Extensions {
         f.debug_struct("Extensions").finish()
     }
 }
+
+/// A gRPC Method info extension.
+#[derive(Debug)]
+pub struct GrpcMethod<'a> {
+    /// gRPC service name
+    pub service: &'a str,
```
LucioFranco, line 79: These should both be moved to `'static` and be made private. We should then expose `service() -> &str` and `method() -> &str` to avoid exposing that 1) these are 'static (incase we want to change that in the future) and 2) in case we want to add new fields we can do that and 3) we remove the `'a` lifetime on the struct. Since, with extensions https://docs.rs/http/latest/http/struct.Extensions.html#method.insert always requires `'static` there is no point to have a lifetime.

tonic-build/src/client.rs
```diff
@@ -309,7 +337,9 @@ fn generate_streaming<T: Method>(
             })?;
             let codec = #codec_name::default();
             let path = http::uri::PathAndQuery::from_static(#path);
-            self.inner.streaming(request.into_streaming_request(), path, codec).await
+            let mut req = request.into_streaming_request();
+            req.extensions_mut().insert(GrpcMethod{service: #service_name, method: #method_name});
```
LucioFranco, line 341:
```suggestion
            req.extensions_mut().insert(GrpcMethod {service: #service_name, method: #method_name});
```

tonic-build/src/client.rs
```diff
@@ -211,15 +220,19 @@ fn generate_methods<T: Service>(
     stream
 }
 
-fn generate_unary<T: Method>(
-    method: &T,
+fn generate_unary<T: Service>(
+    service: &T,
+    method: &<T as Service>::Method,
```
LucioFranco, line 225:
```suggestion
    method: &T::Method,
```

This should work no?

LucioFranco (changes requested):

tonic/src/extensions.rs
```diff
@@ -71,3 +71,26 @@ impl fmt::Debug for Extensions {
...
+/// A gRPC Method info extension.
+#[derive(Debug)]
+pub struct GrpcMethod {
+    service: &'static str,
+    method: &'static str,
+}
+
+impl GrpcMethod {
+    /// Create a new `GrpcMethod` extension.
+    pub fn new(service: &'static str, method: &'static str) -> Self {
```
LucioFranco, line 84: I would like to make this `#[doc(hidden)]` so that codegen can use it but users can't find it in the docs since I am not sure if we want to expose the fact that internally its `&'static`.

LucioFranco (changes requested):
Everything looks good, have one more request then we can merge.

LucioFranco (approved):
Thanks for following through on this!

PR #1281, opened 2023-02-18
chore(tonic): Remove duplicate tls feature gate and re-export http crate instead of using wildcard importing

tottoto:
## Motivation

Simplifies the codes.

## Solution

Removes duplicated tls feature gate and re-exports http crate instead of using wildcard importing.


LucioFranco (approved):
Thanks!

PR #1283, opened 2023-02-18
chore(health): Fix clippy lint

tottoto:
Fixes a clippy lint error. It's strange that this hasn't been detected.

PR #1284, opened 2023-02-19
chore(tonic): Use BoxBody type alias

tottoto:
We can use this type alias.


tottoto:
clippy ci is fixed in #1283.

LucioFranco:
https://github.com/hyperium/tonic/pull/1288 Should fix clippy :)

PR #1309, opened 2023-03-10
chore(tonic): Remove redundant reference to reference

tottoto:
Removes redundant reference to reference.


LucioFranco (approved):
Thanks!

PR #1349, opened 2023-04-11
chore: Update to baptiste0928/cargo-install@v2

tottoto:
Updates to baptiste0928/cargo-install@v2.


LucioFranco (approved):

LucioFranco:
Ah seems there is a CI conflict once we fix that we can merge!

tottoto:
Resolved the conflict.

tottoto:
The failed `cargo-deny` ci would be fixed in #1419.

Issue #1351, opened 2023-04-12
Server traits with `Arc<Self>` as argument.


h33p:
## Feature Request

It would be great to have ability to support defining server traits as accepting `Arc<Self>` argument, instead of just `&self`.

### Crates

`tonic_build`

### Motivation

Handling streaming requests usually requires `tokio::spawn`, thus suddenly you can't use data from self, unless data inside was also wrapped with (unnecessary) arc.

### Proposal

From what I can see, codegen needs to be changed here: https://github.com/hyperium/tonic/blob/master/tonic-build/src/server.rs#L425 to avoid converting Arc<Self> to &self. We could have a codegen config switch to use the new way. I don't see any drawbacks apart from increased verbosity upon flipping that switch, but it would be completely opt-in, and later could also be made more fine-grained.

Finally, I volunteer to go ahead and implement this :)

### Alternatives

- Wrapping variables inside `Self` with another arc.
- Implementing service traits on `Arc<T>`? (is it even possible?)

Both of these alternatives involve additional arc.

PR #1352, opened 2023-04-13
feat(build): Optionally to emit `Arc<Self>` as server receiver

h33p:
This implements an option to emit `Arc<Self>` instead of `&self` in server traits. This enables implementor to reference `Self` data for indefinite duration, which may be useful for streaming requests.

Fixes: #1351

I have a few comments about this:

This impl required me to generate a new `inner_arg` token stream for each method, because I could not find a way to easily coerce between `Arc<T>` and `&T` in a generic way that does not conflict with ambiguous function names.

Tests are very minimal, if this feature looks good otherwise, I'll add more.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,5 +23,6 @@ members = [
   "tests/compression",
   "tonic-web/tests/integration",
   "tests/service_named_result",
+  "tests/use_arc_self",
 ]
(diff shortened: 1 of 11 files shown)


LucioFranco (approved):
LGTM thanks!

PR #1356, opened 2023-04-15
chore: Use taiki-e/install-action to setup protoc

tottoto:
## Motivation

We've had a lot of warnings `Node.js 12 actions are deprecated.` due to aruduino/setup-protoc we've used to setup protoc.  This action does not seem to be actively maintained at the moment.

## Solution

Replaces arduino/setup-protoc with taiki-e/install-action, which is used in tokio.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,10 +40,10 @@ jobs:
       uses: baptiste0928/cargo-install@v1
       with:
         crate: cargo-machete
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):

LucioFranco:
there is a conflict, but can merge once that is fixed

tottoto:
Fixed.

PR #1358, opened 2023-04-16
forward specialized BytesMut methods on {Encode,Decode}Buf

srijs:
Forwards more `Buf`/`BufMut` methods on `EncodeBuf` and `DecodeBuf` to take advantage of specialized implementations on `BytesMut`, where available. 

## Motivation

`BytesMut` has some optimized variants of `Buf` and `BufMut` methods that users of `EncodeBuf` and `DecodeBuf` currently miss out on. In particular, an O(1) zero-copy variant of `Buf::copy_to_bytes` and variants of `BufMut::put` and `BufMut::put_slice` that elide checking `remaining_mut`/`advance_mut`.

## Solution

Provide specialized implementations of `Buf::copy_to_bytes` on `DecodeBuf` as well as `ButMut::put` and `BufMut::put_slice` on `EncodeBuf` which forward these calls onto the underlying `BytesMut` implementations.


tottoto:
Thanks. Seems to be addressed at #1695.

PR #1359, opened 2023-04-16
optimize header name handling in `Grpc::map_response`

srijs:
Avoid parsing reserved header names on every call to `MetadataMap::into_sanitized_headers` as well as in `Status::to_http`. More explanation below. The change should be relatively straight forward and low-complexity, but let me know if you'd like to see an isolated benchmark for this regardless!

## Motivation

The `MetadataMap::into_sanitized_headers` function is called for every response transitively through `Grpc::map_response` and `Response::into_http`.

The way that it currently sanitizes headers by removing the reserved GRPC headers involves calling `HeaderMap::remove` with a `&str` key, which results in the string key being parsed into a header name at runtime for every one of the reserved headers (see the call [here](https://github.com/hyperium/http/blob/3a2f3e01c01e60b894d1f5b30554013d990010c6/src/header/map.rs#L3422)).

This makes it show up on my profiles in the single-digit percentage range, which seemed worthwhile improving to me, especially considering the low complexity of the change.

A similar situation exists in `Status::to_http` where the GRPC status header names are being parsed at runtime when inserted into the header map.

## Solution

The solution implemented is to parse the reserved headers at compile time using the `const` function `HeaderName::from_static`. This eliminates any runtime overhead of parsing the reserved header names.


srijs:
@LucioFranco this should be a relatively simple chance, any indication as to whether you'd be able to merge it or have any feedback would be appreciated, thanks!

PR #1364, opened 2023-04-19
chore: Use cargo-udeps to check unused dependencies

tottoto:
## Motivation

Checks unused dependencies more strictly.

## Solution

Replaces cargo-machete with cargo-udeps, and fixes detected dependencies and features.

### DIFF
--- a/tonic/src/transport/service/discover.rs
+++ b/tonic/src/transport/service/discover.rs
@@ -9,7 +9,7 @@ use std::{
 };
 use tokio::sync::mpsc::Receiver;
 
-use futures_core::Stream;
+use tokio_stream::Stream;
 use tower::discover::Change;
 
 type DiscoverResult<K, S, E> = Result<Change<K, S>, E>;
--- a/tonic/src/transport/service/discover.rs
+++ b/tonic/src/transport/service/discover.rs
@@ -9,7 +9,7 @@ use std::{
 };
 use tokio::sync::mpsc::Receiver;
 
-use tokio_stream::Stream;
+use futures_core::Stream;
 use tower::discover::Change;
 
 type DiscoverResult<K, S, E> = Result<Change<K, S>, E>;
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -23,6 +23,9 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macOS-latest, windows-latest]
+        include:
+        - os: windows-latest
+          option: --exclude-features uds
     steps:
     - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@nightly
@@ -33,7 +36,7 @@ jobs:
       with:
         tool: protoc@3.20.3
     - uses: Swatinem/rust-cache@v2
-    - run: cargo hack udeps --workspace --ignore-private --each-feature
+    - run: cargo hack udeps --workspace --each-feature ${{ matrix.option }}
 
   check:
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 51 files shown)


tonic/src/transport/service/discover.rs
```diff
@@ -9,7 +9,7 @@ use std::{
 };
 use tokio::sync::mpsc::Receiver;
 
-use tokio_stream::Stream;
+use futures_core::Stream;
```
LucioFranco, line 12: why this change?

tottoto: As tonic does not use tokio-stream except Stream which is the re-exporting of futures-core, we could use futures-core directly to reduce dependencies. This is detected by cargo-udeps.

tottoto: Can I ask the purpose of favoring `tokio-stream` with unused additional dependencies over `futures-core`?

LucioFranco: Ideally, we'd like to move off of the futures crates (stream is complicated since we expected it to land in std by now) since they are not as maintained as the tokio ones. Its a bit of a complicated history.

tottoto: Thanks for your answer. I am a bit confused, and would like to ask some more questions to better update this pull request. Even if we depend on `tokio-stream` instead of `futures-core`, as `Stream` we use is still provided by `futures-core`, we effectively depend on `futures-core`. Therefore I think that even if we use `tokio-stream` this time, we just add dependencies that we don't use and can't achieve our goal of not using `futures-core`. I would like to know a little more about this point.

LucioFranco: Eventually, `Stream` will go into `std` so then we will only depend on `tokio-stream` but for now we can't do that since progress has been slow there. That said, I would rather be positioned for that future than live with one extra dependency right now if that makes sense. In addition, we control tokio-stream so its better to depend on that then directly depend on another package if that makes sense. Happy to explain more over discord/email/what ever if you're interested 😄

tottoto: Thank you for your kind explanation!

> we control tokio-stream so its better to depend on that then directly depend on another package if that makes sense

Thanks to this point mainly, I think I reached to understand why we should depend on `tokio-steam` instead of `futures-core`. I will investigate whether it is possible to configure that we should explicitly depend on `tokio-stream`.

tottoto: @LucioFranco Should we replace the existing all `futures_core::Stream` with `tokio_stream::Stream` in this pull request, or should it be done in the another pull request?

tottoto: I found that the `cargo-udeps` checking will success by replacing all `futures_core::Stream` with `tokio_stream::Stream`! Therefore I think we should do it in this pull request.

tottoto:
`cargo-deny` is failed due to duplicate version of `regex-syntax`. It will be fixed in #1373.

tonic/src/transport/service/discover.rs
LucioFranco: Yeah, we should convert everything from futures crates over. I think I originally used it but would like to remove it but I have not had time. Feel free to start on that!

LucioFranco:
I updated your PR and with the commit from master that fixes it all. Should be good to go.

tottoto:
The msrv test is failed because of https://github.com/hyperium/tonic/pull/1364/commits/9d9458d9e632613f35f34c877bc162b3a03967b6. `std::task::ready` (and `core::task::ready`) was stabilized at 1.64.0, but our msrv is 1.60.

LucioFranco (approved):
Sorry for the delay! Thanks!

PR #1367, opened 2023-04-21
Implement Deref for Request

ThouCheese:
Hello! This pull request adds a Deref impl to the Request struct. 

## Motivation
This is convenient for using copy fields that are in the message, or for borrowing from it. The get_ref function also allows this, but it feels intuitive to be able to do `request.field` regardless.

## Solution
Implement Deref<Target = T> for the Request<T> struct.

### DIFF
--- a/tonic/src/request.rs
+++ b/tonic/src/request.rs
@@ -17,6 +17,15 @@ pub struct Request<T> {
     extensions: Extensions,
 }
 
+impl<T> std::ops::Deref for Request<T> {
+    type Target = T;
+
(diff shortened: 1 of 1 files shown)


LucioFranco:
Hi! Thank you for the contribution, but the reason that `Request` doesn't implement deref is because this is bad style in rust to prefer deref over get_ref in this situation. As well the lower level http library has made the same choice on its Request type https://docs.rs/http/latest/http/request/struct.Request.html. For that reason, I am going to close this PR since this is not actually what we want.

PR #1368, opened 2023-04-22
docs: fix broken doc links 

martinabeleda:
## Motivation

See https://github.com/hyperium/tonic/issues/1365

## Solution

Updated the links to refer to https://httpwg.org/specs/rfc9113.html. There's no links to the individual settings, so I've just linked to the section for each. Also added a missing link for `SETTINGS_MAX_FRAME_SIZE` for consistency


tottoto:
Addressed in #1411. Thank you!

PR #1373, opened 2023-04-24
chore: Add regex-syntax to deny skip-tree

tottoto:
## Motivation

Fixes ci failure due to the duplicate version of regex-syntax.

## Solution 

Adds regex-syntax to the deny skip-tree.


LucioFranco:
oh sorry didnt see this, I fixed it here https://github.com/hyperium/tonic/pull/1374

Issue #908, opened 2022-02-07
build: Expose `bytes` config from prost


LucioFranco:

zuston:
Any update on this?

Congee:
Really hope #1376 could be merged. There is a workaround with `compile_with_config` now:

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let mut config = prost_build::Config::new();
    config.bytes(&["."]);

    tonic_build::configure().compile_with_config(config, &["proto/message.proto"], &["."])?;

    Ok(())
}
```

PR #1376, opened 2023-04-26
build: expose `bytes` and `btree_map` config from prost

stamp711:
## Motivation

Since we want to evaluate whether using `Bytes` in protobuf messages can lead to a noticeable performance improvement in our system, I thought it's a good idea to address issue #908.

## Solution

Add the `bytes` and `btree_map` config functions for `tonic_build::Builder`.

Please note that those two functions take iterators to paths as input and **override** previously configured paths for `bytes` or `btree_map`. This is different from the `box` config function, which extends previously configured paths instead of overriding them.

I personally do not like this inconsistency, but I'll leave them as they are so the behaviour matches `prost-build`.

Fixes: #908

### DIFF
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -28,6 +28,8 @@ pub fn configure() -> Builder {
         enum_attributes: Vec::new(),
         type_attributes: Vec::new(),
         boxed: Vec::new(),
+        btree_map: None,
+        bytes: None,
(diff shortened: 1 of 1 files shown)


zuston:
If you have time, could you help review this? @LucioFranco This is really useful for me to develop a storage system.

Thanks for proposing this. Do you have checked this validity? @stamp711

PR #1377, opened 2023-04-28
chore: Remove OptionPin

taiki-e:
## Motivation

`Option` has a method called `as_pin_mut` for pin projection.

## Solution

Remove `OptionPin` and use `Option::as_pin_mut` instead.


LucioFranco:
Thanks!

PR #1397, opened 2023-05-17
tonic-build: Strip service prefix before matching methods

### LINKED ISSUE #1396: Generated server's service should not match full paths
## Feature Request
tonic-build 0.9.2 generates `tonic::codegen::Service` with every call matching full `URI` instead of stripping service prefix

It should strip prefix `/<prefix>/` and match only on method name (or better use hash)

### Crates

```
tonic-build 0.9.2
```

### Motivation

Performance

### Proposal

Generated code could be simple along lines of something like:

```
    impl<T, B> tonic::codegen::Service<http::Request<B>> for TaskServer<T>
    where
        T: Task,
        B: Body + Send + 'static,
        B::Error: Into<StdError> + Send + 'static,
    {
        type Response = http::Response<tonic::body::BoxBody>;
        type Error = std::convert::Infallible;
        type Future = BoxFuture<Self::Response, Self::Error>;
        fn poll_ready(
            &mut self,
            _cx: &mut Context<'_>,
        ) -> Poll<std::result::Result<(), Self::Error>> {
            Poll::Ready(Ok(()))
        }
        fn call(&mut self, req: http::Request<B>) -> Self::Future {
            let inner = self.inner.clone();
            let path = req.uri().path();
            if let Some(method) = path.strip_prefix("/tasks.Task/") {
                match method {
                    "method1" => todo!(),
                    _ => todo!(),
                }
            }
       }
}

```

DoumanAsh:
Fixes #1396


DoumanAsh:
I rebase my PR to resolve conflict and will re-check tests soon

DoumanAsh:
Since `tonic-health` includes generated file, I committed it
Is it ok?

DoumanAsh:
I removed generated code from PR until you decide to merge it as otherwise it is going to conflict

DoumanAsh:
@LucioFranco Do you need this PR?

PR #1399, opened 2023-05-17
fix(tls): encode der to pem

Direktor799:
## Motivation
When using tonic with mtls, we expect the method `peer_certs` return PEM encoded certs since `Certificate` is described as a PEM cert in the doc, but it actually returns a DER encoded certs.

## Solution
- We could return the original `tokio_rustls::rustls::Certificate` with `TlsStream::peer_certs` since the `TlsStream` is import from `tokio_rustls` anyway.

- As for `Request::peer_certs`, we could encode the DER certs with `pem` crate, so that it could be turned into a `tonic::transport::Certificate` properly. Now this method does more than just `Arc::clone`, so I removed the `Arc` in return value.

I doubt that anyone would want a PEM encoded cert form peer, but seems like we've been trying to get rid of `rustls` types, and it's not worthy to change the `tonic::transport::Certificate` to be compatible with DER for just this case.


djc:
I would dislike the requirement to pull in the pem crate only because the `tls` feature is being used. If you don't want to have `rustls::Certificate` in your public API, just yield a `Vec<u8>` or define your own trivial `Certificate` newtype?

(Note that rustls will in the future start using types from the new [rustls-pki-types](https://crates.io/crates/rustls-pki-types) crate with a long-term stable API. Might be a decent alternative?)

tottoto:
Addressed at #1707. Thank you!

PR #1408, opened 2023-06-06
Support for adding services dynamically.

tsroka:
## Motivation

This change allows more flexible creation of grpc server, specifically for use case where multiple grpc services are initiated in different parts of an application. Currently it is pretty hard to do due to Server / Router generic (and consuming value every time new svc is added).

This should also address this [issue](https://github.com/hyperium/tonic/issues/886)

## Solution

Since Routes struct does not have any generic type it can be easily used to add new services. The only complication is that it `add_service` fn is taking self by value and functions are not public. In order to make it work, RoutesBuilder was added that takes mutable reference to self and can be used to add new services, and some methods were made public. 

Please see the provided example. This change is backwards compatible.

## Alternative

Other approach would be to get rid of generic parameter from `Router` and decouple it from `Server` so that Router can be created first and latter passed to `Server` once all services are added. The current coupling between Router/Server is unnecessary complex (Router can be created only through instance of Server, which is cloned value stored as a Router field) and does not seem to provide any benefits.


examples/Cargo.toml
```diff
@@ -253,6 +253,10 @@ name = "richer-error-server-vec"
 path = "src/richer-error/server_vec.rs"
 required-features = ["types"]
 
+[[bin]]
+name = "dynanmic-server"
```
andrewhickman, line 257:
```suggestion
name = "dynamic-server"
```

killix:
I am looking for this feature. When can this be merged? :)

tsroka:
Hello, is there something else I need to do on my side to get it merged?

killix:
@andrewhickman any answer?

andrewhickman:
I'm not a maintainer on this repo, I don't have power to merge it

killix:
> I'm not a maintainer on this repo, I don't have power to merge it

Asked because look you are the reviewer on this PR.

LucioFranco (changes requested):
Overall this looks really good. I would like to see a test that uses this code to ensure we don't break things down the line then we can merge this. Thanks!

tsroka:
Thanks for reviewing @LucioFranco ; added test, lmk if this is what you were looking for.

tsroka:
@LucioFranco I saw that a build is failing due to use of let else, I changed the test to use match instead, should be good now.

tsroka:
Fixed merge conflicts with master and compilation errors introduced by changes in master. @LucioFranco do you think it is ok to merge?

LucioFranco (approved):

PR #1410, opened 2023-06-10
Add Virtio VSOCK support

jul-sh:
## Motivation

gRPC recently merged support for Virtio VSOCK as transport. It would be helpful for [us](https://github.com/project-oak/oak) if tonic were to match gRPC's support for it.

gRPC PR that adds VSOCK support: https://github.com/grpc/grpc/pull/32847

## Solution

Add Virtio VSOCK support, integration test, and example

cc @tiziano88

### DIFF
--- a/examples/src/vsock/client.rs
+++ b/examples/src/vsock/client.rs
@@ -1,0 +1,47 @@
+#![cfg_attr(not(unix), allow(unused_imports))]
+
+pub mod hello_world {
+    tonic::include_proto!("helloworld");
+}
+
+use hello_world::{greeter_client::GreeterClient, HelloRequest};
+use tokio_vsock::VsockStream;
+use tonic::transport::{Endpoint, Uri};
+use tower::service_fn;
+
+// Use vsock-loopback so we don't need to spin up a VM. Must match server.
+static TEST_CID: u32 = vsock::VMADDR_CID_LOCAL;
+// Arbitrarily chosen. Must match server.
+static TEST_PORT: u32 = 8000;
+
+// Virtio VSOCK does not use URIs, hence this URI will never be used.
+// It is defined purely since in order to create a channel, since a URI has to
+// be supplied to create an `Endpoint`.
+static IGNORED_ENDPOINT_URI: &str = "file://[::]:0";
+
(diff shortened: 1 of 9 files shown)


examples/src/vsock/client.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use hello_world::{greeter_client::GreeterClient, HelloRequest};
+use tokio_vsock::VsockStream;
+use tonic::transport::{Endpoint, Uri};
+use tower::service_fn;
+
+// Use vsock-loopback so we don't need to spin up a VM. Must match server.
+static TEST_CID: u32 = vsock::VMADDR_CID_LOCAL;
+// Arbitrarily choosen. Must match server.
+static TEST_PORT: u32 = 8000;
```
jul-sh, line 15: Personally I would prefer to store this in a public variable and import them in both client & server.

However, I decided to repeat them in each place they are used, to match the other examples & tests.

examples/src/vsock/client.rs
```diff
@@ -0,0 +1,47 @@
...
+
+// Virtio VSOCK does not use URIs, hence this URI will never be used.
+// It is defined purely since in order to create a channel, since a URI has to
+// be supplied to create an `Endpoint`.
+static IGNORED_ENDPOINT_URI: &str = "file://[::]:0";
```
jul-sh, line 20: Not proud of this, but this matches the code of the UDS example. 

In the long term it would probably be a good idea to update the endpoint struct to no longer require a URI. However that should be done in a separate PR, which also updates the UDS code.

jul-sh:
cc @LucioFranco for review

LucioFranco:
I believe a lot of this could be achieved outside of tonic. So I don't see why we should add another dependency when you can new type and wrap it in your own code?

I am weary of adding this project as a dependency as well since it states:

> tokio-vsock is for the most part pre-alpha quality, so there are probably sharp edges. Please test it thoroughly before using in production. Happy to receive pull requests and issue reports.

In the readme, while I probably trust the code I wouldn't want to include it in a wide spread project like this. IF for some reason you can't work around it then we should try to fix that problem.

Thanks!

PR #1414, opened 2023-06-16
Update Cargo.toml

Dylan-DPC:
Updates dependency to the patch version of 0.3.17

[Advisory](https://rustsec.org/advisories/RUSTSEC-2023-0034.html)


LucioFranco:
We can merge this but because cargo picks up the latest version I am not gonna rush to get this out into a release. (Also the original bug was uh a bit extreme for a CVE IMHO).

PR #1419, opened 2023-06-22
chore: Add bitflags to deny skip

tottoto:
## Motivation

Fixes a `cargo-deny` error.

## Solution

Allows `bitflags` duplicate version error.


tottoto:
Fixed in #1434.

PR #1421, opened 2023-06-24
chore: Reduce futures-util in favor of tokio-stream

tottoto:
## Motivation

Continues #1364.

## Solution

Reduces `futures-util` in favor of `tokio-stream`.


tottoto:
Rebased.

PR #1423, opened 2023-06-27
feat(core): amortize many ready messages into fewer, larger buffers

jgraettinger:
Introduce an EncodedBytes combinator which encodes multiple ready messages - up to a yield threshold - before yielding the next bytes buffer.

Or, if the message stream polls to pending, yield the available bytes immediately.

These ammortized buffers exhibit far better throughput when streaming a high rate of small messages, because hyper and h2 avoid copying the yielded buffer and dispatch each as a separate, non-vectorized tcp send.

## Motivation

We have observed that tonic exhibits poor CPU performance when sending a high data-rate of small messages. Each message is dispatched as a single `tcp_sendmsg`, and each incurs overhead for routing and other kernel activities in the network path. It's incumbent to increase the size of network sends without adding latency.

## Solution

Introduce a combinator which repeatedly polls its delegate stream for ready messages, extending a current buffer (up to a yield threshold). This approach avoids any new allocation and adds just one "magic" constant: a yield threshold after which the combinator will immediately yield the next chunk of bytes.


LucioFranco (approved):
LGTM! Thanks!

tonic/src/codec/encode.rs
```diff
@@ -47,55 +46,125 @@ where
...
+///  * The encoded buffer surpasses YIELD_THRESHOLD.
+#[pin_project(project = EncodedBytesProj)]
+#[derive(Debug)]
+pub(crate) struct EncodedBytes<T, U>
+where
+    T: Encoder<Error = Status>,
+    U: Stream<Item = Result<T::Item, Status>>,
+{
+    #[pin]
+    source: Fuse<U>,
```
LucioFranco, line 72: We need to drop fuse for implementing that logic internally within this, as we don't want to depend on futures-util for now.

LucioFranco (changes requested):
If we can get the merge conflicts resolved and remove fuse/use `tokio-stream` then we can merge it.

LucioFranco (approved):

PR #1425, opened 2023-07-03
Allow to duplicate `GrpcMethod` extensions

weiznich:
## Motivation

I have a custom middleware implementation that retries requests based on certain conditions. That requires cloning the request including all extensions. On previous tonic versions there were no extensions, so we could simply ignore them. While updating to tonic 0.9 I've noticed that an extension was added to each request. It's unfortunately not easily possible to duplicate the extension, even if the internal details are trivially clonable.

## Solution

Add a `#[derive(Clone)]` to the corresponding type

### DIFF
--- a/tonic/src/extensions.rs
+++ b/tonic/src/extensions.rs
@@ -73,7 +73,7 @@ impl fmt::Debug for Extensions {
 }
 
 /// A gRPC Method info extension.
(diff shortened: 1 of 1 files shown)


weiznich:
@LucioFranco Any chance to get a review on this?

PR #1429, opened 2023-07-12
support protobuf crate for gRPC reflect

woshihaoren4:
## Motivation
The prost crate does not support getting grpc options content by reflection.

## Solution
For the grpc reflection service, an implementation based on protobuf crates is provided.
And let the user choose that way by feature.


woshihaoren4:
yes

LucioFranco:
Tonic has no plans to include `protobuf` as a dependency so I will not accept this PR. I would recommend creating this as an external crate and if you need stuff exposed from the core crates here we can do that. Thanks!

PR #1438, opened 2023-07-26
chore(interop): Replace futures crates with tokio-stream

tottoto:
## Motivation

Continuing #1421.

## Solution

This removes `futures-core` and `futures-util` from `interop`.


LucioFranco (approved):
Thank you!

PR #1452, opened 2023-08-04
chore: Improve cargo doc ci

tottoto:
## Motivation

Fixes unnecessary doc tests being executed and the redundant ci definitions, and make the doc test fail when their errors are thrown. I think we can remove doc test on test crates (they might throw unnecessary document warnings or errors as well. For example, [some warnings](https://github.com/hyperium/tonic/actions/runs/5754277946/job/15599233527#step:6:64) have been appeared).

## Solution

- Changes the doc tests to be run only against those which are aimed to be published.
- Makes the doc test fail when they throw warnings.
- Moves doc test to the msrv job as the most of the test process is same.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,26 +70,6 @@ jobs:
     - name: Check all targets
       run: cargo check --all --all-targets --all-features
 
-  check-docs:
-    name: check docs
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Thanks!

PR #1457, opened 2023-08-09
chore(tonic): Reduce futures-util

tottoto:
## Motivation

Continuing of reducing futures crates.

## Solution

Replaces them with alternative implementations.


tottoto:
The failed cargo-deny check ci would be fixed in #1456.

LucioFranco (approved):

PR #1461, opened 2023-08-11
chore: Refactor ci

tottoto:
## Motivation

Improves readability and maintainability.

## Solution

- Defines `RUSTFLAGS` environment variables at one place which are aimed to use every jobs.
- Moves `QUICKCHECK_TESTS` environment variable at the place to be used.
- Removes redundant matrix variable `rust`, which is only `stable`.
- Reorders steps for the uniform order.
- Updates deprecated cargo option `--all` to `--workspace`.
- Adds msrv comment to be easily found by command like grep.
- Makes `check` and `msrv` job next to each other as they have similar structure.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -21,6 +21,13 @@ jobs:
         components: rustfmt
     - run: cargo fmt --all --check
 
+  deny-check:
+    name: cargo-deny check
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Nice work! Thank you!

PR #1477, opened 2023-08-25
chore(examples): Add `cancellation` example

tcohen01:
## Motivation

There are some valid usecases for being notified of request cancellation and executing clean up logic on client cancellation.
This can be easily done with server streaming and bidi streaming, but there doesn't seem to be a straightforward approach for unary RPCs.

## Solution

Creating a new task for unary requests (like in server/bidi streaming requests), having the unary request' future await on this new task with a CancellationToken drop guard.

Under the assumption that request futures are dropped when client cancels the request, this will notify the task through the drop guard's drop logic

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -268,6 +268,10 @@ name = "cancellation-server"
 path = "src/cancellation/server.rs"
 required-features = ["cancellation"]
 
+[[bin]]
+name = "cancellation-client"
+path = "src/cancellation/client.rs"
+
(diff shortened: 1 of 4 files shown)


tcohen01:
Example is testable using hello world's client - simply let the server finish handling the request or stop the hello world client before the server finishes the request.

Also, not sure if the futures in this example should be fused before selecting on them - will add if so.

LucioFranco (approved):

PR #1478, opened 2023-08-26
chore: Remove futures crates

tottoto:
## Motivation

Completes of removing futures crates.

## Solution

- Replaces `futures_{core, util}::ready` with `std::task::ready`, which was stabilized at Rust 1.64.
- Uses `pin_utils` crate directly instead of via `futures_util`. (When `tonic`'s MSRV be bumped to Rust `1.68` in the future, `std::pin::pin` can be used.)

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -37,7 +37,8 @@ transport = [
   "channel",
   "dep:h2",
   "dep:hyper",
-  "tokio",
+  "tokio/net",
+  "tokio/time",
   "dep:tower",
   "dep:hyper-timeout",
 ]
@@ -50,10 +51,10 @@ channel = []
 [dependencies]
 base64 = "0.21"
 bytes = "1.0"
-pin-utils = "0.1"
 http = "0.2"
 tracing = "0.1"
 
+tokio = "1.0.1"
 http-body = "0.4.4"
 percent-encoding = "2.1"
 pin-project = "1.0.11"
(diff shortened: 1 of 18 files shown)


tonic/Cargo.toml
```diff
@@ -50,7 +50,7 @@ channel = []
 [dependencies]
 base64 = "0.21"
 bytes = "1.0"
-futures-util = {version = "0.3", default-features = false}
+pin-utils = "0.1"
```
LucioFranco, line 53: I think we can use https://docs.rs/tokio/latest/tokio/macro.pin.html

LucioFranco (changes requested):

tonic/Cargo.toml
tottoto: For now, `tokio` has been used behind `transport` feature. To use `tokio::pin` macro, we have to use `tokio` throughout `tonic` implementations. I think `tonic` is mostly used with `tokio` so it seems not to be really matter. However, as I'm not sure whether it is appropriate to always depend on `tokio` only for using `tokio::pin`, I would like to hear your opinion.

LucioFranco: If we depend on tokio's default features (which should be basically empty) then it should be available and pretty light weight.

tottoto: Thanks for reviews! I'll fix to use `tokio`.

tottoto: Fixed.

LucioFranco (approved):

LucioFranco:
Thanks!

PR #1487, opened 2023-08-29
chore(tls): Small refactoring

tottoto:
## Motivation

Small refactorings of `tonic`'s tls config.

## Solution

- Removes `rustls` crate from dependencies as it is not used.
- Updates to `rustls-native-certs` 0.6.3 from 0.6.1 and use its `AsRef<[u8]>` for `Certificate`.
- Uses `AsRef<u8>` for `tonic`'s `Certificate` struct instead of slice syntax.


djc (approved):
Drive-by review: LGTM!

PR #1493, opened 2023-09-01
chore: Remove base64 from deny skip-tree

tottoto:
## Motivation

Reflects the current state.

## Solution

As `headers` 0.3.9 updated to `base64` 0.21, the duplicate version of `base64` is resolved.


tottoto:
Taken over by #1503.

PR #1494, opened 2023-09-01
chore(tonic): Move async-stream crate to transport section

tottoto:
## Motivation

`async-stream` is now used in `transport` feature.

## Solution

Moves `async-stream` crate to `transport` section.


tottoto:
Rebased to cleanup the commit log.

djc (approved):

PR #1497, opened 2023-09-03
chore(transport): Use hyper timeout feature

tottoto:
## Motivation

Simplifies implementations.

## Solution

`hyper` has connection timeout mechanism. `tonic` could use it instead of `hyper-timeout`'s connection when using `hyper` connector. As `connect_with_connector` does not assume only `hyper`'s connection, `hyper-timeout` cannot be removed at this point.


djc (approved):
LGTM, thanks!

PR #1500, opened 2023-09-03
chore: Fix clippy warning

tottoto:
Fixes clippy warnings except for [default_constructed_unit_structs](https://rust-lang.github.io/rust-clippy/master/index.html#/default_constructed_unit_structs), which I am not sure whether this lint is appropriate for `tonic`.


tottoto:
The failed MSRV CI would be fixed in #1549.

PR #1505, opened 2023-09-07
feat: Support thread-local functionality

inq:
This adds a feature "current-thread" which generates a trait with `#[async_trait(?Send]`, like this:

```
#[tonic::async_trait(?Send)]
 impl test_server::Test for Svc {
     async fn test_request(&self, req: Request<SomeData>) -> Result<Response<SomeData>, Status> {
         Ok(Response::new(req.into_inner()))
     }
 }
```

I've tried make the PR as small as possible, but I apologize for the size. But the changes are really simple!

## Summary

1. Add `local_executor` option for tonic-build. So If we activate it, we generate altered code using `async_trait(?Send)`. We can see an example in ‘tests/current_thread`
2. Remove axum dependency: https://github.com/hyperium/tonic/issues/1501
3. We can distinguish the environments by two structs implementing `hyper::Executor`. They are existing `TokioExec` and newly added `LocalExec`.
4. Now some structs and traits receive new generic argument, `Ex`. Its default variable is `TokioExec`, so we can preserve compatibility. Plus, to avoid direct access with `LocalExec`, we need to use type aliases. For example, `Routes` became `Routes<Ex = TokioExec>`, and `LocalRoutes` is added as a type alias for `Routes<LocalExec>`.
5. The only one place with pain is `transport::server::executor::define_serve_with_shutdown` macro. I could not find a better way, but I think we need to wait until https://github.com/rust-lang/rust/issues/63063 is finished.


LucioFranco:
So I think we can reduce the diff here by a lot if we don't go with the feature flag method and just add a config option to tonic-build. If its more complicated that we need to rework send bounds everywhere then that might be more complicated and become out of scope for now.

inq:
Thank you for a good advice. Let me try another way related with trait bound.

inq:
I am sorry for late submission. It's almost done except adding test cases. Would it be okay if we use trait-bound way? I can ensure that:

1. Compatible except axum::Router
2. zero-cost since it is decided on the compile-time
3. Even though there were many changes, it would be easier to manage type bounds since I moved them into tonic::util::body

inq:
Done! I've added some tests (copied from integration tests)

LucioFranco:

PR #1539, opened 2023-09-28
feat: optional `SSLKEYLOGFILE` support

crepererum:
## Motivation
Using `SSLKEYLOGFILE` is helpful when you want to intercept TLS traffic for debugging and is generally supported by many libraries and browsers. Also see: https://wiki.wireshark.org/TLS#using-the-pre-master-secret

## Solution
Add a `use_key_log` option to server and client TLS configs that -- when set -- will enable rustls's `SSLKEYLOGFILE` handling.

### DIFF
--- a/tonic/src/transport/channel/service/tls.rs
+++ b/tonic/src/transport/channel/service/tls.rs
@@ -26,12 +26,14 @@ pub(crate) struct TlsConnector {
 }
 
 impl TlsConnector {
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
         ca_certs: Vec<Certificate>,
         trust_anchors: Vec<TrustAnchor<'static>>,
         identity: Option<Identity>,
         domain: &str,
         assume_http2: bool,
+        use_key_log: bool,
         #[cfg(feature = "tls-native-roots")] with_native_roots: bool,
         #[cfg(feature = "tls-webpki-roots")] with_webpki_roots: bool,
     ) -> Result<Self, crate::BoxError> {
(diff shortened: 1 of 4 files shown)


crepererum:
Since this is mostly config wiring and relies on the already existing rustls features, does this need a test? If so, where should this test be placed?

tottoto:
Seems to be related to #893.

crepererum:
The solution there seems to be: just bypass `tonic::transport` entirely and write 100+ lines of glue-code to wire up all the components (`tower`, `hyper`, `rustls`) yourself. That seems doable, but somewhat defeats the purpose of `tonic`, i.e. having an easy-to-use client&server toolkit.

tottoto (approved):
This option is useful in terms of providing debuggability, and I think the benefits are worth the additional cost of adding this option. However, @LucioFranco has already rejected adding this feature in #893 and #1102, so we need his approval.

@LucioFranco As it's been a while since your previous review, I ask you if you've changed your mind and made this acceptable. If you accept it, I think it would be good to include this in the next v0.13 release.

LucioFranco (approved):

crepererum:
will rebase...

tonic/src/transport/channel/service/tls.rs
```diff
@@ -26,12 +26,14 @@ pub(crate) struct TlsConnector {
 }
 
 impl TlsConnector {
+    #[allow(clippy::too_many_arguments)]
```
crepererum, line 29: That's a bit cheating. If you want, I can refactor this, either using a parameter structure like

```rust
TlsConnectorParameters<'a> {
    ca_certs: Vec<Certificate>,
    trust_anchors: Vec<TrustAnchor<'static>>,
    identity: Option<Identity>,
    domain: &'a str,
    assume_http2: bool,
    use_key_log: bool,
}
```

or a builder pattern.

LucioFranco: this is fine, this is a private fn

PR #1542, opened 2023-10-05
skip debug

tpoliaw:
## Motivation

A recent version of prost-build added a `skip_debug` method to let users
manually implement debug for types generated from protobuf messages. While it
was possible to use this from tonic-build, it was a bit awkward and required an
additional direct dev-dependency on prost-build as well as additional steps to
create a config instance, configure properties of that, then do the same with
tonic's configuration before calling `compile_with_config` to combine them.

## Solution

This adds a new `skip_debug` method to the configuration Builder to handle the
prost config instance for the user.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ members = [
   "codegen",
   "interop", # Tests
   "tests/disable_comments",
+  "tests/manual_debug",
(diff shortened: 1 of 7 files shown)


rustedLightning:
It has been 7 months

PR #1560, opened 2023-10-28
chore: Refactor use of tokio-stream

tottoto:
## Motivation

Refactoring.

## Solution

Refactors the use of `tokio-stream`.


LucioFranco (approved):
Thanks!

PR #1561, opened 2023-10-30
chore(transport): Replace axum with matchit

tottoto:
## Motivation

Simplifies dependency tree, and implementations. `axum` seems to be a relatively too large library for the purpose of implementing `tonic`'s router. Directly using `matchit`, which is a crate used in `axum` seems to be simple for `tonic`.

## Solution

> **warning**
> This proposal contains breaking changes.

Replaces `axum` with `matchit` to implement the router.

### DIFF
--- a/examples/src/h2c/server.rs
+++ b/examples/src/h2c/server.rs
@@ -61,7 +61,7 @@ mod h2c {
 
     impl<S> Service<Request<Body>> for H2c<S>
     where
-        S: Service<Request<Body>, Response = Response<tonic::transport::AxumBoxBody>>
+        S: Service<Request<Body>, Response = Response<tonic::body::BoxBody>>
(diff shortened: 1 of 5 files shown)


LucioFranco:
So I've actually been using the axum support a lot https://github.com/tursodatabase/libsql/blob/main/libsql-server/src/http/user/mod.rs#L417 for example there. So I think we if we move to matchit we loose this ability to mix and match?

tottoto:
I think so. `axum`'s router seems not to be able to be got from the `matchit`'s one by the user side.

LucioFranco:
> I think so. axum's router seems not to be able to be got from the matchit's one by the user side.

Yeah then lets keep it how it is for now. its quite useful to use the axum router actually ive found.

PR #1567, opened 2023-11-08
feat(channel): Add local_address as an option Endpoint builder

jparris:
## Motivation
I need to be able to set the local address of a Channel. This is the source IP address the Channel will be coming from. Currently Tonic's Channel builder doesn't expose a way to do this.

## Solution
Add `local_address` to Endpoint builder.

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -36,8 +36,8 @@ pub struct Endpoint {
     pub(crate) http2_max_header_list_size: Option<u32>,
     pub(crate) connect_timeout: Option<Duration>,
     pub(crate) http2_adaptive_window: Option<bool>,
-    pub(crate) executor: SharedExec,
     pub(crate) local_address: Option<IpAddr>,
+    pub(crate) executor: SharedExec,
 }
 
 impl Endpoint {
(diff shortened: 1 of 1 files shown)


jparris:
@tottoto can you take a look at this PR.

tonic/src/transport/channel/endpoint.rs
```diff
@@ -325,12 +326,30 @@ impl Endpoint {
...
+    ///
+    /// This sets  the IP address the client will use. By default we let hyper select the IP address.
+    /// ```
+    /// # use std::net::IpAddr;
+    /// # use std::str::FromStr;
+    /// # use tonic::transport::Endpoint;
+    /// # let mut builder = Endpoint::from_static("https://example.com");
+    /// # builder.local_address(IpAddr::from_str("127.0.0.1").expect("Unable to parse IP address"));
+    /// ```
+    pub fn local_address(self, addr: IpAddr) -> Self {
```
tottoto, line 339: Can this take`Option<IpAddr>` as the argument.

tonic/src/transport/channel/endpoint.rs
```diff
@@ -325,12 +326,30 @@ impl Endpoint {
         )
     }
 
+    /// Set the local address
```
tottoto, line 329:
```suggestion
    /// Set the local address.
```

tonic/src/transport/channel/endpoint.rs
```diff
@@ -325,12 +326,30 @@ impl Endpoint {
...
+    ///
+    /// This sets  the IP address the client will use. By default we let hyper select the IP address.
+    /// ```
+    /// # use std::net::IpAddr;
+    /// # use std::str::FromStr;
+    /// # use tonic::transport::Endpoint;
+    /// # let mut builder = Endpoint::from_static("https://example.com");
+    /// # builder.local_address(IpAddr::from_str("127.0.0.1").expect("Unable to parse IP address"));
+    /// ```
```
tottoto, line 338: As the contents of this example are all hidden, they will not be displayed. And this example itself is redundant.

```suggestion
```

tottoto (changes requested):

tonic/src/transport/channel/endpoint.rs
```diff
@@ -37,6 +37,7 @@ pub struct Endpoint {
     pub(crate) connect_timeout: Option<Duration>,
     pub(crate) http2_adaptive_window: Option<bool>,
     pub(crate) executor: SharedExec,
+    pub(crate) local_address: Option<IpAddr>,
```
tottoto, line 40:
```suggestion
    pub(crate) local_address: Option<IpAddr>,
    pub(crate) executor: SharedExec,
```

tonic/src/transport/channel/endpoint.rs
jparris: Sure thing.

tottoto (approved):

PR #1570, opened 2023-11-15
chore: Use cargo-no-dev-deps to check msrv

tottoto:
## Motivation

Removes the effects from `dev-dependencies` to MSRV, and makes easier to specify public crates for checking.

## Solution

Uses `cargo-no-dev-deps` to check msrv.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -91,13 +91,14 @@ jobs:
     - uses: hecrj/setup-rust-action@v2
       with:
         rust-version: "1.70"    # msrv
+    - uses: taiki-e/install-action@cargo-no-dev-deps
     - name: Install protoc
       uses: taiki-e/install-action@v2
       with:
         tool: protoc@${{ env.PROTOC_VERSION }}
     - uses: Swatinem/rust-cache@v2
-    - run: cargo check --workspace --all-targets --all-features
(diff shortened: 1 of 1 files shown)


tottoto:
It can remove the effect to `Cargo.lock` from crates in `dev-dependencies`.

djc:
> It can remove the effect to `Cargo.lock` from crates in `dev-dependencies`.

Can you elaborate a little bit? Why is that important (how do we rely upon it here)?

tottoto:
It is important to test in an environment as close as possible to the environment in which it will be actually released.

Let's consider the following example.

A project depends on `crate-a` `0.1` as `dependencies` and `crate-b` as `dev-dependencies`, and `crate-b` depends on `crate-a` `=0.1.0` as dependencies. When `crate-a` has a version `0.1.1`, checking (with the affected `Cargo.lock` by `dev-dependencies`) is targeted to `crate-a` `0.1.0`, not `0.1.1` which is users actually faced. And more thinking the case if `crate-a` bumped the MSRV at `0.1.1` from `0.1.0`, the MSRV checking is broken. 

Although such an example may seem trivial, one of the important roles of testing is to prepare for such risks.

djc:
@LucioFranco any thoughts on the trade-off here? It's nice to avoid dev-deps for MSRV checks but it feels to me like the extra complexity of depending on an external tool is probably not worth it.

LucioFranco (approved):
@djc to me it seems fine to depend on a tool by the author of that one. We can always revert this back pretty easily later. I think this makes a lot of sense.

PR #1571, opened 2023-11-15
chore: Bump msrv to 1.66

tottoto:
`jobserver` 0.1.27, which `tonic` depends on as transitive dependency, requires Rust 1.66.


tottoto:
Taken over by #1587.

PR #1572, opened 2023-11-18
feat(transport): Make service router independent from transport

tottoto:
Allows users to use the service router independent from transport feature.

### DIFF
--- a/tonic/src/transport/service/mod.rs
+++ b/tonic/src/transport/service/mod.rs
@@ -6,7 +6,6 @@ pub(crate) mod executor;
 pub(crate) mod grpc_timeout;
 mod io;
 mod reconnect;
-mod router;
 #[cfg(feature = "tls")]
 mod tls;
 mod user_agent;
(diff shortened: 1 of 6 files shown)


tonic/src/transport/service/mod.rs
```diff
@@ -22,5 +21,5 @@ pub(crate) use self::io::ServerIo;
 pub(crate) use self::tls::{TlsAcceptor, TlsConnector};
 pub(crate) use self::user_agent::UserAgent;
 
-pub use self::router::Routes;
-pub use self::router::RoutesBuilder;
+/// A deprecated re-export. Please use `tonic::service::{Routes, RoutesBuilder}` directly.
+pub use crate::service::{Routes, RoutesBuilder};
```
djc, line 25: Should we put `#[deprecated]` on these deprecated re-exports? Why/why not?

tottoto: This is because of the limitation of `#[deprecated]`.

https://github.com/rust-lang/rust/issues/30827.

tottoto:
Added a change which makes `Routes::prepare` public.

djc (approved):
LGTM!

PR #1587, opened 2023-12-20
chore: Bump msrv to 1.70

tottoto:
`home` 0.5.9, which `tonic` depends on via `prost-build`, requires Rust 1.70.


LucioFranco (approved):
Thanks!

PR #1593, opened 2024-01-05
chore: Replace tokio::pin with std::pin::pin

tottoto:
[`std::pin::pin`](https://doc.rust-lang.org/std/pin/macro.pin.html) can be used now as `tonic`'s MSRV is updated to 1.70.


LucioFranco (approved):
Thanks!

PR #693, opened 2021-06-26
Add client feature flag to support wasm32 targets.

boxdot:
When protobuf is generated without the transport feature, the generated client can be compiled to wasm32 targets and used in browser environment.

To be able to run the client in browser, a `h2` version including the commit 5c72713 has to be used.

## Motivation

This enables tonic clients to compile to `wasm32-unknown-unknown` target and run in a Browser environment. For more information, please see #491.

## Solution

For this to work, we 

1. introduced a new feature flag, which only enables system independent parts of tonic transport,
2. added `spawn` function which either uses `tokio::spawn` or `wasm_bindgen_futures::spawn_local` dependening on the compilation target,
3. set `http2_max_concurrent_reset_streams` to 0 for wasm, to disable the `Instant::now` usage in h2 which is not available in Browser.

Issue #491, opened 2020-11-12
Enable transport over WebSocket (with support of wasm)


boxdot:
I have an experimental implementation of a `Connector` which allows to tunnel gRPC communication over a WebSocket. This makes it possible to implement a gRPC-over-WebSocket server and more importantly also a client. With small patches of h2 and tower-buffer the client also runs in wasm32.

* Implementation incl. examples: https://github.com/boxdot/tonic-ws-transport
* Patch of `tonic`: https://github.com/boxdot/tonic/commit/b06a97dc563eb41f3ff7f4399befd7d872dae090
* Patch of `h2`: https://github.com/boxdot/h2/commit/03838652b2b2a40c46df44a8299ebc1a9236939c
* Path of `tower-buffer`: https://github.com/boxdot/tower/commit/1091340a3eb106da844348bc1b1ef7ce7b4a763d

If you find this useful and my approach as suitable, I would like to discuss how to move forward from here.

### Crates

* tonic (transport module)
* h2
* tower-buffer

### Motivation

Run gRPC in browser.

### Proposal

(High level)

* Expose the client parts that are independent of TCP and work in wasm as a separate feature.
* Use this new feature and the new `tonic-ws-transport` crate to enable implementations of wasm clients.

I am not sure if this should become a separate feature of `tonic` crate though, or we could do it differently. Due to this bug in cargo (https://github.com/rust-lang/cargo/issues/2524), we might run into problems if we choose conditional compilation based on `wasm32` target arch.

### Notes:

* Implementation and patches are based on `tokio` 0.2 branch.
* https://github.com/rust-lang/cargo/issues/2524
* Examples are not part of the experimental crate due to another cargo bug: https://github.com/rust-lang/cargo/issues/1796

LucioFranco:
Sorry for the delayed response on this!

Is there a spec for gRPC over WS? Beyond gRPC web?

boxdot:
> Sorry for the delayed response on this!
> 
> Is there a spec for gRPC over WS? Beyond gRPC web?

I don't think so. The idea was just to use websocket as a tunnel for the TCP transport in environments where TCP is not available but WebSocket is (e.g. browsers). Since gRPC runs on top of HTTP2, and HTTP2 on top of TCP, and my crate allows TCP to run on top of a WebSocket, it is completely transparent to gRPC.

[...]

PR #1594, opened 2024-01-09
Add client feature flag to support connect_with_connector in wasm32 targets

lucasmerlin:
This is an updated version of @boxdot's PR #693. The main purpose is to allow compiling tonic for wasm32-unknown-unknown with the connect_with_connector function, so you could use custom connectors in wasm. 

## Motivation

The main motivation is to allow using something like https://github.com/boxdot/tonic-ws-transport to use tonic in browsers over websockets. grpc-web has a limit on concurrent conections that are open and doesn't allow for client streaming. There is also a related issue for this, #491.

## Solution

The connect_with_connector function, as well as the Endpoint and Channel structs, are gated behind the transport feature. Enabling this feature also enables things that are impossible to compile for wasm. This PR introduces a `client` feature that that enables everything necessary to make connect_with_connector work in wasm and disables anything that fails to compile for wasm. 

An alternative to adding a new `client` feature could also be to gate everything that doesn't compile for wasm behind a #[cfg()] attribute that is disabled for wasm. 

I could also add a test that checks whether compilation to wasm works with the client feature, so this won't break in the future.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -38,8 +38,19 @@ transport = [
   "channel",
   "dep:h2",
   "dep:hyper",
+  "hyper/full",
+  "dep:tokio",
   "tokio/net",
   "tokio/time",
+  "dep:tower",
+  "tower/balance",
+  "dep:hyper-timeout",
+]
+client = [
+  "dep:h2",
+  "hyper/client",
+  "hyper/http2",
+  "dep:tokio",
   "dep:tower",
   "dep:hyper-timeout",
 ]
@@ -55,7 +66,7 @@ bytes = "1.0"
 http = "0.2"
 tracing = "0.1"
 
-tokio = "1.0.1"
+tokio = { version = "1.0.1", default-features = false, optional = true }
 http-body = "0.4.4"
 percent-encoding = "2.1"
 pin-project = "1.0.11"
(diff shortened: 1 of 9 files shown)


flokli:
@lucasmerlin I'm also interested in this. Any chance you can address the conflicts?

lucasmerlin:
Hi, I gave it a shot and got the native version of [tonic-ws-transport working](https://github.com/lucasmerlin/tonic-ws-transport/tree/hyper-1.0) natively, but I'm struggling to get tonic to compile for wasm again. 
hyper-utils e.g. depends on tower/make which in turn depends on tokio features which aren't available on wasm. So I think there will need to be some patches to other crates to get this working again, which I unfortunately don't have the time for right now. 


If someone else wants to give it a try, these are the branches with my rebased changes: 
https://github.com/lucasmerlin/tonic/tree/wasm-client-feature-hyper-1.0
https://github.com/lucasmerlin/tonic-ws-transport/tree/hyper-1.0


That said, I'm still using tonic-ws-transport with tonic 0.10.x in my project and it's working really well for me.

flokli:
IIRC there's some work ongoing in hyper-utils to make some dependencies optional, though tower still seems to be required after https://github.com/hyperium/hyper-util/pull/105. cc @tottoto for visibility.

PR #1598, opened 2024-01-12
feat: Relax GrpcMethod lifetime

### LINKED ISSUE #1547: GrpcMethod's service and method should have "static" lifetime?
### Crates

extensions::GrpcMethod

### Motivation
I'm currently create dynamic client but if create GrpcMethod struct with `&'static str` then there is a memory leak occur. so I want to avoid this happen.

### Proposal
Actually, this is not proposal. it's a question.
In client side generated code, GrpcMethod created with `&'static str`, but is it necessary?
if it's necessary one then, can you explain it for me? and there is a way to avoid this? 
thank you in advance.
```
req.extensions_mut()
    .insert(GrpcMethod::new("helloworld.Greeter", "SayHelloBidiStream"));

----

#[derive(Debug, Clone)]
pub struct GrpcMethod {
    service: &'static str,
    method: &'static str,
}
```

tottoto:
Resolves #1547.

### DIFF
--- a/tonic/src/extensions.rs
+++ b/tonic/src/extensions.rs
@@ -1,14 +1,14 @@
 /// A gRPC Method info extension.
 #[derive(Debug, Clone)]
-pub struct GrpcMethod {
-    service: &'static str,
-    method: &'static str,
+pub struct GrpcMethod<'a> {
+    service: &'a str,
+    method: &'a str,
 }
 
-impl GrpcMethod {
+impl<'a> GrpcMethod<'a> {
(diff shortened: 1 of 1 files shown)


lemma105:
Why is this still open?

djc:
Seems reasonable. I guess adding a lifetime parameter is a semver-incompatible change, though...

LucioFranco:
This change makes sense, we need to bundle it with other breaking changes though.

djc (approved):

PR #1599, opened 2024-01-12
feat(build): Custom codecs for generated code

kvcache:
## Motivation

My company has a pubsub service powered in part by Tonic. It is designed to accept
hundreds of thousands of subscriptions per EC2 instance. Subscriptions normally
receive infrequent, very small messages - to the tune of 200 bytes. It is rare that a
message is over 1kb, and the service has a cap at 4kb.

I observed that memory is the limiting factor for reaching the required memory
density goals. I've tuned our structs and removed as many extra vtable references
as possible. I even gave a concrete stream type up to tonic in the subscription rpc
implementation to save 8 bytes per subscription. These optimizations, per
Amdahl's Law, can only save on the order of 2% memory (in my system) with
16 kilobytes allocated to encoding and decoding per stream.

I'm not the only person at my company who writes services, and our ecosystem that
touches protocol buffers is multi-language. Generated code helps to ensure
consistency in cross-language service vs client definition. For these reasons and
others, we strongly prefer to keep the tonic service traits as generated code. The
Tonic code generator is awesome!

I experimented with changing decode.rs's `const BUFFER_SIZE: usize = 8 * 1024;` to
smaller values, and it did the needful. This PR is an attempt to make it configurable.

## Solution

Broadly, this change does 2 things:
1. Allow the built-in Prost codec to have its buffer sizes customized
2. Allow users to specify custom codecs on the tonic_build::prost::Builder

The Prost codec is convenient, and handles any normal use case. However,
the buffer sizes today are too large in some cases - and they may grow too
aggressively. By exposing BufferSettings, users can make a small custom
codec with their own BufferSettings to control their memory usage - or give
enormous buffers to rpc's, as their use case requires.

While one can define a custom service and methods with a custom codec today
explicitly in Rust, the code generator does not have a means to supply a
custom codec. I've reached for .codec... on the tonic_build::prost::Builder
many times and keep forgetting it's not there. This change adds .codec_path
to the Builder, so people can simply add their custom buffer codec or even
their own full top level codec without reaching for manual service definition.

I've included a hello world example with a customized codec similar to that which
I'm currently compiling in my service.

By exposing ProstEncoder/ProstDecoder, BufferSettings, and a
tonic_build::prost::Builder::codec_path setting, I am able to meet the memory
density requirements for our small item, low throughput streaming rpcs. I am also
able to separate the buffer size settings between small message Services and large
message Services.

### DIFF
--- a/examples/src/json-codec/common.rs
+++ b/examples/src/json-codec/common.rs
@@ -30,6 +30,10 @@ impl<T: serde::Serialize> Encoder for JsonEncoder<T> {
     fn encode(&mut self, item: Self::Item, buf: &mut EncodeBuf<'_>) -> Result<(), Self::Error> {
         serde_json::to_writer(buf.writer(), &item).map_err(|e| Status::internal(e.to_string()))
     }
+
+    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
+        Default::default()
+    }
 }
 
 #[derive(Debug)]
@@ -47,6 +51,10 @@ impl<U: serde::de::DeserializeOwned> Decoder for JsonDecoder<U> {
         let item: Self::Item =
             serde_json::from_reader(buf.reader()).map_err(|e| Status::internal(e.to_string()))?;
         Ok(Some(item))
+    }
+
+    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
+        Default::default()
     }
 }
 
--- a/examples/src/json-codec/common.rs
+++ b/examples/src/json-codec/common.rs
@@ -30,10 +30,6 @@ impl<T: serde::Serialize> Encoder for JsonEncoder<T> {
     fn encode(&mut self, item: Self::Item, buf: &mut EncodeBuf<'_>) -> Result<(), Self::Error> {
         serde_json::to_writer(buf.writer(), &item).map_err(|e| Status::internal(e.to_string()))
     }
-
-    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
-        Default::default()
-    }
 }
 
 #[derive(Debug)]
@@ -51,10 +47,6 @@ impl<U: serde::de::DeserializeOwned> Decoder for JsonDecoder<U> {
         let item: Self::Item =
             serde_json::from_reader(buf.reader()).map_err(|e| Status::internal(e.to_string()))?;
         Ok(Some(item))
-    }
-
-    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
-        Default::default()
     }
 }
 
--- a/tonic-build/src/compile_settings.rs
+++ b/tonic-build/src/compile_settings.rs
@@ -1,0 +1,69 @@
+use std::{
+    marker::PhantomData,
+    mem::take,
+    sync::{Mutex, MutexGuard},
+};
+
+#[derive(Debug, Clone)]
+pub(crate) struct CompileSettings {
+    pub(crate) codec_path: String,
+}
+
+impl Default for CompileSettings {
+    fn default() -> Self {
+        Self {
+            codec_path: "tonic::codec::ProstCodec".to_string(),
+        }
+    }
+}
+
+thread_local! {
+    static COMPILE_SETTINGS: Mutex<Option<CompileSettings>> = Default::default();
+}
+
+/// Called before compile, this installs a CompileSettings in the current thread's
+/// context, so that live code generation can access the settings.
+/// The previous state is restored when you drop the SettingsGuard.
+pub(crate) fn set_context(new_settings: CompileSettings) -> SettingsGuard {
+    COMPILE_SETTINGS.with(|settings| {
+        let mut guard = settings
+            .lock()
+            .expect("threadlocal mutex should always succeed");
+        let old_settings = guard.clone();
+        *guard = Some(new_settings);
+        SettingsGuard {
+            previous_settings: old_settings,
+            _pd: PhantomData,
+        }
+    })
+}
+
+/// Access the current compile settings. This is populated only during
+/// code generation compile() or compile_with_config() time.
+pub(crate) fn load() -> CompileSettings {
+    COMPILE_SETTINGS.with(|settings| {
+        settings
+            .lock()
+            .expect("threadlocal mutex should always succeed")
+            .clone()
+            .unwrap_or_default()
+    })
+}
+
+type PhantomUnsend = PhantomData<MutexGuard<'static, ()>>;
+
+pub(crate) struct SettingsGuard {
+    previous_settings: Option<CompileSettings>,
+    _pd: PhantomUnsend,
+}
+
+impl Drop for SettingsGuard {
+    fn drop(&mut self) {
+        COMPILE_SETTINGS.with(|settings| {
+            let mut guard = settings
+                .lock()
+                .expect("threadlocal mutex should always succeed");
+            *guard = take(&mut self.previous_settings);
+        })
+    }
+}
--- a/tonic-build/src/compile_settings.rs
+++ b/tonic-build/src/compile_settings.rs
@@ -1,11 +1,13 @@
 #[derive(Debug, Clone)]
 pub(crate) struct CompileSettings {
+    #[cfg(feature = "prost")]
     pub(crate) codec_path: String,
 }
 
 impl Default for CompileSettings {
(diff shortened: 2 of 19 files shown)


kvcache:
Force-pushed commit to rebase on the tip of `master`. Force push was used because this hasn't been reviewed yet.

tonic/src/codec/mod.rs
```diff
@@ -21,6 +21,41 @@ pub use self::decode::Streaming;
...
+const DEFAULT_YIELD_THRESHOLD: usize = 32 * 1024;
+
+/// Settings for how tonic allocates and grows buffers.
+#[derive(Clone, Copy, Debug)]
+pub struct BufferSettings {
+    /// Initial buffer size, and the growth unit for cases where the size
+    /// is larger than the buffer's current capacity. Defaults to 8 KiB.
+    ///
+    /// Notably, this is eagerly allocated per streaming rpc.
+    pub buffer_size: usize,
```
LucioFranco, line 45: I would like to avoid exposing public struct fields like this and would rather have setters/getters for them or a new fn that accepts them. I would then move the field docs to the struct as well.

tonic/src/codec/mod.rs
```diff
@@ -63,6 +98,9 @@ pub trait Encoder {
 
     /// Encodes a message into the provided buffer.
     fn encode(&mut self, item: Self::Item, dst: &mut EncodeBuf<'_>) -> Result<(), Self::Error>;
+
+    /// Controls how tonic creates and expands encode buffers.
+    fn buffer_settings(&self) -> BufferSettings;
```
LucioFranco, line 103: to avoid another breaking change here, can we have this be a default fn that just calls `BufferSettings::default()`?

LucioFranco: and the same for the other trait.

LucioFranco (changes requested):
Overall looks really good left some feedback on how we can avoid some breaking changes but I think the idea is following down the right path, thanks!

tonic-build/src/lib.rs
```diff
@@ -137,7 +140,7 @@ pub trait Method {
     /// Identifier used to generate type name.
     fn identifier(&self) -> &str;
     /// Path to the codec.
-    fn codec_path(&self) -> &str;
+    fn codec_path(&self) -> String;
```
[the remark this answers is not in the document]
kvcache: Yeah, I was using a borrowed state (from load()) which only lives on the stack. That state can't outlive static, so I needed to return something owned. I'm looking at your other suggestion to newtype Service and add the configuration directly. Threadlocal is not my preference or remotely the first thing I tried ;-). Thanks for your feedback!

Probably if I get that resolved then the codec path will be able to be borrowed from the struct instead of the environment, which is the root of the ownership mismatch here.

tonic-build/src/prost.rs
```diff
@@ -102,8 +101,17 @@ impl crate::Method for Method {
...
+    ///
+    /// You should set the codec path to an import path that has a free
+    /// function like `fn default()`. The default value is tonic::codec::ProstCodec,
+    /// which returns a default-configured ProstCodec. You may wish to configure
+    /// the codec, e.g., with a buffer configuration.
+    ///
+    /// Though ProstCodec implements Default, it is currently only required that
+    /// the function match the Default trait's function spec.
+    fn codec_path(&self) -> String {
+        compile_settings::load().codec_path
```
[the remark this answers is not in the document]
kvcache: Yeah, this was a good thought. I've got it updated, and the next update will not have threadlocal. Let's all just pretend like I never pr'd a threadlocal 😅

tonic-build/src/compile_settings.rs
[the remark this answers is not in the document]
kvcache: Unneeded, will update.

tonic/src/codec/mod.rs
```diff
@@ -21,6 +21,41 @@ pub use self::decode::Streaming;
 #[cfg(feature = "prost")]
 #[cfg_attr(docsrs, doc(cfg(feature = "prost")))]
 pub use self::prost::ProstCodec;
+#[cfg(feature = "prost")]
+#[cfg_attr(docsrs, doc(cfg(feature = "prost")))]
+pub use self::prost::ProstDecoder;
+#[cfg(feature = "prost")]
+#[cfg_attr(docsrs, doc(cfg(feature = "prost")))]
+pub use self::prost::ProstEncoder;
```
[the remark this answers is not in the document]
kvcache: I dug into this a little deeper and yes, I got it worked out. I think this is strictly better, and I agree with you that keeping the public type surface area smaller is worth some friction on a very niche use case.

tonic/src/codec/mod.rs
kvcache: Of course, I'll adjust that. Thanks

tonic/src/codec/mod.rs
kvcache: That seems very sensible, thank you

LucioFranco:
CI is failing due to
```
  error: unused import: `compile_settings::CompileSettings`
     --> tonic-build/src/lib.rs:101:16
      |
  101 | pub(crate) use compile_settings::CompileSettings;
      |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      |
      = note: `-D unused-imports` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(unused_imports)]`
  
  error: field `codec_path` is never read
   --> tonic-build/src/compile_settings.rs:3:16
    |
  2 | pub(crate) struct CompileSettings {
    |                   --------------- field in this struct
  3 |     pub(crate) codec_path: String,
    |                ^^^^^^^^^^
    |
    = note: `CompileSettings` has derived impls for the traits `Clone` and `Debug`, but these are intentionally ignored during dead code analysis
    = note: `-D dead-code` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(dead_code)]`
```

tonic/benches/decode.rs
```diff
@@ -105,6 +105,10 @@ impl Decoder for MockDecoder {
         buf.advance(self.message_size);
         Ok(Some(out))
     }
+
+    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
+        Default::default()
+    }
```
LucioFranco, line 111: again here we can remove this because we have a default fn on the trait?

LucioFranco (changes requested):
Overall looks really good, I think there are a few warnings/CI fixes needed and I left a few comments. I think we can merge it once its all fixed.

tonic/src/codec/mod.rs
```diff
@@ -22,6 +22,75 @@ pub use self::decode::Streaming;
...
+
+/// Settings for how tonic allocates and grows buffers.
+///
+/// Tonic eagerly allocates the buffer_size per RPC, and grows
+/// the buffer by buffer_size increments to handle larger messages.
+/// Buffer size defaults to 8KiB.
+///
+/// Example:
+/// ```
+/// Buffer start:       | 8kb |
```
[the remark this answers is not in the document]
kvcache: Ah yes, thanks!

examples/src/json-codec/common.rs
```diff
@@ -30,6 +30,10 @@ impl<T: serde::Serialize> Encoder for JsonEncoder<T> {
     fn encode(&mut self, item: Self::Item, buf: &mut EncodeBuf<'_>) -> Result<(), Self::Error> {
         serde_json::to_writer(buf.writer(), &item).map_err(|e| Status::internal(e.to_string()))
     }
+
+    fn buffer_settings(&self) -> tonic::codec::BufferSettings {
+        Default::default()
+    }
```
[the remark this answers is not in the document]
kvcache: That's a great point. I'd love to have no change to this example so people don't have to revisit their custom codecs on upgrade. It's not necessary with the default implementation so away it goes.

tonic/benches/decode.rs
kvcache: yep!

kvcache:
I installed nightly, cargo-hack, and cargo-udeps to reproduce the workflow error locally. I addressed it by gating the codec path on the prost feature. If it's needed outside of that feature, I think it can be consumed and the feature gates removed.

LucioFranco (approved):
Thanks!

Issue #1600, opened 2024-01-12
Add an example client to the `reflection` example


grottohub:
### Crates

`examples`

### Motivation

As a Rust newbie, it took a little bit of time to grok how to set up the `ServerReflectionClient` that's generated in `tonic-reflection`, so having an example would be nice.

### Proposal

Add a simple example for using the client (I have one ready for a PR).

### Alternatives

I don't see a downside to having it documented, but definitely open to changing the example I came up with.

Ashu999:
@grottohub, is there any example available for how to first generate `ServerReflectionClient` using `tonic-reflection`. I'm unable to find one.

I'm using this example as base to implement add `tonic-reflection` on top of: https://github.com/hyperium/tonic/blob/master/examples/helloworld-tutorial.md

but unable to get it to work, here is what I have tried so far, steps mentioned in it's readme: https://github.com/Ashu999/helloworld-tonic
I have made changes in server side only and using `grpcurl ` as client to test the functionality of `tonic-reflection`.

would apricate any help, thanks.

PR #1601, opened 2024-01-12
examples: add client example for `tonic-reflection`

grottohub:
This adds an example usage of the `ServerReflectionClient` using a `ListServices` request.

Fixes: #1600

## Motivation

Took awhile to figure out how to set up the reflection client, so thought I would provide an example.

## Solution

This example includes everything needed to communicate with a server providing reflection:
- setting up the client with an `Endpoint` (since the codegen doesn't provide the `connect` method most clients have as `transport` is an optional feature for `tonic-reflection`)
- configuring a `ServerReflectionRequest` using one of the provided `MessageRequest` options
- putting the request into a stream since reflection is a bidi-streaming procedure
- parsing the response received from the server

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -165,6 +165,11 @@ path = "src/health/server.rs"
 required-features = ["health"]
 
 [[bin]]
+name = "reflection-client"
+path = "src/reflection/client.rs"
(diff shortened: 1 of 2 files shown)


LucioFranco:

PR #1605, opened 2024-01-16
tonic-reflection: feature-gate server impl

A-Manning:
## Motivation

`tonic-reflection` cannot be used in certain environments eg. wasm, due to a dependency on the `tokio` runtime.

The `tokio` runtime is only used in `tonic_reflection::server`, and so it makes sense to feature-gate this module and make it a default feature, for backwards compatibility.

This PR makes it possible to use the `tonic_reflection::pb` module without having to build `tokio`.
This is useful for eg. wasm reflection clients.

### DIFF
--- a/tonic-reflection/Cargo.toml
+++ b/tonic-reflection/Cargo.toml
@@ -2,6 +2,7 @@
 authors = [
   "James Nugent <james@jen20.com>",
   "Samani G. Gikandi <samani@gojulas.com>",
+  "Ash Manning <10554686+A-Manning@users.noreply.github.com>"
 ]
 categories = ["network-programming", "asynchronous"]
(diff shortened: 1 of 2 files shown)


tonic-reflection/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 authors = [
   "James Nugent <james@jen20.com>",
   "Samani G. Gikandi <samani@gojulas.com>",
+  "Ash Manning <10554686+A-Manning@users.noreply.github.com>"
```
LucioFranco, line 5:
```suggestion
```

LucioFranco (approved):
This seems fine to me thanks!

Issue #1611, opened 2024-01-30
Update UDS example with socket paths determined at runtime


kriswuollett:
### Motivation

Workaround and Rust education for closure capture requirement that currently only a static path is used:

https://github.com/hyperium/tonic/blob/c30cb783b9e5eaa4ea05003d9cacac733d135c58/examples/src/uds/client.rs#L16-L28

### Proposal

Use the `ustr` crate to remove reliance on only a static hard-coded path.

Issue #1612, opened 2024-01-30
Support non-static UDS client connection paths


kriswuollett:
## Feature Request

Support non-static UDS client connection paths

### Motivation

Currently the UDS client example is using a static path, I believe because of the ownership requirements of the service function closure:

https://github.com/hyperium/tonic/blob/c30cb783b9e5eaa4ea05003d9cacac733d135c58/examples/src/uds/client.rs#L16-L28

It would be great if `tonic` and its `hyperium` dependencies, as needed, took ownership of the specified path to avoid the user needing to vend a `&'static str`. 

### Related

- Captured path perhaps eventually should be supported through a URI that supports file URLs: https://github.com/hyperium/http/issues/323
- Proposal to update UDS client example with a workaround: #1611

philjb:
I second this and am currently using `Box::leak()` to work around it.

youyuanwu:
> I second this ~and am currently using `Box::leak()` to work around it.~ Some of my rust teammates helped me out: An `Arc` or a cloned `PathBuf` should also meet the 'static requirements.
> 
> like
> 
> ```rust
>     let path = PathBuf::from("/tmp/tonic/helloworld");
> 
>     let channel = Endpoint::try_from("http://[::]:50051")?
>         .connect_with_connector(service_fn(move |_: Uri| {
>             // Connect to a Uds socket
>             UnixStream::connect(path.clone())
>         }))
>         .await?;
> ```

This no longer works on 0.12 version after this change: https://github.com/hyperium/tonic/pull/1670
```rs
    let path = PathBuf::from("/tmp/my.sock");

    let channel = Endpoint::try_from("http://[::]:50051")?
        .connect_with_connector(service_fn(move |_| async {
            // Connect to a Uds socket
            let io = TokioIo::new(UnixStream::connect(path.clone()).await?);
            Ok::<_, std::io::Error>(io)
        }))
        .await?;
```
gives error:
```
   |
25 |           .connect_with_connector(service_fn(move |_| async {
   |  ____________________________________________--------_^
   | |                                            |      |
   | |                                            |      return type of closure `{async block@src/walfs/src/bin/walfsctl.rs:25:53: 29:10}` contains a lifetime `'2`
   | |                                            lifetime `'1` represents this closure's body
26 | |             // Connect to a Uds socket
27 | |             let io = TokioIo::new(UnixStream::connect(path.clone()).await?);
28 | |             Ok::<_, std::io::Error>(io)
29 | |         }))
[...]

PR #1613, opened 2024-01-30
examples: allow configurable uds socket paths

kriswuollett:
Update unix domain socket examples to show configurable paths. It demonstrates a way of configuring the UDS client using the `ustr` crate to meet the closure capture requirements of the connector service_fn.

Fixes: #1611
Refs: #1612

## Motivation

Share a solution of how to configure a UDS client path at runtime.

## Solution

Use the `ustr` crate to vend a `&'static str` to the UDS connection building closure.

### DIFF
--- a/examples/src/uds/client.rs
+++ b/examples/src/uds/client.rs
@@ -18,7 +18,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     // Path needs to be a static str due to service_fn FnMut.
     let path = match env::var("EXAMPLE_UDS_PATH") {
         Ok(path) => ustr(path.as_ref()).as_str(),
-        Err(_) => "/tmp/tonic/helloworld"
+        Err(_) => "/tmp/tonic/helloworld",
     };
     println!("Using path {}", path);
 


examples/src/uds/client.rs
```diff
@@ -5,21 +5,28 @@ pub mod hello_world {
...
 use tonic::transport::{Endpoint, Uri};
 use tower::service_fn;
+use ustr::ustr;
 
 #[cfg(unix)]
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Path needs to be a static str due to service_fn FnMut.
+    let path = match env::var("EXAMPLE_UDS_PATH") {
+        Ok(path) => ustr(path.as_ref()).as_str(),
```
LucioFranco, line 20: I want to say converting a path to a string should be possible without a third party crate?

kriswuollett: Would something like a `OnceLock<String>` work? AFAICT a `&'static str` is required. I used `ustr` because in an actual application one may have multiple copies of the same client but with different endpoints -- I think you'd end up reimplementing `ustr` in that case?

kriswuollett: Perhaps there is something obvious I missed, but when using a `String`, and for the reason I filed that other linked issue...

```
expected a closure that implements the `FnMut` trait, but this closure only implements `FnOnce`
required for `ServiceFn<{closure@examples/src/uds/client.rs:29:44: 29:57}>` to implement `Service<Uri>`
required for `ServiceFn<{closure@examples/src/uds/client.rs:29:44: 29:57}>` to implement `MakeConnection<Uri>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B3%5D?3#file%3A%2F%2F%2FUsers%2Fkris%2FCode%2Fkriswuollett%2Ftonic%2Fexamples%2Fsrc%2Fuds%2Fclient.rs)
client.rs(29, 10): the requirement to implement `FnMut` derives from here
client.rs(31, 33): closure is `FnOnce` because it moves the variable `path` out of its environment
```

PR #1619, opened 2024-02-01
Add connection timeout to connect_with_connector_lazy

Benjscho:
Currently connect_with_connector_lazy doesn't respect connection timeouts. This means that if you have a misbehaving or unresponsive server, a request that's connecting lazily with a custom connector can hang for the client without timing out. This commit adds the timeout to the custom connector if it has been set on the endpoint, bringing the behaviour in line with other connect methods.

## Motivation

This brings the behaviour of `connect_with_connector_lazy` in line with the other connect methods that respect the connection time set on the Endpoint.

This solves instances where users set a connection time out on an endpoint, but use a custom connector to connect lazily. If the server fails to respond to the connection request at all then the connection attempt (and request) will hang without ever failing. 

## Solution

The solution is to use the same `hyper_timeout` connector to add a timeout.

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -390,7 +390,13 @@ impl Endpoint {
         crate::Error: From<C::Error> + Send + 'static,
     {
         let connector = self.connector(connector);
-        Channel::new(connector, self.clone())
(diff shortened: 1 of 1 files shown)


LucioFranco (approved):
Thanks!

PR #1622, opened 2024-02-03
Update Cargo.toml

Dylan-DPC:
Updating h2 to 0.3.24 as per [advisory](https://rustsec.org/advisories/RUSTSEC-2024-0003.html). I would also consider upgrading to 0.4 at some point


dswij (approved):

LucioFranco:
I believe 0.4 would require finishing our upgrade to hyper 1.0 which will take abit.

PR #1630, opened 2024-02-09
feat(transport): Make transport server and channel independent

tottoto:
## Motivation

Allows users to use transport server and channel features independently.

## Solution

> [!warning]
> This proposal includes a breaking change.

Depends on #1574. Adds a new feature `server` to enable server implementation in transport module and makes server feature and channel feature independent.


tonic/src/transport/mod.rs
```diff
@@ -129,7 +129,3 @@ pub use self::server::ServerTlsConfig;
 #[cfg(feature = "tls")]
 #[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
 pub use self::tls::Identity;
-
-#[cfg(feature = "channel")]
```
djc, line 129: I don't understand how your commit message describes this change? It seems to remove this instance and makes another instance private, but this commit doesn't "move" anything in the sense of creating something new.

djc (approved):
Very nice!

tonic/src/transport/mod.rs
tottoto: The definition of BoxFuture used in the implementation of the channel feature is moved to the channel module.

djc:
Ah, sorry, I just missed it for some reason.

PR #1644, opened 2024-02-28
Expose Extensions::from_http to user

DoumanAsh:
This exposes `Extensions::from_http` to allow user to write more efficient tower's Service instead of using Interceptor as it doesn't allow to customize Response


tottoto:
As `http::Extensions` has already been an external type, it sounds good to exposing this API.

tottoto (approved):

LucioFranco (approved):

PR #1648, opened 2024-03-03
Add a `GRPC_CONTENT_TYPE` for "application/grpc".

gibbz00:
Currently defined in `tonic::metadata`. Would need guidance from maintainers for where it might be better located.

Motivation is that I'm currently implementing my own tower-service::Service for handling gRPC calls, and it would be nice if tonic could provide a constant for the media type. (Writing my own because I need to combine multiple gRPC services using generics.)

### DIFF
--- a/tonic/src/metadata/mod.rs
+++ b/tonic/src/metadata/mod.rs
@@ -33,6 +33,9 @@ pub use self::value::MetadataValue;
 
 pub(crate) use self::map::GRPC_TIMEOUT_HEADER;
 
+/// HTTP Header `content-type` value for gRPC calls.
+pub const GRPC_CONTENT_TYPE: &str = "application/grpc";
+
 /// The metadata::errors module contains types for errors that can occur
 /// while handling gRPC custom metadata.
 pub mod errors {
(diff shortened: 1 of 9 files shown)


tottoto:
I think this is good improvement to me.

tottoto:
Looks good to me.

tottoto:
I would think locating this constant at `tonic::metadata` is reasonable.

tonic/src/metadata/mod.rs
```diff
@@ -33,6 +33,9 @@ pub use self::value::MetadataValue;
 
 pub(crate) use self::map::GRPC_TIMEOUT_HEADER;
 
+/// HTTP Header `content-type` value for gRPC calls.
+pub const GRPC_CONTENT_TYPE: &str = "application/grpc";
```
djc, line 37: Could we have this be of type `HeaderValue`?

djc:
(Also, this needs a rebase.)

tonic/src/metadata/mod.rs
ikrivosheev: @djc hello! I fixed it here: https://github.com/hyperium/tonic/pull/1648

gibbz00:
Awesome, thanks for picking this up :)

PR #1658, opened 2024-03-14
add encode/decode explicit logging

jasonwbarnett:
## Motivation

I ran into an error when trying to debug [pants][1] and when looking up the error in tonic's source I found two places with identical error messages which made it more difficult to identify the true error source.

## Solution

Add explicit details to the error message that will help someone identify the cause.

[1]: https://github.com/pantsbuild/pants


tottoto:
Sounds reasonable to me.

tottoto (approved):

djc (approved):

PR #1595, opened 2024-01-10
Update to hyper 1.0 and axum 0.7

ikrivosheev:
Hello! This is continue for: https://github.com/hyperium/tonic/pull/1583

Also I merged:

- https://github.com/allan2/tonic/pull/1
- https://github.com/allan2/tonic/pull/2

TODO:

- [x] update imports to use -util crates
- [x] update AddrStream and AddrIncoming
- [ ] use axum Request and Response in transport (in-progress)
- [ ] add poll_frame to impl http_body::Body (in-progress)
- [ ] remove poll_trailers from impl http_body::Body (in-progress)
- [x] fix hyper-timeout connector
- [ ] fix connect, Connect (no more hyper::client::service::Connect)
- [ ] update version, docs
- [ ] update examples
- [ ] many more (tonic-web, etc. Some of these will be in separate PRs)

Depends on:

- [x] https://github.com/hyperium/hyper/issues/3461

Closed: https://github.com/hyperium/tonic/issues/1579

Issue #102, opened 2019-10-29
tonic-examples build crashes: "No such file or directory"


curiousleo:
## Bug Report

Building the examples in `tonic-examples` in this repository crashes.

### Version

I just checked out this repository, so it's just `master` of `tonic`.

```console
$ git rev-parse HEAD
af5754bd437ffbb5c7a9fbe36af14cd182d48c53
```

### Platform

```console
$ uname -a
Linux think5 4.19.80 #1-NixOS SMP Thu Oct 17 20:45:44 UTC 2019 x86_64 GNU/Linux
$ rustc --version
rustc 1.39.0-beta.7 (23f8f652b 2019-10-26)
$ cargo --version
cargo 1.39.0-beta (1c6ec66d5 2019-09-30)
```

### Description

Following [tonic-examples/README.md](https://github.com/hyperium/tonic/blob/af5754bd437ffbb5c7a9fbe36af14cd182d48c53/tonic-examples/README.md#client), I tried to run `helloworld-client`.

This resulted in the following build error:

```console
$ cargo run --bin helloworld-client
[...]
   Compiling tonic-examples v0.1.0 (/home/leo/Code/other/tonic/tonic-examples)
error: failed to run custom build command for `tonic-examples v0.1.0 (/home/leo/Code/other/tonic/tonic-examples)`

Caused by:
  process didn't exit successfully: `/home/leo/Code/other/tonic/target/debug/build/tonic-examples-c5600591329ec7f8/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:1165:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.37/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:76
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:60
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1030
[...]

PR #1589, opened 2023-12-22
chore: Update to zstd 0.13

tottoto:
Updates to `zstd` 0.13.

PR #1670, opened 2024-03-31
Upgrade to Hyper 1.0 & Axum 0.7

alexrudy:
**Current Status**: This is ready for CI & review!

This is a continuation of PR #1595 by @ikrivosheev.

TODO:

- [X] update imports to use -util crates (done in #1595)
- [X] update AddrStream and AddrIncoming (done in #1595)
- [X] fix hyper-timeout connector (done in #1595)
- [X] use axum Request and Response in transport (via TowerToHyperService in Client)
- [X] change poll_trailers to use `poll_frame`. Still some work to do here about preserving data and trailers if `poll_frame` returns a frame type we aren't expecting.
- [X] Correctly handle data & trailer frames for server.
- [X] fix connect, Connect (no more hyper::client::service::Connect) (using `hyper-util`)
- [X] Get around ["Implementation of From not general enough"](https://github.com/rust-lang/rust/issues/102211) compiler issue.
- [X] Correctly implement graceful shutdown (maybe based on `hyper-util`, but also not too hard to do independently, following `axum`'s example.)
- [X] Pass through `max_pending_accept_reset_streams` (needs a release with PR [#102](https://github.com/hyperium/hyper-util/pull/102) in `hyper-util`)
- [X] Correctly implement `http2_only` (needs PR [#111](https://github.com/hyperium/hyper-util/pull/111) in `hyper-util`.
- [ ] Figure out what to do about hyper_util `ConnectError`
- [X] Add `socket2` dependency to support TCP `SO_NODELAY` and `SO_KEEPALIVE`
- [X] Compatibility of other tonic-* crates (e.g. tonic-web)
- [X] Examples
- [X] Check over Docs
- [ ] Version Update


alexrudy:
I'd like to propose that this be reviewed as-is, and that we add follow-up issues for the remaining todos, which will all require PRs to land and then releases to happen for `hyper` and/or `hyper-utils`.

karuna:
works perfectly after `cargo update`

examples/Cargo.toml
```diff
@@ -332,19 +384,24 @@ rand = { version = "0.8", optional = true }
...
+# warp = { version = "0.3.4", default-features = false, optional = true }
 listenfd = { version = "1.0", optional = true }
 bytes = { version = "1", optional = true }
 h2 = { version = "0.3", optional = true }
-tokio-rustls = { version = "0.24.0", optional = true }
-hyper-rustls = { version = "0.24.0", features = ["http2"], optional = true }
-rustls-pemfile = { version = "1", optional = true }
-tower-http = { version = "0.4", optional = true }
+tokio-rustls = { version = "0.26.0", optional = true }
+hyper-rustls = { version = "0.27.0", features = ["http2"], optional = true }
```
aumetra, line 402: It would be nice if there were feature flags to switch between the crypto backends.  
Defaulting to `aws-lc-rs` pulls in a cmake, nasm requirement, impacts platform support, etc.  
Trading that off for FIPS compliance isn't really needed in most cases.

aumetra: Which crypto provider `tonic` should default to is up to the maintainers, I guess.  
But `reqwest` seems to be going the route of sticking to `ring` as the default crypto backend.

alexrudy: I think that is a good idea, but I don't think it should be included in this PR - this PR is already complicated as is. We can create an issue and work on multi-backend support separately once this lands.

aumetra: Yeah, definitely. I just felt like the comment fit best here for the time being since the PR isn't merged yet and I probably would have forgotten about it if I hadn't written it down

tisonkun:
@alexrudy FYI https://github.com/hyperium/hyper-util/pull/102 is merged now.

hubertshelley:
called `Result::unwrap()` on an `Err` value: hyper::Error(User(ManualUpgrade))

alexrudy:
> @alexrudy Example h2c failed.

Thanks! https://github.com/hyperium/tonic/pull/1670/commits/5a579e63cadbe83c9dbe903472490c1d2397c904 contains the fix.

djc:
@alexrudy and @ikrivosheev thanks for all your work on this!

I'm going to be helping @LucioFranco out with tonic reviews, and obviously this is a desirable change. As such, I've made an initial pass of the changes here. Here are some thoughts:

- Since this is already such a large change, I'd like to avoid intertwining the changes that are necessary to upgrade hyper/axum with any other changes that can be done independently. For example, reformatting of `Cargo.toml` files, upgrading to a different nightly, reordering imports and other small changes like adding type annotations. To facilitate review I think all of those changes should either be reverted in this PR or submitted as part of separate PRs (with one commit or PR per logical change) where we can judge them properly -- the goal is to make this PR as small and focused as possible.
- I think it might make sense for now to just delete any of the examples that rely on warp, probably in a separate PR? While warp was one of the newer options when tonic got started, today it's much less popular than Axum -- and by virtue of doing this in a separate commit/PR it should be relatively straightforward to bring back those examples once warp upgrades to hyper/http 1.
- I would like to avoid further entangling of Axum and Tonic proper. Ideally this change should be neutral in how much of Axum is used in -- if that's hard for some reason let's discuss that separately? Especially for client-only users, we should be working in a direction where it's possible to avoid depending on Axum.

LucioFranco (commented):
I am still working through the PR, I think what @djc makes a lot of sense, I am going to continue reviewing the core web and tonic parts since it seems like those changes are somewhat non trivial. So it will take some time just as an FYI.

tisonkun:
@alexrudy FYI hyper-util v0.1.4 should contain the API you need abovex

ikrivosheev:
@alexrudy as I can see, hyper-util v0.1.4 is released. Can you upgrade the PR?

https://crates.io/crates/hyper-util

djc:
> @djc Thanks so much for your initial comments.
> 
> Overall, I tried to minimally scope the changes here based on what was required to make the tests pass with `hyper` >= 1.
> 
> For instance:
> 
>     * The change in the nightly compiler version from `2024-02-06` to `2024-04-16` is required to make Axum >= 0.7 compile on nightly (it otherwise requires `#![feature(diagnostic_namespace)]` to compile, and we can't change the nightly features requested by a dependency IIUC.

Makes sense. So can we bump this independently from the Axum upgrade or is this required to be "exactly" in sync? IMO this is one of these things that seems conceptually separate enough that it would be nice to deal with in a separate PR if possible.

>     * There are type annotations added where type inference was no longer sufficient to allow code to compile - e.g. in the h2c/server.rs example, due to the move from calling hyper's server directly to calling `TcpListener:bind()`, type inference is no longer able to resolve the `FromStr::Err` type (which will get transformed to `Box<dyn Error>`, and in `tls_rustls` examples, types have to be added because rustls and rustls_pemfile have changed their public API types, which in turn are required upgrades (I think) for compatibility with `hyper-rustls` for `hyper` 1.0.

That's fair! I'll point out detailed examples on my next, more detailed round of review.

> I will fix the `cargo.toml` formatting (this came from my IDE's auto formatter, and I should have caught that before).

👍 

> For `warp` – I'm happy to delete those examples entirely, either in this PR or another one - deleting them is a pre-requisite to the CI tests passing.

Yes, let's take care of this in a separate PR.

> This also brings up @LucioFranco's point about the changed error type, which we might want to leave failing: that sounds fine by me, I don't know of a good way though to leave a test in an expected failure state in Rust. Also, just letting the test fail will be problematic, as CI is currently configured to fail-fast, meaning that one failure will cause a lot of other tests to be skipped entirely. We could change that setting as well, if that is what is desired.
> 
> On Axum: I don't think that this PR increases the degree to which the two libraries are intertwined - there is some private adaptor code added, but the `tonic::transport::service::router` part has always been built on the `axum::Router` class - are you suggesting that we should be implementing our own routing? That feels like a separate endeavor to me. Or was there somewhere else that this PR ends up intertwining the two further? From my perspective, I thought it was pretty neutral about how much of Axum is required by tonic.

Sounds good -- I'll take a look in more detail on my next read through.

> Finally, it seems like you both have a preference for a re-written history here to break commits into more logical chunks - I'll take a stab at that as well, I had defaulted to preserving history to start, but am fine with either strategy.

I think this repo has a squash merge strategy so doing one logical change per PR might be better. In other projects I help maintain I usually prefer to rebase and keep a clean history of atomic commits that deal with one conceptual change each -- exactly because in my opinion this facilitates review which is usually in shortest supply in open source projects. However, this is just fundamentally a large change (and I'm not advocating we split it up by doing piecemeal changes that try to paper over http 0.2/1.0 differences somehow).

> Overall, @djc, I would love specific comments on parts of this PR that are really separable - or the decision that we'd rather merge a series of PRs, some of which contain the pre-requistites to the upgrade to hyper-1.0, and the final one that contains the final upgrade.

alexrudy:
@djc (just about Axum & Nightly)

> Makes sense. So can we bump this independently from the Axum upgrade or is this required to be "exactly" in sync? IMO this is one of these things that seems conceptually separate enough that it would be nice to deal with in a separate PR if possible.

We can bump the nightly version separately, thought it is then a one line change. I'll put up a PR for that - it is just a change which is (IMO) purely motivated by finding the minimum nightly where we can compile Axum.

alexrudy:
@djc [PR to remove warp examples](https://github.com/hyperium/tonic/pull/1708)

alexrudy:
Nightly bump in a separate PR: https://github.com/hyperium/tonic/pull/1709

djc:
The separate PRs have been merged -- will go through this again once it has been rebased.

alexrudy:
I've both rebased this and tried to split the commits into roughly atomic parts - hopefully that makes it easier to review. Because of the large surface of api changes in hyper and http, I'm not sure if any of these remaining commits work as separate PRs since they all require the hyper and http upgrades.

tests/integration_tests/tests/connection.rs
```diff
@@ -52,7 +52,10 @@ async fn connect_returns_err_via_call_after_connected() {
     let res = client.unary_call(Request::new(Input {})).await;
 
     let err = res.unwrap_err();
-    assert_eq!(err.code(), Code::Unavailable);
+    assert_eq!(err.code(), Code::Unknown);
```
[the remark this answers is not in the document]
alexrudy: I took a deeper dive here - `hyper` stopped exposing `is_connect()` since the connection tooling was moved to `hyper_util`. However, `tonic` wraps connection logic anyways, so I stuck in an (internal) wrapper type to identify connection errors and set the correct status.

tottoto:
After a quick look, it appears that there are some unrelated changes included. Although I have not checked everything, here are some of them:

- The version style in `Cargo.toml` manifest such as `0.26` -> `0.26.0`. If you want to propose these kind of changes, they should be issued as a separated pull request (but it is not the maintainer's preffered style: https://github.com/hyperium/tonic/pull/1589#discussion_r1608126077)
- There are some unnecessary version updating which seems not to be unnecessary:  `2.0` -> `2.1.1`, and so on.
- The kind of the changes like dropping `logging` feature of `tokio-rustls` is not related to this migration.
- The rebasing resolution seems to be broken, seeing the tls implementation.

For very large changes of this kind, it is necessary and important to reduce the scope of interest of the changes required in a single review as much as possible.

The impact of updating dependencies must also be considered in conjunction with the existing design. For example, dealing with tonic's `Extensions`, which is one of the effects of the breaking changes to the `http` crate, should not be handled simply, but should be handled by considering the necessity of the type in the first place. By the change of #1710 first, the responsibility for changing the update itself will be slightly reduced.

Or other example, since hyper no longer provides a Body type, I think it would be a good to have tonic's own Body type instead of `BoxBody`, like Axum does, in order to make the handling of the Body type in tonic more robust and ease to maintain, and it is possible to be addressed before updating hyper version. After that adopting it to hyper 1 can be addressed (Of course, whether this is actually possible requires separate consideration).

Basically, large changes like this should be broken up into smaller scope changes. For example, the scale of transport implementations is large, making it difficult to update them incrementally and therefore difficult to review.

The purpose of the following pull request is to improve both the user's convenience and the maintainability of this project itself, and I believe that these should be reflected first before migrating to `hyper` 1.

- #1572 
- #1574
- #1630

alexrudy:
Hi @tottoto,

Thanks for the comments!

This is a particularly painful kind of review for me, since I'm not sure that it consistently moves this PR towards being ready to merge, so I'll address your comments one-by-one.

> The version style in Cargo.toml manifest such as 0.26 -> 0.26.0. If you want to propose these kind of changes, they should be issued as a separated pull request (but it is not the maintainer's preffered style: https://github.com/hyperium/tonic/pull/1589#discussion_r1608126077)

For the sake of getting this to merge, I'll revert that part of the change, despite [the `.0` being meaningful](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) - I'm not sure what @djc meant by that comment originally?

> There are some unnecessary version updating which seems not to be unnecessary:  2.0 -> 2.1.1, and so on.

I've reverted this particular version change. I'm not sure where the "and so on." comes from though - I have really tried to whittle down the dependencies. It is true that a lot of the `tls-specific` libraries get bumped, they are bumped to maintain compatibility with `hyper-rustls` (used in the examples) which requires the minor version bump.

> The kind of the changes like dropping logging feature of tokio-rustls is not related to this migration.

Happy to put this back in. I'm not sure what function it is providing right now, since it isn't used by `tonic`.

> The rebasing resolution seems to be broken, seeing the tls implementation.

I'm not sure what you mean by this - can you point to something specific for me?

Overall, I'm a little frustrated to keep coming back to this PR to find that folks are annoyed that it is so _large_ - it is of course large, the scope of API changes between hyper 0.14 and hyper 1 is pretty huge. I also think that the desire to keep this PR as small as possible is pretty heavily in conflict with the desire to "consider dependencies" re. things like #1710, (which I'm fine with, but @djc mentioned there that they'd rather wait to land that until this lands).

The same applies to a Body type for tonic - that feels like a decent-sized API change which deserves its own PR, not something to be wedged into this giant morass. 

> Basically, large changes like this should be broken up into smaller scope changes. For example, the scale of transport implementations is large, making it difficult to update them incrementally and therefore difficult to review.

This really makes absolutely no sense to me - the transport changes have to land in-sync with the other changes because they are all required by the upgrade to hyper 1. Doing them separately would not make much sense. Alternatively, we could try to introduce some of the changes required for this PR in `transport` ahead of time, but that is rife with potential problems, since the API changes between hyper 0.14 and hyper 1 are so large, that writing the same code for hyper 0.14 is considerable additional effort which doesn't (at least to me, contributing on my free time) feel like it will pay off.

It also seems that you would prefer your list of PRs to be merged before this one:

> The purpose of the following pull request is to improve both the user's convenience and the maintainability of this project itself, and I believe that these should be reflected first before migrating to hyper 1.

That's fine, but perhaps not a useful contribution here. In this thread, I think I've demonstrated that I'm very willing and responsive to rebasing this (massive) PR on top of additional changes to `tonic`. If you have a different view of how the project prioritizes reviews, maybe we could discuss that on Discord?

I think I feel a particular frustration here because you are the 3rd person to "take a quick look" and provide review comments which are difficult to take significant actions on. As well, although I'm probably mis-reading your comment, it comes off to me as pretty condescending to explain, yet again, that changes should be broken into small parts to make it easier on reviewers. I think that @djc and I have been having that conversation back-and-forth considering the tradeoffs up thread. Comments like "basically,  large changes like this should be broken up into smaller scope changes" read like you are trying to educate someone with limited experience in the software world. I suspect you mean well, but in this case, I'd really encourage you to assume that the folks on the other end of your messages are knowledgable and competent developers, especially when they've clearly put a lot of time and effort into what they are building.

Again, thanks for the comments - I'm really looking forward to getting something like this merged, since it does also greatly impact my own user experience with tonic (since it prevents me from upgrading numerous other libraries which depend on hyper 1 now), which is a large part of my own motivation here.

djc:
> This is a particularly painful kind of review for me, since I'm not sure that it consistently moves this PR towards being ready to merge, so I'll address your comments one-by-one.
> 
> > The version style in Cargo.toml manifest such as 0.26 -> 0.26.0. If you want to propose these kind of changes, they should be issued as a separated pull request (but it is not the maintainer's preffered style: [#1589 (comment)](https://github.com/hyperium/tonic/pull/1589#discussion_r1608126077))
> 
> For the sake of getting this to merge, I'll revert that part of the change, despite [the `.0` being meaningful](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) - I'm not sure what @djc meant by that comment originally?

I'm not sure exactly what part of the Cargo documentation leads you to believe that the trailing zero in `0.26.0` is meaningful, but I'm pretty sure it's not in practice. Both `0.26.0` and `0.26` mean the same thing, namely that the accepted version range is `>= 0.26.0, < 0.27.0`. This might be different for the case where the leading number is `0` because Cargo treats `0.x` somewhat specially (in the sense that `0.1.x` and `0.2.x` are treated as semver-incompatible but `1.1.x` and `1.2.x` are treated as compatible).

> > There are some unnecessary version updating which seems not to be unnecessary:  2.0 -> 2.1.1, and so on.
> 
> I've reverted this particular version change. I'm not sure where the "and so on." comes from though - I have really tried to whittle down the dependencies. It is true that a lot of the `tls-specific` libraries get bumped, they are bumped to maintain compatibility with `hyper-rustls` (used in the examples) which requires the minor version bump.

The abundance of examples is unfortunate in my view -- IMO examples have diminishing returns as the number grows, and given that the maintenance efforts of examples is non-zero, increasing numbers of examples over time can have a net-negative value to the project (in particular for examples that drag in dependencies beyond what the libraries need anyway). So I think it would be good to have a discussion on what examples we ought to keep which might reduce the changes needed in this PR, but I also don't necessarily want to block this PR on those changes.

> > The kind of the changes like dropping logging feature of tokio-rustls is not related to this migration.
> 
> Happy to put this back in. I'm not sure what function it is providing right now, since it isn't used by `tonic`.

The `logging` feature is forwarded to rustls in the end, so it provides a feature that downstream users might belong on. I think defaulting to enabled makes sense here, but primarily I agree that this PR should not change whether we enable it or not because the change is unrelated to the change this PR seeks to make.

> > The rebasing resolution seems to be broken, seeing the tls implementation.
> 
> I'm not sure what you mean by this - can you point to something specific for me?
> 
> Overall, I'm a little frustrated to keep coming back to this PR to find that folks are annoyed that it is so _large_ - it is of course large, the scope of API changes between hyper 0.14 and hyper 1 is pretty huge. I also think that the desire to keep this PR as small as possible is pretty heavily in conflict with the desire to "consider dependencies" re. things like #1710, (which I'm fine with, but @djc mentioned there that they'd rather wait to land that until this lands).
> 
> The same applies to a Body type for tonic - that feels like a decent-sized API change which deserves its own PR, not something to be wedged into this giant morass.

Agreed that these changes should go in separate PRs. I think the main question is whether they should be merged before this PR or other -- though at the moment I'm inclined to agree that it probably doesn't make sense to merge them before this PR.

> > Basically, large changes like this should be broken up into smaller scope changes. For example, the scale of transport implementations is large, making it difficult to update them incrementally and therefore difficult to review.
> 
> This really makes absolutely no sense to me - the transport changes have to land in-sync with the other changes because they are all required by the upgrade to hyper 1. Doing them separately would not make much sense. Alternatively, we could try to introduce some of the changes required for this PR in `transport` ahead of time, but that is rife with potential problems, since the API changes between hyper 0.14 and hyper 1 are so large, that writing the same code for hyper 0.14 is considerable additional effort which doesn't (at least to me, contributing on my free time) feel like it will pay off.
> 
> It also seems that you would prefer your list of PRs to be merged before this one:
> 
> > The purpose of the following pull request is to improve both the user's convenience and the maintainability of this project itself, and I believe that these should be reflected first before migrating to hyper 1.

I think it's interesting to consider changes that isolate the impact of the specific dependencies used for the default transport implementation, but given that most downstream users are still likely to end up wanting to use the default transport implementation I feel like this is the more important PR.

> Again, thanks for the comments - I'm really looking forward to getting something like this merged, since it does also greatly impact my own user experience with tonic (since it prevents me from upgrading numerous other libraries which depend on hyper 1 now), which is a large part of my own motivation here.

Thanks for your mature response to comments and sorry for being unable to spend more time on this PR -- tonic is just a small part of my job's dependency stack so while I want to keep things moving forward I have to be deliberate in how I spend my time on this. That said, as mentioned before this is a high priority for me and will definitely keep spending time on reviewing this.

tottoto:
> I think I feel a particular frustration here because you are the 3rd person to "take a quick look" and provide review comments which are difficult to take significant actions on. 

I do not know what makes you frustrated, but I said `a quick look` to make it clear that I did not give a completely comprehensive review. And it is difficult to review for large changes that include changes that seem unrelated  (I read this entire pull request, of course). Additionally, it does not matter that anybody be a `3rd person` for reviewing.  Recommended that you understand that receiving multiple similar points means that that point of view is considered important and needs to be addressed. I cannot do anything about difficulties to take significant action on makes you frustrated. Because your emotions are your own. Since it contains inherently difficult issues, it is inevitable that the content pointed out in reviews will be difficult. What I can say is that expressing frustration over unfavorable comments for you is not a good idea for OSS because it evokes fear that makes people hesitant to review. I believe that OSS contributors participate in the hope that OSS will become better.

When making large-scale changes (which is not a bad thing in itself; there are times when it is inherently inevitable, and I think this will be the case), we should judge the necessity of the changes more strictly than usual as possible to reduce the cognitive area required in the review. In this regard, your suggestions in this pull request are painful to review. I understand that you are tired of what is taking up your time, but I encourage you to realize that you are making unkind suggestions in reviews.

> As well, although I'm probably mis-reading your comment, it comes off to me as pretty condescending to explain, yet again, that changes should be broken into small parts to make it easier on reviewers. 

Yes. It is not the intended meaning. It seems that you are reading too deeply and are reading into things that are not written. It doesn't matter whether you're a seasoned developer or not, and I'm not talking about it. I am commenting on the code. You look  be reluctant to basically make the changes smaller. In any case, at least I'm commenting on what I feel is necessary in the context, even if others have already pointed it out. In other words, at least from my point of view, it doesn't seem to be of sufficient quality in this respect. When we developers write code ourselves, we fully understand the scope of the code, so we tend to misunderstand some changes (including small refactoring) as trivial changes and underestimate the negative impact it will have on reviews. We tend to misunderstand, but the time it takes to write code has little to do with its quality.

> > The rebasing resolution seems to be broken, seeing the tls implementation.

> I'm not sure what you mean by this - can you point to something specific for me?

Can you explain the need for the following changes to this pull request?

`tonic/src/transport/service/tls.rs` of https://github.com/hyperium/tonic/pull/1670/files#diff-f8078e6eaa6dfb7a56bc0ee9d2b6c345c231ebaa4c76d58e54faa95bf43945c4

alexrudy:
@djc Thanks for the clarification about the version numbers - I see your point now. Its too bad that the current state of the `Cargo.toml` files for `tonic` is pretty inconsistent on this front.

> The abundance of examples is unfortunate in my view -- IMO examples have diminishing returns as the number grows, and given that the maintenance efforts of examples is non-zero, increasing numbers of examples over time can have a net-negative value to the project (in particular for examples that drag in dependencies beyond what the libraries need anyway). So I think it would be good to have a discussion on what examples we ought to keep which might reduce the changes needed in this PR, but I also don't necessarily want to block this PR on those changes.

Happy to consider removing `tls_rustls` in a separate PR - that would reduce the changes required to the tls code elsewhere, although this issue is likely to re-emerge as users who want to leverage hyper-rustls start to upgrade.

tonic/src/transport/service/tls.rs
```diff
@@ -47,13 +42,13 @@ impl TlsConnector {
         let mut roots = RootCertStore::empty();
 
         #[cfg(feature = "tls-roots")]
-        roots.add_parsable_certificates(rustls_native_certs::load_native_certs()?);
+        roots.add_parsable_certificates(rustls_native_certs::load_native_certs()?.into_iter());
```
djc, line 45: Seems unrelated?

tonic/src/transport/service/tls.rs
```diff
@@ -47,13 +42,13 @@ impl TlsConnector {
...
 
         #[cfg(feature = "tls-webpki-roots")]
         roots.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
 
         if let Some(cert) = ca_cert {
-            add_certs_from_pem(&mut Cursor::new(cert), &mut roots)?;
+            add_certs_from_pem(std::io::Cursor::new(cert.as_ref()), &mut roots)?;
```
djc, line 51: Seems unrelated?

tonic/src/transport/service/tls.rs
```diff
@@ -65,10 +60,12 @@ impl TlsConnector {
             None => builder.with_no_client_auth(),
         };
 
+        let domain: ServerName<'static> = ServerName::try_from(domain)?.to_owned();
```
djc, line 63: Seems unrelated?

tonic/src/transport/service/tls.rs
```diff
@@ -83,12 +80,19 @@ impl TlsConnector {
 
         // Generally we require ALPN to be negotiated, but if the user has
         // explicitly set `assume_http2` to true, we'll allow it to be missing.
-        let (_, session) = io.get_ref();
-        let alpn_protocol = session.alpn_protocol();
-        if !(alpn_protocol == Some(ALPN_H2) || self.assume_http2) {
-            return Err(TlsError::H2NotNegotiated.into());
-        }
-        Ok(BoxedIo::new(io))
+        let assume_http2 = self.assume_http2;
```
djc, line 83: Mostly formatting/syntax differences?

tonic/src/transport/service/tls.rs
```diff
@@ -165,29 +179,141 @@ impl fmt::Display for TlsError {
 
 impl std::error::Error for TlsError {}
 
-fn load_identity(
-    identity: Identity,
-) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>), TlsError> {
-    let cert = rustls_pemfile::certs(&mut Cursor::new(identity.cert))
-        .collect::<Result<Vec<_>, _>>()
-        .map_err(|_| TlsError::CertificateParseError)?;
+mod rustls_keys {
```
djc, line 182: Why does this need to become a separate module?

tonic/Cargo.toml
```diff
@@ -68,19 +69,22 @@ async-trait = {version = "0.1.13", optional = true}
...
 
 # rustls
-rustls-pemfile = { version = "2.0", optional = true }
+rustls-pemfile = { version = "2.0.0", optional = true }
 rustls-native-certs = { version = "0.7", optional = true }
-tokio-rustls = { version = "0.26", default-features = false, features = ["logging", "tls12", "ring"], optional = true }
-webpki-roots = { version = "0.26", optional = true }
+tokio-rustls = { version = "0.26", optional = true, default-features = false, features = ["logging", "ring", "tls12"] }
+rustls = { version = "0.23.4", optional = true, default-features = false, features = ["ring", "std", "tls12"] }
+webpki-roots = { version = "0.26.1", optional = true }
```
djc, line 87: There should be no need for any of the changes here.

tonic/src/transport/server/mod.rs
```diff
@@ -885,3 +1084,27 @@ where
         future::ready(Ok(svc))
     }
 }
+
+// A future which only yields `Poll::Ready` once, and thereafter yields `Poll::Pending`.
+#[pin_project]
```
djc, line 1089: Isn't there something in futures_util we can use for this?

djc (commented):
Here's a bunch of more detailed feedback. I think there are still a lot of unnecessary (formatting etc) changes in here that it would be good to get rid off to make the diff more focused.

tonic/src/transport/server/mod.rs
svix-jplatte: Looks like this is exactly `futures-util`s `Fuse` type (except that one uses pin-project-lite).

djc: I guess tonic itself doesn't already have a futures-util dependency and does already have a pin-project dependency, so maybe this is reasonable?

tonic/src/codec/decode.rs
```diff
@@ -27,7 +28,7 @@ struct StreamingInner {
     state: State,
     direction: Direction,
     buf: BytesMut,
-    trailers: Option<MetadataMap>,
+    trailers: Option<HeaderMap>,
```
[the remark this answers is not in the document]
alexrudy: This has to do with the change to how polling works in http bodies.

While polling frames, we may encounter the trailers, and it doesn't appear that the http_body API guarantees that we'll encounter them after the end of the data frames. So in `poll_next`, we can't avoid polling the trailers (as was done before) - we have to catch them as they come out of the `poll_frame` method.

Also, the `status::infer_grpc_status` method accepts `&HeaderMap`, so if we immediately consume the headers and store the `MetadataMap`, we can't get back to the GRPC status.

Therefore, we can either start storing the `Status` and the `MetadataMap`, and then only use one of them, depending on whether the end user uses the `Stream` api or calls `.trailers()` directly, or we can store the `HeaderMap` and do the conversion to `Status` or `MetadataMap` at the last moment.

tonic/src/transport/server/mod.rs
```diff
@@ -512,53 +526,226 @@ impl<L> Server<L> {
...
 
         let http2_keepalive_interval = self.http2_keepalive_interval;
         let http2_keepalive_timeout = self
             .http2_keepalive_timeout
             .unwrap_or_else(|| Duration::new(DEFAULT_HTTP2_KEEPALIVE_TIMEOUT_SECS, 0));
         let http2_adaptive_window = self.http2_adaptive_window;
-        let http2_max_pending_accept_reset_streams = self.http2_max_pending_accept_reset_streams;
 
-        let svc = self.service_builder.service(svc);
+        let http2_max_pending_accept_reset_streams = self.http2_max_pending_accept_reset_streams;
```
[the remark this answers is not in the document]
alexrudy: The big chunk of chagnes around the server `Builder` struct are hard to retain in the original format because the API for configuring servers has changed subtly - instead of the `http2` prefixed methods, we need a temporary `Http2Builder`, which means we can't inline all of the confiugration calls into a single statement. Additioanlly, `http2_only` no longer accepts a `bool`, so can't be inlined either.

tonic/src/transport/server/mod.rs
alexrudy: Yes, this is from `futures-util`, vendored here to avoid the dependency - but either approach is fine. I'll add a comment noting that.

tonic-web/src/call.rs
```diff
@@ -160,32 +163,47 @@ where
...
-                    Some(Ok(data)) => this.buf.put(data),
+                match ready!(this.inner.as_mut().poll_frame(cx)) {
+                    Some(Ok(frame)) if frame.is_data() => this.buf.put(frame.into_data().unwrap()),
+                    Some(Ok(frame)) if frame.is_trailers() => {
+                        return Poll::Ready(Some(Err(internal_error(
+                            "malformed base64 request has unencoded trailers",
+                        ))))
+                    }
+                    Some(Ok(_)) => {
+                        unreachable!("unexpected frame type")
```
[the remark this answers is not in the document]
alexrudy: Frame is internally an enum with only Data and Trailers - so this should be unreachable. But I guess to future-proof against http_body API changes, this can be an internal error.

tonic-web/src/layer.rs
```diff
@@ -24,7 +24,7 @@ impl Default for GrpcWebLayer {
 
 impl<S> Layer<S> for GrpcWebLayer
 where
-    S: Service<http::Request<hyper::Body>, Response = http::Response<BoxBody>>,
+    S: Service<http::Request<BoxBody>, Response = http::Response<BoxBody>>,
```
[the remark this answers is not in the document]
alexrudy: I think the alternative is to use `hyper::body::Incoming` here, but that makes testing and re-using `tonic-web` much harder, because `Incoming` cannot be constructed outside of `hyper`.

tonic-web/src/call.rs
```diff
@@ -237,49 +251,61 @@ where
...
-    ) -> Poll<Option<Result<Self::Data, Self::Error>>> {
+    ) -> Poll<Option<Result<http_body::Frame<Self::Data>, Self::Error>>> {
         if self.client && self.direction == Direction::Decode {
             let mut me = self.as_mut();
 
             loop {
-                let incoming_buf = match ready!(me.as_mut().poll_decode(cx)) {
-                    Some(Ok(incoming_buf)) => incoming_buf,
-                    None => {
-                        // TODO: Consider eofing here?
```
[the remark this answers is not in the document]
alexrudy: I think so - because we can continue polling the decoder waiting for trailers, and we can't switch to calling `poll_trailers`, so we no longer bail when there is no more data.

tonic/src/transport/server/incoming.rs
```diff
@@ -164,33 +164,66 @@ impl TcpIncoming {
...
+                set_accept_socketoptions(&stream, self.tcp_nodelay, self.tcp_keepalive_timeout);
+                Some(Ok(stream)).into()
+            }
+            other => Poll::Ready(other),
+        }
+    }
+}
+
+// Consistent with hyper-0.14, this function does not return an error.
+fn set_accept_socketoptions(
```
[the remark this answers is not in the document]
alexrudy: This is heavily inspired by the implementation in hyper 0.14's AddrIncoming, but not crib'd.

tonic/src/transport/server/incoming.rs
```diff
@@ -164,33 +164,66 @@ impl TcpIncoming {
...
 
 impl Stream for TcpIncoming {
-    type Item = Result<AddrStream, std::io::Error>;
+    type Item = Result<TcpStream, std::io::Error>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        Pin::new(&mut self.inner).poll_accept(cx)
+        match ready!(Pin::new(&mut self.inner).poll_next(cx)) {
+            Some(Ok(stream)) => {
+                set_accept_socketoptions(&stream, self.tcp_nodelay, self.tcp_keepalive_timeout);
```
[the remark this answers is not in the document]
alexrudy: Actually, internally, hyper used to set these options on the TcpStream [after accept via AddrIncoming](https://github.com/hyperium/hyper/blob/dedcb674f35eaec765a42b550caabe6f694d86d1/src/server/tcp.rs#L220). I think that isn't exposed because you can't set `nodelay` or `keepalive` on a listening socket because it can't keep alive or delay sending packets to a peer it doesn't yet know about.

tonic/src/transport/mod.rs
```diff
@@ -121,5 +127,8 @@ pub use self::server::ServerTlsConfig;
 #[cfg(feature = "tls")]
 #[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
 pub use self::tls::Identity;
+use crate::body::BoxBody;
 
 type BoxFuture<'a, T> = std::pin::Pin<Box<dyn std::future::Future<Output = T> + Send + 'a>>;
+pub(crate) type Response<B = BoxBody> = http::Response<B>;
+pub(crate) type Request<B = BoxBody> = http::Request<B>;
```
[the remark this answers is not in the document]
alexrudy: I un-moved them, and I agree, they don't add much value. Not sure of the desire to keep them (less diff-noise) or get rid of them.

alexrudy:
@djc Super helpful comments - especially since I've been knee-deep in this PR, and some of it I inherited from the previous attempts. Thanks for taking the time. I've spent some time minimizing the formatting changes etc. that you commented on.

tonic-web/src/layer.rs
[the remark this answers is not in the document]
LucioFranco: probably just makes sense to not rely on incoming tbh

tonic/src/codec/decode.rs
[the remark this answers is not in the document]
LucioFranco: This makes a lot of sense, as long as its not exposed externally its totally fine with me. I can't think of anything metadatamap does that is special so I don't see a problem there anyways.

tonic/src/codec/decode.rs
```diff
@@ -254,9 +255,18 @@ impl StreamingInner {
...
-            self.buf.put(data);
-            Ok(Some(()))
+        Poll::Ready(if let Some(frame) = chunk {
+            match frame {
+                frame if frame.is_data() => {
+                    self.buf.put(frame.into_data().unwrap());
+                    Ok(Some(()))
+                }
+                frame if frame.is_trailers() => {
+                    self.trailers = Some(frame.into_trailers().unwrap());
```
LucioFranco, line 265: Do we want to potentially merge the trailers rather than replace the old ones? This sounds like another lack of guarantee from the Body trait that isn't super clear what we should do. I think the safe bet is to merge?

tonic/src/codec/encode.rs
```diff
@@ -298,22 +298,21 @@ where
...
         match self.role {
-            Role::Client => Ok(None),
+            Role::Client => None,
             Role::Server => {
                 if self.is_end_stream {
-                    return Ok(None);
+                    return None;
                 }
 
+                self.is_end_stream = true;
```
LucioFranco, line 309: This seems to be a change not related to upgrading, can you explain why this was moved out of the if statement?

LucioFranco: Ah I see let me know if this is correct, now that the data/trailers are combined into the same stateless poll fn we can always assume that when we call `trailers` that this is when the incoming stream of bytes has ended so the stream is always over at this point.

tonic/src/transport/mod.rs
```diff
@@ -38,29 +39,31 @@
 //!     .connect()
 //!     .await?;
 //!
-//! channel.call(Request::new(BoxBody::empty())).await?;
+//! channel.call(Request::new(empty_body())).await?;
```
LucioFranco, line 42: nit: should we just fully qualify this instead of hiding the import. Feels a bit confusing as to where this `empty_body` fn comes from.

tonic/src/transport/mod.rs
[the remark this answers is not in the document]
LucioFranco: How bad is the diff noise? Maybe we can change this in a follow up PR? Though not super motivated that we actually need to do that.

LucioFranco:
I've gotten about halfway through the commits and will continue reviewing likely tomorrow depending on time. Left some feedback but mostly questions. Overall looks great. Thank you so much for breaking this up into commits, it has made it so much easier to review!

tonic/src/transport/server/mod.rs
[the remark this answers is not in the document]
djc: Makes sense, thanks!

tonic/src/codec/decode.rs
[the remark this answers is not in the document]
djc: @LucioFranco I wonder, what is the purpose of the `MetadataMap` wrapper? Are there good reasons to not get rid of it in favor of using `HeaderMap` directly (potentially with a custom `HeaderValue` type)?

tonic/src/transport/mod.rs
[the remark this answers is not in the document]
djc: Let's keep them in this PR, I might stage a change to get rid of them after.

tonic-web/src/layer.rs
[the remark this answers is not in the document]
djc: So should we have our own `Incoming` type maybe? Not necessarily a fan of the type erasure approach of just using `BoxBody` all over the place.

tonic-web/src/call.rs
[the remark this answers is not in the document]
djc: I think returning an error might be better here -- avoids another panic path even though I agree with you it seems actually unreachable in practice.

tonic/src/codec/decode.rs
alexrudy: I think that makes sense - I can't find any specific documentation on what the semantics of `poll_frame` should be. The `Body` trait vaugely implies that trailers should only come at the end, but even in that case it might still be legal to end a body with two trailer frames.

tonic-web/src/layer.rs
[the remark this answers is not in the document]
alexrudy: I think the Body story is pretty fragmented. Internally, we are converting to and from `axum::body::Body` when using the `transport`. Axum just wraps a `BoxBody` in their own tuple struct. Effectively so does reqwest. We could add our own body to the mix, and we might get some benefits for enum-dispatch over one layer, but the next layer is gonna hit the v-table.

tonic/src/codec/decode.rs
djc: @seanmonstar do you have any context on this? I'd be happy to submit a docs PR for `http_body` if you can clarify the intended semantics?

tonic/src/codec/encode.rs
alexrudy: Yeah, we only call `trailers` after the stream has returned `None`, so it is safe to assume it is done. Otherwise, we don't ever mark `is_end_stream` when the stream ends normally without an error.

tonic/src/codec/decode.rs
seanmonstar: There _should_ only be one trailers frame. That's what h2 and h3 do. But some things are overly complicated to encode into the type system... So, I guess someone _could_ make a body that does weird things.

tonic/src/codec/decode.rs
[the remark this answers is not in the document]
AmeliasCode: That might be confusing to people coming from other grpc implementations since "metadata" is the "official" nomenclature.

tonic/src/transport/server/mod.rs
```diff
@@ -9,6 +9,13 @@ mod tls;
 #[cfg(unix)]
 mod unix;
 
+use tokio_stream::StreamExt as _;
+use tower::util::BoxCloneService;
+use tower::util::Oneshot;
+use tower::ServiceExt;
+use tracing::debug;
+use tracing::trace;
```
LucioFranco, line 17: nit: I would like the imports to be merged as much as possible. Mostly at the module level.

tonic/src/transport/server/mod.rs
LucioFranco: avoiding the dependency is the right way 👍

tonic/src/transport/server/mod.rs
```diff
@@ -527,42 +536,213 @@ impl<L> Server<L> {
...
+}
+
+type ConnectionBuilder = hyper_util::server::conn::auto::Builder<TokioExecutor>;
+
+/// An adaptor which converts a [`tower::Service`] to a [`hyper::service::Service`].
+///
+/// The [`hyper::service::Service`] trait is used by hyper to handle incoming requests,
+/// and does not support the `poll_ready` method that is used by tower services.
+#[derive(Debug, Copy, Clone)]
+pub struct TowerToHyperService<S> {
```
LucioFranco, line 676: Do we need to make these public? We could in theory just accept `tower::Service` and convert under the hood?

tonic/src/status.rs
```diff
@@ -412,13 +412,7 @@ impl Status {
         // > status. Note that the frequency of PINGs is highly dependent on the network
         // > environment, implementations are free to adjust PING frequency based on network and
         // > application requirements, which is why it's mapped to unavailable here.
-        //
-        // Likewise, if we are unable to connect to the server, map this to UNAVAILABLE.  This is
-        // consistent with the behavior of a C++ gRPC client when the server is not running, and
-        // matches the spec of:
-        // > The service is currently unavailable. This is most likely a transient condition that
-        // > can be corrected if retried with a backoff.
```
LucioFranco, line 420: should we move this comment down to the new spot where we check the connect error?

LucioFranco (changes requested):
Okay I have fully reviewed through everything. This looks really great, I've created a `v0.11.x` branch so that we can merge this and still be able to patch release the old code. 


I have a few comments that should be easy to address, lets do those and lets merge this! Thank you so much! Please when you have updated the PR ping me on discord and here. That way I can take a look right away.

alexrudy:
@LucioFranco I think I have addressed your comments!

tonic/src/codec/decode.rs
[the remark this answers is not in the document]
LucioFranco: @djc its to not expose the internal details and to provide ascii/bin support.

LucioFranco (approved):
LGTM! Thanks to everyone for following up and getting this work completed. This is probably the biggest refactor that tonic has seen and I appreciate the patience. This will get released soon, I am going to collect a few other breaking changes and then ship this all together. There is a `v0.11.x` branch that will keep support for hyper v0.14.

PR #1673, opened 2024-04-04
chore: Update to cargo-deny v2 config

tottoto:
## Motivation

Updates to the latest `cargo-deny` config.

## Solution

Follows instructions provided in https://github.com/EmbarkStudios/cargo-deny/pull/611.


tottoto:
The CI error would be fixed in #1674.

PR #1674, opened 2024-04-04
chore(tests): Fix dead code warning

tottoto:
Fixes dead code warning which causes the ci failure.

PR #1676, opened 2024-04-07
chore(tls): Update to rustls 0.23

tottoto:
Updates to `rustls` 0.23.


tottoto:
The failed CIs would be fixed in #1674 and #1677.

djc (approved):

PR #1681, opened 2024-04-10
chore(tls): Refactor tls alpn protocol check

tottoto:
Refactors tls alpn protocol check.


tottoto:
The failed CIs would be fixed in #1674 and #1677.

djc (approved):

djc:
Thanks!

PR #1682, opened 2024-04-11
Expose Error Status To Caller

zhixinwen:
## Motivation

We were seeing a high rate `Missing request message` error on server side, but we do not know what exactly was going on.
I think the error message may be hidden away in `poll_data`.
I am not sure why tonic choose to drop the error message instead of returning it to the user.
Can people familiar with the code explain why it was done in the current way?

## Solution

return the error to users, so the error is more meaningful


LucioFranco (approved):
LGTM thanks!

bmwill:
One unfortunate outcome of this change seems to be that once an error occurs, the Stream will yield that error forever instead of terminating the stream on error as was done before.

djc:
@bmwill that sounds bad -- want to submit a PR to fix it?

bmwill:
@djc yeah happy to take a look at addressing the issue, assuming the behavior is a regression that the project would like fixed and not an change in behavior that it would like to keep.

djc:
I think that's a good assumption!

bmwill:
@djc perfect, opened https://github.com/hyperium/tonic/pull/1823 to address this!

PR #1684, opened 2024-04-16
Expose formatted service name as a `const fn`

mkatychev:
## Motivation

It is oftentimes cumbersome to use `tonic::server::NamedService::NAME` when `T` has multiple generics:
```rust
impl<T: ServerTrait> tonic::server::NamedService for ServiceServer<T> {
    const NAME: &'static str = "service_api.Service";
}
```

This is particularly egregious when using the `HealthReporter` methods since `HealthReporter::set_serving` requires a fully qualified `ServiceServer<T>`:

```rust
let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
health_reporter.set_serving::<MyServer<MyInner<MyLongQualifiedType, MyOtherQualifiedType, T, T2>>>().await;
```

## Solution

expose a `pub const fn service_name() -> &'static str` that returns the formatted string.


AmeliasCode:
+1 to this! It would make it a lot less verbose to get the service name.

tonic-build/src/server.rs
```diff
@@ -375,8 +375,14 @@ fn generate_named(
     service_name: &str,
 ) -> TokenStream {
     let service_name = syn::LitStr::new(service_name, proc_macro2::Span::call_site());
+    let name_doc = generate_doc_comment(" Generated gRPC service name");
 
     quote! {
+        #name_doc
+        pub const fn service_name() -> &'static str {
```
djc, line 382: I think it would still make sense for this to a `const SERVICE_NAME: &str`?

mkatychev: fair point, should the trait const below be `const NAME: &'static str = SERVICE_NAME;` then?

djc: Might be nice, want to make that change?

mkatychev:
I cannot seem to reproduce this diff locally:
https://github.com/hyperium/tonic/actions/runs/9472891207/job/26134316596#step:7:16

mkatychev:
> @mkatychev you just need to run `cargo run --package codegen`

Hi @LucioFranco  I did that with the given version on macOS and I am not able to produce the comment diff in the run log

djc:
> > @mkatychev you just need to run `cargo run --package codegen`
> 
> Hi @LucioFranco I did that with the given version of protoc (25.1) on macOS and I am not able to produce the comment diff in the run log

Run `cargo update` locally before you run the codegen?

djc (approved):

djc:
Thanks!

PR #1687, opened 2024-04-20
feat: Make async_trait optional

AaronDewes:
## Motivation

Using `async fn` in traits has been [stabilized in recent Rust version](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html). Because of this, the dependency on `async_trait` is no longer necessary with recent Rust versions. Removing unneeded dependencies can reduce compile times and help reduce the risk of supply chain attacks.

## Solution

This implements a feature `async_trait` for `tonic` and `tonic_build`. To avoid breaking any existing things, this feature is enabled by default.


matze:
Doesn't this require #1556 in case that feature is disabled?

AaronDewes:
Yes, you're right, I also noticed that while trying, that's why it's a draft. I'll have a look at integrating these changes soon.

PR #1649, opened 2024-03-04
chore: Update to base64 0.22

tottoto:
Updates to `base64` 0.22.

PR #1689, opened 2024-04-24
Update base64 dependency from 0.21 to 0.22

j-vanderstoep:
Tonic currently depends on two different versions of base64 - directly on 0.21, and transitively through rustls-pemfile on 0.22. Updating tonic to use 0.22 results in just a single version dependency. This is helpful for ecosystems (like Android) where keeping crates on the latest version of dependencies across various projects helps reduce memory usage and disc space.

### DIFF
--- a/tonic-web/Cargo.toml
+++ b/tonic-web/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/hyperium/tonic"
 version = "0.11.0"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


djc (approved):

djc:
Would be nice to have!

tottoto:
Duplicate of #1649.

LucioFranco:
Merged #1649 so closing this thanks!

PR #1697, opened 2024-05-09
feat: support `async_trait`-less generation

jun-sheaf:
## Motivation

`async_trait` is not necessary for most use-cases. This PR provides the option for users.

## Solution

Providing an option to generate async traits without `async_trait`.

Fixed: https://github.com/hyperium/tonic/issues/1651


jun-sheaf:
@LucioFranco Note that the current CI doesn't test all feature combinations. We should use `cargo hack --feature-powerset test` to ensure all features are covered.

djc:
IIRC the use of async methods in traits that are part of public API is still discouraged while there is no way to guarantee auto trait bounds etc. So I'm not completely sure the time is right to start offering this. What's your motivation?

Cargo.toml
```diff
@@ -5,11 +5,12 @@ members = [
   "tonic-health",
   "tonic-types",
   "tonic-reflection",
-  "tonic-web", # Non-published crates
+  "tonic-web",                        # Non-published crates
```
djc, line 8: Please avoid unrelated (formatting) changes.

codegen/Cargo.toml
```diff
@@ -8,4 +8,8 @@ version = "0.1.0"
 
 [dependencies]
 tempfile = "3.8.0"
-tonic-build = {path = "../tonic-build", default-features = false, features = ["prost", "cleanup-markdown"]}
+tonic-build = { path = "../tonic-build", default-features = false, features = [
```
djc, line 11: Please avoid unrelated (formatting) changes.

tonic-build/src/server.rs
```diff
@@ -274,91 +287,187 @@ fn generate_trait_methods<T: Service>(
...
-            method.server_streaming(),
-            generate_default_stubs,
-        ) {
-            (false, false, true) => {
-                quote! {
-                    #method_doc
-                    async fn #name(#self_param, request: tonic::Request<#req_message>)
-                        -> std::result::Result<tonic::Response<#res_message>, tonic::Status> {
-                        Err(tonic::Status::unimplemented("Not yet implemented"))
+        #[cfg(not(feature = "async_trait"))]
```
djc, line 290: That's a lot of code duplication on top of what's already there. Given that the only difference is whether the method is prefixed with `async` or not and wrapping of the return type in `impl std::future::Future<Output = ..>`, can we simplify?

jun-sheaf: Decomposing a method to such an extent may not be advisable. It represents a premature optimization that could lead to developer confusion and complications. Understanding the interdependencies among the components within the decomposition requires extra caution, potentially increasing the risk of errors.

djc: That's not a satisfying answer to me. I don't think the term "premature optimization" applies here, since there's no performance impact to what I'm suggesting (indeed, the performance might well be worse).

I'm suggesting that the increased code duplication makes this code harder to maintain (for example, carefully reviewing these changes is a pain in the ass), and asking you whether that issue could be mitigated.

jun-sheaf: The changes have already been made. I was just expressing an opinion.

jun-sheaf: @djc Gentle ping. Could you please take a look?

LucioFranco:
I think we need to generate the code as `fn foo() -> impl Future + Send` to allow sendable futures. We also need to check if the generated functions will work with the Arc since that also needs Sync. I would like to move in this direction but I don't feel like there is a huge rush. I would like to see some more tests as well checking all the permutations of sendiness/syncness so that we know this will work for users.

PR #1707, opened 2024-05-24
feat(tls): Use rustls_pki_types::CertificateDer to describe DER encoded certificate

tottoto:
## Motivation

Although `Certificate` is for PEM encoded certificate, it is used by `Request::peer_certs`'s return type, which is DER encoded certificate.

## Solution

Adds `CertificateDer` type and changes `Request::peer_certs` to use it.

This is a breaking change.

### DIFF
--- a/tonic/src/request.rs
+++ b/tonic/src/request.rs
@@ -2,13 +2,15 @@ use crate::metadata::{MetadataMap, MetadataValue};
 #[cfg(feature = "transport")]
 use crate::transport::server::TcpConnectInfo;
 #[cfg(feature = "tls")]
-use crate::transport::{server::TlsConnectInfo, CertificateDer};
+use crate::transport::server::TlsConnectInfo;
 use crate::Extensions;
 #[cfg(feature = "transport")]
 use std::net::SocketAddr;
 #[cfg(feature = "tls")]
 use std::sync::Arc;
 use std::time::Duration;
+#[cfg(feature = "tls")]
+use tokio_rustls::rustls::pki_types::CertificateDer;
 use tokio_stream::Stream;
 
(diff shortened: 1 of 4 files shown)


djc:
So I'm guessing the name is inspired by `rustls_pki_types::CertificateDer<'_>`. Given that rustls-pki-types is intended to have a long-term stable API, I would like to suggest that we adopt usage of types like `CertificateDer<'_>` in the tonic API directly instead of copying them here. This would ensure improved type safety and clarity across crates, and should cause no public API churn due to the stability promise.

tottoto:
> So I'm guessing the name is inspired by rustls_pki_types::CertificateDer<'_>. 

Yes. introduced `CertificateDer` is a type for not exposing `rustls-pki-types` crate. I think both exposing `rustls-pki-types` and hiding it are reasonable, so I will follow the maintainer's decision.

tottoto:
> So I'm guessing the name is inspired by rustls_pki_types::CertificateDer<'_>. 

Yes. introduced `CertificateDer` is a type for not exposing `rustls-pki-types` crate. I think both exposing `rustls-pki-types` and hiding it are reasonable, so I am not sure which is more appropriate.

tottoto:
Added a commit to show how the implementation would be when using `rustls_pki_types::CertificateDer` directly.

djc:
LGTM. I think the same reasoning as in #1710 should apply here: rustls-pki-types is expected to have a more stable API than tonic so it should be okay to use rustls-pki-types API in the public tonic API.

tottoto:
Updated this pull request title to reflect the latest status.

LucioFranco (approved):
Using a 1.0 crate API seems fine to me.

PR #1708, opened 2024-05-26
Remove warp examples

alexrudy:
Warp is much less popular than Axum today - 30,000 downloads/day on crates.io vs Axum's 200,000/day, as a rough quantiative measure.

As well, warp does not yet have compatiblity with hyper-1.0, though there is some work ongoing (seanmonstar/warp#1088).

When warp adds compatibility with hyper 1.0, these examples can be restored.


tisonkun (approved):
Thanks for helping drive this progress.

I'm waiting for the final blocker at tonic to bump all our HTTP stack to hyper-1.0 ... It's now blocked by we defining middleware for both tonic-web and axum so we cannot bump axum along and let tonic use the 0.6 version.

djc:
@tisonkun we all want it, no need to keep piling on.

djc:
Thanks!

PR #1709, opened 2024-05-26
Bump nightly to 2024-04-16

alexrudy:
Upgrading to hyper>=1 will reqiure that we upgrade axum to >= 0.7. Bumping nightly is required to support axum >= 0.7 on nightly, since it uses features not stabilized before this date. This is in turn reqiured to run `cargo udeps` against tonic and associated crates.


djc:
Thanks!

PR #1713, opened 2024-05-28
feat: Implement http_body::Body::size_hint for custom body

tottoto:
## Motivation

Provides the `Body` stream's size hint.

## Solution

Implements `Body::size_hint` for custom body types. It is intentional that this is not added to test code.


djc (approved):

djc:
Nice, thanks!

PR #1715, opened 2024-05-29
chore: Improve document about certificate

tottoto:
## Motivation

Adds documents related to #1707 separated from its breaking changes.

## Solution

Adds some documents of certificates for that users may be confused about.

### DIFF
--- a/tonic/src/request.rs
+++ b/tonic/src/request.rs
@@ -256,6 +256,12 @@ impl<T> Request<T> {
     /// and is mostly used for mTLS. This currently only returns
     /// `Some` on the server side of the `transport` server with
     /// TLS enabled connections.
+    ///
+    /// ## NOTE
+    /// Although the returned ['Certificate'] is supposed to contain PEM
+    /// encoded certificate, it actually contains a DER encoded certificate.
+    /// This issue is planned to be addressed.
+    /// See <https://github.com/hyperium/tonic/pull/1707>.
     #[cfg(feature = "tls")]
     #[cfg_attr(docsrs, doc(cfg(feature = "tls")))]
     pub fn peer_certs(&self) -> Option<Arc<Vec<Certificate>>> {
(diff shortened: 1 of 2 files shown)


tonic/src/request.rs
```diff
@@ -256,6 +256,12 @@ impl<T> Request<T> {
     /// and is mostly used for mTLS. This currently only returns
     /// `Some` on the server side of the `transport` server with
     /// TLS enabled connections.
+    ///
+    /// ## NOTE
+    /// Alghouth the returned ['Certificate'] is supposed to contain PEM
```
AmeliasCode, line 261: Typo: "although"

tonic/src/transport/tls.rs
```diff
@@ -15,6 +15,11 @@ impl Certificate {
     /// Parse a PEM encoded X509 Certificate.
     ///
     /// The provided PEM should include at least one PEM encoded certificate.
+    ///
+    /// ## SAFETY
+    /// `pem` must contain valid PEM encoded certificate. It is undefined
+    /// behavior to call this with `pem` that is not valid PEM encoded
```
AmeliasCode, line 21: Is this really undefined behavior? When I hear "UB" I think things like "the compiler is free to do whatever it wants", but in this case it seems like the behavior is pretty clear: You get a valid (in the rust sense) Certificate object containing invalid data (in the business logic sense). Am I missing something?

tonic/src/request.rs
tottoto: Thanks! Fixed it.

tonic/src/transport/tls.rs
tottoto: Thanks! Makes sense. The "UB" should be in the meaning of Rust. I updated it just a detailed document. By the way, now that I think about it, the constructor name is `from_pem` so it might not as confusing as I thought it would be. If this note section seems unnecessary, I will remove it.

djc:
I don't think we'll want to merge this, we should merge (something like) #1707 instead.

tottoto:
I understand. I will close this.

PR #1724, opened 2024-06-12
feat(tls): Add ability to add multiple ca certificates

### LINKED ISSUE #1629: feat: permit multiple ca_certificates
### Motivation

Sometimes there are multiple ca's to be added to the trust store. For instance, when migrating between setups (e.g. going from an internal manual CA to a SPIRE setup). Having only one ca_certificate means a flag-day event is required rather than permitting graceful migration component by component.

### Proposal

Change the config builder to have a list of ca's to add; change the ca_certificate call to appends to the vector.

tottoto:
Resolves #1629.

### DIFF
--- a/tonic/src/transport/channel/tls.rs
+++ b/tonic/src/transport/channel/tls.rs
@@ -51,6 +51,13 @@ impl ClientTlsConfig {
         ClientTlsConfig { certs, ..self }
     }
 
+    /// Sets the multiple CA Certificates against which to verify the server's TLS certificate.
+    pub fn ca_certificates(self, ca_certificates: impl IntoIterator<Item = Certificate>) -> Self {
+        let mut certs = self.certs;
+        certs.extend(ca_certificates);
+        ClientTlsConfig { certs, ..self }
+    }
+
     /// Sets the client identity to present to the server.
     pub fn identity(self, identity: Identity) -> Self {
         ClientTlsConfig {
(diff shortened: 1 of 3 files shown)


djc (approved):
Is the next step here to get rid of the `Certificate` abstraction around `CertificateDer<'_>`? Would be nice!

tottoto:
I would think that `Certificate` interface cannot be removed to support PEM encoded certificate file.

djc:
Well, the caller would rely on rustls-pemfile so it's one less mandatory dependency to carry within tonic.

tottoto:
Considering `rustls-pemfile` has major version release, I would think it would be acceptable to depend on the crate for the purpose of easiness to use.

tottoto:
Added a method to add multiple ca certificates at once, which is useful when having some sets of certificates.

tonic/src/transport/channel/tls.rs
```diff
@@ -51,6 +51,13 @@ impl ClientTlsConfig {
         ClientTlsConfig { certs, ..self }
     }
 
+    /// Sets the multiple CA Certificates against which to verify the server's TLS certificate.
+    pub fn ca_certificates(self, ca_certificates: Vec<Certificate>) -> Self {
```
djc, line 55: Let's take an `impl IntoIterator<Item = Certificate>` here, instead?

djc (approved):

tonic/src/transport/channel/tls.rs
tottoto: Sounds good to me. Updated to use it.

PR #1726, opened 2024-06-14
Add timer to hyper server for http2_keepalive_interval

AmeliasCode:
## Motivation

As described in this issue: https://github.com/hyperium/tonic/issues/1725 the recent upgrade to hyper 1 has broken services that use http2_keepalive_interval. This was due to a missing timer, resulting in panics with the message `You must supply a timer.` See here: https://github.com/hyperium/hyper/blob/master/src/common/time.rs#L36

## Solution

This solution adds in a timer to the server via this method: https://docs.rs/hyper-util/0.1.5/hyper_util/server/conn/auto/struct.Http2Builder.html#method.timer

The timer is a standard tokio timer: https://docs.rs/hyper-util/0.1.5/hyper_util/rt/tokio/struct.TokioTimer.html

Fixes https://github.com/hyperium/tonic/issues/1725

### DIFF
--- a/tests/integration_tests/tests/http2_keep_alive.rs
+++ b/tests/integration_tests/tests/http2_keep_alive.rs
@@ -1,18 +1,15 @@
-use integration_tests::pb::{test_client::TestClient, test_server, Input, Output};
-use std::sync::{Arc, Mutex};
 use std::time::Duration;
+
 use tokio::sync::oneshot;
-use tonic::{
-    transport::{Endpoint, Server},
-    Code, Request, Response, Status,
-};
+
+use integration_tests::pb::{test_client::TestClient, test_server, Input, Output};
+use tonic::{transport::Server, Request, Response, Status};
 
 struct Svc;
 
 #[tonic::async_trait]
 impl test_server::Test for Svc {
     async fn unary_call(&self, _: Request<Input>) -> Result<Response<Output>, Status> {
-
         Ok(Response::new(Output {}))
     }
 }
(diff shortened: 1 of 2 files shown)


djc (approved):
Thanks!

akiradeveloper:
@AmeliasCode Thanks to your work. Unfortunately, my test still fails with the fix. 

```
   Compiling tonic v0.11.0 (https://github.com/hyperium/tonic#98478acc)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 56.86s
     Running `target/debug/tonic-hyper10-debug 1`
test 1 (fail)
thread 'main' panicked at /Users/akira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hyper-1.3.1/src/common/time.rs:37:17:
You must supply a timer.
```

Apparently, your test adds `http2_keepalive_interval` at the server-side but my test add it on client side. I guess this is the reason my test isn't fixed.

```rust
            let endpoint = Endpoint::from_static("http://localhost:50000")
                // commenting out this line turn it into pass.
                .http2_keep_alive_interval(std::time::Duration::from_secs(1))
                .keep_alive_while_idle(true)
                .timeout(std::time::Duration::from_secs(5))
                .connect_timeout(std::time::Duration::from_secs(5));
```

Is enabling keepalive at client-side invalid? If so, it should be removed from the API.

nit: The commit ordering is very strange. The integration test should have been added at first.

AmeliasCode:
@akiradeveloper Sorry for missing that you were experiencing the error on the client side. I was seeing it on the server side so that's what I was focusing on. I have a fix for the client side as well that I'll raise in a moment.

PR #1727, opened 2024-06-14
Allow specifying compression encoding order preference

CapCap:
## Motivation
Although Zstd is significantly more performant (> 5x in our testing), the default chosen encoding remains Gzip when both are enabled. This blocks us from being able to gradually roll out zstd support for existing downstream clients before disabling Gzip entirely.

## Solution
This PR introduces a feature that allows for specifying the order of compression encodings. The key changes include:
	-	Updated EnabledCompressionEncodings to support ordering/prioritizing multiple compression encodings such as Gzip and Zstd.
	-	Modified the enable function to reorder encodings based on the priority, ensuring the most recently enabled encoding has the lowest priority- i.e "first encoding enabled has the highest priority"
	-	Added new tests to validate the correct priority ordering of compression encodings settings.
	-  Make sure GRPC correctly copies this order for each connection as well.
	
The behavior is now "given a list of supported encodings, I will pick the highest priority one that I support"

### DIFF
--- a/tonic/src/codec/compression.rs
+++ b/tonic/src/codec/compression.rs
@@ -2,6 +2,7 @@ use crate::{metadata::MetadataValue, Status};
 use bytes::{Buf, BytesMut};
 #[cfg(feature = "gzip")]
 use flate2::read::{GzDecoder, GzEncoder};
+use http::HeaderValue;
 use std::fmt;
 #[cfg(feature = "zstd")]
 use zstd::stream::read::{Decoder, Encoder};
@@ -9,13 +10,15 @@ use zstd::stream::read::{Decoder, Encoder};
 pub(crate) const ENCODING_HEADER: &str = "grpc-encoding";
 pub(crate) const ACCEPT_ENCODING_HEADER: &str = "grpc-accept-encoding";
 
+/// This should always match the cardinality of the `CompressionEncoding` enum
+pub(crate) const COMPRESSION_ENCODINGS_LENGTH: usize = 2;
+
 /// Struct used to configure which encodings are enabled on a server or channel.
-#[derive(Debug, Default, Clone, Copy)]
+/// Supports setting the priority of each compression
+#[derive(Debug, Default, Clone)]
 pub struct EnabledCompressionEncodings {
-    #[cfg(feature = "gzip")]
-    pub(crate) gzip: bool,
-    #[cfg(feature = "zstd")]
-    pub(crate) zstd: bool,
+    // We have an array so we can keep the order of the encodings (i.e prefer `zstd` over `gzip`)
+    pub(crate) order: Vec<CompressionEncoding>,
 }
 
 impl EnabledCompressionEncodings {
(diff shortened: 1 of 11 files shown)


tonic/src/codec/compression.rs
```diff
@@ -78,7 +120,7 @@ pub(crate) struct CompressionSettings {
 }
 
 /// The compression encodings Tonic supports.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, EnumCountMacro, PartialEq, Eq)]
```
djc, line 123: I don't think we want to take on the extra dependency just for this. Should be pretty easy to just implement methods by hand to do what we need here?

tonic/src/codec/compression.rs
```diff
@@ -2,20 +2,30 @@ use crate::{metadata::MetadataValue, Status};
...
 pub(crate) const ACCEPT_ENCODING_HEADER: &str = "grpc-accept-encoding";
 
+/// This should always match the cardinality of `CompressionEncoding`
+pub(crate) const COMPRESSION_ENCODINGS_LENGTH: usize = CompressionEncoding::COUNT;
+
 /// Struct used to configure which encodings are enabled on a server or channel.
+/// Supports setting the priority of each compression
 #[derive(Debug, Default, Clone, Copy)]
 pub struct EnabledCompressionEncodings {
+    // We keep the encodings as struct values so we get `const fn` compatibility
```
djc, line 22: This all seems pretty roundabout. Can we just change `EnabledCompressionEncodings` to store a `Vec<CompressionEncoding>` instead?

CapCap: Is a slice (as is) ok, or would you prefer a true vec?

tonic/src/codec/compression.rs
CapCap: I have changed it to be manually tracked- when you say by hand, do you intend a macro, or just a regular vec?

tonic/src/codec/compression.rs
```diff
@@ -160,29 +208,23 @@ impl CompressionEncoding {
...
+        #[cfg(any(feature = "gzip", feature = "zstd"))]
         match self {
             #[cfg(feature = "gzip")]
             CompressionEncoding::Gzip => "gzip",
             #[cfg(feature = "zstd")]
             CompressionEncoding::Zstd => "zstd",
         }
+
+        #[cfg(not(any(feature = "gzip", feature = "zstd")))]
+        unreachable!("CompressionEncoding::as_str called without any compression features enabled")
```
djc, line 221: This seems problematic? Why do this?

tonic/src/codec/compression.rs
djc: The way you've implemented the array here seems to cause a bunch of unnecessary complexity. I think a `Vec` would be simpler to use internally (just `contains()` calls on it, `push`, `remove`).

I also think the first element in the list should take priority rather than the last, and not the most recently added element, that seems confusing to me.

tonic/src/codec/compression.rs
CapCap: because rust won't let you have a `match` on an empty enum (which is what happens when both are turned off)

tonic/src/codec/compression.rs
CapCap: The way it's implemented, the most recently added element takes the **lowest** real priority; i.e the way it is currently implemented already matches your request of "the first element in the list should take priority rather than the last, and not the most recently added element". 

I will try to simplify a bit (with vec), but no matter the change, you will ultimately have a disconnect between where you insert things, and what the "priority" is

CapCap: The `array` instead of `Vec` was, originally, to localize/minimize the impact of my changes, and support a `copy` type; moving to Vec means spreading `.clone()` in a few places, but also moving to using refs in others

LucioFranco:
@CapCap I ran CI looks like there are some failures, can you address those and then ping me again when its ready for review and CI is passing? thanks

tonic/src/codec/compression.rs
```diff
@@ -2,69 +2,91 @@ use crate::{metadata::MetadataValue, Status};
...
-            CompressionEncoding::Gzip => self.gzip,
+            CompressionEncoding::Gzip => self.is_gzip_enabled(),
             #[cfg(feature = "zstd")]
-            CompressionEncoding::Zstd => self.zstd,
+            CompressionEncoding::Zstd => self.is_zstd_enabled(),
         }
     }
 
     /// Enable a [`CompressionEncoding`].
+    /// Every time an encoding is enabled, it is given the lowest priority: i.e first added, highest priority
```
djc, line 36: How about we adopt `push_front()`, `push_back()` instead of this?

tonic/src/codec/compression.rs
```diff
@@ -104,19 +126,34 @@ impl CompressionEncoding {
...
+                };
+                if let Some(encoding) = encoding {
+                    enabled_encodings
+                        .priority(encoding)
+                        .map(|priority| (encoding, priority))
+                } else {
+                    None
+                }
+            })
+            .max_by_key(|(_, priority)| *priority)
```
djc, line 149: I'm confused by this. It seems to suggest that we pick the max by priority (which is implemented as index in the array), but in the other comment you said that the array represents highest-to-lowest priority order. So which is it?

tonic/src/codec/compression.rs
djc: See my other comments. Array seems okay to me.

djc:
@CapCap we merged #1757 which is substantially similar to this -- hope that's okay and that it address your issues. If not, please open an issue so we can discuss in more detail what you need.

PR #1729, opened 2024-06-14
chore(transport): Use hyper_util::service::TowerToHyperService

tottoto:
Uses `hyper_util::service::TowerToHyperService` instead of defining the own version with change.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -16,7 +16,10 @@ use tracing::{debug, trace};
 pub use crate::service::{Routes, RoutesBuilder};
 
 pub use conn::{Connected, TcpConnectInfo};
-use hyper_util::rt::{TokioExecutor, TokioIo, TokioTimer};
+use hyper_util::{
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -595,7 +598,9 @@ impl<L> Server<L> {
                     let req_svc = svc
                         .call(&io)
                         .await
-                        .map_err(super::Error::from_source)?;
+                        .map_err(super::Error::from_source)?
```
djc, line 601: Style nit: how do you feel about using an explicit `match` here? I generally feel like `map().map_err()` doesn't give as much clarity.

djc (approved):
Nice, thanks!

tonic/src/transport/server/mod.rs
tottoto: Personally, I feel the style using method chain like this to be a bit more intuitive.

tottoto: I feel the style using method chain like this to be a bit more intuitive.

PR #1731, opened 2024-06-16
feat(tls): Remove tls roots implicit configuration

tottoto:
Adds ability to add CA certificates and removes `tls-webpki-roots` and `tls-roots` features.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -29,9 +29,8 @@ zstd = ["dep:zstd"]
 default = ["transport", "codegen", "prost"]
 prost = ["dep:prost"]
 tls = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"]
-tls-roots = ["tls-roots-common", "dep:rustls-native-certs"]
-tls-roots-common = ["tls", "channel"]
-tls-webpki-roots = ["tls-roots-common", "dep:webpki-roots"]
+tls-roots = ["tls", "channel", "dep:rustls-native-certs"]
+tls-webpki-roots = ["tls", "channel", "dep:webpki-roots"]
 router = ["dep:axum"]
 server = [
   "router",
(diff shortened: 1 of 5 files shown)


djc:
Mentioned this on Discord, just repeating it here so it doesn't get lost: why do you want to remove the TLS roots features? In #1724 you didn't want to require the caller to have to depend on rustls-pemfile directly, so this seems to be reasoning in the opposite direction.

tottoto:
They share the same philosophy of increasing the user's freedom of configuration without compromising ease of use as much as possible, and reducing the overall complexity of tonic to lower maintenance costs.

Considering that using PEM encoded certificate files when using tls is a typical use case, I think it is a redundant interface to leave the user to make that conversion as web framework.

Removing tls roots and adding interfaces to add them (in exchage for increasing internal `Certificate` implementation's complexity a little) simplify the tls feature config the and config implementation by removing `Endpoint`'s `tls_assume_http2` config and reducing feature config and its conditional compilation. And users are free to chose tls roots or versions of crates which provide tls roots (e.g. `rustls-native-certs`, `tls-webpki-roots`) without tonic's releases.

These are intended to balance easiness to use, configurability, and tonic's maintenance costs.

LucioFranco:
I am okay with adding the ability to use your own CA certs but I like that idea that we provide the ability to just work with either the system or webpki root certs. I think removing this kinda makes it more tough for users. If they want more flexibility they should be using their own customer connector imo.

djc:
hyper-rustls offers high-level APIs to use rustls-platform-verifier, rustls-native-certs and webpki-roots. Maybe we can reuse that code here? (I forget if tonic can build on hyper-rustls or whether there's some reason it needs to work with tokio-rustls directly.)

LucioFranco:
probably legacy stuff, though I am inclined to not change this code that much. We have a new transport on the way anyways.

tottoto:
Instead of removing these features, added options to enable tls roots, and removes the implicit configuration of them. I think this change simplifies the implementation with keeping the user friendly features.

djc (approved):

djc:
This makes sense to me.

matze:
Would've been nice to mark this as a breaking change. Was confused that I could not connect after bumping tonic.

PR #1735, opened 2024-06-19
chore: fix merge queue

LucioFranco:


djc:
Are these code changes actually needed for the queue fix?

LucioFranco:
@djc in theory no...but I think because merge queue wasn't running something snuck in that I just merged. So this should enable CI for the merge queue and also fix the issues on master.

PR #1754, opened 2024-06-23
feat: Make boxed function public

tottoto:
Providing the way to create body gives greater flexibility. This change eliminates the need to use axum's features in the case of h2c example.

### DIFF
--- a/examples/src/h2c/server.rs
+++ b/examples/src/h2c/server.rs
@@ -4,7 +4,7 @@ use hyper_util::rt::{TokioExecutor, TokioIo};
 use hyper_util::server::conn::auto::Builder;
 use hyper_util::service::TowerToHyperService;
 use tokio::net::TcpListener;
-use tonic::{transport::Server, Request, Response, Status};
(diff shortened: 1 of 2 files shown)


examples/src/h2c/server.rs
```diff
@@ -100,8 +98,12 @@ mod h2c {
...
         }
 
-        fn call(&mut self, mut req: hyper::Request<Incoming>) -> Self::Future {
-            let svc = self.s.clone();
+        fn call(&mut self, req: hyper::Request<Incoming>) -> Self::Future {
+            let mut req = req.map(tonic::body::boxed);
+            let svc = self
+                .s
+                .clone()
+                .map_request(|req: Request<Incoming>| req.map(tonic::body::boxed));
```
djc, line 106: Does this really need to specify the type of `req`?

djc (approved):
Looks okay to me.

examples/src/h2c/server.rs
tottoto: The body type seems not to be needed. Fixed it.

PR #1757, opened 2024-06-25
Make compression encoding configuration more malleable

djc:
## Motivation

Alternative approach to #1727, enable callers to influence compression encoding priority. Also changes the default order so that `Zstd` comes before `Gzip` if enabled (which happens when the `zstd` Cargo feature is explicitly enabled).

## Solution

Implement an internal array while avoiding making changes to existing public API.

cc @CapCap

### DIFF
--- a/tonic/src/codec/compression.rs
+++ b/tonic/src/codec/compression.rs
@@ -1,5 +1,5 @@
 use crate::{metadata::MetadataValue, Status};
-use bytes::{Buf, BytesMut};
+use bytes::{Buf, BufMut, BytesMut};
 #[cfg(feature = "gzip")]
 use flate2::read::{GzDecoder, GzEncoder};
 use std::fmt;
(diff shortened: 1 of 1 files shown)


tonic/src/codec/compression.rs
```diff
@@ -10,62 +10,79 @@ pub(crate) const ENCODING_HEADER: &str = "grpc-encoding";
...
     }
 
     pub(crate) fn into_accept_encoding_header_value(self) -> Option<http::HeaderValue> {
-        match (self.is_gzip_enabled(), self.is_zstd_enabled()) {
-            (true, false) => Some(http::HeaderValue::from_static("gzip,identity")),
-            (false, true) => Some(http::HeaderValue::from_static("zstd,identity")),
-            (true, true) => Some(http::HeaderValue::from_static("gzip,zstd,identity")),
-            (false, false) => None,
+        let mut value = BytesMut::new();
+        for encoding in self.inner.iter().copied().filter_map(|e| e) {
```
tottoto, line 55: The value copying seems not to be needed. `filter_map` is used just for filtering without mapping of the values, I would think `flatten` might be idiomatic in this case.

```suggestion
        for encoding in self.inner.into_iter().flatten()
```

tonic/src/codec/compression.rs
```diff
@@ -157,8 +174,6 @@ impl CompressionEncoding {
         }
     }
 
-    #[allow(missing_docs)]
-    #[cfg(any(feature = "gzip", feature = "zstd"))]
     pub(crate) fn as_str(&self) -> &'static str {
         match self {
```
tottoto, line 178: I think derefenrencing here can solve the non-exhaustive error.

```suggestion
        match *self {
```

tottoto (commented):
Thanks! I feel the design and implementation is readable and extensible. I commented on small points.

tonic/src/codec/compression.rs
tottoto: The value copying seems not to be needed. `filter_map` is used just for filtering without mapping of the values, I would think `flatten` might be idiomatic in this case.

```suggestion
        for encoding in self.inner.into_iter().flatten() {
```

tottoto, line 55: The value copying seems not to be needed. `filter_map` is used just for filtering without mapping of the values, I would think `flatten` might be idiomatic in this case.

```suggestion
        for encoding in self.inner.into_iter().flatten() {
```

tonic/src/codec/compression.rs
spiderbiggen: wouldn't that have more to do with the `#[non_exhaustive]` on `CompressionEncoding`?

neoeinstein: The match can't see through the `&` to realize that the match is otherwise exhaustive and in the same crate as the `#[non_exhaustive]` enum is defined in. By dereferencing the `Copy` type, the exhaustiveness check can succeed.

tonic/src/codec/compression.rs
```diff
@@ -10,62 +10,79 @@ pub(crate) const ENCODING_HEADER: &str = "grpc-encoding";
...
+                Some(e) => {
+                    let removed = *e;
+                    *entry = None;
+                    return Some(removed);
+                }
+                _ => continue,
+            }
         }
+
+        None
```
tottoto, line 50: This could be written without looping.

```suggestion
        if let Some(entry) = self.inner.iter_mut().rev().find(|entry| entry.is_some()) {
            entry.take()
        } else {
            None
        }
```

tonic/src/codec/compression.rs
```diff
@@ -10,62 +10,79 @@ pub(crate) const ENCODING_HEADER: &str = "grpc-encoding";
...
-    #[cfg(not(feature = "zstd"))]
-    const fn is_zstd_enabled(&self) -> bool {
-        false
+impl Default for EnabledCompressionEncodings {
+    fn default() -> Self {
+        Self {
+            inner: [None, None],
+        }
     }
 }
```
tottoto, line 87: This could be implemented with derive macro.

tonic/src/codec/compression.rs
djc: I don't like `flatten()` in this kind of context, but took your suggestion to use `into_iter()`.

tonic/src/codec/compression.rs
djc: Changed it to

```rust
self.inner
            .iter_mut()
            .rev()
            .find(|entry| entry.is_some())?
            .take()
```

tonic/src/codec/compression.rs
djc: Done.

tonic/src/codec/compression.rs
djc: Changed the function to take `self`.

djc:
Not completely sure whether this actually solves the original issue described in #1727. IMO it would be better if the config wrapper allowed overwriting the `EnabledCompressionEncodings` instead of appending to it.

tonic/src/codec/compression.rs
```diff
@@ -157,9 +159,7 @@ impl CompressionEncoding {
         }
     }
 
-    #[allow(missing_docs)]
-    #[cfg(any(feature = "gzip", feature = "zstd"))]
-    pub(crate) fn as_str(&self) -> &'static str {
+    pub(crate) fn as_str(self) -> &'static str {
```
tottoto, line 162: I am not sure about the appropriateness of taking ownership where it is not needed.

When taking the ownership, the naming convention suggests to use `into_`.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

tottoto: Thanks. I have overlooked that it implements `Copy` trait. I am convinced with the performance aspect. Sounds reasonable to use `as_str`.

tonic/src/codec/compression.rs
tottoto: Why would not you like `flatten()` here?

My opinion is that it would be better to follow the recommendations of the Rust documentation and tooling.

The official document describes it as the typical use case.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten

And clippy warns it.

https://rust-lang.github.io/rust-clippy/master/index.html#/filter_map_identity

djc: Okay, changed it. I don't like `flatten()` because IMO it's counterintuitive to treat `Option` as an iterator.

tottoto (approved):

PR #1760, opened 2024-06-26
feat: add debug spans for decoding requests

### LINKED ISSUE #1759: Add debug spans for codec tracing
### Crates

tonic

### Motivation

We've had some clients that have, for some reason, been really slow at sending up body bytes. This would be easier to confirm if we could see that the decoding process was taking up a certain amount of time, and could be done by providing some tracing spans at the debug level that could be enabled to expose the time spend in request decoding, which often spends some time waiting for data to arrive over the wire.

### Proposal

Add some tracing debug spans for the codecs so that consumers can use tracing to identify potential stalls in decoding of a request. This could add a bit more weight to the state of the decoder, but the overall cost is generally a couple of bytes per connection, and if debug spans aren't enabled, these will always be disabled values. 

### Alternatives

The only other option has been inference, but the time spent here is difficult to separate out from other causes.

neoeinstein:
Closes: #1759

## Motivation

Adds optional debug spans into the decoder state. At the same time, it resolves a size regression in the `State` struct, which got much larger because of the inclusion of the `Status` in the error. The size change overall is from 16 bytes (v0.11.0) to 176 (master) to 56 (this PR).

## Solution

Adds tracing spans at the debug level for tracking decoding of streaming messages. This should have negligible overhead when debug spans are not enabled, and provide some more context when they are enabled.

Boxes the `Status` in the `Error` variant to avoid size bloat for the common case.

If desired, this functionality could be gated behind a feature, but I start off by avoiding the extra complexity of adding conditional compilation code.

### DIFF
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -57,6 +57,7 @@ struct StreamingInner {
     decompress_buf: BytesMut,
     encoding: Option<CompressionEncoding>,
     max_message_size: Option<usize>,
+    span: tracing::Span,
 }
 
 impl<T> Unpin for Streaming<T> {}
@@ -65,10 +66,34 @@ impl<T> Unpin for Streaming<T> {}
 enum State {
     ReadHeader,
     ReadBody {
+        span: tracing::Span,
         compression: Option<CompressionEncoding>,
         len: usize,
     },
     Error(Option<Status>),
+}
+
+impl State {
+    fn read_body(compression: Option<CompressionEncoding>, len: usize) -> Self {
+        let span = tracing::debug_span!(
+            "read_body",
+            body.compression = compression.map(|c| c.as_str()).unwrap_or("none"),
+            body.bytes.compressed = compression.is_some().then_some(len),
+            body.bytes.uncompressed = compression.is_none().then_some(len),
+        );
+        Self::ReadBody {
+            span,
+            compression,
+            len,
+        }
+    }
+
+    fn span(&self) -> Option<&tracing::Span> {
+        match self {
+            Self::ReadBody { span, .. } => Some(span),
+            Self::ReadHeader | Self::Error(_) => None,
+        }
+    }
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -163,6 +188,7 @@ impl<T> Streaming<T> {
                 decompress_buf: BytesMut::new(),
                 encoding,
                 max_message_size,
+                span: tracing::debug_span!("streaming"),
             },
         }
     }
@@ -173,6 +199,7 @@ impl StreamingInner {
         &mut self,
         buffer_settings: BufferSettings,
     ) -> Result<Option<DecodeBuf<'_>>, Status> {
+        let _guard = self.span.enter();
         if let State::ReadHeader = self.state {
             if self.buf.remaining() < HEADER_SIZE {
                 return Ok(None);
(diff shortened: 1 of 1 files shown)


tonic/src/codec/decode.rs
```diff
@@ -142,7 +173,19 @@ impl StreamingInner {
...
+                return Ok(None);
+            }
+
+            let span = span.get_or_insert_with(|| {
+                tracing::debug_span!(
+                    "read_header",
+                    compression = tracing::field::Empty,
+                    body.bytes = tracing::field::Empty,
+                )
+            });
```
neoeinstein, line 187: This avoids creating a vestigial `read_header` span for unary endpoints, which have only a single message. This does mean that there may be a small gap if there isn't any pending header data for a streaming endpoint until the first bytes of the next header are available in the buffer.

tonic/src/codec/decode.rs
```diff
@@ -38,12 +38,43 @@ impl<T> Unpin for Streaming<T> {}
...
+    ReadHeader {
+        span: Option<tracing::Span>,
+    },
     ReadBody {
+        span: tracing::Span,
         compression: Option<CompressionEncoding>,
         len: usize,
     },
-    Error(Status),
+    Error(Box<Status>),
```
neoeinstein, line 49: Boxing here significantly reduces the size of `State` and `StreamingInner` as a result, even with the addition of `tracing::Span` to the other variants.

neoeinstein:
I wonder if there's value in more explicit boxing of `Status`, or if `Status` should internally box in order to reduce its size.

tonic/src/codec/decode.rs
```diff
@@ -142,7 +173,19 @@ impl StreamingInner {
         &mut self,
         buffer_settings: BufferSettings,
     ) -> Result<Option<DecodeBuf<'_>>, Status> {
-        if let State::ReadHeader = self.state {
+        if let State::ReadHeader { span } = &mut self.state {
+            if !self.buf.has_remaining() {
```
djc, line 177: Why add this guard here? Can we just do the `self.buf.remaining() < HEADER_SIZE` check here instead?

djc (approved):

djc:
This seems reasonable to me, thanks! (Sorry for the delay in reviewing.)

> I wonder if there's value in more explicit boxing of `Status`, or if `Status` should internally box in order to reduce its size.

Could make sense to evaluate this next, in a separate PR? How/why is the size of `Status` causing issues for you?

tonic/src/codec/decode.rs
marcus-griep-simplisafe: I had considered that. The reason for the guard at all is that, even with unary requests, Tonic still polls the stream after reading the first message. That was generating `read_header`, `read_body`, `read_header` for these unary requests, with the last `read_header` really being vestigial, since the stream would then be dropped after the message. I opted to try to remove this vestigial span, but we could simplify by re-introducing it.

With that in mind, if there are still bytes in the stream, then that would indicate that there's a header already partially in the input, so opening a new span is indicated. If not, then we'd delay opening a new span (but that also means that a new span wouldn't capture the time from the empty poll until the next header started appearing in the buffer).

Changing it so that the span is only opened once 5 bytes are in the buffer would mean that the span would only open once we've already completely read the header, making this really just timing the extraction of the `u8` compression encoding and `u32` payload size, which probably then isn't all that valuable.

Another option here is to have a debug span one level higher for the entire stream itself, skip the span for the header, and keep a span for the body.

marcus-griep-simplisafe (commented):
> Could make sense to evaluate this next, in a separate PR? How/why is the size of `Status` causing issues for you? 

I'm not opposed to taking this out. There's probably not a lot of issue here, other than the fact that it now requires more data move instructions to copy up the `Result` on every return. Since this is in the hot-path for deserialization, and `Status` in this case is an error condition, optimizing the size of the non-error case seemed appropriate, but I don't have any data that indicates it is required.

nathanielford:
I've pushed a merge of the master branch changes since this was opened, and I think also addressed the requested items from the PR here. @LucioFranco do you mind doing a quick review? 

(Just moving through the backlog looking for likely things to clean up and close out.)

PR #1761, opened 2024-06-27
Apply clippy suggestions

djc:


tonic/src/transport/channel/service/add_origin.rs
```diff
@@ -54,8 +54,8 @@ where
             // Split the request URI into parts.
             let mut uri: http::uri::Parts = head.uri.into();
             // Update the URI parts, setting hte scheme and authority
-            uri.scheme = self.scheme.clone();
-            uri.authority = self.authority.clone();
```
tottoto, line 58: I think these fixes are good. 

As `assigning_clones` rule will be moved to `pedantic` and is scheduled to be allowed by default, when we adopt this rule, setting it as `warn` might be desirable.

https://github.com/rust-lang/rust-clippy/pull/12779

tottoto:
Regarding to `default_constructed_unit_structs` rule, some of them seem not to be detected, but I am not sure the reason.

And I think we have other option to fix the warning, by making these types non-exhaustive. How do you think? I would think both of them are reasonable.

djc:
> And I think we have other option to fix the warning, by making these types non-exhaustive. How do you think? I would think both of them are reasonable.

Given that all of these seem to be in examples or test code, I don't think making these non-exhaustive is worth it.

tonic/src/transport/channel/service/add_origin.rs
```diff
@@ -54,8 +54,8 @@ where
...
+            uri.scheme.clone_from(&self.scheme);
+            uri.authority.clone_from(&self.authority);
```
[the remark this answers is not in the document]
tottoto: What I wrote setting it as `warn` means setting `assigning_clones` as `warn` explicitly in crate wide config preparing for the future consistency as it will be allowed as default and will not be detected in the future.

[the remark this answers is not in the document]
tottoto: What I wrote setting it as `warn` was intended to mean setting `assigning_clones` as `warn` explicitly in crate wide config preparing for the future consistency as it will be allowed as default and will not be detected in the future.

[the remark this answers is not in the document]
djc: I understood what you meant and disagreed -- I was convinced by the reasoning upstream (which applies here, too) that there's often not an actual performance benefit and it's harder to understand. So alternatively we could maybe add a local `#[allow()]` here.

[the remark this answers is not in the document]
tottoto: What I have said assumes that you suggest to apply this change since you are proposing it in this pull request. I am not proposing adopting this rule. What I mean is just it seems no problem to me and if we adopt this change as you are proposing here, I think that we should maintain consistency in applying the rule. From your description it seems like you are saying that this change should not be applied, but that seems to contradict what is suggested actually in the code. Could you please clarify what you are proposing?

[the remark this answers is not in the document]
djc: I've rolled back this change and added a crate-level `allow(clippy::assigning_clones)`.

examples/src/multiplex/server.rs
```diff
@@ -23,7 +23,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let addr = "[::1]:50051".parse().unwrap();
 
     let greeter = GreeterServer::new(MyGreeter::default());
-    let echo = EchoServer::new(MyEcho::default());
+    let echo = EchoServer::new(MyEcho);
```
tottoto, line 26: For some reason not all `default_constructed_unit_structs` targets are detected, resulting in partial changing and a lack of consistency throughout the code. If it seems difficult to resolve, I think we can allow this rule for the examples as the crate level setting.

tottoto, line 26: It seems some supposed `default_constructed_unit_structs` targets are not detected, resulting in partial changing and a lack of consistency throughout the code. If it seems difficult to resolve, I think we can allow this rule for the examples as the crate level setting.

djc: What targets aren't detected? Personally I prefer being a little consistent about this over disabling a lint that we think points in the right direction.

tottoto: > What targets aren't detected?

The commented place here is one of them. `MyGreeter` is not detected.

I figured out the difference. The detected structs are defined like `struct MyEcho;` but not detected structs are done such as `struct MyGreeter {}` with empty brace style. I prefer using default trait to construct them, replacing the style with the empty brace `{}` one as the struct definitions, since it explicitly expresses constructing the value in the crates which do not effect on the efficiency.

tottoto: > What targets aren't detected?

The commented place here is one of them. `MyGreeter` is not detected.

I figured out the difference. The detected structs are defined like `struct MyEcho;` but not detected structs are done such as `struct MyGreeter {}` with empty brace style. I prefer using default trait to construct them in the places not affecting the efficiency, replacing the style with the empty brace `{}` one as the struct definitions, since it explicitly expresses constructing the value.

djc: Given that there is an actual difference, IMO further changes are off-topic for this PR.

tottoto: I do not agree just applying this clippy suggestion with introducing code inconsistency. When these changes are reverted, I agree merging this.

tottoto: It does not seem off-topic to me as the issues this change causes are relevant, and I do not agree just applying this clippy suggestion with introducing code inconsistency. When these changes are reverted, I agree merging this.

PR #1765, opened 2024-07-01
chore(codec): Use as_str method to implement display for CompressionEncoding

tottoto:
Reduces duplicate similar implementations for the similar purpose.


djc:
Thanks for these PRs! Would have welcomed these suggestions as feedback on my original PR, too.

PR #1768, opened 2024-07-01
feat(codec): Make error when not utf8 value in compression encoding

tottoto:
Makes error in the case when not utf8 value is in the compression encoding.

### DIFF
--- a/tonic/src/codec/compression.rs
+++ b/tonic/src/codec/compression.rs
@@ -129,26 +129,30 @@ impl CompressionEncoding {
             return Ok(None);
         };
 
-        let header_value_str = if let Ok(value) = header_value.to_str() {
-            value
-        } else {
-            return Ok(None);
-        };
-
-        match header_value_str {
+        match header_value.as_bytes() {
             #[cfg(feature = "gzip")]
-            "gzip" if enabled_encodings.is_enabled(CompressionEncoding::Gzip) => {
+            b"gzip" if enabled_encodings.is_enabled(CompressionEncoding::Gzip) => {
                 Ok(Some(CompressionEncoding::Gzip))
             }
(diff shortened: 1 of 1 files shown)


djc (approved):
Good idea!

tonic/src/codec/compression.rs
```diff
@@ -129,26 +129,31 @@ impl CompressionEncoding {
...
+                    Err(_) => {
+                        other_str_tmp = format!("{:?}", other);
+                        &other_str_tmp
+                    }
+                };
+
                 let mut status = Status::unimplemented(format!(
                     "Content is compressed with `{}` which isn't supported",
-                    other
+                    other_str,
```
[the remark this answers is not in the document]
tottoto: Updated.

tonic/src/codec/compression.rs
```diff
@@ -129,26 +129,30 @@ impl CompressionEncoding {
...
+                // https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html#extending-automatic-temporary-lifetime-extension
+                let other_debug_string;
+
                 let mut status = Status::unimplemented(format!(
                     "Content is compressed with `{}` which isn't supported",
-                    other
+                    match std::str::from_utf8(other) {
+                        Ok(s) => s,
+                        Err(_) => {
+                            other_debug_string = format!("{:?}", other);
```
djc, line 152: I actually meant just inlining the `other_str` in the string, but this works, too!

Just now noticing we could also inline the `other` argument, here.

djc (approved):

tonic/src/codec/compression.rs
tottoto: Updated.

djc (approved):

PR #1772, opened 2024-07-06
feat: Move Request server util method to RequestExt

tottoto:
## Motivation

These utility functionalities seem to be more practical APIs than the other basic ones. Moving them to an extension trait can keep the request implementation more basic. The drawback is that users need to import the trait to use these functionalities.

## Solution

Adds an extension trait for request and moves these functionalities to the trait.

### DIFF
--- a/examples/src/autoreload/server.rs
+++ b/examples/src/autoreload/server.rs
@@ -1,4 +1,7 @@
-use tonic::{transport::Server, Request, Response, Status};
+use tonic::{
+    transport::{RequestExt, Server},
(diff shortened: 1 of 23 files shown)


djc:
This doesn't seem like an improvement to me.

LucioFranco:
I don't think we want to take this path right now, I think this adds too much complexity for the added benefit. So for now and the interest of the direction of where the project is going I am going to close this. Once, we start merging in the new foundational code we can reconsider this in the larger group setting.

PR #1778, opened 2024-07-08
Keep TLS transport items together in a single module

djc:
Counterproposal to #1763.


tottoto:
The modules are already nicely separated correspoinding to their roles, this change feels like it breaks that.

djc:
> The modules are already nicely separated correspoinding to their roles, this change feels like it breaks that.

What's the difference in role between `transport::service::tls` and `transport::tls`? Having those separate doesn't really make sense to me either conceptually or in terms of complexity (both are very small modules, so having them in separate files just makes you have to bounce between modules to make sense of what's going on).

tottoto:
`transport::service::tls` module is the utilities used by the channel and server service modules and `transport::tls` is used by the transport wide implementations related to tls.

LucioFranco:
@djc I am not totally following the motivation for this? Could you expand more and if its a counter proposal to the other PR it would be good to get a summary of why its a counter proposal.

djc:
> @djc I am not totally following the motivation for this? Could you expand more and if its a counter proposal to the other PR it would be good to get a summary of why its a counter proposal.

The main difference IMO is that this PR merges more of the TLS infrastructure (which is currently kind of spread over 4 different `tls.rs` modules) to live in a single module, and to expose a smaller interface between that module (`transport::tls`) and the modules that depend on it.

LucioFranco (commented):
I think I am more infavor of https://github.com/hyperium/tonic/pull/1763 due to two things 1) its a smaller change that does less and 2) I kinda agree with @tottoto that the way things are laid out is done on purpose. To me the main transport::tls file is designed to only contain the public tls infrastructure while the rest is designed to contain the internal transport things like the tls error for example.

PR #1785, opened 2024-07-09
build: add removed `prost_build::{Service, Method}` trait implementations

mbrobbel:
## Motivation

#1599 removed the `tonic_build::Service` and `tonic_build::Method` 
implementations for `prost_build::Service` and `prost_build::Method`.
These impls are used in https://github.com/neoeinstein/protoc-gen-prost.

## Solution

This PR re-adds the implementations that were removed in #1599.

### DIFF
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -62,8 +62,90 @@ pub fn compile_protos(proto: impl AsRef<Path>) -> io::Result<()> {
     Ok(())
 }
 
+const PROST_CODEC_PATH: &str = "tonic::codec::ProstCodec";
+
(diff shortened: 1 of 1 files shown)


djc (approved):
This looks reasonable to me. @LucioFranco thoughts?

neoeinstein:
It's possible that this may not be required _inside_ of Tonic, and we can instead use the new-type pattern inside of `protoc-gen-tonic` to do this ourselves instead. I'll take a look, and if so, that will encourage the pattern of having implementers provide their own values.

mbrobbel:
Closing as this is indeed not needed.

PR #1788, opened 2024-07-10
 chore(server): Vendor hyper-util graceful shutdown feature

tottoto:
Uses `hyper_util::server::graceful::GracefulShutdown` to implement graceful shutdown.


djc:
Why/how is this an improvement?

tottoto:
Centralizing the graceful shutdown implementation and using it at server implementation can split their responsibility and improve the maintainability.

djc:
But this is duplicating the entire thing... that doesn't seem like centralizing in a meaningful way.

PR #1790, opened 2024-07-11
Update Cargo.toml

ExHuzaifa:
Changes made:

Removed unnecessary triple quotes around the description. Indented the features and dependencies sections for better readability. Removed unnecessary whitespace between sections.
Kept the same formatting for the dependencies and dev-dependencies sections for consistency.

### DIFF
--- a/tonic-health/Cargo.toml
+++ b/tonic-health/Cargo.toml
@@ -1,9 +1,7 @@
 [package]
 authors = ["James Nugent <james@jen20.com>"]
 categories = ["network-programming", "asynchronous"]
-description = """
-Health Checking module of `tonic` gRPC implementation.
-"""
+description = "Health Checking module of `tonic` gRPC implementation."
 documentation = "https://docs.rs/tonic-health/0.12.0"
 edition = "2021"
 homepage = "https://github.com/hyperium/tonic"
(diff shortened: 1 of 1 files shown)


tonic-health/Cargo.toml
```diff
@@ -21,11 +19,16 @@ transport = []
 [dependencies]
 async-stream = "0.3"
 prost = "0.13"
-tokio = {version = "1.0", features = ["sync"]}
+tokio = { version = "1.0", features = ["sync"] }
```
djc, line 22: While I like this change and believe it is broadly idiomatic in matching the ecosystem at large, I believe the other `Cargo.toml` in this repo do not follow this style. If we're going to do this, I'd prefer to do it in a separate PR (or a separate commit) that does this consistently across all `Cargo.toml` files in the repo.

tonic-health/Cargo.toml
```diff
@@ -21,11 +19,16 @@ transport = []
...
 tokio-stream = "0.1"
-tonic = { version = "0.12", path = "../tonic", default-features = false, features = ["codegen", "prost"] }
+tonic = {
```
djc, line 24: On the other hand, not a big fan of this. IMO keeping it on one line makes it the list of dependencies easier to comprehend.

PR #1802, opened 2024-07-15
tonic-reflection: Restructure module to prep reintroducing v1alpha

ttkjesper:
To simplify API verification, this is a prepatory change to shift around the v1 support to allow for a cleaner v1alpha reintroduction.

* Created a multi-file version of the server.rs module
  * Added mod.rs containing the base `Error` with associated imports / exports
  * Added v1.rs containing the v1 version of the reflection server

## Motivation

The current `v1` reflection API support is lacking in various existing tools (like *Postman* and *Kreya*).

From feedback in https://github.com/hyperium/tonic/pull/1787, we should split the change into multiple PRs.

## Solution

This change splits the server module to enable us to re-add the `v1alpha` support in a later PR a cleaner fashion.

### DIFF
--- a/tonic-reflection/src/server/mod.rs
+++ b/tonic-reflection/src/server/mod.rs
@@ -1,0 +1,36 @@
+pub use crate::pb::v1::server_reflection_server::{ServerReflection, ServerReflectionServer};
+
+use prost::DecodeError;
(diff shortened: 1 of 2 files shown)


djc (approved):
Looks good to me, thanks!

PR #1804, opened 2024-07-16
Naming collisions with std::marker::Sync/Send

LockedThread:
## Motivation

Please see this issue: https://github.com/hyperium/tonic/issues/1803.

TLDR: Unable to name proto service Sync or Send due to naming conflict with [Send](https://doc.rust-lang.org/std/marker/trait.Send.html) and [Sync](https://doc.rust-lang.org/std/marker/trait.Sync.html).

## Solution

Instead of using the Sync and Send traits directly, I wrote out their imports in the codegen.

## Example Use Case

Checkout my minimal example with fix [here](https://github.com/LockedThread/tonic_example_broken_project).


djc (approved):

LockedThread:
Thanks @djc, looks like the codegen CI failed, as expected. The failed tests seem safe to ignore.

djc:
> Thanks @djc, looks like the codegen CI failed, as expected. The failed tests seem safe to ignore.

Uhh, I'm guessing you need to run codegen so the tests start passing again?

LockedThread:
> > Thanks @djc, looks like the codegen CI failed, as expected. The failed tests seem safe to ignore.
> 
> Uhh, I'm guessing you need to run codegen and include the changes in this PR so the tests start passing again?

Let's see what happens, just pushed changes.

LockedThread:
I am not sure how to fix that test. @djc Can you assist?

tottoto:
It can be caused by generating file descriptor set with protoc of which version is different from the one used in the test.

LockedThread:
> It can be caused by generating file descriptor set with protoc of which version is different from the one used in the test.

Okay, thanks. Which version do y'all use?

djc:
CI seems to be configured to use 3.25.1 (see `PROTOC_VERSION`).

LockedThread:
> CI seems to be configured to use 3.25.1 (see `PROTOC_VERSION`).

Thanks, I will re-run the tests later today.

LockedThread:
```
vscode ➜ /workspaces/tonic (fix/std-naming-collision) $ protoc --version
libprotoc 25.1
```
Pushed update with new bin file

LockedThread:
@djc Tests passed :partying_face:, thanks for the help!

djc (approved):

PR #1821, opened 2024-07-25
fix(tonic-build): cargo build script outputs are not functional

d3rpp:
## Motivation

I was looking through and debugging some services and noticed that the `println!()` statements in the `tonic-build` crate were not correctly formatted.

Example 1: https://github.com/hyperium/tonic/blob/255a885d316392a0135fa810ecfd5ae87135a329/tonic-build/src/prost.rs#L651

Example 2:
https://github.com/hyperium/tonic/blob/255a885d316392a0135fa810ecfd5ae87135a329/tonic-build/src/prost.rs#L658

They all have a single `:` when they should have 2 as shown in the cargo book:

https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed

## Solution

Just added the extra `:`, did a full workspace search to ensure I got all of them :+1:

---

If there's anything else you'd like me to do to properly follow any requirements please let met know :smile:

### DIFF
--- a/interop/build.rs
+++ b/interop/build.rs
@@ -4,5 +4,5 @@ fn main() {
     tonic_build::compile_protos(proto).unwrap();
 
     // prevent needing to rebuild if files (or deps) haven't changed
-    println!("cargo:rerun-if-changed={}", proto);
+    println!("cargo::rerun-if-changed={}", proto);
 }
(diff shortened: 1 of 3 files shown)


djc (approved):
Thanks!

PR #1822, opened 2024-07-26
tonic-reflection: Refactor descriptor parser to be separate

ttkjesper:
As part of reintroducing the v1alpha reflection support, this extracts the parser into a separate module.

## Motivation

If we were to naively reintroduce the v1alpha support, there would be a lot of code duplication.

## Solution

The parser is extracted into a separate module that can be shared between both protocols.


djc:
This should be superseded by #1888. @ttkjesper please close if you find #1888 satisfactory.

PR #1823, opened 2024-07-26
streaming: terminate stream after an error has been returned once

bmwill:
PR #1682 introduced a change to return the error Status to the user of Streaming but resulted in a change in behavior such that if an error occurs the Stream will never end and instead always yeild the error. This patch adjusts the semantics to only return the error one time and then terminates the stream by returning Poll::Ready(None) on subseqent calls.

### DIFF
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -43,7 +43,7 @@ enum State {
         compression: Option<CompressionEncoding>,
         len: usize,
     },
-    Error(Status),
+    Error(Option<Status>),
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


tonic/src/codec/decode.rs
```diff
@@ -393,8 +393,8 @@ impl<T> Stream for Streaming<T> {
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         loop {
-            if let State::Error(status) = &self.inner.state {
-                return Poll::Ready(Some(Err(status.clone())));
+            if let State::Error(status) = &mut self.inner.state {
+                return Poll::Ready(status.take().map(Err));
```
djc, line 397: Maybe add a short comment here on what the desired behavior is?

djc (approved):

tonic/src/codec/decode.rs
bmwill: of course!

djc (approved):
Thanks, LGTM!

Issue #1825, opened 2024-07-26
Make the TimeoutExpired a public tonic type rather than a transport type


FrankReh:
## Feature Request

The TimeoutExpired type is trivial and carries no identifying information other than its type. It seems to have been introduced before the transport module could be carved out of the tonic build. But as it stands, the tonic error handling cannot report a GRPC timeout from the server when the crate's user has proved their own transport.

It may or may not be considered a breaking change but the error could be a tonic::TimeoutExpired type which its own error handling can test for (attempt to downcast to) regardless of the transport/server feature having been enabled, and the builtin transport could instantiate it in its Future implementation and other versions of transport could do the same.

### Crates

tonic

### Motivation

There is a test, 'picks_server_timeout_if_thats_sorter', that could then pass, even when a different transport crate was used. And as the test is designed to show, it would allow a client to specify a timeout on the server side, and still be able to differentiate the error from other h2 internal errors.

### Proposal

Move the few lines that define the type to either a new tonic module or the existing 'error' tonic module.

### Alternatives

Leave the status quo, where the builtin transport has better support from tonic.

PR #1826, opened 2024-07-27
chore(tonic): Move TimeoutExpired out of transport

FrankReh:
By moving the TimeoutExpired type from the transport module to the the tonic engine proper, the type is made available to other transport implementations and tonic can be built without the "server" feature while still returning a Status indicating the timeout.

Alternative transport implementations can use the type, now that it is moved, when they want the tonic Status error handling to recognize the error as having been triggered by a timeout in the transport logic. The tonic Status error will have a code of Cancelled with a message of "Timeout expired".

There is already a test for this:

    cargo test picks_server_timeout_if_thats_sorter

which worked the original way and contines to work; but now a new transport implementation can get the same behavior.

Addresses #1825.

### DIFF
--- a/tonic/src/lib.rs
+++ b/tonic/src/lib.rs
@@ -120,7 +120,7 @@ pub use extensions::GrpcMethod;
 pub use http::Extensions;
 pub use request::{IntoRequest, IntoStreamingRequest, Request};
 pub use response::Response;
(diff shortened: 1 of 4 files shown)


djc (approved):
Makes sense, thanks!

PR #1828, opened 2024-07-28
chore(tonic): Move ConnectError type from transport

FrankReh:
Move the ConnectError type from transport to tonic.

## Motivation

Making `ConnectError` a tonic type allows the tonic Status to convert a transport generated connection error into a Status::code() of Status::unavailable. As the comment in status.rs already says, this is consistent with the behavior of a C++ gRPC client when the server is not running, and matches the spec.

The only other error that results in a Status::unavailable is a timeout error when the crate is built with its 'transport' module, allowing it to recognize when an error is a hyper timeout error. But a possible loosening of that relationship is not being addressed here.

## Solution

Move the type from the transport module, where it can only be used by the Status code when tonic is compiled with its own transport, to the status code in tonic itself, so the built in transport can use it and proprietary transports can use it.

There are already two tests that look for this Status::unavailable result. They passed before and they pass with this change as well. (`connect_lazy_reconnects_after_first_failure` and `connect_returns_err_via_call_after_connected`)

### DIFF
--- a/tonic/src/lib.rs
+++ b/tonic/src/lib.rs
@@ -120,7 +120,7 @@ pub use extensions::GrpcMethod;
 pub use http::Extensions;
 pub use request::{IntoRequest, IntoStreamingRequest, Request};
 pub use response::Response;
-pub use status::{Code, Status};
+pub use status::{Code, Status, TimeoutExpired};
 
 pub(crate) type Error = Box<dyn std::error::Error + Send + Sync>;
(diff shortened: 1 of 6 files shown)


djc (approved):
LGTM. @tottoto any thoughts?

tottoto (approved):

tottoto:
Looks good to me.

PR #1830, opened 2024-08-01
chore: Update to EmbarkStudios/cargo-deny-action@v2

tottoto:
Allows `workspace-duplicates` which is introduced at `cargo-deny` [0.15](https://github.com/EmbarkStudios/cargo-deny/releases/tag/0.15.0), and `BSD-2-Clause` which is newly pulled in dependencies.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: EmbarkStudios/cargo-deny-action@v1
+    - uses: EmbarkStudios/cargo-deny-action@v2
 
   codegen:
     runs-on: ubuntu-latest


deny.toml
```diff
@@ -3,6 +3,7 @@ all-features = true
 
 [bans]
 multiple-versions = "deny"
+workspace-duplicates = "allow"
```
djc, line 6: What is this for?

tottoto: cargo-deny started to detect the same crates are used in the workspace but they are not defined by using workspace inheritance from version 0.15. In tonic project, a lot of crates are detected.

djc:
> 0.15 seems not to be released at cargo-deny-action.

Filed https://github.com/EmbarkStudios/cargo-deny-action/issues/79.

tottoto:
The license part is resolved at #1831.

tottoto:
cargo-deny 0.16 is released and it is released at cargo-deny-action 2. In this version, the workspace dependencies are changed not to be checked as default.

https://github.com/EmbarkStudios/cargo-deny/releases/tag/0.16.0

Issue #1834, opened 2024-08-01
long headers from the server hang tonic


ajwerner:
### Version

0.12.1

### Platform
`Linux pc 6.5.0-42-generic #42-Ubuntu SMP PREEMPT_DYNAMIC Mon Jun 10 09:28:55 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

### Description

Long error messages corrupt h2 streams in some way. I noticed this primarily because the upgrade in 9c1f2f9402d97cfa4c9d9065b5af2fb99d2ef521 / https://github.com/hyperium/tonic/pull/1670 changed the default maximum header size from 16MiB to 16KiB. For bidirectional streaming, I first get the error message, `"h2 protocol error: http2 error"`, but later streams on that connection seem somewhat busted. For unary RPCs, it seems that even the first request just hangs as opposed to the client getting an error. When I've debugged this, I've found that the client resets the stream, but then things just hang.

I've written a simple test that just hangs to demonstrate the issue:

`tests/integration_tests/tests/status.rs`
```rust
#[tokio::test]
async fn long_error_message() {
    struct Svc;

    const N: usize = 100_000;

    #[tonic::async_trait]
    impl test_server::Test for Svc {
        async fn unary_call(&self, _: Request<Input>) -> Result<Response<Output>, Status> {
            Err(Status::internal("a".repeat(N)))
        }
    }

    let svc = test_server::TestServer::new(Svc);

    let (tx, rx) = oneshot::channel::<()>();

    let jh = tokio::spawn(async move {
        Server::builder()
            .add_service(svc)
            .serve_with_shutdown("127.0.0.1:1340".parse().unwrap(), async { drop(rx.await) })
            .await
            .unwrap();
    });

    tokio::time::sleep(Duration::from_millis(100)).await;

    let mut channel = test_client::TestClient::connect("http://127.0.0.1:1340")
        .await
        .unwrap();

    // !!! This call never finishes!
    let err = channel
        .unary_call(Request::new(Input {}))
        .await
        .unwrap_err();

    assert_eq!(err.message(), "a".repeat(N));

    tx.send(()).unwrap();

    jh.await.unwrap();
}
```

ajwerner:
On some level, the problem happens in h2:
```
[DEBUG h2::proto::streams::recv] stream error REQUEST_HEADER_FIELDS_TOO_LARGE -- recv_headers: frame is over size; stream=StreamId(1)
[...]

PR #1835, opened 2024-08-02
transport: allow setting max_header_list_size

ajwerner:
## Motivation

There is a bug such that if the client sends a response with a header value that exceeds the max_header_list_size, then RPCs just hang (#1834). When tonic upgraded to hyper 1, it picked up [hyper#3622] which changed the default from 16MiB to 16KiB for this configuration value. Error messages in gRPC use headers. That means that services which ever sent error messages in excess of 16KiB (including in their error details!) will just hang.

## Solution

This commit adds the ability for the client to configure this value to something larger (perhaps the old default of 16MiB) to mitigate the above-referenced bug. 

[hyper#3622]: https://github.com/hyperium/hyper/pull/3622

### DIFF
--- a/tests/integration_tests/tests/http2_max_header_list_size.rs
+++ b/tests/integration_tests/tests/http2_max_header_list_size.rs
@@ -1,0 +1,70 @@
+use std::time::Duration;
+
+use integration_tests::pb::{test_client, test_server, Input, Output};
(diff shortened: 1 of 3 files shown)


tottoto (approved):
Sounds reasonable to me.

PR #1852, opened 2024-08-04
chore(examples): Use tonic::Result util

tottoto:
Uses `tonic::Result` util in examples.


djc:
I'd prefer to avoid the use of `Result` aliases, especially in example code where it can be helpful to explicitly see the error type being used. I think the changes here are an improvement for the cases where a "custom" `Result` alias is being replaced with tonic's, but I'm not a fan of the introduction of the use of `tonic::Result` in favor of `std::Result` with an explicit error type.

tottoto:
Makes sense. I understand the benefit.

PR #1855, opened 2024-08-06
feat(router): Add RoutesBuilder constructor

tottoto:
Adds `RoutesBuilder` constructor.


tonic/src/service/router.rs
```diff
@@ -27,6 +27,11 @@ pub struct RoutesBuilder {
 }
 
 impl RoutesBuilder {
+    /// Create a new empty builder.
+    pub fn empty() -> Self {
```
djc, line 31: I don't see much value in this, and it's definitely not necessary for the `builder()` method.

tottoto: Yes. In terms of whether this is necessary for a `Routes::builder` method, the answer is no, but it is there to provide a constructor that indicates that intention, and is not provided for implementing a `Routes::builder` method.

tottoto: In terms of whether this is necessary for a Routes::builder method, it is not, but it is there to provide a constructor that indicates that intention, and is not provided for implementing a Routes::builder method.

tottoto: Removed it as the preferred style for creating builder is `Routes::builder`.

djc (approved):

PR #1857, opened 2024-08-06
chore(examples): Call Routes::prepare

tottoto:
Calls `Routes::prepare` in examples.

### DIFF
--- a/examples/src/h2c/server.rs
+++ b/examples/src/h2c/server.rs
@@ -39,7 +39,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("GreeterServer listening on {}", addr);
 
     let incoming = TcpListener::bind(addr).await?;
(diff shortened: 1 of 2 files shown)


djc:
Why? What does it do/why is it valuable? If it works without it, it looks like pure noise.

tottoto:
No, it is not noise. This is recommended because it uses `axum`'s router internally. Examples should show how it should be used.

djc:
You haven't answered my question. I mean "noise" not in a disparaging but in a technical sense, as in, code that is unnecessary (since it apparently compiles just fine without it). So why is it better to call `prepare()` on these?

djc:
Okay, can you link that recommendation here? Essentially I'm asking why Axum recommends this.

tottoto:
https://docs.rs/axum/0.7.5/axum/routing/struct.Router.html#a-note-about-performance

djc (approved):

PR #1864, opened 2024-08-08
feat(router): Add missing unimplemented fallback to RoutesBuilder

tottoto:
Adds missing unimplemented fallback routing when constructing `Routes` via `Routes::default` or `RoutesBuilder`.

### DIFF
--- a/tonic/src/service/router.rs
+++ b/tonic/src/service/router.rs
@@ -14,7 +14,7 @@ use std::{
 use tower::{Service, ServiceExt};
 
 /// A [`Service`] router.
-#[derive(Debug, Default, Clone)]
+#[derive(Debug, Clone)]
 pub struct Routes {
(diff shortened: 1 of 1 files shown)


tonic/src/service/router.rs
```diff
@@ -47,6 +47,14 @@ impl RoutesBuilder {
         self.routes.unwrap_or_default()
     }
 }
+
+impl Default for Routes {
+    fn default() -> Self {
+        let router = axum::Router::new().fallback(unimplemented);
+        Self { router }
```
djc, line 54: Suggest inlining the `router` binding directly.

djc (approved):

tonic/src/service/router.rs
tottoto: Updated.

PR #1865, opened 2024-08-08
Added support for grpc max_connection_age_ms

ionut-slaveanu:
## Motivation
See:

- https://github.com/hyperium/tonic/issues/1428

## Related PRs:

- https://github.com/hyperium/hyper/pull/3729

## Solution
Added a new configuration parameter: max_connection_age_ms.
Example:
 
Server::builder()
        .max_connection_age_ms(5)
        .add_service(GreeterServer::new(greeter))
        .serve(addr)
        .await?;

I'm using max_connection_age_ms set to 5 ms only for test purposes. In production it should be set to a greater value.

See:

- https://github.com/ionut-slaveanu/tonic/tree/master/examples/src/max_connection_age

We try to gracefully close each connection to the server after max_connection_age_ms milliseconds have passed.

## Testing
Run:
./target/release/max-connection-age-server as grpc server.
As a client, i'm using ghz: 

- https://ghz.sh/ 

Example of generating requests with:
./ghz --insecure   --proto /opt/work3/tonic/examples/proto/helloworld/helloworld.proto --call helloworld.Greeter.SayHello  -d '{"name":"Tonic"}' [::1]:50051 -n 100000 --connections=500 --cpus=16 --concurrency=500

Summary:
  Count:	100184
  Total:	2.62 s
  Slowest:	1.11 s
  Fastest:	0.04 ms
  Average:	8.30 ms
  Requests/sec:	38309.90

Response time histogram:
  0.038    [1]     |
  111.425  [99908] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  222.813  [26]    |
  334.201  [0]     |
  445.588  [0]     |
  556.976  [0]     |
  668.363  [0]     |
  779.751  [0]     |
  891.139  [0]     |
  1002.526 [0]     |
  1113.914 [65]    |

Latency distribution:
  10 % in 0.38 ms 
  25 % in 0.88 ms 
  50 % in 2.09 ms 
  75 % in 8.35 ms 
  90 % in 25.62 ms 
  95 % in 34.60 ms 
  99 % in 55.86 ms 

Status code distribution:
  [OK]            100000 responses   
  [Unavailable]   184 responses      

Error distribution:
  [91]   rpc error: code = Unavailable desc = the connection is draining   
  [93]   rpc error: code = Unavailable desc = transport is closing

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -45,6 +45,7 @@ use http::{Request, Response};
 use http_body_util::BodyExt;
 use hyper::{body::Incoming, service::Service as HyperService};
 use pin_project::pin_project;
+use std::future::pending;
 use std::{
     convert::Infallible,
     fmt,
@@ -57,6 +58,7 @@ use std::{
     time::Duration,
 };
 use tokio::io::{AsyncRead, AsyncWrite};
+use tokio::time::sleep;
 use tokio_stream::Stream;
 use tower::{
     layer::util::{Identity, Stack},
@@ -99,6 +101,7 @@ pub struct Server<L = Identity> {
     max_frame_size: Option<u32>,
     accept_http1: bool,
     service_builder: ServiceBuilder<L>,
+    max_connection_age: Option<Duration>,
 }
 
 impl Default for Server<Identity> {
@@ -122,6 +125,7 @@ impl Default for Server<Identity> {
             max_frame_size: None,
             accept_http1: false,
             service_builder: Default::default(),
+            max_connection_age: None,
         }
     }
 }
@@ -230,6 +234,27 @@ impl<L> Server<L> {
     pub fn max_concurrent_streams(self, max: impl Into<Option<u32>>) -> Self {
         Server {
             max_concurrent_streams: max.into(),
+            ..self
+        }
+    }
+
+    /// Sets the maximum time option in milliseconds that a connection may exist
+    ///
+    /// Default is no limit (`None`).
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use tonic::transport::Server;
+    /// # use tower_service::Service;
+    /// # use std::time::Duration;
+    /// # let builder = Server::builder();
+    /// builder.max_connection_age(Duration::from_secs(60));
+    /// ```
+    #[must_use]
+    pub fn max_connection_age(self, max_connection_age: Duration) -> Self {
+        Server {
+            max_connection_age: Some(max_connection_age),
             ..self
         }
     }
@@ -498,6 +523,7 @@ impl<L> Server<L> {
             http2_max_header_list_size: self.http2_max_header_list_size,
             max_frame_size: self.max_frame_size,
             accept_http1: self.accept_http1,
+            max_connection_age: self.max_connection_age,
         }
     }
 
@@ -538,6 +564,7 @@ impl<L> Server<L> {
             .unwrap_or_else(|| Duration::new(DEFAULT_HTTP2_KEEPALIVE_TIMEOUT_SECS, 0));
         let http2_adaptive_window = self.http2_adaptive_window;
         let http2_max_pending_accept_reset_streams = self.http2_max_pending_accept_reset_streams;
+        let max_connection_age = self.max_connection_age;
 
         let svc = self.service_builder.service(svc);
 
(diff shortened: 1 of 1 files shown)


djc:
Marking this as draft until upstream changes have been merged.

ionut-slaveanu:
Fixed workflow:
[CI](https://github.com/ionut-slaveanu/tonic/actions/workflows/CI.yml)
Also removed the dependency on:
[patch.crates-io]
hyper = {git = "https://github.com/ionut-slaveanu/hyper.git" }

, so this pull request can be merged independently of the PR on hyper.

examples/src/max_connection_age/server.rs
```diff
@@ -0,0 +1,42 @@
+use tonic::{transport::Server, Request, Response, Status};
```
djc, line 1: I don't think this is worth adding more examples.

djc:
So is this even needed if https://github.com/hyperium/hyper/pull/3729 is merged?

tottoto:
Can this config take `Duration` instead of `u64` ms?

djc:
> Can this config take `Duration` instead of `u64` ms?

Good idea -- if we do that (pending whether it's needed after hyper implements this), should also get rid of the `_ms` suffix.

examples/src/max_connection_age/server.rs
ionut-slaveanu: Removed that example code.

ionut-slaveanu:
> > Can this config take `Duration` instead of `u64` ms?
> 
> Good idea -- if we do that (pending whether it's needed after hyper implements this), should also get rid of the `_ms` suffix.

Fixed,

ionut-slaveanu:
> So is this even needed if [hyperium/hyper#3729](https://github.com/hyperium/hyper/pull/3729) is merged?

The changes present in  [hyperium/hyper#3729](https://github.com/hyperium/hyper/pull/3729) are on my fork.
So, the following code is the same as the one from hyper PR:
[patch.crates-io]
hyper = {git = "https://github.com/ionut-slaveanu/hyper.git" }

The problem with hyper's graceful_shutdown function that I'm trying to fix in that PR is:
- if the graceful_shutdown function is called during the http2 handshake, no GOAWAY frame is sent by the server to the clients. Basically, clients don't know that the http2 channel has been closed. Because of this some responses are lost.

tottoto:
By the way, as another approach, could this be solved in a more general way by adding an ability to specify a `Future` of signal that cancels the graceful shutdown? This allows the user to have options that are not limited to timeout.

djc:
> By the way, as another approach, could this be solved in a more general way by adding an ability to specify a `Future` of signal that cancels the graceful shutdown? This allows the user to have options that are not limited to timeout.

I'm not sure how cancelling graceful shutdown is involved? Seems to me this PR is about closing idle connections after a timeout.

tottoto:
I misunderstood the process a little at the time. I think you are right. I consider whether there is any other useful Futures besides timeouts, but I could not find out, so if we add this functionality, a timeout seems like a good idea.

tonic/src/transport/server/mod.rs
```diff
@@ -654,6 +695,10 @@ fn serve_connection<B, IO, S, E>(
                         }
                         break;
                     },
+                    _ = &mut sleep  => {
+                            conn.as_mut().graceful_shutdown();
```
djc, line 699: Looks like this is indented too much.

tonic/src/transport/server/mod.rs
```diff
@@ -621,13 +646,26 @@ impl<L> Server<L> {
...
+    let _result = match sleep_time {
+        Some(time) => {
+            let sleep = tokio::time::sleep(time);
+            sleep.await;
+        }
+        None => {
+            future::pending::<()>().await;
+        }
+    };
+}
```
djc, line 659: Nits: spell it (as much as possible -- importing `sleep()` and `pending()`) as below, move below `serve_connection()`.

```rust
async fn sleep_or_pending(wait_for: Option<Duration>) {
    match wait_for {
        Some(wait) => sleep(wait).await,
        None => pending().await,
    }
}
```

ionut-slaveanu: Fixed!

tonic/src/transport/server/mod.rs
ionut-slaveanu: Fixed! Thanks!

djc (approved):

djc:
Thanks, I think this makes sense.

ionut-slaveanu:
> Thanks, I think this makes sense.

Thank you, too!

PR #1866, opened 2024-08-09
fix(tonic-build,tonic) Add back TLS handling in genereated `Client::connect` code

mkatychev:
## Motivation
The recent breaking change in https://github.com/hyperium/tonic/pull/1731 removed the ability for generated client functions to connect to HTTPS endpoints through strings alone.
For example the doctest below  would fail in 0.12.x `tonic-build` connection code:
https://github.com/hyperium/tonic/blob/03913b987ea75ac6902799f0cfa36b583b5d40e1/tonic/src/transport/channel/endpoint.rs#L79-L82

## Solution

This PR introduces a best attempt at picking up tls feature flags in `Endpoint::new` to allow successfully connecting to strings that contain `https://` once more

### DIFF
--- a/tests/integration_tests/tests/connection.rs
+++ b/tests/integration_tests/tests/connection.rs
@@ -22,9 +22,13 @@ impl test_server::Test for Svc {
 #[tokio::test]
 async fn connect_returns_err() {
     let res = TestClient::connect("http://thisdoesntexist").await;
-    dbg!(&res);
 
     assert!(res.is_err());
+}
+
+#[tokio::test]
+async fn connect_handles_tls() {
+    TestClient::connect("https://example.com").await.unwrap();
 }
 
 #[tokio::test]
--- a/tests/integration_tests/tests/connection.rs
+++ b/tests/integration_tests/tests/connection.rs
@@ -22,6 +22,7 @@ impl test_server::Test for Svc {
 #[tokio::test]
 async fn connect_returns_err() {
     let res = TestClient::connect("http://thisdoesntexist").await;
+    dbg!(&res);
 
     assert!(res.is_err());
 }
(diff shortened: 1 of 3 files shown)


mkatychev:
@tottoto @djc  Let me know if there is a better approach to handle generated clients. 0.12 `::connect` does not work with TLS anymore.

djc (approved):
Seems a little ugly but probably still better to fix the regression for now.

tottoto:
Personally, I do not like this kind of implicit configuration and would prefer that users explicitly configure TLS. It seems better to update the documentation instead.

djc:
> Personally, I do not like this kind of implicit configuration and would prefer that users explicitly configure TLS. It seems better to update the documentation instead.

I'm sympathetic to this but I think regressing user's code in an upgrade is worse, and I don't think adding documentation enough. We could improve this via the type system instead.

mkatychev:
@tottoto  I would be happy to hear of alternatives, https://github.com/hyperium/tonic/pull/1731 for better or worse took a lot of convenience out of a TLS approach that "just works".
Long term, `From<Uri> for Endpoint` should be where a TLS config decision is made, this should enable getting rid of `Endpoint::new` entirely.
A [`Uri`'s scheme](https://docs.rs/http/latest/src/http/uri/mod.rs.html#108) should determine whether TLS is enabled if some feature flag like `auto_tls` is enabled. It's fine to be explicit about configuring TLS but I would rather not do it on each _connection_ and a non-default feature flag seems explicit enough.

tottoto:
> It's fine to be explicit about configuring TLS but I would rather not do it on each connection and a non-default feature flag seems explicit enough.

You can define your own constructor such as function to construct connection with your config.

mkatychev:
> > It's fine to be explicit about configuring TLS but I would rather not do it on each connection and a non-default feature flag seems explicit enough.
> 
> You can define your own constructor such as function to construct connection with your config.

That is not the case for generated code.

djc (approved):

PR #1870, opened 2024-08-14
server: allow setting max_header_list_size

BugenZhao:
Signed-off-by: Bugen Zhao <i@bugenzhao.com>

## Motivation

This is basically the dual of #1835 on the server side. Can refer to the motivation of #1835.

## Solution

Add the configuration for the server with the same interface as the client (transport). Also refine the documentation.


tottoto (approved):
Sounds reasonable to me.

Issue #1871, opened 2024-08-15
Remove trait bound on generated server struct


jhpratt:
### Crates

`tonic-build`

### Motivation

I would like to have a single struct containing any of a number of different backend servers. This is currently not reasonable to implement due to the trait bounds present on the generated server struct.

```rust
pub struct KvServer<T: Kv> { // <-- unnecessary trait bound
    inner: Arc<T>,
    accept_compression_encodings: EnabledCompressionEncodings,
    send_compression_encodings: EnabledCompressionEncodings,
    max_decoding_message_size: Option<usize>,
    max_encoding_message_size: Option<usize>,
}
```

### Proposal

The trait bound on the generated server struct should be removed. This is in line with standard practice, including in the standard library (see [`HashMap`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html), for example). The trait bounds would remain in place for methods where it is necessary for correctness.

PR #1872, opened 2024-08-15
Remove unnecessary trait bounds from server

jhpratt:
See #1871 for details. This is a straightforward implementation of removing the trait bounds where possible.


djc (approved):
LGTM, thanks!

PR #1874, opened 2024-08-18
Prevent server from exiting on ECONNABORTED

grembo:
## Motivation

On FreeBSD, accept can fail with `ECONNABORTED`, which means that *"A connection arrived, but it was closed while	 waiting on the listen queue."* (see [`man 2 accept`](https://man.freebsd.org/cgi/man.cgi?accept(2) )), which [is in line with POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/accept.html).

Without this patch, a server built without the `tls` feature exits returning 0 in case of an aborted connection, which makes it vulnerable not only to intentional denial of service, but also to unintentional crashes, e.g., by haproxy TCP health checks.

The problem can be reproduced on any FreeBSD system by running the tonic "helloworld" example (when building without feature `tls`) and then sending a packet using nmap:

    cd examples
    cargo run --bin helloworld-server --no-default-features &
    nmap -sT -p 50051 -6 ::1
    # server exits

When running the example with the feature `tls` enabled, it won't exit (as the tls event loop in [tonic/src/transport/server/incoming.rs](https://github.com/hyperium/tonic/blob/v0.12.1/tonic/src/transport/server/incoming.rs#L67-L69) handles errors gracefully):

    cd examples
    cargo run --bin helloworld-server --no-default-features \
      features=tls &
    nmap -sT -p 50051 -6 ::1
    # server keeps running

## Solution

The patch checks if the error returned in the accept loop of tcp_incoming is `ECONNABORTED`. If it is, it stays in the loop instead of exiting with an error (which subsequently exits the main event loop in `serve_with_shutdown`).

~~**This patch is not optimal** - it removes some generic error parameters to gain access to `std::io::Error::kind()`.~~ The logic itself should be sound.

Edit: Removed a couple of things after helpful input in review.

See also:
- https://man.freebsd.org/cgi/man.cgi?accept(2)
   FreeBSD `accept(2)` man page
- https://pubs.opengroup.org/onlinepubs/9799919799/functions/accept.html
   IEEE Std 1003.1-2024 on `accept`/`accept4`
- https://github.com/giampaolo/pyftpdlib/issues/105
   https://github.com/giampaolo/pyftpdlib/commit/0f822329e3431ec1bc6250c03e938c65a61b2eb4
   https://bugs.squid-cache.org/show_bug.cgi?id=4889
   https://github.com/golang/go/issues/3395
   Basically the same issue (and its fix) in other projects

### DIFF
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -2,6 +2,7 @@ use super::service::ServerIo;
 #[cfg(feature = "tls")]
 use super::service::TlsAcceptor;
 use std::{
+    io,
     net::{SocketAddr, TcpListener as StdTcpListener},
     pin::{pin, Pin},
     task::{ready, Context, Poll},
@@ -27,18 +28,18 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            if let Err(e) = item {
-                if let Some(e) = Into::<crate::Error>::into(e).downcast_ref::<std::io::Error>() {
-                    tracing::debug!(message = "Accept loop error.", error = %e);
-                    if e.kind() == std::io::ErrorKind::ConnectionAborted {
-                        continue;
+            yield match item {
+                Ok(_) => item.map(ServerIo::new_io)?,
+                Err(e) => {
+                    let e = e.into();
+                    tracing::debug!(error = %e, "accept loop error");
+                    if let Some(e) = e.downcast_ref::<io::Error>() {
+                        if e.kind() == io::ErrorKind::ConnectionAborted {
+                            continue;
+                        }
                     }
-                } else {
-                    tracing::debug!(message = "Accept loop error (unknown error).");
+                    Err(e)?
                 }
-                break;
-            } else {
-                yield item.map(ServerIo::new_io)?
             }
         }
     }
@@ -77,7 +78,7 @@ where
                 }
 
                 SelectOutput::Err(e) => {
-                    tracing::debug!(message = "Accept loop error.", error = %e);
+                    tracing::debug!(error = %e, "accept loop error");
                 }
 
                 SelectOutput::Done => {
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -27,12 +27,16 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            if item.is_err() {
-                let e: crate::Error = item.err().unwrap().into();
-                tracing::debug!(message = "Accept loop error.", error = %e);
-                if !e.to_string().contains("os error 53") {
-                    break;
+            if let Err(e) = item {
+                if let Some(e) = Into::<crate::Error>::into(e).downcast_ref::<std::io::Error>() {
+                    tracing::debug!(message = "Accept loop error.", error = %e);
+                    if e.kind() == std::io::ErrorKind::ConnectionAborted {
+                        continue;
+                    }
+                } else {
+                    tracing::debug!(message = "Accept loop error (unknown error).");
                 }
+                break;
             } else {
                 yield item.map(ServerIo::new_io)?
             }
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -16,17 +16,22 @@ use tokio_stream::{Stream, StreamExt};
 use tracing::warn;
 
 #[cfg(not(feature = "tls"))]
-pub(crate) fn tcp_incoming<IO, IE>(
-    incoming: impl Stream<Item = Result<IO, IE>>,
+pub(crate) fn tcp_incoming<IO>(
+    incoming: impl Stream<Item = Result<IO, std::io::Error>>,
 ) -> impl Stream<Item = Result<ServerIo<IO>, crate::Error>>
 where
     IO: AsyncRead + AsyncWrite + Unpin + Send + 'static,
-    IE: Into<crate::Error>,
 {
     async_stream::try_stream! {
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
+            if let Some(e) = item.as_ref().err() {
+                   if e.kind() == std::io::ErrorKind::ConnectionAborted {
+                       tracing::debug!(message = e.to_string(), error = %e);
+                       continue;
+                   }
+            }
             yield item.map(ServerIo::new_io)?
         }
     }
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -16,23 +16,26 @@ use tokio_stream::{Stream, StreamExt};
 use tracing::warn;
 
 #[cfg(not(feature = "tls"))]
-pub(crate) fn tcp_incoming<IO>(
-    incoming: impl Stream<Item = Result<IO, std::io::Error>>,
+pub(crate) fn tcp_incoming<IO, IE>(
+    incoming: impl Stream<Item = Result<IO, IE>>,
 ) -> impl Stream<Item = Result<ServerIo<IO>, crate::Error>>
 where
     IO: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    IE: Into<crate::Error>,
 {
     async_stream::try_stream! {
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            if let Some(e) = item.as_ref().err() {
-                   if e.kind() == std::io::ErrorKind::ConnectionAborted {
-                       tracing::debug!(message = e.to_string(), error = %e);
-                       continue;
-                   }
+            if item.is_err() {
+                let e: crate::Error = item.err().unwrap().into();
+                tracing::debug!(message = "Accept loop error.", error = %e);
+                if !e.to_string().contains("os error 53") {
+                    break;
+                }
+            } else {
+                yield item.map(ServerIo::new_io)?
             }
-            yield item.map(ServerIo::new_io)?
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -810,16 +805,15 @@ impl<L> Router<L> {
     /// This method discards any provided [`Server`] TCP configuration.
     ///
     /// [`Server`]: struct.Server.html
-    pub async fn serve_with_incoming_shutdown<I, IO, IE, F, ResBody>(
+    pub async fn serve_with_incoming_shutdown<I, IO, F, ResBody>(
```
djc, line 808: Same here.

tonic/src/transport/server/mod.rs
```diff
@@ -775,15 +774,11 @@ impl<L> Router<L> {
     /// This method discards any provided [`Server`] TCP configuration.
     ///
     /// [`Server`]: struct.Server.html
-    pub async fn serve_with_incoming<I, IO, IE, ResBody>(
-        self,
-        incoming: I,
-    ) -> Result<(), super::Error>
+    pub async fn serve_with_incoming<I, IO, ResBody>(self, incoming: I) -> Result<(), super::Error>
```
djc, line 777: We can't add additional constraints in a public interface like this without making a semver-incompatible release, and I don't think we want to do that for this issue.

djc (commented):
I agree that we need to improve this, but unfortunately I don't think this change is a good idea in its current form.

tonic/src/transport/server/mod.rs
grembo: Yeah, I figured that this isn't good.

One "solution" I offered in the description is:

```diff
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -27,6 +27,10 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
+            if item.is_err() {
+                    tracing::debug!(message = "Accept loop error.");
+                    continue
+            }
             yield item.map(ServerIo::new_io)?
         }
     }
```

which will work, but strikes me as too basic.

An alternative would be to change the function to look more like what the TLS enabled version uses (but without the TLS parts). This would make things more complex though:

```rust
#[cfg(not(feature = "tls"))]
pub(crate) fn tcp_incoming<IO, IE>(
    incoming: impl Stream<Item = Result<IO, IE>>,
) -> impl Stream<Item = Result<ServerIo<IO>, crate::Error>>
where
    IO: AsyncRead + AsyncWrite + Unpin + Send + 'static,
    IE: Into<crate::Error>,
{
    async_stream::try_stream! {
        let mut incoming = pin!(incoming);

        let mut tasks = tokio::task::JoinSet::new();

        loop {
            match select(&mut incoming, &mut tasks).await {
                SelectOutput::Incoming(stream) => {
                    yield ServerIo::new_io(stream);
                }

                SelectOutput::Io(io) => {
                    yield io;
                }

                SelectOutput::Err(e) => {
                    tracing::debug!(message = "Accept loop error.", error = %e);
                }

                SelectOutput::Done => {
                    break;
                }
            }
        }
    }
}
```
(untested)

Any other ideas?

grembo:
@djc I reduced the change to something basic. The string search isn't nice, but works given the constraints.

Also, the tracing example makes testing this nicer:

```
cargo run --bin tracing-server --no-default-features --features="tracing"
```

tonic/src/transport/server/incoming.rs
```diff
@@ -27,7 +27,15 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            yield item.map(ServerIo::new_io)?
+            if item.is_err() {
```
djc, line 30: Can't we do `if let Err(e) = item {` here?

IMO we should try a `downcast_ref()` to `io::Error` and then look for the right `io::ErrorKind`, would that be feasible?

djc (commented):
This is compatible at least! I'd like to be more precise, though.

grembo:
Also mentioning https://github.com/hyperium/hyper/discussions/3597, as it seems related.

grembo:
@djc Thank you for your help.

Personally I would be be fine with it to land at the current iteration (it seems tightly scoped and low-risk now).

That said, I'm wondering about a few things:
- The "feature tls" accept loop seems not to exit at all on any error (so the two work differently), which means it could loop fast under low resource conditions
- Other daemons I looked at handle `errcode == ENFILE || errcode == EMFILE` as transient problems, so they would delay and try again (but not exit) 
- It might also make sense to handle `ECONNABORTED` directly in https://github.com/tokio-rs/tokio/blob/5b9a290acd25d88416e70c73c49a54106fcf02b5/tokio/src/net/tcp/listener.rs#L171-L192 (and other poll_accept functions in tokio/net (I don't remember any code that actually made any use of ECONNABORTED beyond logging it, so it might be worth considering - but that would be outside of the scope of this change).

tonic/src/transport/server/incoming.rs
```diff
@@ -27,7 +27,19 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            yield item.map(ServerIo::new_io)?
+            if let Err(e) = item {
+                if let Some(e) = Into::<crate::Error>::into(e).downcast_ref::<std::io::Error>() {
```
djc, line 31: I think this is broadly acceptable but I have bunch of stylistic suggestions.

First, use a match for the top-level error extraction:

```rust
let err = match item {
    Ok(io) => yield Ok(ServerIo::new_io(io)), // Or however this is spelled
    Err(err) => err,
};

..
```

This reduces rightward drift.

Second, I would expect we can just call `err.into()` and don't need the whole `Into::<crate::Error>::into(e)` dance?

Third, for the debug messages, I think we should not use an explicit `message =` name but use the general form which is having the message after the "structured" argument (so `tracing::debug!(error = %e, "accept loop error");`). Note that error messages should be all-lowercase and not have any `.` at the end.

Fourth, please import `io` at the top of the file so we can drop the `std::` prefixes here.

djc:
> @djc Thank you for your help.
> 
> Personally I would be be fine with it to land at the current iteration (it seems tightly scoped and low-risk now).
> 
> That said, I'm wondering about a few things:
> 
>     * The "feature tls" accept loop seems not to exit at all on any error (so the two work differently), which means it could loop fast under low resource conditions

Agreed that it would be good to keep `tls` and non-tls loops in sync on this front.

>     * Other daemons I looked at handle `errcode == ENFILE || errcode == EMFILE` as transient problems, so they would delay and try again (but not exit)
> 
>     * It might also make sense to handle `ECONNABORTED` directly in https://github.com/tokio-rs/tokio/blob/5b9a290acd25d88416e70c73c49a54106fcf02b5/tokio/src/net/tcp/listener.rs#L171-L192 (and other poll_accept functions in tokio/net (I don't remember any code that actually made any use of ECONNABORTED beyond logging it, so it might be worth considering - but that would be outside of the scope of this change).

I think it makes sense to do more fine-tuned handling like this but would suggest keeping it out of scope for this PR.

tonic/src/transport/server/incoming.rs
grembo: > I think this is broadly acceptable but I have bunch of stylistic suggestions.
> 
> First, use a match for the top-level error extraction:
> 
> ```rust
> let err = match item {
>     Ok(io) => yield Ok(ServerIo::new_io(io)), // Or however this is spelled
>     Err(err) => err,
> };
> 
> ..
> ```
> 
> This reduces rightward drift.
> 
> Second, I would expect we can just call `err.into()` and don't need the whole `Into::<crate::Error>::into(e)` dance?
> 
> Third, for the debug messages, I think we should not use an explicit `message =` name but use the general form which is having the message after the "structured" argument (so `tracing::debug!(error = %e, "accept loop error");`). Note that error messages should be all-lowercase and not have any `.` at the end.
> 
> Fourth, please import `io` at the top of the file so we can drop the `std::` prefixes here.

Personally I don't have much of an opinion on these style issues, so I just changed things as requested.

There is one thing I really didn't like though, which was error extraction, which would have looked like this:

```
let err = match item {
    Ok(io) => { yield Ok(ServerIo::new_io(io)); continue },
    Err(err) => err,
};

do more stuff with err, then break;
```

So instead, I went with clearly defined Ok and Err branches (horizontal drift is IMHO acceptable). I'm quite happy with the readability of the outcome (especially having only one continue statement, pointing out the exception, and no break statements). This should also be relatively easy to parse to people who are not familiar with all language features + the diff to the master branch is very easy to understand.

tonic/src/transport/server/incoming.rs
```diff
@@ -27,7 +28,19 @@ where
         let mut incoming = pin!(incoming);
 
         while let Some(item) = incoming.next().await {
-            yield item.map(ServerIo::new_io)?
+            yield match item {
+                Ok(_) => item.map(ServerIo::new_io)?,
```
grembo, line 32: Alternative to this line I had before was:

```
                Ok(io) => Ok::<ServerIo<IO>, crate::Error>(ServerIo::new_io(io))?,
```

Better ideas?

tonic/src/transport/server/incoming.rs
```diff
@@ -27,7 +28,19 @@ where
...
+                    if let Some(e) = e.downcast_ref::<io::Error>() {
+                        if e.kind() == io::ErrorKind::ConnectionAborted {
+                            continue;
+                        }
+                    }
+                    Err(e)?
+                }
+            }
         }
     }
```
grembo, line 45:
```
Ok(_) => item.map(ServerIo::new_io)?
```

could also be written as

```
Ok(io) => Ok::<ServerIo<IO>, crate::Error>(ServerIo::new_io(io))?,
```

but in terms of readability and diff, I opted for the former (any better ideas?).

djc (approved):

djc:
This is looking much better, let's get it merged and iterate from here.

grembo:
> This is looking much better, let's get it merged and iterate from here. Thanks!

Thank you for helping to make this mergeable.

Issue #1512, opened 2023-09-12
build: Support deprecated attribute for rpc


alexheretic:
## Feature Request
It would be useful to mark a deprecated rpc with the #[deprecated] attr on the rust method. In particular for the client method (but makes sense also for the server trait method).

### Crates
_tonic-build_

### Motivation

To allow deprecating rpc and communicating that in rust code.

### Proposal

```protobuf
service Foo {
    // Deprecated. Use `Baz`.
    rpc Bar(Msg) returns (Msg) {
        option deprecated = true;
    }
    rpc Baz(Msg) returns (Msg);
}

message Msg {}
```

Currently generates `FooClient` code without any deprecated attr.

Instead _tonic-build_ should add the attr.
```rust
impl<T> FooClient<T> {
    /// Deprecated. Use `Baz`.
    #[deprecated]
    pub async fn bar(
        &mut self,
        request: impl tonic::IntoRequest<super::Msg>,
    ) -> Result<tonic::Response<super::Msg>, tonic::Status> {
        ...
    }
}
```

PR #1879, opened 2024-08-20
feat(build): Add `#[deprecated]` to deprecated client methods

bytedream:
## Motivation

When an rpc method is marked as deprecated, tonic currently doesn't forward it to the user (#1512).

## Solution

This PR adds the `#[deprecated]` attribute to generated rust client methods that are marked as deprecated in a protobuf file. I've added `Method::deprecated` which returns if the method is deprecated and is checked when the client code is generated. ~~I'd consider this a breaking change, as the `Method` trait is publicly exposed and the code will now emit warnings if a deprecated generated method is called.~~ `Method::deprecated` has a default implementation which always returns `false`.

### DIFF
--- a/tests/deprecated_methods/tests/deprecated_methods.rs
+++ b/tests/deprecated_methods/tests/deprecated_methods.rs
@@ -3,6 +3,11 @@ use std::{fs, path::PathBuf};
 #[test]
 fn test() {
     let path = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("test.rs");
-    let s = fs::read_to_string(path).unwrap();
-    assert_eq!(s.match_indices("#[deprecated]").count(), 1);
+    let s = fs::read_to_string(path)
+        .unwrap()
+        .split_whitespace()
+        .collect::<Vec<_>>()
+        .join(" ");
+    assert!(s.contains("#[deprecated] pub async fn deprecated("));
+    assert!(!s.contains("#[deprecated] pub async fn not_deprecated("));
 }
--- a/tests/deprecated_methods/tests/deprecated_methods.rs
+++ b/tests/deprecated_methods/tests/deprecated_methods.rs
@@ -1,0 +1,8 @@
+use std::{fs, path::PathBuf};
+
+#[test]
+fn test() {
+    let path = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("test.rs");
+    let s = fs::read_to_string(path).unwrap();
+    assert_eq!(s.match_indices("#[deprecated]").count(), 1);
+}
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -144,6 +144,8 @@ pub trait Method {
     fn server_streaming(&self) -> bool;
     /// Get comments about this item.
     fn comment(&self) -> &[Self::Comment];
+    /// Method is deprecated.
+    fn deprecated(&self) -> bool;
     /// Type name of request and response.
     fn request_response_name(
         &self,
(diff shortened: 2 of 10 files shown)


tonic-build/src/lib.rs
```diff
@@ -144,6 +144,8 @@ pub trait Method {
     fn server_streaming(&self) -> bool;
     /// Get comments about this item.
     fn comment(&self) -> &[Self::Comment];
+    /// Method is deprecated.
+    fn deprecated(&self) -> bool;
```
djc, line 148: Could we give this a default implementation that yields `false`? That would make this a non-breaking change, which might be nice, and it seems conceptually reasonable.

tests/deprecated_methods/tests/deprecated_methods.rs
```diff
@@ -0,0 +1,8 @@
+use std::{fs, path::PathBuf};
+
+#[test]
+fn test() {
+    let path = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("test.rs");
+    let s = fs::read_to_string(path).unwrap();
+    assert_eq!(s.match_indices("#[deprecated]").count(), 1);
+}
```
tottoto, line 8: I would think that just detecting the `#[deprecated]` string seems not enough to test this functionality. Is it possible to verify that this part is generated in the correct place, or to check that the entire generated code matches what is expected?

tonic-build/src/lib.rs
tottoto: For now I agree to make this setting false by default, and I think it would be good to set it to true in the future breaking release.

tests/deprecated_methods/tests/deprecated_methods.rs
bytedream: I've oriented myself to the [disable comments test](https://github.com/hyperium/tonic/blob/4c9356c10eb7838e97bbf303f9868aee2364b1af/tests/disable_comments/tests/disable_comments.rs). Testing if `#[deprecated]` is generated above the correct method should be doable by checking `#[deprecated] pub fn somefunction`.

tonic-build/src/lib.rs
bytedream: That's a good idea, I've added it

tests/deprecated_methods/tests/deprecated_methods.rs
bytedream:
```suggestion
#[test]
fn test() {
    let path = PathBuf::from(std::env::var("OUT_DIR").unwrap()).join("test.rs");
    let s = fs::read_to_string(path)
        .unwrap()
        .split_whitespace()
        .collect::<Vec<_>>()
        .join(" ");
    assert!(s.contains("#[deprecated] pub async fn deprecated("));
    assert!(!s.contains("#[deprecated] pub async fn not_deprecated("));
}
```
This is only alternative I can think of atm

tottoto: I am not sure if the way is appropriate for testing code generating, but I feel it can test the attributes.

@djc How do you think?

tonic-build/src/lib.rs
djc: > For now I agree to make this setting false by default, and I think it would be good to set it to true in the future breaking release.

Why? That doesn't make much sense to me.

tottoto: What I intended to mean was to not mark the code generated from the proto's deprecated as deprecated by default this time, and to enable it in the next major release. And I just noticed that I was commenting on something different than the perspective of your review.

djc: Ah, so you mean removing the default implementation when we release 0.13? That makes sense to me.

tests/deprecated_methods/tests/deprecated_methods.rs
djc: @tottoto not entirely sure what you mean. IMO the [proposed test](https://github.com/hyperium/tonic/pull/1879/files#r1729465879) is good enough for now.

tottoto: > IMO the [proposed test](https://github.com/hyperium/tonic/pull/1879/files#r1729465879) is good enough for now.

Thanks. I was not confident so I wanted other's opinions.

djc (approved):

tottoto (approved):

PR #1880, opened 2024-08-21
fix(tonic-web): fix panic caused in trailer parsing when there is more than one trailer

zakhenry:
…

## Motivation

Attempting to parse a real grpc-web message response from an envoy server that added a few other trailers resulted in a panic

## Solution

Root issue was a simple failure to advance the cursor in a buffer read. It happened to work for the first trailer so this wasn't already caught in tests

### DIFF
--- a/tonic-web/src/call.rs
+++ b/tonic-web/src/call.rs
@@ -393,11 +393,15 @@ fn decode_trailers_frame(mut buf: Bytes) -> Result<Option<HeaderMap>, Status> {
     let mut cursor_pos = 0;
 
     for (i, b) in buf.iter().enumerate() {
+        // if we are at a trailer delimiter (\r\n)
         if b == &b'\r' && buf.get(i + 1) == Some(&b'\n') {
+            // read the bytes of the trailer passed so far
             let trailer = temp_buf.copy_to_bytes(i - cursor_pos);
-            cursor_pos = i;
+            // increment cursor beyond the delimiter
+            cursor_pos = i + 2;
             trailers.push(trailer);
             if temp_buf.has_remaining() {
+                // advance buf beyond the delimiters
                 temp_buf.get_u8();
                 temp_buf.get_u8();
             }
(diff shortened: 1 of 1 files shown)


tonic-web/src/call.rs
```diff
@@ -548,14 +548,14 @@ mod tests {
...
         // Byte version of this:
-        // b"\0\0\0\0L\n$975738af-1a17-4aea-b887-ed0bbced6093\x1a$da609e9b-f470-4cc0-a691-3fd6a005a436\x80\0\0\0\x0fgrpc-status:0\r\n"
-        let buf = [
-            0, 0, 0, 0, 76, 10, 36, 57, 55, 53, 55, 51, 56, 97, 102, 45, 49, 97, 49, 55, 45, 52,
-            97, 101, 97, 45, 98, 56, 56, 55, 45, 101, 100, 48, 98, 98, 99, 101, 100, 54, 48, 57,
-            51, 26, 36, 100, 97, 54, 48, 57, 101, 57, 98, 45, 102, 52, 55, 48, 45, 52, 99, 99, 48,
-            45, 97, 54, 57, 49, 45, 51, 102, 100, 54, 97, 48, 48, 53, 97, 52, 51, 54, 128, 0, 0, 0,
-            15, 103, 114, 112, 99, 45, 115, 116, 97, 116, 117, 115, 58, 48, 13, 10,
-        ];
+        let buf = b"\0\0\0\0L\n$975738af-1a17-4aea-b887-ed0bbced6093\x1a$da609e9b-f470-4cc0-a691-3fd6a005a436\x80\0\0\0\x0fgrpc-status:0\r\ngrpc-message:\r\nrequest_id:d0e56d4d10ab4c85b0882c3467049522\r\na:1\r\nb:2\r\n";
```
zakhenry, line 551: oops revert

tonic-web/src/call.rs
```diff
@@ -395,7 +395,7 @@ fn decode_trailers_frame(mut buf: Bytes) -> Result<Option<HeaderMap>, Status> {
     for (i, b) in buf.iter().enumerate() {
         if b == &b'\r' && buf.get(i + 1) == Some(&b'\n') {
             let trailer = temp_buf.copy_to_bytes(i - cursor_pos);
-            cursor_pos = i;
+            cursor_pos = i + 2;
```
djc, line 398: Could you please add a few comments around this loop that explain the conceptual logic here?

zakhenry: yea I agree, I'm not entirely convinced this is the most idiomatic rust to parse the trailers, but I wanted to do the minimal change. Comments added to my understanding of what is going on, PTAL

djc: Thanks. Trying to reason about this code hurts my head, but I suppose this is an improvement at least. Will probably try to rewrite this.

djc (approved):

PR #1881, opened 2024-08-21
chore: prepare 0.12.2 release

djc:
Prepare release for #1866, which fixes a regression from #1731.

Would like to get this merged before releasing:

- [ ] #1888

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,26 @@
+# [0.12.2](https://github.com/hyperium/tonic/compare/v0.12.1...v0.12.2) (2024-08-23)
+
+### Features
+
+* Move TimeoutExpired out of transport (#1826)
+* Move ConnectError type from transport (#1828)
+* **channel:** allow setting max_header_list_size (#1835)
+* **router:** Add RoutesBuilder constructor (#1855)
+* **tls:** Rename tls-roots feature with tls-native-roots (#1860)
+* **router:** Rename Routes::into_router with into_axum_router (#1862)
+* **router:** Implement from axum::Router for Routes (#1863)
+* **channel:** Re-enable TLS based on Cargo features in generated clients (#1866)
+* **server:** allow setting max_header_list_size (#1870)
+* **build:** Expose formatted service name (#1684)
+* **reflection:** add back support for v1alpha reflection protocol (#1888)
+
+### Bug Fixes
+
+* **router:** Add missing unimplemented fallback to RoutesBuilder (#1864)
+* **server:** Prevent server from exiting on ECONNABORTED (#1874)
+* **web:** fix panic in trailer parsing on multiple trailers (#1880)
+* **web:** fix empty trailer parsing causing infinite parser loop (#1883)
+
 # [0.12.1](https://github.com/hyperium/tonic/compare/v0.12.0...v0.12.1) (2024-07-17)
 
 ### Bug Fixes
--- a/tonic-build/Cargo.toml
+++ b/tonic-build/Cargo.toml
@@ -4,7 +4,7 @@ categories = ["network-programming", "asynchronous"]
 description = """
 Codegen module of `tonic` gRPC implementation.
 """
-documentation = "https://docs.rs/tonic-build/0.12.1"
+documentation = "https://docs.rs/tonic-build/0.12.2"
 edition = "2021"
 homepage = "https://github.com/hyperium/tonic"
 keywords = ["rpc", "grpc", "async", "codegen", "protobuf"]
@@ -12,7 +12,7 @@ license = "MIT"
 name = "tonic-build"
 readme = "README.md"
 repository = "https://github.com/hyperium/tonic"
-version = "0.12.1"
+version = "0.12.2"
 
 [dependencies]
 prettyplease = { version = "0.2" }
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -70,7 +70,7 @@
     html_logo_url = "https://raw.githubusercontent.com/tokio-rs/website/master/public/img/icons/tonic.svg"
 )]
 #![deny(rustdoc::broken_intra_doc_links)]
-#![doc(html_root_url = "https://docs.rs/tonic-build/0.12.1")]
+#![doc(html_root_url = "https://docs.rs/tonic-build/0.12.2")]
 #![doc(issue_tracker_base_url = "https://github.com/hyperium/tonic/issues/")]
 #![doc(test(no_crate_inject, attr(deny(rust_2018_idioms))))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
--- a/tonic-health/Cargo.toml
+++ b/tonic-health/Cargo.toml
@@ -4,7 +4,7 @@ categories = ["network-programming", "asynchronous"]
 description = """
 Health Checking module of `tonic` gRPC implementation.
 """
-documentation = "https://docs.rs/tonic-health/0.12.1"
+documentation = "https://docs.rs/tonic-health/0.12.2"
 edition = "2021"
 homepage = "https://github.com/hyperium/tonic"
 keywords = ["rpc", "grpc", "async", "healthcheck"]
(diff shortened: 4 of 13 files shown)


tottoto:
I think #1866 should be treated as a breaking change.

djc:
> I think #1866 should be treated as a breaking change.

What code do you think it breaks? It seems to strictly make more code work, in particular code that trivially worked pre-#1731 and for which there is no obvious/simple alternative. Some of my downstream code turns out to be affected, too, in that if call `SomeServiceClient::new()` and start calling methods on it, that just fails if I give it a https URL. How should I rewrite that code, in your opinion?

LucioFranco (approved):

djc:
I've revised the changelog and added more things to it. I think this is ready. @LucioFranco can you get it released?

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+* Implement from axum::Router for Routes (#1863)
+* server: allow setting max_header_list_size (#1870)
+* tonic-reflection: add back support for v1alpha reflection protocol (#1888)
+
+### Bug Fixes
+
+* Re-enable TLS based on Cargo features in generated clients (#1866)
+* Prevent server from exiting on ECONNABORTED (#1874)
+* tonic-web: fix panic in trailer parsing on multiple trailers (#1880)
+* web: fix empty trailer parsing causing infinite parser loop (#1883)
```
tottoto, line 14: Added some changes to changelog, conformed the prefixes to the existing style, and moved `Re-enable TLS based on Cargo features in generated clients (#1866)` to `Features` section. This change is the same kind of `tonic-reflection: add back support for v1alpha reflection protocol (#1888)`. Both were implemented correctly and with intent, and are bug-free. They were introduced this time as a result of yet another choice. This type of change would be a bug fix only if we unintentionally removed functionality and then re-added it to fix them, but it was an intentional choice at the time. Bugs should be limited to something like unintended deviations from spec, changes in behavior or API, or breaking changes, which should treated as bugs.

```suggestion
### Features

* Move TimeoutExpired out of transport (#1826)
* Move ConnectError type from transport (#1828)
* **channel:** allow setting max_header_list_size (#1835)
* **router:** Add RoutesBuilder constructor (#1855)
* **tls:** Rename tls-roots feature with tls-native-roots (#1860)
* **router:** Rename Routes::into_router with into_axum_router (#1862)
* **router:** Implement from axum::Router for Routes (#1863)
* **channel:** Re-enable TLS based on Cargo features in generated clients (#1866)
* **server:** allow setting max_header_list_size (#1870)
* **build:** Expose formatted service name (#1684)
* **reflection:** add back support for v1alpha reflection protocol (#1888)

### Bug Fixes

* **router:** Add missing unimplemented fallback to RoutesBuilder (#1864)
* **server:** Prevent server from exiting on ECONNABORTED (#1874)
* **web:** fix panic in trailer parsing on multiple trailers (#1880)
* **web:** fix empty trailer parsing causing infinite parser loop (#1883)
```

tottoto (changes requested):

CHANGELOG.md
mkatychev: > Re-enable TLS based on Cargo features in generated clients 

Apologies but could you point out where was the removal of TLS  for generated clients documented? I'm looking through https://github.com/hyperium/tonic/blob/master/CHANGELOG.md but I can't seem to spot it.

djc: It's "tls: Remove tls roots implicit configuration".

mkatychev: Given that generated clients don't have a choice between explicit/implicit configuration (since they do not handle `Endpoint`s directly) I would surmise that the TLS behaviour being affected is a side effect of the TLS config changes. 

If there was mention of codegen being impacted in https://github.com/hyperium/tonic/pull/1731 then it would be far easier to sell disabling TLS altogether as an intentional change, otherwise the argument that adding TLS back in as a feature is not being made in good faith.

djc: I think you're arguing that you disagree with the way either @tottoto or me have phrased the changelog line for this change? Do you want to make a proposal that would better convey what you need to know?

mkatychev: @djc I don't have an opinion one way or the other as to how the fix shows up in the changelog. I'm fine with the proposed changes.

TLS and reflection (for things like Postman and `grpcui` that still use `v1alpha`) broke for my codebase moving to 0.12, that happens when huge changes to the dependency tree are made (such as the move to `http` 1.0) and that was a risk I took on.

I suspect that most users care more about TLS working again in 0.12 genrated code than how a commit is categorized.

These kinds of disucssions are best conducted after a fix is out.

tottoto: >>  Re-enable TLS based on Cargo features in generated clients
>
> Apologies but could you point out where the initial removal of TLS for generated clients is documented? I'm looking through https://github.com/hyperium/tonic/blob/master/CHANGELOG.md but I can't seem to spot it.

As @djc answered, it is "tls: Remove tls roots implicit configuration". At least at the time, @djc and I selected this explanation. If you think it is bug based on this explanation, you misunderstand and the documentation may be insufficient in this case. Basically, if you notice a difference between pieces of information that should be the same, you should not assume that only one is correct, but rather consider various possibilities, such as which is unintentional, which is a total inconsistency, or it is a misunderstanding of you.

> Given that generated clients don't have a choice between explicit/implicit configuration (since they do not handle Endpoints directly) I would surmise that the TLS behaviour being affected is a side effect of the TLS config changes.

It does not matter if it is a side effect, it is whether it is intended or not.

> If there was mention of codegen being impacted in https://github.com/hyperium/tonic/pull/1731 then it would be far easier to sell disabling TLS altogether as an intentional change, otherwise the argument that adding TLS back in is a feature is not being made in good faith.

Documentation, including changelog, can sometimes be insufficient. We are always in limited resources, so we cannot guarantee that all of the documentation is in a completely satisfactory state for everyone. If you find something difficult or insufficient to understand and you can solve it, you can help the project by contributing with proposing enhanced document instead of just complaining about it.

> TLS and reflection (for things like Postman and grpcui that still use v1alpha) broke for my codebase moving to 0.12, that happens when huge changes to the dependency tree are made (such as the move to http 1.0) and that was a risk I took on.
>
> I suspect that most users care more about having a version of 0.12.x that has generated code with working TLS than how a commit is categorized.

From your comments it seems that you are picking up pieces of information and judging whether something is a bug or not based on your own preferences.

The 0.12.0 release is a major release that assumes the existence of breaking changes, so it is legitimate for it to include any breaking changes. Whether or not you like a change has no bearing upon whether or not it is a bug. It is a technical problem.

> These kinds of disucssions are best conducted after a fix is out.

Again, it is intended and not a bug, so the change the functionality which is not a bug cannot be a fix. It is just adding a new feature after the last breaking change.

djc:
> It does not matter if it is a side effect, it is whether it is intended or not.

So when you wrote #1731, did you consider (and intend) that `SomeServiceClient::new()` would no longer be able to handle `https` endpoints? I did not consider that while reviewing, and would have objected if I did. From that perspective, it is not entirely intentional.

> From your comments it seems that you are picking up pieces of information and judging whether something is a bug or not based on your own preferences.
> 
> The 0.12.0 release is a major release that assumes the existence of breaking changes, so it is legitimate for it to include any breaking changes. Whether or not you like a change has no bearing upon whether or not it is a bug. It is a technical problem.

IMO it is reasonable to consider it a bug that `SomeServiceClient::new()` still compiled in 0.12.0 but failed to work with `https` endpoint URLs. While it may be technically allowed by semver, I think the effect of the changes in #1731 on that scenario are problematic for this reason. In general, if one (or two) maintainers think something is reasonable but most of the downstream users disagree, I'm inclined to think the users are right by default.

tottoto:
> So when you wrote #1731, did you consider (and intend) that `SomeServiceClient::new()` would no longer be able to handle `https` endpoints?

It works with adding appropriate TLS configuration after that change. It is intended and the one of the main purposes of the changes.

djc:
This was published: https://crates.io/crates/tonic/0.12.2.

PR #1883, opened 2024-08-22
fix(web): fix empty trailer parsing causing infinite parser loop

zakhenry:
## Motivation

Attempting to parse a grpc response from an envoy proxy that inserts trailers causes an infinite loop

## Solution
Digging in to the code I found that there was an attempt to read more data after parsing trailers but the buffer was empty, ending up in the same place again. Guarding this fixes the issue and I can successfully fetch a message.

Note this fix needs to be combined with #1880 to get a valid grpc response back (otherwise it panics earlier).

### DIFF
--- a/tonic-web/src/call.rs
+++ b/tonic-web/src/call.rs
@@ -304,9 +304,10 @@ where
                         }
                     }
                     FindTrailers::IncompleteBuf => continue,
-                    FindTrailers::Done(len) => {
-                        Poll::Ready(Some(Ok(Frame::data(buf.split_to(len).freeze()))))
-                    }
+                    FindTrailers::Done(len) => Poll::Ready(match len {
+                        0 => None,
+                        _ => Some(Ok(Frame::data(buf.split_to(len).freeze()))),
+                    }),
                 };
             }
         }


zakhenry:
I want to add a test for this fix, however I got stuck trying to actually create trailers for a unary request with tonic? It seems to me that trailers are not actually supported by the server, and so therefore it becomes pretty difficult to test this functionality. I tried also with a streaming call to see if I could work around it that way, but no joy there either.

tonic-web/src/call.rs
```diff
@@ -305,7 +305,11 @@ where
                     }
                     FindTrailers::IncompleteBuf => continue,
                     FindTrailers::Done(len) => {
-                        Poll::Ready(Some(Ok(Frame::data(buf.split_to(len).freeze()))))
+                        if len > 0 {
```
djc, line 308: Nit: please spell it like this:

```rust
Poll::Ready(match len {
    0 => None,
    _ => Some(Ok(Frame::data(buf.split_to(len).freeze()))),
})
```

djc (approved):

tonic-web/src/call.rs
zakhenry: good idea, done.

zakhenry: cargo fmt ended up inlining this expression with `FindTrailers::Done(len) =>`, which is neat enough I think

tottoto:
Rerun the CI and seems successful.

tottoto:
I rerun the CI and it seems successful.

PR #1884, opened 2024-08-23
feat(grpc-web): make public some methods to enable grpc-web client proxying

zakhenry:
## Motivation

I want to create a middleware grpc service that can listen to regular grpc calls, and forward them to another service as grpc-web calls. In order to do this fully generically, there needs to be made public some internals of tonic that are normally called via the `Grpc` service that the generated service code uses.

## Solution

Publish the minimal set of symbols that are required

---

If there is interest, I can create a new example in the examples dir to demonstrate grpc-web proxying. I've only tested unary calls locally so far.

### DIFF
--- a/tonic/src/codec/compression.rs
+++ b/tonic/src/codec/compression.rs
@@ -269,7 +269,7 @@ pub(crate) fn decompress(
 }
 
 #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum SingleMessageCompressionOverride {
+pub enum SingleMessageCompressionOverride {
     /// Inherit whatever compression is already configured. If the stream is compressed this
     /// message will also be configured.
     ///
--- a/tonic/src/codec/decode.rs
+++ b/tonic/src/codec/decode.rs
@@ -54,7 +54,9 @@ enum Direction {
 }
 
 impl<T> Streaming<T> {
-    pub(crate) fn new_response<B, D>(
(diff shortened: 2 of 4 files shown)


zakhenry:
relates to #31 but doesn't necessarily close it

zakhenry:
Marked as ready for review, but note it is not ready for merge as it needs docs on the newly public interfaces and possible an example of use. Just want to make sure the concept is sound before I put further effort in.

djc:
I think these changes are okay (once CI passes).

zakhenry:
@tottoto thanks for the review, I've fixed up the CI with some docs, and added a new example demonstrating a super basic proxy that just passes a request on. Of course my real world proxy does more things like grpc-web translation but I wanted to keep it smallish.

djc:
I don't think we want to add yet another example here. Would be good to get the documentation additions squashed into the initial commit, though.

zakhenry:
@djc I understand the resistance to the proliferation of too many examples though since this one touches these newly public api do you reckon I'd be better of reformulating the example as an integration test?

zakhenry:
For now, I've dropped the example from this PR entirely so we can get this in and continue discussion in another PR if that's a preferred option?

zakhenry:
I think the remaining CI failure is fixed by #1906

zakhenry:
@tottoto any chance you could have another look at this PR? Thanks!

tottoto (approved):

PR #1885, opened 2024-08-23
server: ignore more error kinds in incoming socket stream

djc:
As discussed in https://github.com/hyperium/tonic/pull/1882#issuecomment-2302371292.

cc @grembo

### DIFF
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -91,7 +91,13 @@ fn handle_accept_error(e: impl Into<crate::Error>) -> ControlFlow<crate::Error> 
     let e = e.into();
     tracing::debug!(error = %e, "accept loop error");
     if let Some(e) = e.downcast_ref::<io::Error>() {
-        if e.kind() == io::ErrorKind::ConnectionAborted {
+        if matches!(
+            e.kind(),
+            io::ErrorKind::ConnectionAborted
+                | io::ErrorKind::Interrupted
+                | io::ErrorKind::InvalidData // Raised if TLS handshake failed
+                | io::ErrorKind::WouldBlock
+        ) {
             return ControlFlow::Continue(());
         }
     }


grembo:
Looks good to me. In theory, WouldBlock is already filtered in TcpStream, but having all the transient ones checked in the same place is better :+1:

I wonder if it would be a good idea to add/move these checks to tokio/net in the long run (around https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/tcp/listener.rs#L186, where it already handled WouldBlock).

djc:
> I wonder if it would be a good idea to add/move these checks to tokio/net in the long run (around https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/tcp/listener.rs#L186, where it already handled WouldBlock).

I could see why Tokio wouldn't want to be so opinionated -- `WouldBlock` is more clear-cut because that is part of the whole async stack.

grembo (approved):
FWIW, here's my approval.

tonic/src/transport/server/incoming.rs
```diff
@@ -91,7 +91,12 @@ fn handle_accept_error(e: impl Into<crate::Error>) -> ControlFlow<crate::Error>
     let e = e.into();
     tracing::debug!(error = %e, "accept loop error");
     if let Some(e) = e.downcast_ref::<io::Error>() {
-        if e.kind() == io::ErrorKind::ConnectionAborted {
+        if matches!(
+            e.kind(),
+            io::ErrorKind::ConnectionAborted
+                | io::ErrorKind::Interrupted
+                | io::ErrorKind::WouldBlock
```
zier-one, line 98:
```suggestion
                | io::ErrorKind::WouldBlock
                | io::ErrorKind::InvalidData
```
When we got a TLS handshake error, A `InvalidData` error will happen here. see: 1897

zier-one:
Maybe we should add more test about TLS? It is really a critical bug that the server process exits when a TLS handshake error happened. Clients can easily perform denial-of-service attacks.

djc:
Might be nice to get this into 0.12.3.

tottoto (approved):

lap1nou:
> Should we maybe add more tests about TLS? It is really a critical bug that the server process exits when a TLS handshake error happened. Clients can easily perform denial-of-service attacks.

Hello,

Sorry to intervene in that merged PR, but shouldn't that bug have it's own CVE number ? I encountered that bug while developping my application and indeed this seems like a nice primitive to crash any gRPC server running Tonic 0.12.2 (or less ?).

Regards.

LucioFranco:
Its been published as a https://github.com/hyperium/tonic/security/advisories/GHSA-4jwc-w2hc-78qv and is a low sev CVE. RustSec PR will follow.

PR #1886, opened 2024-08-23
Add example axum-share-tcp-port

### LINKED ISSUE #1878: Share tonic server port with other http routes
I'd like to serve non-grpc routes alongside tonic grpc same server tcp port. Is there an existing way to do this?

alexheretic:
Add example of combining tonic grpc routes with regular axum http routes served on a single tcp port.

I also added notes around the downsides of not using the tonic server impl & how another server could be similarly configured.

## Motivation
Resolves #1878

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -260,6 +260,11 @@ path = "src/codec_buffers/server.rs"
 name = "codec-buffers-client"
 path = "src/codec_buffers/client.rs"
 
+[[bin]]
+name = "axum-share-tcp-port"
+path = "src/axum-share-tcp-port/server.rs"
+required-features = ["axum"]
+
 
 [features]
 gcp = ["dep:prost-types", "tonic/tls"]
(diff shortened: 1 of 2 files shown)


djc:
Thanks for putting this together. I'm inclined to avoid merging this, because in my experience adding a lot of examples can cause substantial overhead.

alexheretic:
An alternative would be to add something similar to `into_axum_router` docs.

But coming to the examples can be more natural to a tonic user wondering how to do this and not necessarily knowing about that fn.

tbh I'd be surprised if this example caused much extra ci overhead. But ofc up to you to include this or not.

tottoto:
I agree with djc. As the number of examples increases, maintenance costs also increase, so we should consider the development overhead which added by examples and the benefit. In this case, I feel like it is not worth the cost for now.

Regarding adding API documentation to `into_axum_router`, after converting to the axum router, it will be about how to use axum, so there does not seems to be need to explain it in tonic's documentation.

imotai:
This example is really helpful

PR #1888, opened 2024-08-23
tonic-reflection: add back support for v1alpha reflection protocol

djc:
## Motivation

In tonic 0.12, we [upgraded to v1 reflection](https://github.com/hyperium/tonic/pull/1701). However, some tools (like Postman and Kreya) still use the v1alpha1 version of the reflection schema. This PR tries to add back the v1alpha reflection schema with minimal code duplication, by moving code around so that the `Builder` builds a schema-independent `ReflectionServiceState` which can be wrapped in a server following the requested schema version.

This should supersede #1787 and #1822 (thanks @ttkjesper for the input -- I've stolen your tests) while reducing code duplication.

Would be happy to get feedback on whether this works for people affected by this issue, as I'd like to get this merged into the [pending 0.12.2](https://github.com/hyperium/tonic/pull/1881) release.

### DIFF
--- a/tonic-reflection/src/server/mod.rs
+++ b/tonic-reflection/src/server/mod.rs
@@ -1,10 +1,10 @@
-pub use crate::pb::v1::server_reflection_server::{ServerReflection, ServerReflectionServer};
-
-use prost::DecodeError;
 use std::fmt::{Display, Formatter};
 
-mod v1;
+use prost::DecodeError;
 
+pub use crate::pb::v1::server_reflection_server::{ServerReflection, ServerReflectionServer};
+
+mod v1;
 pub use v1::Builder;
 
 /// Represents an error in the construction of a gRPC Reflection Service.
--- a/tonic-reflection/src/server/mod.rs
+++ b/tonic-reflection/src/server/mod.rs
@@ -76,7 +76,15 @@ impl<'b> Builder<'b> {
     }
 
     /// Build a v1 gRPC Reflection Service to be served via Tonic.
-    pub fn build(mut self) -> Result<v1::ServerReflectionServer<impl v1::ServerReflection>, Error> {
+    #[deprecated(since = "0.12.2", note = "use `build_v1()` instead")]
+    pub fn build(self) -> Result<v1::ServerReflectionServer<impl v1::ServerReflection>, Error> {
+        self.build_v1()
+    }
+
+    /// Build a v1 gRPC Reflection Service to be served via Tonic.
+    pub fn build_v1(
+        mut self,
+    ) -> Result<v1::ServerReflectionServer<impl v1::ServerReflection>, Error> {
         if self.include_reflection_service {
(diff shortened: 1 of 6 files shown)


tonic-reflection/src/server/mod.rs
```diff
@@ -1,11 +1,307 @@
...
+
+use prost::{DecodeError, Message};
+use prost_types::{
+    DescriptorProto, EnumDescriptorProto, FieldDescriptorProto, FileDescriptorProto,
+    FileDescriptorSet,
+};
+use tonic::Status;
 
 mod v1;
+pub use v1::{ServerReflection, ServerReflectionServer}; // For backwards compatibility
```
tottoto, line 13: How about adding a document to guide to the new place of these reexport.

tonic-reflection/src/server/mod.rs
```diff
@@ -1,11 +1,307 @@
...
+mod v1alpha;
```
tottoto, line 14: Should this be located below `mod v1;` line?

tottoto (approved):
Looks good to me. Although the code become relatively complicated, it would be nice to support both versions until the ecosystem moves to v1.

tonic-reflection/src/server/mod.rs
djc: Personally I prefer to keep the `use v1::` imports close to the `mod v1` declaration.

tonic-reflection/src/server/mod.rs
djc: I guess we currently don't have such a new place. I guess we can make `v1 and `v1alpha` `pub` in `server`?

BButner:
From a client perspective how should this properly be handled, differentiating between when to use `v1` and `v1alpha`?

For instance: If I'm going to attempt to invoke a `ServerReflectionRequest` against a gRPC server that I do not know if its implemented `v1` or `v1alpha`, what's the proper way of handling that? Should I attempt it first with using `v1`, and then if I get an `Unimplemented` error attempt to fallback on `v1alpha`?

djc:
@BButner I don't have enough context on the nuances of gRPC versioning to answer this with any certainty. I suspect most clients will already know which version of the reflection their peer supports? If you (or anyone else) comes up with a good solution for this, kindly share it.

BButner:
@djc Yeah, that's a fair point. Specifically my use-case is with developing a tool to test gRPC servers, so being able to automatically tell which version they have is important. That _does_ sound more like a me problem, and not a you/library problem.

Right now my plan is:

1. Make the `ServerReflectionRequest` request with `v1` first.
    a. If that succeeds, done!
    b. If it fails with `Unimplemented` as the error, proceed to 2.
2. Make the `ServerReflectionRequest` with `v1alpha`.
    a. If that succeeds, done!
    b. If it fails, go to 3.
3. Assume that the server just doesn't implement Server Reflection.

djc:
Sounds about right to me.

PR #1893, opened 2024-08-26
feat(Router): permit overriding the registered route path for custom service implementation

zakhenry:
## Motivation

I am writing a generic service handler that can operate over a number of known routes (not using prost compile grpc services, but forwarding to service handlers elsewhere in the stack). 

The route builder methods all require NamedService to be implemented, and this trait is used to derive the route. This approach doesn't work for a generic service that needs to be routed many different routes. 

## Solution

Move the derivation of route path from router into the a trait method of NamedService, to change no behaviour for normal services, but allow custom services to override the method and implement it's own routing scheme.

### DIFF
--- a/tonic/src/server/mod.rs
+++ b/tonic/src/server/mod.rs
@@ -23,4 +23,11 @@ pub trait NamedService {
     ///
     /// [here]: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
     const NAME: &'static str;
+
+    /// The route path that gets registered with the underlying Axum router. This trait method
+    /// can be overridden by implementors of NamedService in order to customise the http path
+    /// routing to the service.
+    fn route_path() -> String {
+        format!("/{}/*rest", Self::NAME)
+    }
 }
--- a/tonic/src/service/router.rs
+++ b/tonic/src/service/router.rs
@@ -88,7 +88,7 @@ impl Routes {
         S::Error: Into<crate::Error> + Send,
     {
         self.router = self.router.route_service(
-            &format!("/{}/*rest", S::NAME),
+            &S::route_path(),
             svc.map_request(|req: Request<axum::body::Body>| req.map(boxed)),
         );
         self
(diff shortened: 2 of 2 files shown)


zakhenry:
Alternative approach that is worth considering. This means changing the signatures of the router to accept a new supertrait `RoutedService` Since it is a supertrait this is not a breakign change to the public API:

```rs
pub trait RoutedService {

    /// The route path that gets registered with the underlying Axum router. This trait method
    /// can be overridden by implementors of RoutedService in order to customise the http path
    /// routing to the service.
    fn route_path() -> String;
}

/// A trait to provide a static reference to the service's
/// name. This is used for routing service's within the router.
pub trait NamedService: RoutedService {
    /// The `Service-Name` as described [here].
    ///
    /// [here]: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests
    const NAME: &'static str;


}

impl<T> RoutedService for T
where
    T: NamedService,
{
    fn route_path() -> String {
        format!("/{}/*rest", Self::NAME)
    }
}
```

alexrudy:
What does this get you that can't be achieved by `into_axum_router()`, at which point you can do whatever you'd like with routing?

zakhenry:
Unless I'm mistaken, I would have to reimplement a lot of what tonic does in it's own router like the large method `serve_with_shutdown`, unless there is a way to turn an axum router back into a tonic router?

djc:
> Unless I'm mistaken, I would have to reimplement a lot of what tonic does in it's own router like the large method `serve_with_shutdown`, unless there is a way to turn an axum router back into a tonic router?

tonic's `Routes` has both `into_axum_router()` and `From<axum::Router>` (I think) implemented. Would that be sufficient?

zakhenry:
Thanks yes I was able to work around it, some fiddling with boxing of response types needed to make it the right kind for service for axum, but otherwise neat enough. I think that is good enough to not warrant a change in tonic (plus gives me more flexibility), so I'll close this.

PR #1898, opened 2024-08-28
Revert "fix(tonic-build): cargo build script outputs are not functional"

### LINKED ISSUE #1894: tonic-build can no longer build code with rust < 1.77.0, due to newer build-script syntax
### Version
0.12.2

### Platform

x86_64 GNU/Linux

### Crates

tonic-build

### Description
Since the following merge (i think)
https://github.com/hyperium/tonic/pull/1821

Rust 1.75 (which I am using) can no longer build,

The issue I get is:
```
cargo::rerun-if-changed=proto/pts_hardware.proto`
Found a `cargo::key=value` build directive which is reserved for future use.
Either change the directive to `cargo:key=value` syntax (note the single `:`) or upgrade your version of Rust.
See https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script for more information about build script outputs.
warning: build failed, waiting for other jobs to finish...
```

I believe the double :: was introduced in Cargo 1.77, if this is the case this change should be reverted or MSRV should be bumped?

[https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script)

Best Regards
Daniel

tottoto:
Reverts #1821.

Resolves #1894.

### DIFF
--- a/interop/build.rs
+++ b/interop/build.rs
@@ -4,5 +4,5 @@ fn main() {
     tonic_build::compile_protos(proto).unwrap();
 
     // prevent needing to rebuild if files (or deps) haven't changed
-    println!("cargo::rerun-if-changed={}", proto);
+    println!("cargo:rerun-if-changed={}", proto);
 }
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -551,7 +551,7 @@ impl Builder {
     }
 
     /// Enable or disable emitting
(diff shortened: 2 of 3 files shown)


djc:
Huh, weird. Thanks for fixing!

djc (approved):

PR #1899, opened 2024-08-28
chore: prepare 0.12.3 release

tottoto:


kushudai:
Could we please get https://github.com/hyperium/tonic/pull/1938 into this release as well?

djc:
More, importantly, we should consider:

- #1924 

(Which tweaks newly public API.)

Sh4d1:
@djc actually, in the current latest tonic release, if the TLS server is configured, and the server is exposed publicly (which is the case for some setup I guess), a external non authenticated user can very easily bring down the tonic server without any issues (HTTP requests or TCP packets to the TLS server). IMHO 0.12.3 should be released quickly with the fix (and maybe adding the impacted versions on rustsec? Or at least a warning somewhere)

kushudai:
> @djc actually, in the current latest tonic release, if the TLS server is configured, and the server is exposed publicly (which is the case for some setup I guess), a external non authenticated user can very easily bring down the tonic server without any issues (HTTP requests or TCP packets to the TLS server). IMHO 0.12.3 should be released quickly with the fix (and maybe adding the impacted versions on rustsec? Or at least a warning somewhere)

+1, it's a pretty serious regression from 0.12.1

djc:
The risk/severity was adequately explained in #1938 -- I agree that we should merge and release this but unfortunately I don't have the privileges to release and there's some governance discussions between the maintainers. So no need for +1 comments (which are actually universally unhelpful).

Sh4d1:
Gotcha, wasn't aware of the governance issues!

LucioFranco (approved):

LucioFranco:
I am going to try to get the other PRs we want for this release and get this out this week

Tuetuopay:
Hi, I just opened #1948 that's a continuation of #1938. While less severe, it still is the same issue, and it would be great to include it. Thanks :)

niebayes:
Looks forward for this release since the #1948 helps a lot!

zakhenry:
@LucioFranco thanks for getting this merged, what has to happen for the 0.12.3 release to be pushed to crates.io ? I see the build passed on master but I don't know the process from there.

LucioFranco:
@zakhenry for my meeting to end so I could publish :)

LucioFranco:
@zakhenry its been published, let me know if you run into any issues.

PR #1900, opened 2024-08-28
feat: plumb skip_debug through prost builder

yuandrew:
## Motivation

[skip_debug](https://docs.rs/prost-build/0.13.1/prost_build/struct.Config.html#method.skip_debug) is a method supported for prost `Config`. tonic doesn't support this method today.

## Solution

Added a new `skip_debug()` method to prost `Builder` and added a corresponding test.


djc (approved):

djc:
Makes sense, thanks!

PR #1902, opened 2024-08-29
fix(Router): add Router::into_axum_router

zakhenry:
Required as there is no replacement for the deprecated Router::into_router since the inner `Routes` is private with no getter

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -770,8 +770,13 @@ impl<L> Router<L> {
     }
 
     /// Convert this tonic `Router` into an axum `Router` consuming the tonic one.
-    #[deprecated(since = "0.12.2", note = "Use `Routes::into_axum_router` instead.")]
+    #[deprecated(since = "0.12.2", note = "Use `into_axum_router` instead.")]
     pub fn into_router(self) -> axum::Router {
+        self.routes.into_axum_router()
+    }
+
+    /// Convert this tonic `Router` into an axum `Router` consuming the tonic one.
+    pub fn into_axum_router(self) -> axum::Router {
         self.routes.into_axum_router()
     }
 


tottoto:
When users want to get axum's router, it should be created from `Routes` directly. And the confusing Router's API will be removed in the future breaking change.

zakhenry:
This would be workable if we had an `add_routes` method on `impl<L> Router<L>`, but because we don't (add_routes exists only on `impl<L> Server<L>`), you can only have explicit Routes registered *before* tonic services, which doesn't work if the Routes addition contains wildcard services intended as fallback.

Would the alternative option of adding `impl<L> Router<L>::add_routes(&mut self, routes: Routes) -> Self` be a better change to be raising?

tottoto:
I am not sure what you want to do. Do you mean that the axum's router obtained via `Router` is different from the one obtained via `Routes`?

zakhenry:
@tottoto no it is the same router, the issue is that ordering of service registration matters and with the current (non deprecated) methods available it is only possible to call `add_routes` _before_ any other `add_service` call. 

For my purposes that doesn't work as the service I'm defining with `add_routes` is a catchall fallback service for when my actual regular code-generated tonic services haven't handled the route. 

For now I can work around that by using the existing deprecated `into_router` call after using the server builder to register  the tonic services first, but that method is deprecated without an alternative.

zakhenry:
to use an example from the example in #1884 , what I *want* to do is the following:

```rs
#[tokio::main]

    let router = axum::Router::new().route_service(
        "/*catchall",
        ProxyService::new("http://[::1]:50051".parse().unwrap()),
    );

    Server::builder()
        .add_service(GreeterServer::new(greeter))
        .add_routes(Routes::from(router))
        .serve("[::1]:50050".parse().unwrap())
        .await?;

    Ok(())
}
```

however it doesn't compile:
```
error[E0599]: no method named `add_routes` found for struct `tonic::transport::server::Router` in the current scope
   --> examples/src/proxy/proxy.rs:196:10
    |
194 |       Server::builder()
    |       -----------------
    |       |
    |  _____method `add_routes` is available on `&mut Server`
    | |
195 | |         .add_service(GreeterServer::new(greeter))
196 | |         .add_routes(Routes::from(router))
    | |         -^^^^^^^^^^ method not found in `Router`
    | |_________|
    |

```

So instead I need to do 

```rs
#[tokio::main]

    let router = Server::builder()
        .add_service(GreeterServer::new(greeter))
        .into_router()
        .route_service(
            "/*catchall",
            ProxyService::new("http://[::1]:50051".parse().unwrap()),
        );

    Server::builder()
        .add_routes(Routes::from(router))
        .serve("[::1]:50050".parse().unwrap())
        .await?;

    Ok(())
}
```

This compiles, and has the service ordering as desired (concrete service before proxy fallback service). Unfortunately it relies on the deprecated `into_router` call.

zakhenry:
Here's the alternative change; to add `add_routes` to both `Router` & `Routes`. This performs an internal merge on the axum router. This is kinda neat too more generally as it allows composing larger sets of services together (including tonic and non-tonic tower services.

```patch
diff --git a/tonic/src/service/router.rs b/tonic/src/service/router.rs
index 7329d00..f3b415a 100644
--- a/tonic/src/service/router.rs
+++ b/tonic/src/service/router.rs
@@ -94,6 +94,11 @@ impl Routes {
         self
     }
 
+    pub fn add_routes(mut self, routes: Routes) -> Self {
+        self.router = self.router.merge(routes.router);
+        self
+    }
+
     /// This makes axum perform update some internals of the router that improves perf.
     ///
     /// See <https://docs.rs/axum/latest/axum/routing/struct.Router.html#a-note-about-performance>
diff --git a/tonic/src/transport/server/mod.rs b/tonic/src/transport/server/mod.rs
index 62006e1..acebb2f 100644
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -748,6 +748,11 @@ impl<L> Router<L> {
         self
     }
 
+    pub fn add_routes(mut self, routes: Routes) -> Self {
+        self.routes = self.routes.add_routes(routes);
+        self
+    }
+
     /// Add a new optional service to this router.
     ///
     /// # Note

```

tottoto:
I still don't understand what your problem is. The following do not work?

```rs
let router = Routes::new(GreeterServer::new(greeter))
    .into_axum_router()
    .route_service(
        "/*catchall",
        ProxyService::new("http://[::1]:50051".parse().unwrap()),
    );
```

tottoto:
I still don't understand what your problem is. The following do not work?

```rs
let router = Routes::new(GreeterServer::new(greeter))
    .into_axum_router()
    .route_service(
        "/*catchall",
        ProxyService::new("http://[::1]:50051".parse().unwrap()),
);
```

tottoto:
I still don't understand what your problem is. The following do not work?

```rs
let router = Routes::new(GreeterServer::new(greeter))
    .into_axum_router()
    .route_service(
        "/*catchall",
        ProxyService::new("http://[::1]:50051".parse().unwrap()),
    );
```

zakhenry:
@tottoto ok yes thank you this does work without having to use deprecated API. I do still think it would be neater and more intuitive to have the `add_routes` available on Router, so that custom routes can be built off a single chain from `Server::builder()` as in all the examples.

Closing for now, as the `add_routes` suggestion is quite separate to this original PR.

PR #1907, opened 2024-08-31
Revert "chore(Dev): ignore .idea directory that Intellij IDEs add automatically"

tottoto:
Removes a git config which is not related to tonic project.

Reverts hyperium/tonic#1903.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@ target/
 **/*.rs.bk
 Cargo.lock
 tags
-.idea


zakhenry:
Worth discussing, it is not uncommon to have editor ignores in other projects. Examples:

https://github.com/rust-lang/rust/blob/master/.gitignore
https://github.com/rust-lang/cargo/blob/master/.gitignore

djc:
I argued against this in #1903.

PR #1910, opened 2024-09-02
chore(ci): Add clippy job to check error

tottoto:
Adds `clippy` job to check only error to detect some kind of error like #1908 which is difficult to find.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -28,6 +28,19 @@ jobs:
     - uses: actions/checkout@v4
     - uses: EmbarkStudios/cargo-deny-action@v2
 
+  clippy:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


tottoto:
@LucioFranco Although I think you would not like running clippy on CI as you removed it #1288, I would like to ask your opinion. This time, setting it to only detect errors in order to reduce the possibility of blocking CI. I would think it is valuable because it allows us to notice important aspects.

djc (approved):
I like it -- in most projects we use `-D warnings` for clippy in CI, and have a weekly scheduled job running on Friday mornings to try and find issues from new releases quickly.

tottoto:
Let's try applying it at this stage and see what the impact is. If it becomes an overhead for development, I think we can easily remove it again.

PR #1915, opened 2024-09-03
feat(build): Add api to generate from file descriptor set

### LINKED ISSUE #1489: Offer optional support for protox in tonic-build
## Feature Request

If I understand correctly, using https://github.com/andrewhickman/protox should make it possible to avoid the use of protoc binaries. It would be great if tonic-build offered built-in API (using a Cargo feature) to leverage protox for schema compilation.

### Crates

tonic-build

### Motivation

Avoid the need for installing protoc on developer systems and in build containers.

tottoto:
Adds APIs to generate codes from file descriptor set.

### DIFF
--- a/tonic-build/Cargo.toml
+++ b/tonic-build/Cargo.toml
@@ -18,12 +18,13 @@ version = "0.12.2"
 prettyplease = { version = "0.2" }
 proc-macro2 = "1.0"
 prost-build = { version = "0.13", optional = true }
+prost-types = { version = "0.13", optional = true }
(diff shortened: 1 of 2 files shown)


djc:
IMO this doesn't really solve that issue -- I think we should at least have API that depends on protox directly (optionally enabled via a Cargo feature) and uses it (instead of protoc).

tottoto:
Updated the description to make this a separate pull request from #1489.

djc (approved):

PR #1917, opened 2024-09-04
chore(tls): Update to rustls-native-certs 0.8

tottoto:
Updates to `rustls-native-certs` 0.8.

### DIFF
--- a/tonic/src/transport/channel/service/tls.rs
+++ b/tonic/src/transport/channel/service/tls.rs
@@ -34,7 +34,15 @@ impl TlsConnector {
 
         #[cfg(feature = "tls-native-roots")]
         if with_native_roots {
-            roots.add_parsable_certificates(rustls_native_certs::load_native_certs()?);
+            let rustls_native_certs::CertificateResult { certs, errors, .. } =
+                rustls_native_certs::load_native_certs();
+            if !errors.is_empty() {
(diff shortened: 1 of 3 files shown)


tonic/src/transport/channel/service/tls.rs
```diff
@@ -34,7 +34,10 @@ impl TlsConnector {
 
         #[cfg(feature = "tls-native-roots")]
         if with_native_roots {
-            roots.add_parsable_certificates(rustls_native_certs::load_native_certs()?);
+            let rustls_native_certs::CertificateResult { certs, errors, .. } =
```
djc, line 37: Suggest yielding an error if `certs.is_empty()`, at least.

tottoto: Updated to return a error when the certs is empty.

tonic/src/transport/service/tls.rs
```diff
@@ -23,6 +25,8 @@ impl fmt::Display for TlsError {
         match self {
             #[cfg(feature = "channel")]
             TlsError::H2NotNegotiated => write!(f, "HTTP/2 was not negotiated."),
+            #[cfg(feature = "tls-native-roots")]
+            TlsError::NativeCertsNotFound => write!(f, "native certs not found"),
```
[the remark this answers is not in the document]
tottoto: Updated to the suggestion.

PR #1918, opened 2024-09-04
feat(tls): Add config to set TrustAnchor

tottoto:
Adds configs to set `TrustAnchor`.

### DIFF
--- a/tonic/src/transport/channel/tls.rs
+++ b/tonic/src/transport/channel/tls.rs
@@ -4,12 +4,14 @@ use crate::transport::{
     Error,
 };
 use http::Uri;
+use tokio_rustls::rustls::pki_types::TrustAnchor;
 
 /// Configures TLS settings for endpoints.
 #[derive(Debug, Clone, Default)]
 pub struct ClientTlsConfig {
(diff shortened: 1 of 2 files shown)


tonic/src/transport/channel/tls.rs
```diff
@@ -46,6 +48,32 @@ impl ClientTlsConfig {
...
+    /// Adds the multiple trust anchors which to verify the server's TLS certificate.
+    pub fn trust_anchors(
+        self,
+        trust_anchors: impl IntoIterator<Item = TrustAnchor<'static>>,
+    ) -> Self {
+        let trust_anchors = {
+            let mut tmp = self.trust_anchors;
+            tmp.extend(trust_anchors);
+            tmp
+        };
```
djc, line 70: Seems like a lot of syntactic overhead for the benefit of not having `mut` on `self`? I would just write `self.trust_anchors.extend(trust_anchors)`.

tottoto: Updated.

djc (approved):

PR #1925, opened 2024-09-05
feat(build): Reexport types related to prost

tottoto:
Reexports types which are used with builder based on `prost`.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -81,6 +81,10 @@ use quote::TokenStreamExt;
 /// Prost generator
 #[cfg(feature = "prost")]
 mod prost;
+#[cfg(feature = "prost")]
+pub use prost_build::Config;
+#[cfg(feature = "prost")]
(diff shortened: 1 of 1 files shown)


tonic-build/src/lib.rs
```diff
@@ -81,6 +81,10 @@ use quote::TokenStreamExt;
 /// Prost generator
 #[cfg(feature = "prost")]
 mod prost;
+#[cfg(feature = "prost")]
+pub use prost_build::Config as ProstBuildConfig;
+#[cfg(feature = "prost")]
+pub use prost_types::FileDescriptorSet;
```
djc, line 87: Are these already used in tonic-build's public API? Where/how?

djc: Given that `Config` is used as `Config` in those functions and given that there is no type definition called `Config` anywhere in the tonic repo, IMO it would be clearer to just import `Config` without aliasing it.

djc (approved):

PR #1926, opened 2024-09-05
chore(doc): Add document about how prost-build config is used

tottoto:
Adds document about how `prost-build`'s config is used.

### DIFF
--- a/tonic-build/src/prost.rs
+++ b/tonic-build/src/prost.rs
@@ -610,8 +610,8 @@ impl Builder {
         self.compile_protos(protos, includes)
     }
 
-    /// Compile the .proto files and execute code generation using a
-    /// custom `prost_build::Config`.
+    /// Compile the .proto files and execute code generation using a custom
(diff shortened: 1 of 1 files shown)


tonic-build/src/prost.rs
```diff
@@ -610,8 +610,8 @@ impl Builder {
         self.compile_protos(protos, includes)
     }
 
-    /// Compile the .proto files and execute code generation using a
-    /// custom `prost_build::Config`.
+    /// Compile the .proto files and execute code generation using a custom
+    /// `prost_build::Config`. The provided config will be updated by this builder's config.
```
djc, line 614: Nit: suggest replacing "by" with "with" (also for the other method).

djc (approved):

tonic-build/src/prost.rs
tottoto: Applied the suggestion.

PR #1927, opened 2024-09-05
chore(test): Move integration test for tonic-web to tests directory

tottoto:
Moves integration test for `tonic-web` to tests directory where other integration tests are located.


tests/integration_web/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 authors = ["Juan Alvarez <j@yabit.io>"]
 edition = "2021"
-name = "integration"
+name = "integration_web"
```
djc, line 4: I'm not a fan of the `integration_` prefix here. Everything in `/tests` is an integration test by default, and adding the prefix here doesn't seem useful -- just call it `/tests/web`?

djc (approved):

tests/integration_web/Cargo.toml
tottoto: Changed to `/tests/web`.

PR #1933, opened 2024-09-09
Use constants for header names

djc:


LucioFranco:
> making them associated constants for Status doesn't seem like the good place to do so, considering that Status itself is not a header

I actually don't agree, we convert these headers in Status and its good to encapsulate the encoding portion of these header names within the status.rs. So to me this makes sense. Metadata is much more abstract than specific extraction of headers. In addition, the metadata shouldn't contain the status headers etc that should all be extracted from the Status functions.

tonic/src/status.rs
```diff
@@ -588,6 +584,13 @@ impl Status {
         self.add_header(response.headers_mut()).unwrap();
         response
     }
+
+    /// Header name for the `grpc-status` header
+    pub const GRPC_STATUS: HeaderName = HeaderName::from_static("grpc-status");
+    /// Header name for the `grpc-message` header
+    pub const GRPC_MESSAGE: HeaderName = HeaderName::from_static("grpc-message");
+    /// Header name for the `grpc-status-details-bin` header
+    pub const GRPC_STATUS_DETAILS: HeaderName = HeaderName::from_static("grpc-status-details-bin");
```
LucioFranco, line 593: Im kinda torn on exposing these publically,I know we need to use them in tonic-web but I wonder if we could just do doc(hidden)?

djc: `#[doc(hidden)]` seems fine to me, will change.

tottoto:
@LucioFranco

>> making them associated constants for Status doesn't seem like the good place to do so, considering that Status itself is not a header
>
> I actually don't agree, we convert these headers in Status and its good to encapsulate the encoding portion of these header names within the status.rs. So to me this makes sense. Metadata is much more abstract than specific extraction of headers. In addition, the metadata shouldn't contain the status headers etc that should all be extracted from the Status functions.

I understand that the abstraction of metadata is a high level one. Then, I think it's fine to use Status's associated constants, considering that these will be `doc(hidden)` and will not actually be made into public API.

djc:
Made them `#[doc(hidden)]` as requested.

LucioFranco (approved):

PR #1937, opened 2024-09-09
build: suppress `clippy::wildcard_imports` in generated code

thecaralice:
## Motivation

The `clippy::wildcard_imports` is not enabled by default, but some projects might enable it, causing tonic-build to generate code failing to lint.

## Solution

`#[allow(clippy::wildcard_imports)]` in generated code

### DIFF
--- a/tonic-build/src/client.rs
+++ b/tonic-build/src/client.rs
@@ -49,6 +49,7 @@ pub(crate) fn generate_internal<T: Service>(
                 unused_variables,
                 dead_code,
                 missing_docs,
+                clippy::wildcard_imports,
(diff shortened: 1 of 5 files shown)


djc:
This makes sense to me, thanks!

thecaralice:
I just noticed there is another lint in the generated code. Should I create a separate PR or I can do everything here?

djc:
> I just noticed there is another lint in the generated code. Should I create a separate PR or I can do everything here?

A separate commit in this PR would be great. 👍

thecaralice:
That second lint turned out to be an issue in prost-build, nothing to do with tonic

PR #1938, opened 2024-09-12
fix(tls): do not shutdown server on unexpected Eof

Sh4d1:
## Motivation

When receiving non TLS packets (like TCP healtchecks), the server shuts down due to the UnexpectedEof returned by rustls.

## Solution

Ignore this error


Sh4d1:
cc @djc if we can get that into 0.12.3

djc (approved):

djc:
LGTM!

Tuetuopay:
Hi! Just chiming in to say thank you!

We're hitting this issue, which is even worse for us as we have long-lived gRPC streams, that prevent the tonic server from exiting, effectively DoS-ing our APIs if for whatever reason we get non-tls tcp connection.

LucioFranco (approved):

PR #1952, opened 2024-09-21
feat(build): Make cleanup-markdown feature additive

tottoto:
Makes cleanup-markdown feature additive.

### DIFF
--- a/tonic-build/Cargo.toml
+++ b/tonic-build/Cargo.toml
@@ -25,7 +25,7 @@ syn = "2.0"
 [features]
 default = ["transport", "prost"]
 prost = ["prost-build", "dep:prost-types"]
-cleanup-markdown = ["prost", "prost-build/cleanup-markdown"]
+cleanup-markdown = ["prost-build?/cleanup-markdown"]
 transport = []
 
 [package.metadata.docs.rs]
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -4,7 +4,8 @@
 //! # Feature flags
 //!
 //! - `cleanup-markdown`: Enables cleaning up documentation from the generated code. Useful
-//!   when documentation of the generated code fails `cargo test --doc` for example.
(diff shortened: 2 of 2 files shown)


tonic-build/Cargo.toml
```diff
@@ -25,7 +25,7 @@ syn = "2.0"
 [features]
 default = ["transport", "prost"]
 prost = ["prost-build", "dep:prost-types"]
-cleanup-markdown = ["prost", "prost-build/cleanup-markdown"]
+cleanup-markdown = ["prost-build?/cleanup-markdown"]
```
djc, line 28: Why is the previous setup non-additive?

tottoto: The `cleanup-markdown` feature does not assume `prost` in terms of meaning, so I used the term additive here in the sense that it will remove the behavior that implicitly enables `prost` and make it possible to be compatible with other codecs in the future. In fact, it is additive at present, so there may be a more appropriate representation.

djc: Should there be some added documentation to make it clearer that `cleanup-markdown` will do nothing unless `prost` is also enabled?

tonic-build/src/lib.rs
```diff
@@ -4,7 +4,8 @@
 //! # Feature flags
 //!
 //! - `cleanup-markdown`: Enables cleaning up documentation from the generated code. Useful
-//!   when documentation of the generated code fails `cargo test --doc` for example.
+//!   when documentation of the generated code fails `cargo test --doc` for example. `prost`
```
djc, line 7: I'd suggest adding "The" here: "The `prost` feature [..]".

djc (approved):

tonic-build/Cargo.toml
tottoto: Updated the document.

tonic-build/src/lib.rs
tottoto: Applied the suggestion.

djc (approved):

PR #1953, opened 2024-09-23
chore: Make codegen independent from protoc

tottoto:
Makes `codegen` independent from `protoc` by using `protox`. This eliminates the need to match `protoc` versions in development, and separates the public crates from `protoc`.


codegen/src/main.rs
```diff
@@ -105,3 +114,10 @@ fn codegen(
         std::fs::copy(&path, &to).unwrap();
     }
 }
+
+fn write_fds(fds: &FileDescriptorSet, path: &Path) {
+    let mut writer = BufWriter::new(std::fs::File::create(path).unwrap());
```
djc, line 119: Nit: I'd suggest importing `File` instead.

djc (approved):
Nice!

codegen/src/main.rs
tottoto: Applied the suggestion.

tottoto: Applied the suggestion.

PR #1956, opened 2024-09-25
chore: Generate file descriptor set constant directly

tottoto:
Changes to generate file descriptor set constants directly instead of exporting them as files and reincluding them.


codegen/src/main.rs
```diff
@@ -117,8 +118,19 @@ fn codegen(
...
+    let fds_raw = buf.as_ref();
+    let tokens = quote! {
+        /// Byte encoded FILE_DESCRIPTOR_SET.
+        pub const FILE_DESCRIPTOR_SET: &[u8] = &[#(#fds_raw),*];
+    };
+
+    let content = format!("{GENERATED_COMMENT}\n{tokens}\n",);
+
+    let mut writer = BufWriter::new(File::create(path).unwrap());
+    writer.write_all(content.as_bytes()).unwrap();
```
djc, line 135: Maybe do some formatting, either with prettyplease or `rustfmt`?

codegen/Cargo.toml
```diff
@@ -10,3 +10,4 @@ version = "0.1.0"
 tempfile = "3.8.0"
 protox = "0.7"
 tonic-build = {path = "../tonic-build", default-features = false, features = ["prost", "cleanup-markdown"]}
+quote = "1"
```
djc, line 13: Nit: keep these in alphabetic order?

djc (approved):
Nice!

codegen/src/main.rs
tottoto: I've considered and tried formatting, but in this case, it does not seem helpful for up as they are mere slices of bytes resulting in a lot of newlines by each elements.

codegen/Cargo.toml
tottoto: Applied the suggestion.

codegen/src/main.rs
djc: Okay. IMO reviewing future changes to these files will probably still be easier if they're formatted more like "normal" Rust files.

tottoto: Sounds reasonable to me. Changed to format them.

djc (approved):

GitHub23114:
> Changes to generate file descriptor set constants directly instead of exporting them as files and reincluding them.

Ghost

PR #1958, opened 2024-09-26
feat(tonic-web): Create enable_with_cors method

mrmeku:
## Motivation

I've not had any luck using the recommended ways to enable custom CORS for GRPC-web using the recommended mechanisms.

## Solution

 This helper method is a much easier escape hatch for custom CORS layering. Instead of calling `tonic_web::enable(service)` which uses a default CORS layer you instead call `tonic_web::enable_with_cors(service, cors_layer)` with your own cors layer.

### DIFF
--- a/tonic-web/src/lib.rs
+++ b/tonic-web/src/lib.rs
@@ -127,7 +127,7 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
 
 /// Enable a tonic service to handle grpc-web requests with the default configuration.
 ///
-/// You can customize the CORS configuration composing the [`GrpcWebLayer`] with the cors layer of your choice.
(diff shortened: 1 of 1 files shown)


djc (approved):
Seems reasonable.

tottoto:
For the next breaking change, considering making `tower` and `tower-http` crates internal dependencies to make `tonic` project less impactful when updates are made to the dependencies. Can the config be made to be `tower_layer::Layer` instead of `tower_http::cors::CorsLayer`?

tottoto:
For the future breaking change, considering making `tower` and `tower-http` crates internal dependencies to make `tonic` project less impactful against the updates of the dependencies. Can the config be made to be `tower_layer::Layer` instead of `tower_http::cors::CorsLayer`?

PR #1966, opened 2024-10-01
feat(service): Add Layered service to propagate NamedService implementation

tottoto:
Adds `Layered` service to propagate `NamedService` implementation.

Resolves #1958.


tottoto:
@mrmeku Does this meet your requirements?

tonic/src/service/layered.rs
```diff
@@ -0,0 +1,48 @@
...
+
+/// A layered service to propagate [`NamedService`] implementation.
+#[derive(Debug, Clone)]
+pub struct Layered<S, T> {
+    inner: S,
+    _ty: PhantomData<T>,
+}
+
+/// Create a new service from the service and the layer.
+pub fn layered_service<S, L>(service: S, layer: L) -> Layered<L::Service, S>
```
djc, line 19: It feels more idiomatic to me to make rename this to `Layered::new()`, but I'm not very familiar with the conventions in the tower ecosystem.

tottoto: The current design is a result of the type system. It can also be the constructor of the associated function, but in that case `Layered` requires one more type parameter to constrain the type of the `Layer`.

mrmeku:
> @mrmeku Does this meet your requirements?

Absolutely. Great solution

tonic/src/service/layered.rs
djc: Not sure I understand this. Can you demonstrate?

tottoto:
@mrmeku Thanks for confirming.

tonic/src/service/layered.rs
tottoto: Added a commit to show it.

tottoto: Considering that a common pattern for applying a layer to a service to create a new service is to use `layer.layer(service)` rather than providing a constructor for the service, it might be more appropriate to provide a trait that wraps the service with Layered when applying a layer to the service rather than providing a constructor.

djc (approved):

djc (approved):

tottoto:
Squashed the commits.

PR #1967, opened 2024-10-01
feat: Reexport fundamental tower trait

tottoto:
Reexports fundamental `tower`'s traits.


djc:
Why?

tottoto:
As tonic is based on tower's ecosystem and these types are already exposed, it is useful when users use with tower's functionality.

tottoto:
The initial motivation was that I felt combining services and layers is frequent, but this doesn't seem convenient as I felt since a pattern using `tower::ServiceBuilder` and `LayerExt` seems simpler in practice.

PR #1973, opened 2024-10-05
feat(server): Remove concurrency limit built in feature

tottoto:
Removes concurrency limit build in feature. As the server is composable with tower's services, this seems not to be needed to be built in.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -173,6 +173,10 @@ impl<L> Server<L> {
     /// builder.concurrency_limit_per_connection(32);
     /// ```
     #[must_use]
+    #[deprecated(
+        since = "0.12.4",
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -173,6 +173,10 @@ impl<L> Server<L> {
     /// builder.concurrency_limit_per_connection(32);
     /// ```
     #[must_use]
+    #[deprecated(
+        since = "0.12.4",
+        note = "Use `layer()` with a layer which provide concurrency limit feature such as `tower::limit::ConcurrencyLimitLayer` used in this config."
```
shikhar, line 178: I don't think that will provide 'per connection' concurrency-limiting semantics like this purports to.

tottoto: Makes sense. Seems necessary to add functionality to set layers for the purpose.

shikhar: I think that would be helpful, see e.g. https://github.com/hyperium/tonic/pull/1616

PR #1974, opened 2024-10-05
feat(tls): Add rustls-platform-verifier support

tottoto:
Adds `rustls-platform-verifier` support.


tottoto:
These are needed for macos and windows.

https://github.com/rustls/rustls-platform-verifier/pull/133
https://github.com/rustls/rustls-platform-verifier/pull/135

tobz:
It would be good to get @djc's thoughts here, since the benefits listed by `rustls-platform-verifier` itself seem good, albeit small... and it's not clear to me if there are more and/or worse platform support caveats using this vs sticking with `rustls-native-certs`.

djc:
As @tottoto suggested in a previous comment (thought it might not be quite clear), released versions of rustls-platform-verifier don't have all the API in place to support additivity with native-roots and webpki-roots across platforms, so might want to wait for that.

https://github.com/rustls/rustls-platform-verifier/pull/148

PR #1977, opened 2024-10-06
docs: improve comment around cloning middleware in examples

shikhar:
~tower does not seem to use the buffer middleware (anymore?) so the existing comment is invalid~, but the pattern is still sensible given https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services


djc (approved):

tottoto:
Could you conforms the commit's message to conventional-commits which conventional-changelog bases on, used in this project.

shikhar:
> tower does not seem to use the buffer middleware (anymore?) so the existing comment is invalid

I see it is actually used on the client-side, https://github.com/hyperium/tonic/blob/be4e0e772112b7fdd879e95398a7b50d305bad3e/tonic/src/transport/channel/mod.rs#L68

But I think the reference to tower docs is probably more useful.

tottoto (approved):

PR #1978, opened 2024-10-06
Make `Status::into_http()` generic over any kind of `Default`-able Body

shikhar:
## Motivation

We have middleware that works for both gRPC and non-gRPC requests where it would be useful to leverage `Status::into_http()` in case the request content-type indicates it was a gRPC request, but don't want to couple to tonic's `BoxBody`.

For creating an empty body, relying on `Default` seems pretty reasonable and is an existing pattern in the ecosystem e.g. https://github.com/tower-rs/tower-http/blob/66dd321a73f87ab204011f07853e156f81cfeb7b/tower-http/src/timeout/service.rs#L109

## Solution

Make `into_http` generic over the body so long as it implements `Default`.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -1,5 +1,5 @@
 use crate::metadata::MetadataMap;
-use crate::{body::BoxBody, metadata::GRPC_CONTENT_TYPE};
+use crate::metadata::GRPC_CONTENT_TYPE;
 use base64::Engine as _;
 use bytes::Bytes;
 use http::{
@@ -579,8 +579,8 @@ impl Status {
     }
 
     /// Build an `http::Response` from the given `Status`.
-    pub fn into_http(self) -> http::Response<BoxBody> {
-        let mut response = http::Response::new(BoxBody::default());
+    pub fn into_http<B: Default>(self) -> http::Response<B> {
+        let mut response = http::Response::new(B::default());
         response
             .headers_mut()
             .insert(http::header::CONTENT_TYPE, GRPC_CONTENT_TYPE);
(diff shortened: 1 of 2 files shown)


tottoto (commented):
Looks good to me.

tonic/src/status.rs
```diff
@@ -579,8 +579,8 @@ impl Status {
     }
 
     /// Build an `http::Response` from the given `Status`.
-    pub fn into_http(self) -> http::Response<BoxBody> {
-        let mut response = http::Response::new(crate::body::empty_body());
+    pub fn into_http<B: Default>(self) -> http::Response<B> {
```
djc, line 582: Maybe it makes sense to add another method for this instead, so that this change could be made in a semver-compatible release? (I guess it depends on your context whether that's useful/feasible.)

shikhar: That is useful and feasible, but I think breakage is relatively unlikely, since `BoxBody` will be type-inferred in most cases. Even if there is breakage, it is going to be in more advanced usage e.g. middleware.

djc: This PR is currently marked as a breaking change. If you think that is not justified, I think it would be good to make that argument in a bit more detail.

tottoto: According to [RFC 1105](https://rust-lang.github.io/rfcs/1105-api-evolution.html), this may be treated as a minor change.

tottoto: According to [RFC 1105](https://rust-lang.github.io/rfcs/1105-api-evolution.html), this might be treated as a minor change.

tottoto (approved):

djc (approved):

PR #1981, opened 2024-10-07
chore: Refactor empty body constructor

tottoto:
Refactors empty body constructor.

### DIFF
--- a/tonic/src/body.rs
+++ b/tonic/src/body.rs
@@ -16,7 +16,5 @@ where
 
 /// Create an empty `BoxBody`
 pub fn empty_body() -> BoxBody {
-    http_body_util::Empty::new()
-        .map_err(|err| match err {})
-        .boxed_unsync()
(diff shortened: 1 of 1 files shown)


djc (approved):

tonic/src/body.rs
```diff
@@ -16,7 +16,5 @@ where
 
 /// Create an empty `BoxBody`
 pub fn empty_body() -> BoxBody {
-    http_body_util::Empty::new()
-        .map_err(|err| match err {})
-        .boxed_unsync()
+    BoxBody::default()
```
djc, line 19: I would suggest we also `#[deprecate]` this function in favor of using `Default`.

djc (approved):

tonic/src/body.rs
tottoto: Deprecated it.

djc (approved):

PR #1982, opened 2024-10-07
feat(web): Remove enable api

tottoto:
Removes `enable()` api.


tonic-web/src/service.rs
```diff
@@ -255,6 +256,27 @@ mod tests {
...
+        )
+        .named_layer(service)
+    }
+
+    fn get_name_of_named_service<S: NamedService>(_: &S) -> &'static str {
+        S::NAME
+    }
+
+    #[test]
+    fn named_service_is_propagated_by_enabled_helper_function() {
```
djc, line 275: What are we trying to test here? I don't see the value of this test (especially considering that it needs the trivial utility function).

djc (approved):

PR #1983, opened 2024-10-07
feat: Add tcp setting getters to endpoint

Benjscho:
Add getters for tcp settings to Endpoints. The purpose for this is that `connect_with_connector` doesn't set TCP settings on the socket, because it's using a provided connector. In the cases where a library is creating a custom TCP connection, we want to be able to respect user tcp settings, in particular connection timeout. Adding these getters allows this.

## Motivation

The purpose for this is that `connect_with_connector` doesn't set TCP settings on the socket, because it's using a provided connector. In the cases where a library is creating a custom TCP connection, we want to be able to respect user tcp settings, in particular connection timeout. Adding these getters allows this.

## Solution

Adds simple getters for existing tcp settings on an endpoint.

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -415,6 +415,25 @@ impl Endpoint {
     pub fn uri(&self) -> &Uri {
         &self.uri
     }
+
+    /// Get the value of `TCP_NODELAY` option for accepted connections.
+    pub fn get_tcp_nodelay(&self) -> bool {
+        self.tcp_nodelay
+    }
+
+    /// Get the connect timeout.
+    pub fn get_connect_timeout(&self) -> Option<Duration> {
+        self.connect_timeout
+    }
+
+    /// Get whether TCP keepalive messages are enabled on accepted connections.
+    ///
+    /// If `None` is specified, keepalive is disabled, otherwise the duration
+    /// specified will be the time to remain idle before sending TCP keepalive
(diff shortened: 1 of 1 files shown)


tonic/src/transport/channel/endpoint.rs
```diff
@@ -415,6 +415,25 @@ impl Endpoint {
     pub fn uri(&self) -> &Uri {
         &self.uri
     }
+
+    /// Get the value of `TCP_NODELAY` option for accepted connections.
+    pub fn get_tcp_nodelay(&self) -> bool {
```
djc, line 420: Adding these makes sense to me, though we should skip the `get_()` prefixes per https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter.

ikrivosheev: @djc I also wanted add comment like this, but the looked here: https://github.com/hyperium/tonic/blob/b4ebfc3e820dab8e4bc1b41e28bdf0965b53c441/tonic/src/transport/channel/endpoint.rs#L260.

djc: Fair point. I guess this is okay, then.

djc (approved):

ikrivosheev (approved):

tottoto (approved):

djc:
Thanks!

PR #1990, opened 2024-10-10
server: swallow TLS errors in the accept loop

PDXKimani:
## Motivation

https://github.com/hyperium/tonic/issues/1989

A [previous commit](https://github.com/hyperium/tonic/commit/c3be20c86e1a6dfa3523b2d77e8c503d0f5b2ce3) altered the handling of errors during TLS handshakes. Prior to that change, the various and sundry error IO errors which can arise during a handshake were ignored. The commit in question changed the behavior to instead exit the accept loop (and thus shut down the server) if any IO error which is not explicitly listed occurs. Since that commit, several error kinds have been added to the list piecemeal to address shutdowns, but the tail of errors which might need to be added is likely long.

## Solution

This change aims to return the handling of TLS errors to the previous state where all errors from `tls.accept` would be swallowed.

This "de-unifies" the handling for those two classes of error, but I believe this to be the correct approach to preserve server stability - TLS handshake errors should not generally be fatal to the server.

### DIFF
--- a/tonic/src/transport/server/incoming.rs
+++ b/tonic/src/transport/server/incoming.rs
@@ -74,12 +74,12 @@ where
                     yield io;
                 }
 
-                SelectOutput::TcpAcceptErr(e) => match handle_tcp_accept_error(e) {
+                SelectOutput::TcpErr(e) => match handle_tcp_accept_error(e) {
                     ControlFlow::Continue(()) => continue,
                     ControlFlow::Break(e) => Err(e)?,
                 }
 
-                SelectOutput::TlsAcceptErr(e) => {
+                SelectOutput::TlsErr(e) => {
                     tracing::debug!(error = %e, "tls accept error");
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/incoming.rs
```diff
@@ -87,7 +92,7 @@ where
     }
 }
 
-fn handle_accept_error(e: impl Into<crate::Error>) -> ControlFlow<crate::Error> {
+fn handle_tcp_accept_error(e: impl Into<crate::Error>) -> ControlFlow<crate::Error> {
```
krispraws, line 95: Since this function only handles TCP accept errors, we can get rid of the error codes that cannot be returned by `accept`

tonic/src/transport/server/incoming.rs
```diff
@@ -148,7 +153,8 @@ where
 enum SelectOutput<A> {
     Incoming(A),
     Io(ServerIo<A>),
-    Err(crate::Error),
+    TcpAcceptErr(crate::Error),
+    TlsAcceptErr(crate::Error),
```
djc, line 157: Nit: let's just call these `Tcp` and `Tls`?

tonic/src/transport/server/incoming.rs
PDXKimani: I've removed the ones which were obviously TLS-related. I know ConnectionAborted, WouldBlock, and TimeOut are possible from `accept(2)` and are supposed to be treated as retryable. The others are uncertain to me, but I do not think that their inclusion here is explicitly harmful.

krispraws (approved):
LGTM

djc (approved):

tottoto (approved):

PR #1996, opened 2024-10-12
feat(server): Remove into_service api

tottoto:
Removes `into_service()` api. This is the same as the one composed of the layers and the `Routes`.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -895,6 +895,7 @@ impl<L> Router<L> {
     }
 
     /// Create a tower service out of a router.
+    #[deprecated(since = "0.12.4", note = "compose the layers and the `Routes`")]
     pub fn into_service<ResBody>(self) -> L::Service
     where
         L: Layer<Routes>,


tonic/src/transport/server/mod.rs
```diff
@@ -897,6 +897,7 @@ impl<L> Router<L> {
     }
 
     /// Create a tower service out of a router.
+    #[deprecated(since = "0.12.4", note = "compose the layers and the `Routes`")]
```
djc, line 900: So how would you spell this in 0.13?

tottoto: What is the purpose? Should this api be kept in 0.13?

djc: I'm asking what the replacement looks like, in order to better understand how much boilerplate this call is hiding.

tottoto: I see. In the case where this API is used, users can build `Routes` in advance, so they can set some layers to the `Routes` with some means, for example `tower::ServiceBuilder`.

tottoto: I see. In the case where this API is used, users can build `Routes` in advance, so they can set some layers to the `Routes` with some means, for example `tower`'s `ServiceBuilder`.

djc: I understand -- do you have a code snippet showing what that would look like?

tottoto:
> At a high level, I wonder if it makes more sense to think about deprecating either Routes or Server entirely.

I don't think so. `Routes` is responsible for configuring the service router, while `Server` builds a server on top of `Routes`, so their roles are different. It would make sense if `Server` could handle other `Routes` equivalents and allow us to change the service router to be used, but that would be a relatively big change.

tobz (approved):

Issue #2003, opened 2024-10-15
Update tonic to tower 0.5


kanpov:
### Crates

tonic

### Motivation

Currently, an empty tokio + tonic project pulls in both tower 0.5 and tower 0.4, which is undesirable.

### Proposal

Bump to tower 0.5

### Alternatives

kanpov:
Fixed in #2004

hatomist:
+1 on this, bumping tower fixes builds on riscv-musl architecture (or anything that is built with -Zbuild-std) as per this ticket https://github.com/tower-rs/tower/issues/771

kanpov:
> +1 on this, bumping tower fixes builds on riscv-musl architecture (or anything that is built with -Zbuild-std) as per this ticket https://github.com/tower-rs/tower/issues/771

Well as far as understand it, #1892 fixing the issue has been open for a while but doesn't seem to be getting merged any time soon.

PR #2004, opened 2024-10-15
chore: Update from tower 0.4 to 0.5

kanpov:
## Motivation

#2003

## Solution

Bumped dependencies

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -263,11 +263,27 @@ path = "src/codec_buffers/client.rs"
 
 [features]
 gcp = ["dep:prost-types", "tonic/tls"]
-routeguide = ["dep:async-stream", "tokio-stream", "dep:rand", "dep:serde", "dep:serde_json"]
+routeguide = [
+    "dep:async-stream",
+    "tokio-stream",
+    "dep:rand",
+    "dep:serde",
+    "dep:serde_json",
+]
 reflection = ["dep:tonic-reflection"]
 autoreload = ["tokio-stream/net", "dep:listenfd"]
 health = ["dep:tonic-health"]
-grpc-web = ["dep:tonic-web", "dep:bytes", "dep:http", "dep:hyper", "dep:hyper-util", "dep:tracing-subscriber", "dep:tower", "dep:tower-http", "tower-http?/cors"]
+grpc-web = [
+    "dep:tonic-web",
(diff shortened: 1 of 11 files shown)


kanpov:
I never want to do this ever again, but here we are. Ready for review, everything has been upgraded to tower 0.5 and towe-http 0.6, including tests and examples so that they compile just fine.

**TL;DR for reviewers**: all changes except for the following were just bumps, the main pain was the Channel since the entire Buffer type is vastly different in its generics in 0.5, so the Svc type I removed completely and rewrote the Channel internals like so using rustc hints:
image

I'd appreciate this getting into a release so that every tonic project doesn't need to duplicate tower.

djc:
This was done already in #1892. Unfortunately it is a change in the public API so we'd need everyone to migrate. Not sure if this change is worth doing an incompatible bump on its own.

kanpov:
> This was done already in #1892. Unfortunately it is a change in the public API so we'd need everyone to migrate. Not sure if this change is worth doing an incompatible bump on its own.

Most of the ecosystem including axum has already migrated to tower 0.5 so I'm not sure how it's a choice of whether to do it or not. Not doing this locks tonic users out of using any tower 0.5-based middleware and any progress done in tower-http 0.6+.

djc:
There's already a PR for that one, too: #1556. For now, I don't think we can enable it by default for now given our MSRV, and I think the stance of the Rust Project is that RPITIT is "discouraged for general use in public traits" due to semver hazards while we don't have RTN.

kanpov:
Anyway, if the async_trait removal is not justifiable, I still think either this needs to be shipped or if we're really committing for tonic holding back its tower version, hold back the axum version with a `<=` version bound so that incompatible axum and tonic don't get compiled together (which I think is an absolutely awful idea, but technically not breaking?).

tottoto:
This pull request is a duplicate of an existing pull request. And since "async fn in traits" is not related to updating of tower dependency, it doesn't seem like an appropriate place to discuss it.

kanpov:
> This pull request is a duplicate of an existing pull request. And since "async fn in traits" is not related to updating of tower dependency, it doesn't seem like an appropriate place to discuss it.

Sure, but I would like for at least the original to get merged so that tonic isn't seemingly purposefully lagging behind tower, while axum which it uses as an internal framework already has upgraded.

PR #2008, opened 2024-10-16
feat(tls): AWS Libcrypto Support

jenr24-architect:
## Motivation

I want to use `aws-lc-rs` with tonic. 

## Solution

adds a feature flag `tls-aws-lc` in parallel to `tls`, removed `tls` feature flag dependency from `tls-*-roots` features


tonic/Cargo.toml
```diff
@@ -28,27 +28,60 @@ gzip = ["dep:flate2"]
 zstd = ["dep:zstd"]
 default = ["transport", "codegen", "prost"]
 prost = ["dep:prost"]
-tls = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"]
-tls-roots = ["tls-native-roots"] # Deprecated. Please use `tls-native-roots` instead.
-tls-native-roots = ["tls", "channel", "dep:rustls-native-certs"]
-tls-webpki-roots = ["tls", "channel", "dep:webpki-roots"]
+tls = [
```
djc, line 31: Please avoid reformatting the `Cargo.toml` and assume that your changes should be consistent with the existing formatting.

tonic/src/request.rs
```diff
@@ -1,15 +1,15 @@
 use crate::metadata::{MetadataMap, MetadataValue};
 #[cfg(feature = "server")]
 use crate::transport::server::TcpConnectInfo;
-#[cfg(all(feature = "server", feature = "tls"))]
+#[cfg(all(feature = "server", any(feature = "tls", feature = "tls-aws-lc")))]
```
djc, line 4: I think we should add some lowest-common-denominator feature that replaces `tls` (maybe `tls-any`) so that we don't need the proliferation of `any()`s all over the place.

djc (approved):
Changes look good to me, thanks! Need to address the CI failures.

tonic/Cargo.toml
```diff
@@ -28,10 +28,13 @@ gzip = ["dep:flate2"]
 zstd = ["dep:zstd"]
 default = ["transport", "codegen", "prost"]
 prost = ["dep:prost"]
-tls = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"]
+tls = ["tls-any", "tls-ring"] # Deprecated. Please use `tls-ring` or `tls-aws-lc` instead.
```
tottoto, line 31: Can this depend only on `tls-ring` which `tls` aliases to?

tonic/Cargo.toml
```diff
@@ -28,10 +28,13 @@ gzip = ["dep:flate2"]
...
+tls-any = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"]
```
tottoto, line 32: Is there the use case for using the `tls-any` feature alone? If not, I think we want to make it clear that it is an internal feature by adding an underscore to the prefix.

tonic/Cargo.toml
```diff
@@ -28,10 +28,13 @@ gzip = ["dep:flate2"]
...
+tls-ring = ["tls-any", "tokio-rustls/ring"]
+tls-aws-lc = ["tls-any", "tokio-rustls/aws-lc-rs"]
 tls-roots = ["tls-native-roots"] # Deprecated. Please use `tls-native-roots` instead.
-tls-native-roots = ["tls", "channel", "dep:rustls-native-certs"]
-tls-webpki-roots = ["tls", "channel", "dep:webpki-roots"]
+tls-native-roots = ["channel", "dep:rustls-native-certs"]
+tls-webpki-roots = ["channel", "dep:webpki-roots"]
```
tottoto, line 37: This seems to make a breaking change.

tottoto (commented):
Thanks. I left some comments.

Could you update the documentation regarding the features.

tonic/Cargo.toml
jenr24-architect: this will break users who depend on tonic like
```toml
tonic = { version = "0.12", features = ["tls-webpki-roots"]}
```
to imply the `tls` feature. I thought this way was better than creating separate `tls-aws-lc-webpki-roots` and `tls-aws-lc-native-roots` features. Open to change here

tonic/Cargo.toml
tottoto: I confirmed that `tls-any` is renamed to `_tls-any`.

tonic/Cargo.toml
djc: We just merged some breaking changes, so we should probably do what's best without regards for backwards compatibility at this point.

tottoto: As another breaking change was merged after my comment, I opened #2014 to confirm if it's okay to make other changes. I think the proposal removing `tls` from these features is good when we can make breaking changes.

LucioFranco:
@jenr24-architect could you try to force push again to see if we can get CI to kick off again?

jenr24-architect:
> @jenr24-architect could you try to force push again to see if we can get CI to kick off again?

Does it have to be a force push?? It doesn't look like my last push triggered it

tonic/Cargo.toml
```diff
@@ -28,10 +28,13 @@ gzip = ["dep:flate2"]
...
-tls = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"]
+_tls-any = ["dep:rustls-pemfile", "dep:tokio-rustls", "dep:tokio", "tokio?/rt", "tokio?/macros"] # Internal. Please choose on of `tls-ring` or `tls-aws-lc`
+tls = ["tls-ring"] # Deprecated. Please use `tls-ring` or `tls-aws-lc` instead.
+tls-ring = ["_tls-any", "tokio-rustls/ring"]
+tls-aws-lc = ["_tls-any", "tokio-rustls/aws-lc-rs"]
 tls-roots = ["tls-native-roots"] # Deprecated. Please use `tls-native-roots` instead.
-tls-native-roots = ["tls", "channel", "dep:rustls-native-certs"]
-tls-webpki-roots = ["tls", "channel", "dep:webpki-roots"]
+tls-native-roots = ["_tls-any", "channel", "dep:rustls-native-certs"]
+tls-webpki-roots = ["_tls-any","channel", "dep:webpki-roots"]
```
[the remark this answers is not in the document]
tottoto: I think this is a good idea. However, since the scope is different from this pull request, I think it would be good to discuss it in a separate pull request rather than changing them here.

tonic/src/lib.rs
```diff
@@ -24,8 +24,11 @@
 //! - `router`: Enables the [`axum`] based service router. Enabled by default.
 //! - `codegen`: Enables all the required exports and optional dependencies required
 //!   for [`tonic-build`]. Enabled by default.
-//! - `tls`: Enables the [`rustls`] based TLS options for the `transport` feature. Not
-//!   enabled by default.
+//! - `tls`: Deprecated. An alias to `tls-ring`
+//! - `tls-ring`: Enables the [`rustls`] based TLS options for the `transport` feature using
+//!   the [`ring`] libcrypto provider. Not enabled by default.
```
tobz, line 29: It's not actually super clear to me _why_ this is failing... perhaps because the `ring` crate isn't a direct dependency? Although then we might expect it to also show a failure for `aws-lc-rs` since that's not a direct dependency either... 🤔

tobz (changes requested):
Looks like this needs to be run through `cargo fmt` and there's the broken intra-doc link stuff to sort out.

tonic/src/lib.rs
jenr24-architect: I believe this is because `aws-lc-rs` is pulled in by `rustls` as a direct dependency. I added a link to `ring` in the docs to resolve this issue

tottoto:
With this change, users can now enable both `ring` and `aws-lc-rs`, and when they enable both at the same time, users will need to specify which of these they want to use as the default provider. This setting was already an exposed dependency of `rustls` as an interface for setting the default provider for `rustls`, but this was not an issue until now as only `ring` was supported. I think it would be needed to be able to re-expose these in `tonic` so that users can set which crypto provider to use when both are enabled in order to allow us to use `rustls` as an internal dependency. It also seems necessary to state that setting the default provider using `rustls` directly is not intended to be stable for `tonic`.

tottoto:
With this change, users can enable both `ring` and `aws-lc-rs`, and when they enable both at the same time, users will need to specify which of these they want to use as the default provider. This setting was already an exposed dependency of `rustls` as an interface for setting the default provider for `rustls`, but this was not an issue until now as only `ring` was supported. I think it would be needed to be able to re-expose these in `tonic` so that users can set which crypto provider to use when both are enabled in order to allow us to use `rustls` as an internal dependency. It also seems necessary to state that setting the default provider using `rustls` directly is not intended to be stable for `tonic`.

tottoto:
With this change, users can enable both `ring` and `aws-lc-rs`, and when they enable both at the same time, users will need to specify which of these they want to use as the default provider. This setting was already an exposed dependency of `rustls` as an interface for setting the default provider for `rustls`, but this was not an issue until now as only `ring` was supported. I think it would be needed to wrap these configuration APIs in `tonic` so that users can set which crypto provider to use when both are enabled in order to allow us to use `rustls` as an internal dependency. It also seems necessary to state that setting the default provider using `rustls` directly is not intended to be stable for `tonic`.

jenr24-architect:
> With this change, users can enable both `ring` and `aws-lc-rs`, and when they enable both at the same time, users will need to specify which of these they want to use as the default provider. This setting was already an exposed dependency of `rustls` as an interface for setting the default provider for `rustls`, but this was not an issue until now as only `ring` was supported. I think it would be needed to wrap these configuration APIs in `tonic` so that users can set which crypto provider to use when both are enabled in order to allow us to use `rustls` as an internal dependency. It also seems necessary to state that setting the default provider using `rustls` directly is not intended to be stable for `tonic`.

I had thought about this -- I initially had written some code to add `ClientTlsConfig::provider(provider: rustls::CryptoProvider)` and `ServerTlsConfig::provider(provider: rustls::CryptoProvider)` so that I could write both a `connect_handles_tls_ring` and `connect_handles_tls_aws_lc` integration test, but the scope of that change started to get out of hand.

I would be willing to add that feature, in either this or a separate PR if we're interested in adding that functionality :)

LucioFranco (approved):

LucioFranco:
@jenr24-architect looks like there are some more CI failures

tottoto (approved):
@jenr24-architect 

> but the scope of that change started to get out of hand.

Makes sense to me. I suppose there are several possible ways to address this, so it seems reasonable to narrow the scope of this pull request.

tottoto:
@LucioFranco The CI failure does not seem to be caused by this pull request's change. I opened a pull request to address it #2021.

tottoto:
@LucioFranco The CI failure does not seem to be caused by this pull request's change. I opened #2021 to address it.

jenr24-architect:
merged master after #2021 merged, this PR is ready for CI to re-run and to enter the merge queue

PR #2009, opened 2024-10-17
chore: Avoid duplicate box on constructing BoxBody

tottoto:
Avoids duplicate box on constructing `BoxBody`.

### DIFF
--- a/tonic/src/body.rs
+++ b/tonic/src/body.rs
@@ -11,7 +11,14 @@ where
     B: http_body::Body<Data = bytes::Bytes> + Send + 'static,
     B::Error: Into<crate::Error>,
 {
-    body.map_err(crate::Status::map_error).boxed_unsync()
+    let mut body = Some(body);
+    if let Some(body) = <dyn std::any::Any>::downcast_mut::<Option<BoxBody>>(&mut body) {
+        body.take().unwrap()
+    } else {
+        body.unwrap()
+            .map_err(crate::Status::map_error)
+            .boxed_unsync()
+    }
 }
 
 /// Create an empty `BoxBody`


tonic/src/body.rs
```diff
@@ -11,7 +11,14 @@ where
     B: http_body::Body<Data = bytes::Bytes> + Send + 'static,
     B::Error: Into<crate::Error>,
 {
-    body.map_err(crate::Status::map_error).boxed_unsync()
+    let mut body = Some(body);
+    if let Some(body) = <dyn std::any::Any>::downcast_mut::<Option<BoxBody>>(&mut body) {
```
djc, line 15: Is there evidence of this being an actual problem in the wild? Why do we need the `Option`-wrapping for this?

LucioFranco, line 15: I don't think ive ever seen this syntax before...feels really odd. Is there any way we can clean this up a bit? I need to sit on this idea though I am not a huge fan of downcasting. Do we know if downcasting is going to be cheaper than actually just allocating and erasing?

tottoto: > Is there evidence of this being an actual problem in the wild?

The motivation is to reduce the effect of the changes like #2001. I have not measured whether duplicate boxes actually affect performance.

> Why do we need the Option-wrapping for this?

This is because of Rust's ownership system. It appears that this is necessary to allow the argument to be reused when the downcasting of the body fails.

tottoto: > Is there any way we can clean this up a bit?

I tried it but I could not find another way.

> Do we know if downcasting is going to be cheaper than actually just allocating and erasing?

I haven't measured it. When I was concerned about the effect of #2001, I investigated and found that `axum` had dealt with the same problem. `axum` seems to evaluate that downcasting is cheaper.

https://github.com/tokio-rs/axum/blob/axum-core-v0.4.5/axum-core/src/body.rs

LucioFranco: Yeah, that is valid. I think this is fine seems like axum has been using it for a while and there is no issue there.

LucioFranco (approved):

PR #2015, opened 2024-10-18
feat: make pub `Channel` fns that allow creating it from a custom connector

shikhar:
The connector is required to be a `Service` that accepts a `http::Uri` and returns a connection implementing hyper's IO traits, which is a very reasonable requirement that can be fulfilled by advanced users that need to customize any mechanics that don't fit into the paved path for HTTP and HTTPS with tonic.

One such requirement is customizing `rustls::ClientConfig`, which has come up for a number of users. They have been pointed in the direction of a custom connector, but the APIs on `Endpoint` are not always suitable:

`Endpoint::connect_(lazy_)with_connector` always wraps the provided connector with `transport::channel::service::Connector`. This wrapper raises an error `HttpsUriWithoutTlsSupport` if (the scheme is `https://` + the `tls` feature is enabled + and `Endpoint::tls_config()` was not used) [ref](https://github.com/hyperium/tonic/blob/758d4f9a95d9752465baa7db10ab6cc069fef00b/tonic/src/transport/channel/service/connector.rs#L62-L70). This is a good safety feature in general, but not if the wrapped connector is taking care of TLS.

As a side bonus, `transport::channel::service::io::BoxedIo` can be avoided.

### DIFF
--- a/tonic/src/transport/channel/mod.rs
+++ b/tonic/src/transport/channel/mod.rs
@@ -16,7 +16,6 @@ use http::{
     uri::{InvalidUri, Uri},
     Request, Response,
 };
-use hyper_util::client::legacy::connect::Connection as HyperConnection;
 use std::{
     fmt,
     future::Future,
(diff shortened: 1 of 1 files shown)


tottoto (approved):
Sounds reasonable to me.

tobz (approved):

PR #2016, opened 2024-10-19
feat(web): Relax GrpcWebService request body type

tottoto:
Relaxes `GrpcWebService`'s request body type.

### DIFF
--- a/tonic-web/src/service.rs
+++ b/tonic-web/src/service.rs
@@ -1,3 +1,4 @@
+use core::fmt;
 use std::future::Future;
 use std::pin::Pin;
 use std::task::{ready, Context, Poll};
@@ -61,9 +62,11 @@ where
     }
 }
 
-impl<S> Service<Request<BoxBody>> for GrpcWebService<S>
+impl<S, B> Service<Request<B>> for GrpcWebService<S>
 where
     S: Service<Request<BoxBody>, Response = Response<BoxBody>>,
+    B: http_body::Body<Data = bytes::Bytes> + Send + 'static,
+    B::Error: Into<crate::BoxError> + fmt::Display,
 {
     type Response = S::Response;
     type Error = S::Error;
@@ -73,7 +76,7 @@ where
         self.inner.poll_ready(cx)
     }
 
-    fn call(&mut self, req: Request<BoxBody>) -> Self::Future {
+    fn call(&mut self, req: Request<B>) -> Self::Future {
         match RequestKind::new(req.headers(), req.method(), req.version()) {
             // A valid grpc-web request, regardless of HTTP version.
             //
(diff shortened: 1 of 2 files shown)


tonic-web/src/service.rs
```diff
@@ -113,7 +116,7 @@ where
                 debug!(kind = "other h2", content_type = ?req.headers().get(header::CONTENT_TYPE));
                 ResponseFuture {
                     case: Case::Other {
-                        future: self.inner.call(req),
+                        future: self.inner.call(req.map(tonic::body::boxed)),
```
tobz, line 119: Hmm, if we have to box the request no matter, what are we gaining with this change?

tottoto: The benefit is the `CorsGrpcWeb` can handle any kind of body types which has the trait bounds.

tobz: Fair enough. 👍🏻

tobz (commented):
While the change itself seems fine, I'm still left with the thought: should we avoid all of the constraints that require a service that accepts `Request<BoxBody>`?

It seems like if we're going to bother to be totally generic over the request body type in `GrpcWebService`, we should also do so in `CorsWebLayer`? Maybe there's a very obvious reason why we can't, but looking at `tower_http::cors::Cors` shows that it's _also_ generic over the request body... so it feels like we should be able to?

tottoto:
> While the change itself seems fine, I'm still left with the thought: should we avoid all of the constraints that require a service that accepts Request<BoxBody>?

> It seems like if we're going to bother to be totally generic over the request body type in GrpcWebService, we should also do so in CorsWebLayer? Maybe there's a very obvious reason why we can't, but looking at tower_http::cors::Cors shows that it's also generic over the request body... so it feels like we should be able to?

Since `CorsWebLayer` does not exist, I respond with information about other parts.

I assume your main question is whether we can avoid requiring a `BoxBody` in the internal services.

I guess so too. When I was working on this, I was also looking into the implementation of `tower_http`'s `Cors` service to see if it is possible to avoid the `BoxBody`. However, it would require an internal design change to the implementation, so I opened this partial change which can be easily done for now.

tottoto:
> While the change itself seems fine, I'm still left with the thought: should we avoid all of the constraints that require a service that accepts Request<BoxBody>?

> It seems like if we're going to bother to be totally generic over the request body type in GrpcWebService, we should also do so in CorsWebLayer? Maybe there's a very obvious reason why we can't, but looking at tower_http::cors::Cors shows that it's also generic over the request body... so it feels like we should be able to?

Since `CorsWebLayer` does not exist, I respond with information about other parts.

I assume your main question is whether we can avoid requiring a `BoxBody` in the internal service.

I guess so too. When I was working on this, I was also looking into the implementation of `tower_http`'s `Cors` service to see if it is possible to avoid the `BoxBody`. However, it would require an internal design change to the implementation, so I opened this partial change which can be easily done for now.

tottoto:
@tobz Are there any other concerns block this? If there are no problems with this, I will merge it at this stage.

tobz (approved):
Sorry for the delay here.

I do think it's fine for the moment and we can always relax the bounds further in the future.

PR #2033, opened 2024-11-02
chore(server): Set nodelay and tcp timeout on constructing TcpIncoming

tottoto:
Sets nodelay and tcp timeout on constructing `TcpIncoming`.


tobz (commented):
I'm not sure this is a good idea.

I know _some_ socket options can be set on a listener and will be inherited by the accepted sockets, but I couldn't find anything in the POSIX specification that implies that they should always be inherited, or which ones should be inherited.

Unless you have a reference that demonstrates this as a stable, long-term behavior (on the major operating systems, such as Linux and Windows), I think we should leave it as-is.

tottoto:
I've looked into this for a while, but as you pointed out the specification is not found, I agree that we should leave it as it is.

PR #2034, opened 2024-11-03
use ring if no default crypto provider is found

conradludgate:
## Motivation

Tonic hard codes the `ring` feature in rustls. When another crate depends on default rustls features they will additionally have the `aws-lc-rs` feature enabled in rustls. Having both
of these features enabled makes rustls not work "out of the box" as there will no longer
be a default crypto provider.

## Solution

Make tonic use the ring provider as a fallback if no default provider is found.

### DIFF
--- a/tonic/src/transport/channel/service/tls.rs
+++ b/tonic/src/transport/channel/service/tls.rs
@@ -5,6 +5,7 @@ use hyper_util::rt::TokioIo;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_rustls::{
     rustls::{
+        crypto,
         pki_types::{ServerName, TrustAnchor},
         ClientConfig, RootCertStore,
     },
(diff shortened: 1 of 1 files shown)


tonic/src/transport/channel/service/tls.rs
```diff
@@ -33,7 +34,14 @@ impl TlsConnector {
         #[cfg(feature = "tls-native-roots")] with_native_roots: bool,
         #[cfg(feature = "tls-webpki-roots")] with_webpki_roots: bool,
     ) -> Result<Self, crate::BoxError> {
-        let builder = ClientConfig::builder();
+        let crypto_provider = match crypto::CryptoProvider::get_default() {
+            Some(default) => default.clone(),
```
tobz, line 38: Can we just change `default` to `existing`?

Would be nice to avoid the clash with the `default` keyword given it's use in specialization.

tobz (changes requested):

conradludgate:
Looks like the assumptions are no longer valid since #2008 merged.

conradludgate:
Now that it's configurable, I'm not sure there is still much use in this feature.

tobz:
While that's a fair point, I do still think there's value here.

If `tonic` is configured with one of the two provider options, and the other option is transitively configured, then there's no default for us to get at all, so we still end up having to make an explicit choice on which one to use. That's fine for `rustls`, but the user already told us they wanted to use AWS-LC or Ring, so we should honor that and use it directly.

tobz (approved):

PR #2042, opened 2024-11-08
feat(tls): Add support for rustls ignore_client_order

emuellen:
## Motivation

When running the TLS Connector with the https://testssl.sh/ testing tool, it showed that the server cipher order was not respected. This can be exploited by attacks such as POODLE. It would be good if we can instruct the TlsAdapter to enforce the server cipher order.

## Solution

The change in this pull request adds an additional builder function to enable the "ignore_client_order" parameter of the RustTLS Server config. By default, this parameter is disabled to ensure backward compatibility.


shikhar:
rustls only supports TLS1.2 and TLS1.3, so it is not vulnerable to POODLE, AFAIU

emuellen:
Yes, it's correct that it's not vulnerable for this specific attack, but maybe something similar might happen in the future with other cipher suites (?). 
Controlling the cipher suite order is considered best practice by our corporate security department and therefore enforced in our security baselines. Thus it would be good if we could incorporate this non-breaking change, so that I can check some boxes without forking tonic :-)

tonic/src/transport/server/tls.rs
```diff
@@ -56,11 +58,24 @@ impl ServerTlsConfig {
         }
     }
 
+    /// Sets whether the server's cipher preferences are followed instead of the client's.
+    /// It prevents attacks such as POODLE
```
tottoto, line 62: I don't think that tonic should maintain the document about some specific use cases same as rustls does not do.

emuellen: @tottoto , ok sure. Shall I just remove line 62?

emuellen: @tottoto , I removed the line. Would it be possible to merge this change now please?

tottoto (approved):

Issue #2072, opened 2024-11-28
Add TLS connect timeout support


honsunrise:
### Crates

`tonic`

### Motivation

Implement timeout controls in the TLS connection process to prevent the client from getting stuck due to the server becoming unresponsive while handling TLS.

### Proposal

Add a `timeout` field in the `ClientTlsConfig` struct, and impl timeout control in `TlsConnector::connect`.

PR #2073, opened 2024-11-28
feat(tls): Add tls handshake timeout support

honsunrise:
Impl: #2072


a1ien:
Wow, thank you! This aligns closely with what we've been observing recently. @tottoto, could you take a look at this? I believe this issue deserves more attention, as it seems that if something goes wrong during the TLS connection, the client could hang indefinitely.

LucioFranco (approved):

PR #2076, opened 2024-11-28
examples: Minor fix in routeguide example

arjan-bal:
## Motivation

1. The following logs line prints a feature but the logs mentions `NOTE`. `NOTE` messages are printed later in bidirectional stream code, so this may be a copy/paste error.
```rust
println!("NOTE = {:?}", feature);
```
2. The following line in the markdown file declares `tx` as `mut` which and detected by `rust-analyzer`. The actual client file doesn't declare `tx` as `mut`.
```rust
let (mut tx, rx) = mpsc::channel(4);
```

Client code:
https://github.com/hyperium/tonic/blob/a7030a7fb7f5e2b270fb47c3f9701a798d81513f/examples/src/routeguide/server.rs#L47-L48

## Solution

1. Make the print statement accurate.
2. Remove the unnecessary `mut`.

### DIFF
--- a/examples/routeguide-tutorial.md
+++ b/examples/routeguide-tutorial.md
@@ -401,7 +401,7 @@ async fn list_features(
     &self,
     request: Request<Rectangle>,
 ) -> Result<Response<Self::ListFeaturesStream>, Status> {
-    let (mut tx, rx) = mpsc::channel(4);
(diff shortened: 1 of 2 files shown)


tottoto:
Thanks.

PR #2092, opened 2024-12-07
fix: Fill GRPC_CONTENT_TYPE when handing recover errors

Manphil:
## Motivation

Fix https://github.com/hyperium/tonic/issues/2090.

## Solution
 Filling http content-type in responses when services are cancelled by tonic.

### DIFF
--- a/tonic/src/transport/server/service/recover_error.rs
+++ b/tonic/src/transport/server/service/recover_error.rs
@@ -1,3 +1,4 @@
+use crate::metadata::GRPC_CONTENT_TYPE;
 use crate::Status;
 use http::Response;
 use http_body::Frame;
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/service/recover_error.rs
```diff
@@ -67,6 +68,8 @@ where
             Err(err) => match Status::try_from_error(err) {
                 Ok(status) => {
                     let mut res = Response::new(MaybeEmptyBody::empty());
+                    res.headers_mut()
+                        .insert(http::header::CONTENT_TYPE, GRPC_CONTENT_TYPE);
```
tottoto, line 72: We can use `Status::into_http` instead of reimplementing the logic.

tottoto: `Status::into_http` can be used instead of reimplementing the logic.

PR #2107, opened 2024-12-25
Fix percent-encoding of percent sign in `grpc-message` header

AndiDog:
## Motivation

I got an unexpected JavaScript client-side gRPC-Web error `Uncaught URIError: malformed URI sequence` and traced it down to the `%` character not being encoded by the tonic-based server. My error message included a percent sign 😏.

## Solution

Correct usage of the dependent crate's `AsciiSet` structure. Otherwise, the `%` character doesn't get changed at all.


LucioFranco (approved):
Nice catch!

PR #2115, opened 2025-01-02
Add `use_generic_streaming_requests` option to make testing server impls easier

yotamofek:
## Motivation
Currently, testing the server trait implementations of client-streaming methods (i.e. requests that receive a stream) is quite difficult, and requires jumping through hoops in order to construct a `Streaming` object (see #462 for one way of doing it).

## Solution
Adds a codegen option that makes the server trait go from this:
```rust
impl test_server::Test for Svc {
    async fn test_request(
        &self,
        req: Request<Streaming<Message>>,
    ) -> Result<Response<Message>, Status>;
}
```
to this:
```rust
impl test_server::Test for Svc {
    async fn test_request(
        &self,
        req: impl IntoStreamingRequest<Message = Result<Message, Status>, Stream: Unpin> + Send,
    ) -> Result<Response<Message>, Status>;
}
```

Since `Request<Streaming<Message>>` impls `IntoStreamingRequest<Message = Result<Message, Status>>`, the grpc server can continue invoking these handlers in the same way, but it also allows any object implementing `Stream<Item = Result<Message, Status>>` to be used.
This makes testing much easier.


yotamofek:
Just realized this uses [associated type bounds](https://github.com/rust-lang/rust/pull/122055/) which were only stabilized in [1.79.0](https://releases.rs/docs/1.79.0/#language).
How often is the MSRV raised?
The `Unpin` bound in `IntoStreamingRequest<Stream: Unpin>` isn't strictly necessary (and might actually make the trait less flexible), and can be either removed or emulated with a named generic param (i.e. `test_request<S: Unpin>(req: impl IntoStreamIterator<Stream = S>)` or so)

tonic-build/src/code_gen.rs
```diff
@@ -46,7 +47,7 @@ impl CodeGenBuilder {
         self
     }
 
-    /// Enable compiling well knonw types, this will force codegen to not
+    /// Enable compiling well known types, this will force codegen to not
```
tottoto, line 50: Can this be a separated pull request.

tonic-build/src/server.rs
```diff
@@ -257,92 +263,61 @@ fn generate_trait_methods<T: Service>(
             quote!(&self)
         };
 
-        let method = match (
```
tottoto, line 260: I'm not sure if the benefits of adding this feature are worth the increased complexity that comes with adding more features. However, the refactoring of this part that was done to add this feature seems to improve the readability and maintainability of the code, so if it be separated into a separate pull request, I think we can proceed with the discussion separately.

yotamofek: Sure, the refactoring is in a separate commit so I'll put it in its own PR. Thanks!

tonic-build/src/code_gen.rs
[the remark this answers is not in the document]
yotamofek: https://github.com/hyperium/tonic/pull/2142

jovanpavl-db:
Thanks for working on this; currently, it is incredibly painful!

PR #2143, opened 2025-01-13
chore(build): Refactor server trait gen to dedup some code

yotamofek:
## Motivation

A lot of the code for building the server trait is duplicated.

## Solution

Refactor to reduce said duplication.

### DIFF
--- a/tonic-build/src/server.rs
+++ b/tonic-build/src/server.rs
@@ -268,43 +268,49 @@ fn generate_trait_methods<T: Service>(
             async fn #name(#self_param, request: #req_param_type)
         };
 
-        let body_or_semicolon = if generate_default_stubs {
-            quote! {
-                {
-                    Err(tonic::Status::unimplemented("Not yet implemented"))
-                }
+        let default_body = quote! {
+            {
+                Err(tonic::Status::unimplemented("Not yet implemented"))
             }
-        } else {
-            quote!(;)
         };
 
         let result = |ok| quote!(std::result::Result<#ok, tonic::Status>);
         let response_result = |message| result(quote!(tonic::Response<#message>));
 
-        let method = if !method.server_streaming() {
-            let return_ty = response_result(res_message);
-            quote! {
-                #partial_sig -> #return_ty #body_or_semicolon
+        let method = match (method.server_streaming(), generate_default_stubs) {
+            (false, true) => {
+                let return_ty = response_result(res_message);
+                quote! {
+                    #partial_sig -> #return_ty #default_body
+                }
             }
-        } else if generate_default_stubs {
-            let return_ty = response_result(quote!(BoxStream<#res_message>));
-            quote! {
-                #partial_sig -> #return_ty #body_or_semicolon
+            (false, false) => {
+                let return_ty = response_result(res_message);
+                quote! {
+                    #partial_sig -> #return_ty;
+                }
             }
-        } else {
-            let stream = quote::format_ident!("{}Stream", method.identifier());
-            let stream_doc = generate_doc_comment(format!(
-                " Server streaming response type for the {} method.",
-                method.identifier()
-            ));
-            let stream_item_ty = result(res_message);
-            let stream_ty = quote!(tonic::codegen::tokio_stream::Stream<Item = #stream_item_ty> + std::marker::Send + 'static);
-            let return_ty = response_result(quote!(Self::#stream));
-            quote! {
-                #stream_doc
-                type #stream: #stream_ty;
+            (true, true) => {
+                let return_ty = response_result(quote!(BoxStream<#res_message>));
+                quote! {
+                    #partial_sig -> #return_ty #default_body
+                }
+            }
+            (true, false) => {
+                let stream = quote::format_ident!("{}Stream", method.identifier());
+                let stream_doc = generate_doc_comment(format!(
+                    " Server streaming response type for the {} method.",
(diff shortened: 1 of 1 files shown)


tonic-build/src/server.rs
```diff
@@ -257,92 +257,54 @@ fn generate_trait_methods<T: Service>(
             quote!(&self)
         };
 
-        let method = match (
-            method.client_streaming(),
-            method.server_streaming(),
-            generate_default_stubs,
-        ) {
```
tottoto, line 264: I'd like to keep the current styles for the conditions about whether it is client streaming or server streaming or not. This refactoring is well designed to avoid reimplementing logic, but at the same time, I feel that it has become more difficult to grasp the outline of the processing from the perspective of code generation. Taking the trade-off into consideration, how about introducing `body_or_semicolon` to the conditional branch of `generate_default_stubs` to reduce the number of branches?

yotamofek: Yeah I agree, refactored a bit more to look more like the older code. There's a bit of duplication here but I think it's a reasonable trade-off because it makes the branching easier to follow. What do you think?

tottoto: What I tried to say is that using `switch` with client streaming and/or server streaming, and then in the each branch using `body_or_semicolon` variable to handle the default stub option.

And, I would like to write `partial_sig` and `return_ty` inline directly without making them variables.

tottoto:
> @tottoto BTW, I think the whole server-trait generation logic could go in its own module and be split up into functions. I didn't want to make such an intrusive rewrite without seeing if the crate's maintainers are onboard with the idea, but while we're already cleaning this up - I can give it a go and see if it turns out well. If you think that's a good idea 😊

A detailed explanation of the proposed content may help understanding.

tottoto:
> @tottoto BTW, I think the whole server-trait generation logic could go in its own module and be split up into functions. I didn't want to make such an intrusive rewrite without seeing if the crate's maintainers are onboard with the idea, but while we're already cleaning this up - I can give it a go and see if it turns out well. If you think that's a good idea 😊

If a specific explanation of the proposal is provided, it might help understanding.

tonic-build/src/server.rs
yotamofek: Not sure I understand what you mean. I don't really see a point in doing something like `match (method.client_streaming(), method.server_streaming())` since those two settings don't impact one another.

yotamofek:
> If a specific explanation of the proposal is provided, it might help understanding.

I don't haha, just thought I'd start by moving everything into a separate module and start splitting it up into smaller functions. Don't have a specific proposal in mind. I can take a crack at it some time.

tottoto:
> Though to be honest, not sure how much effort I'd wanna put into this, my main goal was to get https://github.com/hyperium/tonic/pull/2115 in, that would be really helpful for my project @ work, this cleanup was meant to make that feature easier to develop.

Even if such a refactoring were to be merged, there is no guarantee that #2115 would be merged. That's a different discussion point.

yotamofek:
> > Though to be honest, not sure how much effort I'd wanna put into this, my main goal was to get #2115 in, that would be really helpful for my project @ work, this cleanup was meant to make that feature easier to develop.
> 
> Even if such a refactoring were to be merged, there is no guarantee that #2115 would be merged. That's a different discussion point.

Yeah, I know, what I'm saying is I don't know how much effort I want to put into this PR *because* it won't help #2115 (or something similar) get merged.

tottoto:
> Yeah, I know, what I'm saying is I don't know how much effort I want to put into this PR because it won't help https://github.com/hyperium/tonic/pull/2115 (or something similar) get merged.

This pull request is an extraction of a part from #2115 that seems complex and independently discussable and applicable, so it is essentially a part of it in the sense that it is a precursor to #2115. We had the option of reviewing it directly in #2115, but the inclusion of a different perspective, that of adding functionality, would make it more complicated, and since this part itself would be independently useful if successful, proposed splitting it into a separate it as this pull request for discussion.

As a result of the experiments and discussions here, it seems that refactoring is not that easy, so it will be difficult to apply #2115 while maintaining maintainability with the current design.

yotamofek:
Refactoring is quite easy, agreeing on what constitutes "better" in matters of style and taste is futile :)

Feel free to take this PR and modify it in any way that suits you. Seeing as you're not too inclined on adding the feature I suggested in #2115 (though I do think people other than me will find it beneficial), I don't have the motivation to spend more time on this, especially since it seems to me like the changes you're suggesting would be easier to make than describe.

tottoto:
> Refactoring is quite easy, agreeing on what constitutes "better" in matters of style and taste is futile :)

It is important to consider not only adding features but also future maintenance, so it is important to consider what kind of refactoring is appropriate. It is not a mere matter of style.

> Feel free to take this PR and modify it in any way that suits you. I agree it's a good idea to review and discuss it separately. But seeing as you're not too inclined on adding the feature I suggested in https://github.com/hyperium/tonic/pull/2115 (though I do think people other than me will https://github.com/hyperium/tonic/pull/2115#issuecomment-2592605888), I don't have the motivation to spend more time on this, especially since it seems to me like the changes you're suggesting would be easier to make than describe.

Since this is your suggestion and you are not going to proceed it will be closed. Next time, be careful not to submit a pull request that you are not motivated to pursue yourself. The community has limited resources and no one has the resources to put effort into it.

PR #2145, opened 2025-01-17
feat: Expose Status as a Response extension

tamasfe:
Add the `Status` itself as an extension to a `http::Response` so that it is available in middleware.

## Motivation

I'd like to implement service-level error inspection. `Status` contains a `source` that allows returning pretty much any error, however this is lost when it is converted into a response. The only way to pass information to middleware is headers which is expensive and requires serializable types.

## Solution

Simply added the status to the response extensions in `Status::into_http()`. How a `Status` is turned into a `Response` is not specified in the documentation so I did not add additional notes regarding this feature either.


tonic/src/status.rs
```diff
@@ -580,6 +580,7 @@ impl Status {
             .headers_mut()
             .insert(http::header::CONTENT_TYPE, GRPC_CONTENT_TYPE);
         self.add_header(response.headers_mut()).unwrap();
+        response.extensions_mut().insert(self);
```
LucioFranco, line 583: Am I missing something (its been a while since I looked at this code) but can you not extract the information directly from the headers?

tamasfe: You can pass information to the middleware via headers with additional effort when the status is created. Like I said in the PR description, passing more complex types always involves a serialization step (since headers are just bytes) and a deserialization step if a middleware accesses this information.

LucioFranco (approved):

Massimiliano-solutiontech:
I was looking for this! (Seems impossible that nobody thought about it)

@LucioFranco when you think will be merge into main and then released?

LucioFranco:
looks like we need CI to pass

tamasfe:
> looks like we need CI to pass

I changed a single line that adds a feature nothing relies on yet, all the CI failures were (are) unrelated.

theunwisewolf:
Any update on this? This is really useful.

PR #2162, opened 2025-01-25
feat: Expose GrpcTimeout middleware

tottoto:
Exposes the GrpcTimeout middleware. This allows users to use independent of the transport server.


LucioFranco:
I don't think we really want to expose this as it adds more surface area to maintain publicly. I would suggest that users copy/vendor this code specifically if they need it.

PR #2189, opened 2025-02-24
add support for tower's load-shed layer

joeroback:
picking up from https://github.com/hyperium/tonic/pull/1616 as that branch was out of date.

@abizjak

Text from original PR:

Add an option to enable load shedding for the service constructed for the connection.

### Motivation
In the server we are implementing we want to load-shed when the service is overloaded, and we want to limit the number of concurrent requests per connection. We use tower::load_shed layer for load shedding, but this only applies to the entire service.

If we additionally apply the per-connection concurrency limit the issue is that this limit is typically hit first and so the service responds with "not ready". As a result the load-shed layer of the service is never reached meaning that requests just queue up.

### Solution
The solution is to optionally add a load-shed layer to the per-connection service constructed by tonic.


LucioFranco (changes requested):
Would be interesting to see if we could write a test for this?

joeroback:
> Would be interesting to see if we could write a test for this?

i took my first shot on adding a integration test for resource_exhausted. not sure how much i should or should not expand on this.

LucioFranco (approved):
Sorry for the delay, well get this into a release soon.

Thynix:
> Sorry for the delay, well get this into a release soon.

@LucioFranco I don't see a milestone beyond 0.13 in the issue tracker. Are there things I can do to help with getting to the next release? Load shedding would be great for my use case.

PR #2192, opened 2025-02-28
Create place for grpc crate and initial contents

dfawley:
This PR creates the rough layout for the upcoming grpc crate and license notices.  The biggest notable piece of API is the `load_balancing` package, and it includes a nearly-complete `ChildManager` implementation to show its usage.  Non-LB types that are included are important for the LB API, but left as TODO.

cc @easwars @arjan-bal

### DIFF
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -1,11 +1,11 @@
 [package]
 name = "grpc"
-version = "0.9.0"
+version = "0.9.0-alpha.1"
 edition = "2021"
 authors = ["gRPC Authors"]
 license = "Apache-2.0"
 
 [dependencies]
 url = "2.5.0"
-tokio = { version = "1.37.0", features = ["sync", "full" ] }
-tonic = "0.11.0"
+tokio = { version = "1.37.0", features = ["sync"] }
+tonic = { version = "0.13.0", path = "../tonic", default-features = false, features = ["codegen"] }
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "grpc"
+version = "0.9.0"
+edition = "2021"
+authors = ["gRPC Authors"]
+license = "Apache-2.0"
+
+[dependencies]
+url = "2.5.0"
+tokio = { version = "1.37.0", features = ["sync", "full" ] }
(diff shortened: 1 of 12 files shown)


LucioFranco (approved):

grpc/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "grpc"
+version = "0.9.0"
+edition = "2021"
+authors = ["gRPC Authors"]
+license = "Apache-2.0"
+
+[dependencies]
+url = "2.5.0"
+tokio = { version = "1.37.0", features = ["sync", "full" ] }
+tonic = "0.11.0"
```
LucioFranco, line 11: this needs to be a path dep with a version (look at how the other crates do it)

grpc/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "grpc"
+version = "0.9.0"
+edition = "2021"
+authors = ["gRPC Authors"]
+license = "Apache-2.0"
+
+[dependencies]
+url = "2.5.0"
+tokio = { version = "1.37.0", features = ["sync", "full" ] }
```
LucioFranco, line 10: we should reduce the feature flags here, if we need a runtime add it as a dev-dep

grpc/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "grpc"
+version = "0.9.0"
```
LucioFranco, line 3:
```suggestion
version = "0.9.0-alpha.1"
```

dfawley: Done

grpc/Cargo.toml
dfawley: Is `sync` right here, then?

grpc/Cargo.toml
dfawley: Fixed (I think?) - PTAL please.

grpc/Cargo.toml
LucioFranco: Yeah, if the library item you need is within the sync crate then you can just include that feature. There should be a feature flag overview in the main docs of tokio.

LucioFranco (approved):

PR #2199, opened 2025-03-04
fix(critical): `tonic::Streaming` enters infinite loop on response error

Suficio:
## Motivation

`tonic::Streaming` enters an infinite loop if an HTTP error is received during stream operation.

Typically this happens if the server cuts the connection.

This was not caught as users typically implement re-connect logic when the first error is emitted by `tonic::Streaming`, however, this does not hold if the implementation expects the stream to exit.

## Solution

Correctly handle response errors within the state machine.


LucioFranco (approved):

shikhar:
@LucioFranco @tottoto this one is probably worth a bugfix release

LucioFranco:
@shikhar do you mean a backport into 0.12? it is released with https://github.com/hyperium/tonic/releases/tag/v0.13.0

shikhar:
@LucioFranco No I mean a 0.13.1, I don't believe this was in 0.13.0?

PR #2208, opened 2025-03-06
Reconnect `tonic::Connection` on denied gRPC request due to server error

Suficio:
## Motivation

While operating in production and interfacing with a third-party API I noticed that `tonic::Connection` was not being reconnected after receiving an unrecoverable HTTP server error.

Likely the third party is using an HTTP2 load-balancing proxy that keeps our TCP connection alive but refuses to serve any HTTP after a connection lifetime has been exceeded. 

## Solution

Inspect the output of `reconnect::ResponseFuture`, if it returned an HTTP server error we use an atomic flag to request a reconnect during the next invoke of `Reconnect::poll_ready` assuming that it is currently connected.

---

This PR _may_ address a related issue: https://github.com/hyperium/tonic/issues/1254

I was not able to come up with an integration test case, however, its likely possible to catch all remaining error cases by reconnecting in the error branch in `reconnect::ResponseFuture`.


sonthonaxrk:
This would be extremely useful

humanagent:
thks!

tonic/src/transport/channel/service/reconnect.rs
```diff
@@ -186,31 +218,42 @@ enum Inner<F> {
...
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        //self.project().inner.poll(cx).map_err(Into::into)
         let me = self.project();
         match me.inner.project() {
-            InnerProj::Future(fut) => fut.poll(cx).map_err(Into::into),
+            InnerProj::Future(fut) => match ready!(fut.poll(cx)) {
+                Ok(res) => {
+                    // Server errors map to `Status::UNAVAILABLE` and are unrecoverable
+                    if res.status().is_server_error() {
```
LucioFranco, line 249: This doesn't feel correct, what if you are connected to a proxy and want to keep the connection alive but the downstream from the proxy returned a 500 but it was transient. So now you did all that work to reconnect even though keeping the connection would have been better?

LucioFranco (changes requested):
I think that this PR limits how the user can react to certain status codes and makes some choices that may not work for every one.

I also found that I can fix the test by doing something like:
```
diff --git a/tests/integration_tests/tests/connection.rs b/tests/integration_tests/tests/connection.rs
index c802cf5..1924d8b 100644
--- a/tests/integration_tests/tests/connection.rs
+++ b/tests/integration_tests/tests/connection.rs
@@ -210,6 +210,8 @@ async fn connect_lazy_reconnects_after_connection_denied() {
     client.unary_call(Input {}).await.unwrap();
     client.unary_call(Input {}).await.unwrap_err();

+    let mut client = TestClient::connect(format!("http://{addr}")).await.unwrap();
+
     // Expect reconnect, without reconnect we would see the error condition persist
     client.unary_call(Input {}).await.unwrap();
     client.unary_call(Input {}).await.unwrap_err();

tonic/src/transport/channel/service/reconnect.rs
Suficio: My understanding is that if we receive a non-successful code then we should be expecting a 200 status with `grpc-header` set.

Other status codes would imply that we have lost the gRPC connection.

To put it into context, TCP proxies do not implement gRPC and would not be expected know to signal transient conditions using 500 error codes.

Suficio:
> I think that this PR limits how the user can react to certain status codes and makes some choices that may not work for every one.
> 
> I also found that I can fix the test by doing something like:
> 
> ```
> diff --git a/tests/integration_tests/tests/connection.rs b/tests/integration_tests/tests/connection.rs
> index c802cf5..1924d8b 100644
> --- a/tests/integration_tests/tests/connection.rs
> +++ b/tests/integration_tests/tests/connection.rs
> @@ -210,6 +210,8 @@ async fn connect_lazy_reconnects_after_connection_denied() {
>      client.unary_call(Input {}).await.unwrap();
>      client.unary_call(Input {}).await.unwrap_err();
> 
> +    let mut client = TestClient::connect(format!("http://{addr}")).await.unwrap();
> +
>      // Expect reconnect, without reconnect we would see the error condition persist
>      client.unary_call(Input {}).await.unwrap();
>      client.unary_call(Input {}).await.unwrap_err();
> ```

This works in the common case.

It is more difficult to decide whether this is the correct behavior if we take into account balanced connections.

LucioFranco:
> It is more difficult to decide whether this is the correct behavior if we take into account balanced connections.

I think my unfortunate answer right now is that if you want to add more logic to balancing connections for the moment you may want to take what is within tonic now and use the tower components yourself. The current load balance stuff was actually just a stop gap. We are also working on new transport's that will be much more closely aligned with what other grpc crates do but this is still a ways out. Which is another reason I don't really want to touch the current code that much.

LucioFranco:
Based on my comment above, I am going to close this for now. Work has already started in the new `grpc` folder.

PR #2209, opened 2025-03-06
chore: Expose `TlsError`

ratankaliani:
## Motivation

When parsing errors from `tonic`, we currently apply conditional logic to notify users if their machine lacks native root certificates. We currently do this based off of the string returned from the error.

If `tonic` exposed the `TlsError` type, we'd be able to do this in a more type-safe way.

## Solution

Expose `TlsError` from `tonic::transport`.


ratankaliani:
@tottoto What's the process for a review here?

PR #2213, opened 2025-03-08
chore(server): Refactor default http2 keepalive timeout config

tottoto:
Refactors the default http2 keepalive timeout config.


LucioFranco:
Whats the motivation for this change?

tottoto:
@LucioFranco

> Whats the motivation for this change?

Refactoring.

tonic/src/transport/server/mod.rs
```diff
@@ -122,7 +120,7 @@ impl Default for Server<Identity> {
             tcp_keepalive: None,
             tcp_nodelay: false,
             http2_keepalive_interval: None,
-            http2_keepalive_timeout: None,
+            http2_keepalive_timeout: Duration::from_secs(20),
```
LucioFranco, line 123: I think this is fine can we add back the const just to make it easier to discover the value in the code?

LucioFranco (approved):

PR #2223, opened 2025-03-16
chore: Remove unnecessary docs.rs config

tottoto:
This config is set as default.

https://docs.rs/about/builds#detecting-docsrs

### DIFF
--- a/tonic-reflection/Cargo.toml
+++ b/tonic-reflection/Cargo.toml
@@ -20,7 +20,6 @@ rust-version = { workspace = true }
 
 [package.metadata.docs.rs]
 all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 server = ["dep:prost-types", "dep:tokio", "dep:tokio-stream"]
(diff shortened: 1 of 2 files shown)


LucioFranco (commented):
Do we also need to remove https://github.com/hyperium/tonic/blob/master/tonic/src/lib.rs#L96? And does docrs now by default do the feature flag doc stuff?

tottoto:
@LucioFranco

> Do we also need to remove https://github.com/hyperium/tonic/blob/master/tonic/src/lib.rs#L96?

No, it needs to be enabled when docsrs is specified.

> And does docrs now by default do the feature flag doc stuff?

It is done with the config.

tottoto:
@LucioFranco 

> ```
> #![cfg_attr(docsrs, feature(doc_auto_cfg))]
> ```

LucioFranco:
Right my question was does docs.rs auto set this flag now? If not then removing this config option from Cargo.toml will now disable the feature flag stuff we get in the docs?

tottoto:
@LucioFranco It is set by docs.rs. tonic-build shows that it works, which doesn't have the config.

LucioFranco (approved):

PR #2229, opened 2025-03-24
chore: Update cargo-deny config

adamcavendish:
The paste dependency in aws-lc-rs is removed after v1.12.6 so the only paste dependency is in tonic's tests.
The change migrates the current usage of paste to pastey package.

I also cherry-picked your change in #2210 so the `cargo deny check` will pass.

Reference: https://github.com/aws/aws-lc-rs/releases/tag/v1.12.6


adamcavendish:
@tottoto would you help review the change? Thanks!

tottoto:
We are addressing this at #2210.

adamcavendish:
> We are addressing this at #2210.

Yeah, as I said in the description that I cherry-picked your change too. You can see that the change is slightly a bit different than that one. Feel free to take any of the path to unblock the CI pipelines.

PR #2231, opened 2025-03-24
chore: Use cargo-deny exceptions config

tottoto:
Uses cargo-deny exceptions config.


LucioFranco:
deny has been removed so im gonna close this

PR #2234, opened 2025-03-25
chore(server): Remove import sleep and pending function

tottoto:
Removes import sleep and pending function.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -50,7 +50,6 @@ use http::{Request, Response};
 use http_body_util::BodyExt;
 use hyper::{body::Incoming, service::Service as HyperService};
 use pin_project::pin_project;
-use std::future::pending;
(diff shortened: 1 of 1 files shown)


LucioFranco:
This just seems like a style change and maybe we don't need this? Not sure what the exact motivation is here.

tottoto:
@LucioFranco The motivation is refactoring. In general, importing functions directly is avoided as it reduces readability. In particular, in this case there is a `sleep` variable in the file, and since these used only once in this file, having the full path to it in the place where it is used improves readability.

LucioFranco (approved):

tottoto:
@LucioFranco I think this motivation is sufficiently obvious, but it is possible you want some explanation as this case. It is very difficult to decide what should be explained in advance. I think it would be good if that could be clarified during the discussion in each case. I think it is good to ask it when having some question.

PR #2238, opened 2025-03-26
Allow direct access to modify the TLS config

n1ght-hunter:
## Motivation
For work we need direct access to the tls config to customise the `ServerCertVerifier`.
Also direct access allow any further  config changes to made without  any changes to tonic

## Solution
To solve access to the tls i have added a method to `ClientTlsConfig` called `modify_config`(could be named somthing else maybe).
```rs
 pub fn modify_config<F>(self, f: F) -> Self
    where
        F: Fn(&mut tokio_rustls::rustls::ClientConfig) + Send + Sync + 'static,
    {
        let modify_config = ModdifyConfigFn(std::sync::Arc::new(f));
        ClientTlsConfig {
            modify_config: Some(modify_config),
            ..self
        }
    }
```
this allows users to directly modify `tokio_rustls::rustls::ClientConfig` to fit their needs.
this function is called in `TlsConnector::new` after all tonic config changes are set

### DIFF
--- a/tonic/src/transport/channel/service/tls.rs
+++ b/tonic/src/transport/channel/service/tls.rs
@@ -13,10 +13,13 @@ use tokio_rustls::{
 };
 
 use super::io::BoxedIo;
-use crate::transport::{channel::tls::ModdifyConfigFn, service::tls::{
-    convert_certificate_to_pki_types, convert_identity_to_pki_types, TlsError, ALPN_H2,
-}};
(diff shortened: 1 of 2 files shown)


LucioFranco:
I would suggest that you actually make your own custom transport and in that one you can set rustls (you can even just use tonic's channel but inject the connector/acceptor to be custom and there you can set rustls how you would like).

PR #2250, opened 2025-04-10
feat: preserve request user-agent

### LINKED ISSUE #970: Have full control over the User-Agent
### Motivation

The User-Agent header is populated, by default, with `tonic/x.y.z` (x.y.z. being the current tonic version number) \
There is a [user_agent](https://docs.rs/tonic/latest/tonic/transport/channel/struct.Endpoint.html#method.user_agent) method to somehow customize this by adding a prefix to this `tonic/x.y.z` part.

### Proposal

In some use cases, users could want to _totally_ customize the user-agent. Not suffixing with `tonic/x.y.z` would probably be a good idea.

I'm not sure why this suffix has been made compulsory anyway (but maybe I'm missing something? Is this required by some spec? I feel that any server _requiring_ to know the client version is probably poorly designed)

Thanks

dbolduc:
## Motivation

Currently, tonic will drop an existing `User-Agent` header on a request. This means middleware and applications built on top of tonic will not be able to add their own user agent strings.

Fixes #970, although we go with a different solution than proposed in the issue.

Part of the work for https://github.com/googleapis/google-cloud-rust/issues/1612

## Solution

Append tonic's user agent string to the `User-Agent` header, if the `User-Agent` has already been set in the request.

## Testing

Add unit tests that exercise the behavior with a fake `Service`.


mrfaizal (approved):

dbolduc:
ping?

LucioFranco (approved):

PR #2252, opened 2025-04-11
fix: tls config overwrite in endpoint

vigneshs-12:
## Motivation

PR #1866 fixed the breaking change introduced in #1731, but resets the TLS config without checking if `tls` is set. This patch resolves the regression and restores expected behaviour.

### Description

 `x::connect()` creates a new `Endpoint` using `Endpoint::new` and it creates a fresh  `ClientTlsConfig` for setting enabled roots without any checks. This causes any endpoint with `ClientTlsConfig` set before to be reset when calling `Endpoint::new`. 
 
Also, `with_enabled_rootfs`, creates a fresh config to add roots instead of modifying the existing `ClientTlsConfig` instance.   This could cause any fields set in `ClientTlsConfig` to be reset when this method is called.

## Solution

Check if `tls` (TlsConnector) is not set before creating a fresh `ClientTlsConfig`.


vigneshs-12:
FYI, This is found when upgrading our project from 0.11.x to 0.12.x. We use self-signed certificate for mTLS authentication and upgrading to 0.12.x caused `UnknownIssuer` when connecting to the server from our client.

LucioFranco (approved):

PR #2258, opened 2025-04-28
chore(server): Remove unnecessary await service ready

tottoto:
This service can be `call`ed immediately because its `poll_ready` is guaranteed to always return `Ready(Ok())`.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -52,7 +52,7 @@ use hyper::{body::Incoming, service::Service as HyperService};
 use pin_project::pin_project;
 use std::{
     fmt,
-    future::{self, poll_fn, Future},
+    future::{self, Future},
     marker::PhantomData,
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -725,10 +725,6 @@ impl<L> Server<L> {
 
                     trace!("connection accepted");
 
-                    poll_fn(|cx| svc.poll_ready(cx))
-                        .await
-                        .map_err(super::Error::from_source)?;
-
```
LucioFranco, line 731: technically speaking a user can pass in a service that requires calling poll_ready so removing this could break that? I don't think we need to actually remove this?

tottoto: Whatever service users pass in will not affect this process. This process does not process the user's service.

LucioFranco (approved):

PR #2274, opened 2025-05-12
grpc: add initial grpc/service definition and Channel type

dfawley:


grpc/Cargo.toml
```diff
@@ -9,3 +9,4 @@ license = "Apache-2.0"
 url = "2.5.0"
 tokio = { version = "1.37.0", features = ["sync"] }
 tonic = { version = "0.13.0", path = "../tonic", default-features = false, features = ["codegen"] }
+futures-core = "0.3.31"
```
LucioFranco, line 12: we should depend on `tokio-stream` rather than any `futures-*` crates.

LucioFranco (approved):
LGTM we can fix the dep in a follow up

Issue #2253, opened 2025-04-14
`tonic::Result` and `tonic::Status` trigger `clippy::result_large_err` on rust 1.87


eric-seppanen:
## Bug Report

While testing the rust 1.87 beta toolchain, I noticed that clippy now raises warnings on any code that uses `tonic::Result` or otherwise contains a `tonic::Status` in its `Err` variant. The `result_large_err` lint is enabled by default.

I don't have an opinion about whether clippy's thresholds are correct; just wanted to give the tonic developers a heads-up as this will probably trigger warnings in a lot of downstream code that uses tonic.

### Version

tonic v0.13.0 (also happens with tonic 0.12.3).

### Description

Here's a trivial example that shows the warnings:
```rust
pub fn thing1() -> tonic::Result<()> {
    Ok(())
}
```

`cargo +beta clippy` says:
```plaintext
warning: the `Err`-variant returned from this function is very large
 --> src/lib.rs:1:19
  |
1 | pub fn hello() -> tonic::Result<()> {
  |                   ^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 176 bytes
  |
  = help: try reducing the size of `tonic::Status`, for example by boxing large elements or replacing it with `Box<tonic::Status>`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
  = note: `#[warn(clippy::result_large_err)]` on by default
```

Or alternatively, without explicitly using `tonic::Result` but propagating the error type:
```rust
pub enum MyError {
    Oops,
    Tonic(tonic::Status),
}

pub fn thing2() -> Result<(), MyError> {
    Ok(())
}
```

```plaintext
warning: the `Err`-variant returned from this function is very large
 --> src/lib.rs:6:20
  |
3 |     Tonic(tonic::Status),
  |     -------------------- the largest variant contains at least 176 bytes
...
6 | pub fn thing2() -> Result<(), MyError> {
  |                    ^^^^^^^^^^^^^^^^^^^
  |
  = help: try reducing the size of `MyError`, for example by boxing large elements or replacing it with `Box<MyError>`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
  = note: `#[warn(clippy::result_large_err)]` on by default
```

LucioFranco:
We can add this lint to codegen. I would welcome a PR.

tustvold:
[...]

PR #2282, opened 2025-05-20
chore: box Status contents (#2253)

tustvold:
## Motivation

See #2253

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -37,12 +37,6 @@ const ENCODING_SET: &AsciiSet = &CONTROLS
 #[derive(Clone)]
 pub struct Status(Box<StatusInner>);
 
-impl From<StatusInner> for Status {
-    fn from(value: StatusInner) -> Self {
-        Self(Box::new(value))
-    }
-}
-
 /// Box the contents of Status to avoid large error variants
 #[derive(Clone)]
 struct StatusInner {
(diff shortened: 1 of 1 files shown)


tonic/src/status.rs
```diff
@@ -35,7 +35,17 @@ const ENCODING_SET: &AsciiSet = &CONTROLS
 /// assert_eq!(status1.code(), status2.code());
 /// ```
 #[derive(Clone)]
-pub struct Status {
+pub struct Status(Box<StatusInner>);
+
+impl From<StatusInner> for Status {
```
tustvold, line 40: This was marginally more ergonomic than creating a constructor, but let me know if you would prefer a constructor instead

LucioFranco: I think we need to make this a private function on status so that it doesn't show up in the docs? I don't want the docs to even reference `StatusInner` at all.

LucioFranco (changes requested):

tonic/src/status.rs
tustvold: Good point, although `#[doc(hidden)]` would also work

okkero (approved):
Looks good to me, though someone with better insight into the project than me might have more to say. Addresses the issue straight forwardly. 🎉

LucioFranco (approved):

PR #2283, opened 2025-05-21
feat(tonic): rpit future for generated server traits

kvc0:
## Motivation
Probably the main motivation here is that when I use vs code to auto-generate fn's for a tonic stub, it spews out the guts of async_trait. It's educational, but it's quite heavy and unnecessary since Rust 1.75. I'm pretty confident that `Future + Send` is about as strict as tonic intends to be toward generated service stubs, but of course let's chat about it.

Secondarily, I count microseconds at both my day job and my home hobby work. I also spend a lot of time stepping in a debugger. Box types are pretty annoying to deal with in the vs code debugger, and they do tend to cost a little time. I use `fn f() -> impl Future<Output = _> + Send` in all my async traits nowadays. It works great at home and at work.

## Solution
Supported at 1.75, which is tonic's current MSRV, return position impl trait (RPIT) makes user code cleaner, friendlier in a debugger, and possibly a couple nanoseconds faster due to fewer boxed Future indirections. There is still one Box Future at the tower_service interface due to the switch from RPIT style "simple" traits and the fancier static associated type Future style from tower.

This unwraps 2 layers of Box indirection into static types: 1 in the framework's generated GRPC code below tower, and 1 in user code specified by async_trait.

It's similar to https://github.com/hyperium/tonic/pull/1697 but instead I went for the full change, to evaluate the expected impact more fully.

## Impact

In the expected use cases where people generate their code, they would remove `#[tonic::async_trait]` from their implementations of Server stubs. You can see in all the tests and examples, that's the only typical user-facing delta.

In the use cases where people write their own non-generated Server implementations, they would need to move the trait impl's Future type into the call function's return type. If they're reimplementing UnaryService, they would possibly need to explicitly narrow any wrapped tower_service::Future to Send as the UnaryService (and friends) now explicitly does.

It is a breaking change, but it's a simplification that seemed worthwhile to me with the burden of 1-line deletion per service. The vs code auto-generated todo!() implementations are a little nicer this way too, which is a quality of life boost for me :-)

Relates to: https://github.com/hyperium/tonic/issues/1651

________________
**UX before:**
```rust
 #[tonic::async_trait]
 impl Greeter for MyGreeter {
```

**UX after:**
```rust
 impl Greeter for MyGreeter {
```

________________

**Testing**

Due to the nature of the change, a wide range of tests exercise it. It is a change to the existing infrastructure rather than a new feature, so there's no duplication or parameterization of tests that would increase coverage. The tests were quite reassuring, in that they only required the removal of `#[tonic::async_trait]`, which was by design.


kvc0:
> [Good luck](https://github.com/hyperium/tonic/pull/1556), I tried to land it a while ago but it was blocked on raising MSRV.

[MSRV is currently 1.75](https://github.com/hyperium/tonic?tab=readme-ov-file#rust-version), which is sufficient for this change.

LucioFranco (changes requested):
I took a quick look and overall it looks pretty good. I think if we are going to land this though I would want this to stick as an opt-in only and let users migrate to this codegen version when they want.

tonic-build/src/server.rs
```diff
@@ -265,41 +264,41 @@ fn generate_trait_methods<T: Service>(
             (false, false, true) => {
                 quote! {
                     #method_doc
-                    async fn #name(#self_param, request: tonic::Request<#req_message>)
-                        -> std::result::Result<tonic::Response<#res_message>, tonic::Status> {
-                        Err(tonic::Status::unimplemented("Not yet implemented"))
+                    fn #name(#self_param, request: tonic::Request<#req_message>)
+                        -> impl std::future::Future<Output = std::result::Result<tonic::Response<#res_message>, tonic::Status>> + Send {
+                        async { Err(tonic::Status::unimplemented("Not yet implemented")) }
```
TroyKomodo, line 269: We should just return `std::future::ready()` instead of a async block, less work for the optimizer to do.

nrabulinski: As a consumer of this library, I don't think generating `ready(Err(...))` is particularly useful... It's not meant to be deployed, just a placeholder code for you to jump off of. In fact I'd even argue for generating `async move` block, but alas. Point being, I think the optimizer point is moot for a piece of code you'll replace before you even compile, and you'll probably replace it with an async block anyway

nathanielford:
@kvc0 Do you still have an interest in getting this change landed? We might need to get it updated and re-assessed, but in the process of transitioning to grpc-rust from tonic we'd like to go through and clean up outstanding PRs. Thanks for your time!

kvc0:
> @kvc0 Do you still have an interest in getting this change landed? We might need to get it updated and re-assessed, but in the process of transitioning to grpc-rust from tonic we'd like to go through and clean up outstanding PRs. Thanks for your time!

Oh hey! Yes, I still want it and I think it will improve tonic and any fork thereof. I have a conference this week which makes my availability to update the pr a bit latent but I’ll take a pass on it as soon as I can.

nathanielford:
No huge rush!

As note, however, `grpc-rust` is a floor-up rebuild to track the gRPC spec more tightly, and not technically a fork - though there are currently some shared crates. When you get a chance to take a look let me know if you think it's worth the time to bring things up to date!

nathanielford:
Thanks for working on this! Active development on legacy Tonic is winding down, and the new official `grpc-rust` library (`protoc-gen-rust-grpc`) natively supports modern async traits without `#[async_trait]` boxing. Because maintaining a dual code generation pipeline in `tonic-build` adds substantial complexity during this transition, we won't be able to land this. We really appreciate the effort and look forward to your feedback on the new library. Let us know if you have any outstanding concerns. Thanks again!

nathanielford:
Apologies for the close - too many tabs open!

PR #2291, opened 2025-06-02
chore(build): Make empty client and server modules private

ADSteele916:
Currently, `tonic_build` exports the `client` and `server` modules, both of which contain only private items. Exporting these modules only serves to clutter up the crate's docs. This PR makes the two modules private.


tottoto:
Sounds reasonable to me. It looks needed to be waited until the next release with breaking changes.

LucioFranco (approved):

JASON27333 (approved):

PR #2293, opened 2025-06-04
fix(tonic): make `Streaming` `Sync` again

Twey:
## Motivation

The boxed `Decoder` inside `Streaming` need not be `Sync` since https://github.com/hyperium/tonic/pull/804.  Unfortunately, that makes `Streaming` non-`Sync`, meaning that all the generated `tonic` futures cannot be awaited in `Sync` futures.

## Solution

In fact, the only times we use the `Decoder`, we have a `&mut` unique reference to it, so we are guaranteed not to require synchronization.

The `sync_wrapper` crate encodes this reasoning, allowing us to safely make the `Streaming` type `Sync` regardless of whether the contained `Decoder` is `Sync` or not.


LucioFranco (approved):

LucioFranco:
Thanks

PR #2297, opened 2025-06-11
fix outdated limit in comment

darrasse:
This mention was missed when the default changed in #1335

## Motivation

Avoid confusion.

## Solution

Update the comment to mention the correct value.

### DIFF
--- a/tonic/src/client/grpc.rs
+++ b/tonic/src/client/grpc.rs
@@ -181,7 +181,7 @@ impl<T> Grpc<T> {
     ///     .await
     ///     .unwrap();
     ///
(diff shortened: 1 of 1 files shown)


tonic/src/client/grpc.rs
```diff
@@ -181,7 +181,7 @@ impl<T> Grpc<T> {
     ///     .await
     ///     .unwrap();
     ///
-    /// // Set the limit to 2MB, Defaults to 4MB.
+    /// // Set the limit to 2MB, Defaults to usize::MAX.
```
tottoto:
```suggestion
    /// // Set the limit to 2MB, Defaults to `usize::MAX`.
```

PR #2315, opened 2025-06-27
Merge changes from next to master branch

dfawley:


dfawley:
cc @cjqzhao @arjan-bal @easwars

grpc/src/byte_str.rs
```diff
@@ -0,0 +1,51 @@
+/*
```
LucioFranco, line 1: probably before we merge we should update the license in accordance to https://github.com/cncf/foundation/issues/985#issuecomment-2949404196

LucioFranco (changes requested):

grpc/src/byte_str.rs
dfawley: Done

grpc/Cargo.toml
```diff
@@ -3,9 +3,29 @@ name = "grpc"
...
+hickory-server =  "0.25.2"
+
+[features]
+default = ["dns"]
+dns = ["dep:hickory-resolver"]
+
+[package.metadata.cargo_check_external_types]
+allowed_external_types = [
+    "tonic::*",
+    "futures_core::stream::Stream",
```
LucioFranco, line 30: We should pull from `tokio_stream` instead

LucioFranco: @arjan-bal @dfawley @easwars if one of you can get to this simple fix that would be good, I am going to merge this anyways now.

dfawley: `tokio_stream` re-exports `futures_core::Stream`, so are these just equivalent?

https://docs.rs/tokio-stream/latest/tokio_stream/#reexports

LucioFranco: Yes, they are except that the tokio org controls the tokio_stream crate and I would rather depend on that than on the futures set of crates. (I have publish access for tokio_stream etc)

PR #2320, opened 2025-07-02
feat(grpc): Add `protobuf` codegen

arjan-bal:
This PR includes the following:
1. A tonic codec that uses protobuf-rust.
2. A protoc plugin that generates tonic compatible client code.
3. A `grpc-build` crate that helps generate code during cargo builds.
4. Interop tests that use the new client codgen running against a Go server. This is used as an integration test for the new codegen.

To keep the CI fast, the protoc plugin binary for each OS is cached. The plugin binaries are re-built only if the plugin's code or build files are updated.

## Notes
* The codegen is generating code for tonic. It will be changed when the gRPC API design is finalized and the channel implementation is complete.


compiler/README.md
```diff
@@ -0,0 +1,29 @@
+## Usage example
```
LucioFranco: Should we rename this containg folder? Compiler seems quite ambiguous?

interop/test.sh
```diff
@@ -59,6 +59,9 @@ sleep 1
 
 ./target/debug/client --test_case="${JOINED_TEST_CASES}" "${ARG}"
 
+# Test a grpc rust client against a Go server.
+./target/debug/client_grpc --test_case="${JOINED_TEST_CASES}" ${ARG}
```
LucioFranco: should probably find some better names here too

compiler/README.md
arjan-bal: Yes, it is. I was using the same name as [Java](https://github.com/grpc/grpc-java/tree/master/compiler) and the [core](https://github.com/grpc/grpc/tree/master/src/compiler) repo. I've changed it to the name of protoc plugin, `protoc-gen-rust-grpc` now.

interop/test.sh
arjan-bal: I've updated the file names to `client_new_codegen`. Open to better suggestions if you have any in mind.

arjan-bal:
@dfawley please have a look

grpc/src/codec/protobuf.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+}
+
+impl<T: Message> Encoder for ProtoEncoder<T> {
+    type Item = T;
+    type Error = Status;
+
+    fn encode(&mut self, item: Self::Item, buf: &mut EncodeBuf<'_>) -> Result<(), Self::Error> {
+        let serialized = item.serialize().map_err(from_decode_error)?;
+        buf.put_slice(&serialized.as_slice());
```
arjan-bal: @LucioFranco I can think of the following ways to avoid a copy here:

1. Request the protobuf crate to expose an API that allows deserialization into a destination buffer.
2. Change the codec API to return a `Vec<u8>` instead of accepting the destination as a parameter. Tonic can then create an `EncodedBuf` from the returned `Vec<u8>`.
3. Make the constructor of `EncodedBuf` public. This would allow a codec to create its own `EncodedBuf` and `std::mem::swap` it with the one passed to the `encode` method, avoiding a copy.

I think option 3 is a reasonable workaround that doesn't require API changes. Wanted to get your thoughts on this.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
```diff
@@ -0,0 +1,138 @@
...
+      }
+    }
+
+    if (grpc_opts.message_module_path.empty()) {
+      grpc_opts.message_module_path = "self";
+    }
+
+    // The kernel isn't used by gRPC, it is there to pass Rust protobuf's
+    // validation.
+    protobuf_options.emplace_back("kernel", "upb");
```
acozzette: It seems awkward that you have to do this just to make our helper code happy. I think we should figure out a way to avoid needing to do this.

As far as I can tell, the only thing you need `google::protobuf::compiler::rust::Context` for is to be able to use the naming helper function `RsTypePath`. The reason that helper needs a `Context` is so that it can figure out how to refer to the symbol unambiguously based on the current position in the code (see logic [here](https://github.com/protocolbuffers/protobuf/blob/527b2ab41e785149a6189dba3052a210959b1372/src/google/protobuf/compiler/rust/naming.cc#L193)).

arjan-bal: Yes, I needed to create a context to call `protobuf::compiler::rust::RsTypePath`. The context has information about the crate mapping that it uses for calling `GetCrateName` [here](https://github.com/protocolbuffers/protobuf/blob/527b2ab41e785149a6189dba3052a210959b1372/src/google/protobuf/compiler/rust/naming.cc#L223C23-L223C36).

arjan-bal: I've copied the symbol resolution code and the crate mapping parsing code to avoid using `rust::Context`.

protoc-gen-rust-grpc/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```sh
+# Build the plugin
+bazel build //src:protoc-gen-rust-grpc
+
+# Set the plugin path
+PLUGIN_PATH="$(pwd)/bazel-bin/src/protoc-gen-rust-grpc"
+
+# Run protoc with the Rust and Rust gRPC plugins
+protoc \
```
acozzette, line 26: I think we should probably advise users to use `grpc_build::CodeGen` and/or `protobuf_codegen::CodeGen` instead of directly invoking protoc. If you directly invoke protoc and check in the generated code, then it's at risk of becoming stale if/when you upgrade your protobuf dependency.

arjan-bal: Added a note discouraging the direct use of the plugin. I've also added a README in the `tonic_protobuf_build` (formerly `grpc_build`) crate.

protoc-gen-rust-grpc/README.md
```diff
@@ -0,0 +1,70 @@
...
+# Build the plugin
+bazel build //src:protoc-gen-rust-grpc
+
+# Set the plugin path
+PLUGIN_PATH="$(pwd)/bazel-bin/src/protoc-gen-rust-grpc"
+
+# Run protoc with the Rust gRPC plugin
+protoc \
+  --plugin=protoc-gen-grpc-rust="$PLUGIN_PATH" \
+  --rust_opt="experimental-codegen=enabled,kernel=upb" \
```
acozzette, line 28: I think with your recent changes this line should be unnecessary now, right?

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// List of .proto file names whose codegen is present in the crate.
+    pub proto_files: Vec<String>,
+}
+
+impl From<&Dependency> for protobuf_codegen::Dependency {
+    fn from(val: &Dependency) -> Self {
+        protobuf_codegen::Dependency {
+            crate_name: val.crate_name.clone(),
+            proto_import_paths: val.proto_import_paths.clone(),
+            // TODO: Is this useful to expose the following field? It's not used
```
acozzette: Good point, I made a note that I should delete this now that we're no longer using it.

Is there any reason for creating your own `Dependency` struct other than avoiding this unnecessary field? Because it might be nice to be able to reuse the same struct. My thought was that we would eventually have protoc generate the `Dependency` struct so that you could just call `my_dependency::get_dependency()` to get ahold of it without having to create it manually.

interop/src/client_new_codegen.rs
```diff
@@ -0,0 +1,436 @@
...
+
+        assertions.push(test_assert!(
+            "body must be 314159 bytes",
+            payload_len == LARGE_RSP_SIZE as usize,
+            format!("mem::size_of_val(&body)={:?}", mem::size_of_val(&body))
+        ));
+    }
+}
+
+// pub async fn cachable_unary(client: &mut Client, assertions: &mut Vec<TestAssertion>) {
```
dfawley: Why is this commented out?

protoc-gen-rust-grpc/README.md
dfawley: I'm not opposed to recommending it that way, but is this a real concern?  I thought the protobuf library would support older generated code, but not vice-versa?  I'm fairly sure that's true for Go at least.

protoc-gen-rust-grpc/README.md
```diff
@@ -0,0 +1,70 @@
...
+```sh
+# Build the plugin
+bazel build //src:protoc-gen-rust-grpc
+
+# Set the plugin path
+PLUGIN_PATH="$(pwd)/bazel-bin/src/protoc-gen-rust-grpc"
+
+# Run protoc with the Rust gRPC plugin
+protoc \
+  --plugin=protoc-gen-grpc-rust="$PLUGIN_PATH" \
```
dfawley: `protoc-gen-rust-grpc`?

protoc-gen-rust-grpc/src/BUILD
```diff
@@ -0,0 +1,12 @@
+cc_binary(
+    name = "protoc-gen-rust-grpc",
+    srcs = [
+    "grpc_rust_plugin.cc",
+    "grpc_rust_generator.h",
+    "grpc_rust_generator.cc",
```
dfawley: Indent?

protoc-gen-rust-grpc/src/BUILD
```diff
@@ -0,0 +1,12 @@
+cc_binary(
```
dfawley, line 21: I think we want copyrights on BUILD files too.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
```diff
@@ -0,0 +1,101 @@
...
+      rust_grpc_generator::GenerateService(printer, service, grpc_opts);
+    }
+    return true;
+  }
+};
+
+int main(int argc, char *argv[]) {
+  RustGrpcGenerator generator;
+  return protobuf::compiler::PluginMain(argc, argv, &generator);
+  return 0;
```
dfawley:
```suggestion
```

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace rust_grpc_generator {
+
+namespace impl {
+namespace protobuf = google::protobuf;
+} // namespace impl
+
+class GrpcOpts {
+public:
+  /// Path the module containing the generated message code. Defaults to
```
dfawley:
```suggestion
  /// Path of the module containing the generated message code. Defaults to
```

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+  /// "self", i.e. the message code and service code is present in the same
```
dfawley:
```suggestion
  /// "self", i.e. the message code and service code are present in the same
```

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+
+        cmd.arg(format!("--rust-grpc_out={}", self.output_dir.display()));
+        cmd.arg(format!(
+            "--rust-grpc_opt=crate_mapping={}",
+            crate_mapping_path.display()
+        ));
+        if let Some(message_path) = &self.message_module_path {
+            cmd.arg(format!(
```
dfawley, line 241: Please fix the clippy warnings.

tonic-protobuf-build/src/lib.rs
arjan-bal: > Is there any reason for creating your own Dependency struct other than avoiding this unnecessary field? 

I did this to avoid adding an external struct type (Dependency) to the public API of this crate as a best practice.

interop/src/client_new_codegen.rs
arjan-bal: It's commented out in the original file too: https://github.com/hyperium/tonic/blob/cde96ac0f695b0ebdd11166414e9b87fa8ee9506/interop/src/client.rs#L70-L82

The test case is incomplete. Removing it from the new file.

protoc-gen-rust-grpc/README.md
arjan-bal: Fixed.

protoc-gen-rust-grpc/README.md
arjan-bal: I removed the `--rust-grpc_opt` flag with these options. These are the options passed to protobuf rust.

protoc-gen-rust-grpc/src/BUILD
arjan-bal: Added.

protoc-gen-rust-grpc/src/BUILD
arjan-bal: Indented.

tonic-protobuf-build/src/lib.rs
arjan-bal: Fixed throughout.

protoc-gen-rust-grpc/README.md
acozzette: We are going to make an exception for some libraries, but we do generally plan to require the generated code to match the runtime version. In most other languages we are more relaxed about this but we have found that it's more difficult to maintain protobuf when we have to accommodate stale gencode.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+#ifndef NET_GRPC_COMPILER_RUST_GENERATOR_H_
```
drfloob: You likely want the header guard to match the file path. https://google.github.io/styleguide/cppguide.html#Comment_Style

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+#define NET_GRPC_COMPILER_RUST_GENERATOR_H_
+
+#include "absl/log/absl_log.h"
```
drfloob, line 25: The project could benefit from a C++ formatter & linter. System included should precede other libraries' headers. See https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+#include <google/protobuf/compiler/rust/context.h>
+#include <google/protobuf/descriptor.h>
+#include <google/protobuf/io/zero_copy_stream.h>
+
+namespace rust_grpc_generator {
+
+namespace impl {
+namespace protobuf = google::protobuf;
```
drfloob: Why do you need the namespace alias here? Exposing `impl::protobuf` in the public API seems unnecessary.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+} // namespace impl
+
+class GrpcOpts {
```
drfloob, line 30: This class should have hidden member variables. Structs are intended to be passive objects with public member variables, but structs generally souldn't have methods. See https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+public:
+  /// Path of the module containing the generated message code. Defaults to
+  /// "self", i.e. the message code and service code are present in the same
+  /// module.
+  std::string message_module_path = "self";
+  absl::flat_hash_map<std::string, std::string> import_path_to_crate_name = {};
+  std::vector<const impl::protobuf::FileDescriptor *> files_in_current_crate =
+      {};
+
+  absl::string_view get_crate_name(absl::string_view import_path) const {
```
drfloob: C++ class method names are PascalCase unless they are simple variable accessors. See https://google.github.io/styleguide/cppguide.html#Function_Names

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+
+namespace rust_grpc_generator {
+
+namespace impl {
+namespace protobuf = google::protobuf;
+} // namespace impl
+
+class GrpcOpts {
+public:
+  /// Path of the module containing the generated message code. Defaults to
```
drfloob: There's some inconsistent use of `//` and `///`. The latter is generally used in auto doc generation, which I'm not sure is happening for this code.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  bool is_file_in_current_crate(const impl::protobuf::FileDescriptor &f) const {
+    return std::find(files_in_current_crate.begin(),
+                     files_in_current_crate.end(),
+                     &f) != files_in_current_crate.end();
+  }
+};
+
+// Writes the generated service interface into the given ZeroCopyOutputStream
```
drfloob, line 80: If auto doc generation is going to happen for this code (e.g, doxygen), use `///` for public function documentation.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
```diff
@@ -0,0 +1,92 @@
...
+// This function parses a .rust_crate_mapping file generated by a build system.
+// The file contains:
+//
+//    <crate_name>\n
+//    <number of .proto files covered by the proto_library with that name>\n
+//    <import path of the first .proto file of the crate\n
+//    ...
+//    <import path of the last .proto file of the crate\n
+absl::StatusOr<absl::flat_hash_map<std::string, std::string>>
+GetImportPathToCrateNameMap(const std::string &mapping_file_path);
```
drfloob: This can be a string_view parameter.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
+/*
```
drfloob: Consider using the `//` comment syntax for consistency throughout. See https://google.github.io/styleguide/cppguide.html#Comment_Style

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+#include "src/grpc_rust_generator.h"
```
drfloob, line 21: See the style guide recommendations on include order: https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+#include <google/protobuf/compiler/rust/naming.h>
+#include <google/protobuf/descriptor.h>
+#include <google/protobuf/descriptor.pb.h>
+#include <google/protobuf/io/zero_copy_stream.h>
+#include <google/protobuf/stubs/common.h>
+
+#include <vector>
+
+namespace rust_grpc_generator {
+namespace protobuf = google::protobuf;
```
drfloob, line 35: This namespace alias is already made public in the header (which I argue should not be the case, but it is at the moment)

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+namespace rust = protobuf::compiler::rust;
+
+using google::protobuf::FileDescriptor;
```
drfloob: Given the namespace alias above, this should just be `using protobuf::FileDescriptor`.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+using protobuf::Descriptor;
+using protobuf::MethodDescriptor;
+using protobuf::ServiceDescriptor;
+using protobuf::SourceLocation;
+using protobuf::io::Printer;
+
+template <typename DescriptorType>
+static std::string
```
drfloob: Rather than mark them static, the more modern C++ way of creating internal linkage is to put functions in an anonymous namespace. See https://google.github.io/styleguide/cppguide.html#Internal_Linkage

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+GrpcGetCommentsForDescriptor(const DescriptorType *descriptor) {
+  SourceLocation location;
+  if (descriptor->GetSourceLocation(&location)) {
+    return location.leading_comments.empty() ? location.trailing_comments
+                                             : location.leading_comments;
+  }
+  return std::string();
```
drfloob: `return ""`;

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+    depth = 0;
+  }
+  std::string prefix = "";
+  for (int i = 0; i < depth; ++i) {
+    prefix += "super::";
+  }
+  return prefix + path_to_message_module + std::string(path_within_module);
+}
+
+struct File {
```
drfloob: I don't think this struct needs to exist, it has no state. You can move the function into an anonymous namespace.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+  static absl::Status ReadFileToString(const std::string &name,
+                                       std::string *output, bool text_mode) {
+    char buffer[1024];
+    FILE *file = fopen(name.c_str(), text_mode ? "rt" : "rb");
+    if (file == nullptr)
+      return absl::NotFoundError("Could not open file");
```
drfloob: Note that there may be other issues here besides a missing file. Consider using `absl::ErrnoToStatus(errno)` to print a more relevant error. See https://linux.die.net/man/3/fopen

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+    if (error != 0) {
+      return absl::InternalError(absl::StrCat("Failed to read the file ", name,
+                                              ". Error code: ", error));
+    }
+    return absl::OkStatus();
+  }
+};
+
+absl::StatusOr<absl::flat_hash_map<std::string, std::string>>
+GetImportPathToCrateNameMap(const std::string &mapping_file_path) {
```
drfloob: I believe this can be an absl::string_view as well. `string_view::data()` should get you the char* data you need later.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+          "Couldn't parse number of import paths in mapping file");
+    }
+    for (size_t i = 0; i < files_cnt; ++i) {
+      mapping.insert({std::string(lines[idx++]), std::string(crate_name)});
+    }
+  }
+  return mapping;
+}
+
+/**
```
drfloob: Suggest consistent use of `//` for comments.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+ * Method generation abstraction.
+ *
+ * Each service contains a set of generic methods that will be used by codegen
+ * to generate abstraction implementations for the provided methods.
+ */
+class Method {
+private:
```
drfloob: See https://google.github.io/styleguide/cppguide.html#Declaration_Order for declaration order within a class. Private should be below public.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+  const MethodDescriptor *method_;
+
+public:
+  Method() = delete;
+
+  Method(const MethodDescriptor *method) : method_(method) {}
```
drfloob: Make single-argument constructors `explicit` unless you intend to use implicit casts when constructing Method objects.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+
+  /// The name of the method in Rust style.
+  std::string name() const {
```
drfloob: Function names should be PascalCase. See https://google.github.io/styleguide/cppguide.html#Function_Names

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+                                                 {">", "\\>"},
+                                             });
+
+  return sanitized;
+}
+
+static std::string ProtoCommentToRustDoc(absl::string_view proto_comment) {
+  std::string rust_doc;
+  std::vector<std::string> lines = absl::StrSplit(proto_comment, '\n');
+  for (const std::string &line : lines) {
```
drfloob: `line` can be a string_view.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+  return rust_doc;
+}
+
+static void GenerateDeprecated(Printer &ctx) { ctx.Emit("#[deprecated]\n"); }
+
+namespace client {
+
+static void GenerateMethods(Printer &printer, const Service &service,
+                            const GrpcOpts &opts) {
+  static std::string unary_format = R"rs(
```
drfloob: I expect you want these to be const as well.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+              }
+
+              $methods$
+          }
+      })rs");
+}
+
+} // namespace client
+
+namespace server {} // namespace server
```
drfloob: This can be deleted.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -0,0 +1,568 @@
...
+
+// Writes the generated service interface into the given
+// ZeroCopyOutputStream.
+void GenerateService(protobuf::io::Printer &printer,
+                     const ServiceDescriptor *service_desc,
+                     const GrpcOpts &opts) {
+  const Service service = Service(service_desc);
```
drfloob: The variable will be compiled out. You could either pass the service_desc itself to the `generate_client` method, or call `generate_client(Service(service_desc), ...)`, and have the same effect.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
```diff
@@ -0,0 +1,100 @@
...
+    // Return early to avoid creating an empty output file.
+    if (file->service_count() == 0) {
+      return true;
+    }
+    std::vector<std::pair<std::string, std::string>> options;
+    protobuf::compiler::ParseGeneratorParameter(parameter, &options);
+
+    rust_grpc_generator::GrpcOpts grpc_opts;
+    absl::StatusOr<absl::flat_hash_map<std::string, std::string>>
+        import_path_to_crate_name;
```
drfloob: This does not appear to be used anywhere.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
```diff
@@ -0,0 +1,100 @@
...
+    for (auto opt : options) {
+      if (opt.first == "message_module_path") {
+        grpc_opts.message_module_path = opt.second;
+      } else if (opt.first == "crate_mapping") {
+        absl::StatusOr status =
```
drfloob: Recommended usage of StatusOr here would be `absl::StatusOr<absl::flat_hash_map<std::string, std::string>> crate_name_map`. See https://abseil.io/tips/181

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
```diff
@@ -0,0 +1,100 @@
...
+            rust_grpc_generator::GetImportPathToCrateNameMap(opt.second);
+        if (!status.ok()) {
+          *error = std::string(status.status().message());
+          return false;
+        }
+        import_path_to_crate_name = status.value();
```
drfloob: `import_path_to_crate_name = *crate_name_map`

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed to single line comments.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Ordered the imports according to the style guide.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Renamed to `PROTOC_GEN_RUST_GRPC_GRPC_RUST_GENERATOR_H_`.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: I copied it from [Java's plugin](https://github.com/grpc/grpc-java/blob/01bd63d88fbe07f06a7c28854d6db090a554ee94/compiler/src/java_plugin/cpp/java_generator.h#L52-L54). I thought it was a way to save 4 characters by typing `impl::protobuf` instead of `google::protobuf`. Removed it now.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Made the members private. Added getters and setters.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Changed to `//` throughout.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Changed to PascalCase.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: I don't think we need to generate docs for this code, changed to `//`.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Changed.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Removed.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Shortened as suggested.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Yes, changed to consts.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Removed the namespace alias from the header file.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Replaced.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: I had copied the class from the protobuf rust code which in turn copied a partial implementation of the struct from file/base in g3. Removed the struct now.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: I see, changed as suggested.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Yes, I avoided this because I needed a `char*` to call `fopen` later. Changed now.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Removed block comment throughout.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed throughout.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed throughout.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Changed throughout.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: Moved to an anonymous namespace and remove the static keyword.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
arjan-bal: That's a bug. It should have been set in `grpc_opts`. Fixed it now.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
arjan-bal: Changed.

protoc-gen-rust-grpc/src/grpc_rust_plugin.cc
arjan-bal: Changed.

protoc-gen-rust-grpc/src/grpc_rust_generator.h
arjan-bal: Acknowledged. I'll raise a follow-up PR with scripts to run clang-format and buildifier, both locally and on CI. I don't want to make this PR bigger.

LucioFranco:
@arjan-bal we need to get CI passing before we can merge. I think we can just bump the msrv to 1.79 but lets do that in a separate PR. 

As for the semver CI job I think we can just add a publish = false for now to that crate and it should resolve the issue.

.github/workflows/CI.yml
```diff
@@ -25,14 +25,70 @@ jobs:
...
+        uses: bazel-contrib/setup-bazel@0.15.0
+        with:
+          # Avoid downloading Bazel every time.
+          bazelisk-cache: true
+          # Store build cache per workflow.
+          disk-cache: ${{ github.workflow }}
+          # Share repository cache between workflows.
+          repository-cache: true
+          module-root: ./protoc-gen-rust-grpc
+      - name: Build protoc plugin
```
LucioFranco, line 59: Does this step take a long time? Generally, speaking I would rather that we build the plugin in each test job to ensure we are building the latest code and there is no caching issues. Could you explain the reasoning for breaking it up like this?

grpc/Cargo.toml
```diff
@@ -28,4 +28,5 @@ dns = ["dep:hickory-resolver"]
 allowed_external_types = [
     "tonic::*",
     "futures_core::stream::Stream",
-]
\ No newline at end of file
+    "protobuf::codegen_traits::Message",
```
LucioFranco: We need to add a todo here that we want to remove this. This should all be abstracted by the new protobuf crate. (aka its exposed as a Codec rather than needing the grpc crate to know about a message trait).

grpc/src/macros.rs
```diff
@@ -0,0 +1,90 @@
...
+///
+/// If you have modified the output directory or message path, you should
+/// include the generated code manually instead of using this macro.
+///
+/// The following example assumes the message code is imported using `self`:
+///
+/// ```rust,ignore
+/// mod protos {
+///     // Include message code.
+///     include!("/protobuf/directory/protos/generated.rs");
```
LucioFranco: The leading `/` seems really awkward here, I would say that we don't allow it since it makes me think this is pointing to an absolute path rather than a relative one.

interop/Cargo.toml
```diff
@@ -25,6 +29,10 @@ tokio-stream = "0.1"
 tonic = {path = "../tonic", features = ["tls-ring"]}
 tower = "0.5"
 tracing-subscriber = {version = "0.3"}
+grpc = {path = "../grpc"}
+protobuf = { version = "4.31.1-release"}
```
LucioFranco:
```suggestion
protobuf = { version = "4.31.1-release" }
```

We should probably set a toml style and enforce it via some toml formatter if that exists. No need to do anything now just thinking out loud.

interop/src/bin/client_new_codegen.rs
```diff
@@ -0,0 +1,169 @@
+/*
```
LucioFranco: I think we should rename these to `client-prost` and `client-protobuf` etc, thoughts? Or actually another thought can we just merge the client files together and make it an option which protobuf library it uses? That way its really easy to make some sort of test matrix?

interop/src/lib.rs
```diff
@@ -9,6 +10,10 @@ pub mod pb {
     include!(concat!(env!("OUT_DIR"), "/grpc.testing.rs"));
 }
 
+pub mod grpc_pb {
+    grpc::include_proto!("", "test");
```
LucioFranco: This `""` feels slightly awkward as well

tonic-protobuf-build/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "tonic-protobuf-build"
+version = "0.9.0-alpha.1"
```
LucioFranco: Whats the versioning scheme here? Should we just follow tonic and start this at 0.14?

tonic-protobuf-build/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+Compiles proto files via protobuf rust and generates service stubs and proto
+definitions for use with tonic.
+
+## Features
+
+Required dependencies
+
+```toml
+[dependencies]
```
LucioFranco, line 11: Users will need to also import `tonic-protobuf`? Which I think also would allow us to avoid to have them pullin protobuf since they would just need to align the `tonic-protobuf-build` and `tonic-protobuf` crate?

tonic-protobuf-build/README.md
```diff
@@ -0,0 +1,94 @@
...
+Or configure the generated code deeper via
+
+```rust,no_run
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+   tonic_protobuf_build::configure()
+        .generate_message_code(false)
+        .inputs(["proto/helloworld/helloworld.proto"])
+        .include("external")
+        .message_module_path("super::proto")
+        .dependencies(vec![tonic_protobuf_build::Dependency {
```
LucioFranco: Oh interesting, this is a big difference from prost (probably for the good) in that prost will search for dependencies while this asks you to explicitly include them.

tonic-protobuf-build/README.md
```diff
@@ -0,0 +1,94 @@
...
+pub mod grpc_pb {
+    grpc::include_proto!("proto/helloworld", "helloworld");
+}
+```
+
+Or if you want to save the generated code in your own code base,
+you can uncomment the line `.output_dir(...)` above, and in your lib file
+config a mod like this:
+```rust,ignore
+pub mod generated {
```
LucioFranco, line 87: I believe we still use `include!` for things generated like this, why do we need users to use `path`?

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+#[derive(Debug, Clone)]
+pub struct Dependency {
+    /// Name of the external crate.
+    pub crate_name: String,
+    /// List of paths .proto files whose codegen is present in the crate. This
+    /// is used to re-run the build command if required.
+    pub proto_import_paths: Vec<PathBuf>,
+    /// List of .proto file names whose codegen is present in the crate.
+    pub proto_files: Vec<String>,
+}
```
LucioFranco: We should avoid public fields, and use a builder instead and then provide getters if needed. Even if the protobuf crate exposes stuff like this, I don't want to follow that anti pattern in public apis for tonic.

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// Sets path of the module containing the generated message code. This is
+    /// "self" by default, i.e. the service code expects the message structs to
+    /// be present in the same module. Set this if the message and service
+    /// codegen needs to live in separate modules.
+    pub fn message_module_path(&mut self, message_path: &str) -> &mut Self {
+        self.message_module_path = Some(message_path.to_string());
+        self
+    }
+
+    pub fn generate_and_compile(&self) -> Result<(), String> {
```
LucioFranco: Curious, why you call this generate and compile and not just compile? To the user this is just compiling unless they can individually invoke each step and this just combines that?

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+            protobuf_codegen::CodeGen::new()
+                .inputs(self.inputs.clone())
+                .output_dir(self.output_dir.clone())
+                .includes(self.includes.iter())
+                .dependency(self.dependencies.iter().map(|d| d.into()).collect())
+                .generate_and_compile()
+                .unwrap();
+        }
+        let crate_mapping_path = if self.generate_message_code {
+            self.output_dir.join("crate_mapping.txt")
```
LucioFranco, line 209: is it possible to have multiple codegens override the same file? Do we need to add a warning incase say you do that by accident and its surprising?

tonic-protobuf-build/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+        } else {
+            self.generate_crate_mapping_file()
+        };
+
+        // Generate the service code.
+        let mut cmd = std::process::Command::new("protoc");
```
LucioFranco, line 215: We probably want to allow users to set the path to protoc via an env var as well. We do this in the prost code so we should follow how it does that.

tonic-protobuf/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "tonic-protobuf"
+version = "0.9.0-alpha.1"
+edition = "2021"
+authors = ["gRPC Authors"]
+license = "MIT"
+
+[dependencies]
+tonic = { version = "0.14.0", path = "../tonic", default-features = false, features = ["codegen"] }
+bytes = "1.10.1"
```
LucioFranco:
```suggestion
bytes = "1"
```

If we don't need a specific feature from a minor/patch version its better to be as general as possible.

tonic-protobuf/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+        Self { _pd: PhantomData }
+    }
+}
+
+impl<T: Message> Encoder for ProtoEncoder<T> {
+    type Item = T;
+    type Error = Status;
+
+    fn encode(&mut self, item: Self::Item, buf: &mut EncodeBuf<'_>) -> Result<(), Self::Error> {
+        let serialized = item.serialize().map_err(from_decode_error)?;
```
LucioFranco, line 90: We should add a todo here mentioning that we want to figure out how to remove this extra copy that happens.

tonic-protobuf/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl<U: Message + Default> Decoder for ProtoDecoder<U> {
+    type Item = U;
+    type Error = Status;
+
+    fn decode(&mut self, buf: &mut DecodeBuf<'_>) -> Result<Option<Self::Item>, Self::Error> {
+        let slice = buf.chunk();
+        let item = U::parse(slice).map_err(from_decode_error)?;
+        buf.advance(slice.len());
```
LucioFranco, line 116: I think its possible that the incoming buf slice is larger than the actual message, we should see if we can pull the len from the decoded amount of bytes.

LucioFranco (changes requested):
Overall LGTM, lets get these last few things addressed and CI passing then lets merge.

.github/workflows/CI.yml
arjan-bal: Added a comment in the yaml. Building the protoc plugin from scratch takes 6–14 minutes, depending on the OS. This delays the execution of workflows that use the plugin in build.rs files. 

Example workflow execution: https://github.com/hyperium/tonic/actions/runs/16217230891/job/45789317103

grpc/Cargo.toml
arjan-bal: Removed it. Initially the codec was in the `grpc` crate. I missed reverting this change after the codec was moved out.

grpc/src/macros.rs
arjan-bal: Replaced the leading `/` in the doc with `relative/`. I copied and modified [the rust doc of the tonic macro](https://github.com/hyperium/tonic/blob/cde96ac0f695b0ebdd11166414e9b87fa8ee9506/tonic/src/macros.rs#L17-L20).

interop/src/lib.rs
arjan-bal: I was trying to include both the message and service generated code using a single declarative macro. The message code is located in `dir/generated.rs`, and the service code in `dir/message_grpc.pb.rs`. I couldn't find a way to determine the parent directory from the proto message path (e.g., `dir/message.proto`), so I decided to accept the directory and filename as separate parameters. I'm open to suggestions for improving this approach.

tonic-protobuf-build/Cargo.toml
arjan-bal: I copied the [version from the grpc crate](https://github.com/hyperium/tonic/blob/cde96ac0f695b0ebdd11166414e9b87fa8ee9506/grpc/Cargo.toml#L3). Changed it to 0.14.0 now.

tonic-protobuf-build/README.md
arjan-bal: > Users will need to also import `tonic-protobuf`?

You're right, added that to the `dependencies`.

> Which I think also would allow us to avoid to have them pullin `protobuf` since they would just need to align the `tonic-protobuf-build` and `tonic-protobuf` crate?

If I understand correctly, users will still need to add a `protobuf` dependency if they want to create new request or response messages. However, they can call methods on the protobuf messages to the client or passed to the server handler without the `protobuf` dependency in their `Cargo.toml` file.

tonic-protobuf-build/README.md
arjan-bal: The dependencies are used to refer to code that is generated outside this protoc invocation. In bazel, this would happen by running another build target for the proto dependency. In cargo, this can happen by running another `tonic_protobuf_build.configure()`. I think this can also happen when an external crate has the generated code. The codegen will use the `crate_name` specified here to refer to the request/response message types from these protos.

tonic-protobuf-build/README.md
arjan-bal: I was describing a similar use case as the [tonic-build crate](https://github.com/hyperium/tonic/blob/cde96ac0f695b0ebdd11166414e9b87fa8ee9506/tonic-build/README.md?plain=1#L121-L132) without referring to Google APIs. I think things would work with `include!` as well.

tonic-protobuf-build/src/lib.rs
arjan-bal: Changed to use the builder pattern with private fields.

tonic-protobuf-build/src/lib.rs
arjan-bal: I copied the name of the [`protobuf-codegen` function](https://docs.rs/protobuf-codegen/4.31.1-release/protobuf_codegen/struct.CodeGen.html#method.generate_and_compile) that generates the code. However, their `CodeGen` type has a function named [`compile_only`](https://docs.rs/protobuf-codegen/4.31.1-release/protobuf_codegen/struct.CodeGen.html#method.compile_only) that compiles C++ code, assuming the rust code already exists.

Renamed the function to `compile` and updated the docs.

tonic-protobuf-build/src/lib.rs
arjan-bal: > is it possible to have multiple codegens override the same file?

Yes it's possible. A bigger problem when having multiple `protoc` invocations using the same output directory is the `generated.rs` file getting overwritten. The `generated.rs` file exports the generated message symbols for all the proto files that were part of the input protos list.

> Do we need to add a warning incase say you do that by accident and its surprising?

@acozzette wanted to get your views on this. Have you considered this issue?

tonic-protobuf-build/src/lib.rs
arjan-bal: Here are the relevant prost docs: https://docs.rs/prost-build/latest/prost_build/index.html#sourcing-protoc

The message codegen also [doesn't support](https://github.com/protocolbuffers/protobuf/blob/045486256a5e8f194710ab9126b89b86a2545041/rust/release_crates/protobuf_codegen/src/lib.rs#L156) setting the path to protoc using an env var. I think `protobuf-codegen` should support this before gRPC. Do you want me to file a feature request for protobuf?

tonic-protobuf/src/lib.rs
arjan-bal: Added a comment. Do you want me to file an issue for the protobuf-rust?

tonic-protobuf/src/lib.rs
arjan-bal: From my understanding, each field in an encoded proto message either has a length prefix or a fixed length. The parser should continue reading fields until it reaches the end of the buffer. Fields not present in the proto descriptor should be ignored to ensure forward compatibility—for example, when a new field is added to the proto, but the receiver is using an older version. A parsing error should be returned if the bytes cannot be parsed. Since there is no length prefix for the entire message, the parser must consume all the bytes it is given.

> I think its possible that the incoming buf slice is larger than the actual message

Based on the above, I believe it may be incorrect to pass a slice to a parser but not consume the entire buffer. Can you clarify when this situation arises?

interop/src/bin/client_new_codegen.rs
arjan-bal: I'm still working on this.

arjan-bal: I made the changes to have a single client binary that chooses to test a prost or protobuf client based on a command line flag. I introduced a trait with all the test scenarios and added implementations for prost and protobuf.

interop/src/lib.rs
arjan-bal: I can have two different macros, one which allows specifying the directory path relative to `env!("OUT_DIR")` and one which assumes the generated code to be in `env!("OUT_DIR")` to avoid passing "". WDYT?

.github/workflows/CI.yml
LucioFranco: Okay yeah if this works its fine for now, we probably want to improve this in the future..

interop/src/lib.rs
LucioFranco: I think you can have macros take different amounts of arguments so you can have a macro that is `include_proto!("test")` and then if you do `include_proto!("dir", "test")` it will hit a different branch in the macro. I think that is enough to solve this problem?

tonic-protobuf-build/README.md
LucioFranco: I think maybe what we want to do is to re-export the protobuf crate at the root of the `tonic-protobuf` crate, that way all our codegen can reference `tonic_protobuf::protobuf` and then users ide's can auto import traits etc via that path. This doesn't stop users from importing the correct version and referencing to that. And it might help with issues like https://github.com/hyperium/tonic/issues/1858. I think I will go this way for `tonic-prost`.

tonic-protobuf-build/README.md
LucioFranco: Yeah, I want to avoid the `path` stuff as much as possible. Ide integration isn't the best there as far as I remember.

tonic-protobuf-build/src/lib.rs
LucioFranco: Oh nice, yeah I think aligning with what the prost vesrion is best just for alignment for existing users.

tonic-protobuf-build/src/lib.rs
LucioFranco: Yeah, that would be good I think it shouldn't be too contentious.

tonic-protobuf/src/lib.rs
LucioFranco: That might be good, we discussed this with @acozzette and co last week but probably good to track it somewhere as well. I would be okay if this was also just a tonic issue.

tonic-protobuf/src/lib.rs
LucioFranco: Hmm that is a good question, I will look into it, maybe we also want to actually provide some type safe wrapper that enforces this invariant. I wouldn't worry about this for now I think your assumption is probably right on what we actually do.

tonic-protobuf/src/lib.rs
arjan-bal: Opened a tonic issue with context about why protobuf doesn't provide the required API: https://github.com/hyperium/tonic/issues/2345

tonic-protobuf-build/src/lib.rs
arjan-bal: Filed https://github.com/protocolbuffers/protobuf/issues/22718

interop/src/lib.rs
arjan-bal: Oh yes, having two branches removes the need for passing an empty string. Updated this usage and the macro's rust doc.

tonic-protobuf-build/README.md
arjan-bal: I attempted to export `protobuf` in `tonic_protobuf` and remove the `protobuf` dependency from the interop crate. This resulted in the build failure:
```
DEP_UPB_VERSION should have been set, make sure that the Protobuf crate is a dependency: NotPresent
```
This panic is raised by [protobuf_codegen here](https://github.com/protocolbuffers/protobuf/blob/d772df9ce2d32e009cab65796722a27fc0b2a60d/rust/release_crates/protobuf_codegen/src/lib.rs#L134). `DEP_UPB_VERSION` is set by the [`protobuf` crate's build.rs](https://github.com/protocolbuffers/protobuf/blob/d772df9ce2d32e009cab65796722a27fc0b2a60d/rust/release_crates/protobuf/build.rs#L26). The [link docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key) state the following:
> Note that metadata is only passed to immediate dependents, not transitive dependents.

This means that:
1. `protobuf` must be a direct dependency in the user's crate.
2. The version of the `protobuf` crate must match the version of the the version of the `protobuf_codegen` crate used by `tonic_protobuf_build`.

LucioFranco: @dfawley ^ something we should chat with them about

LucioFranco (commented):
LGTM, I tested this locally, the absolute paths is what solved it for me. We can merge this but I want to merge https://github.com/hyperium/tonic/pull/2321 first and then rebase this ontop of that before we merge this.

tonic-protobuf-build/README.md
dfawley: cc @acozzette

acozzette: I think we can probably remove this check. This made more sense when the codegen crate needed to generate some C code that included upb headers, but we're no longer doing that. I'm going to try to submit a change that removes this.

acozzette: I submitted that change and I'm going to cherry-pick it into our 32.x branch so that it makes it into the upcoming 32.0 release in the next few weeks or so. You might need a temporary workaround if you don't want to wait for that release, though.

tonic-protobuf-build/README.md
arjan-bal: Changed to `include!`.

tonic-protobuf-build/README.md
arjan-bal: Thank you @acozzette!

arjan-bal: @LucioFranco there's another issue while avoiding a direct dependency on `prost`/`protobuf-rust`: The generated code is referring to symbols from these crates using absolute paths like `::prost::Message` and `::protobuf::Message`. This requires a direct dependency to be present in the `Cargo.toml` file.

Update: I found that prost has a [way to configure](https://docs.rs/prost-build/latest/prost_build/struct.Config.html#method.prost_path) the path for the prost crate. Does this mean we need `protobuf-codegen` to expose something similar?

arjan-bal: We decided to export the `protobuf` crate from the `tonic-protobuf`, keeping the `protobuf` dependency for now. We'll remove the requirement to have a direct `protobuf` dependency when support is available in `protobuf-codegen`.

arjan-bal:
Hi @dfawley, @LucioFranco can you please have another look?

LucioFranco (approved):

dfawley (approved):

PR #2321, opened 2025-07-02
feat: Extract prost into its own tonic based crates

LucioFranco:


.github/workflows/CI.yml
```diff
@@ -149,14 +149,13 @@ jobs:
...
 
-  semver:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v4
-    - uses: obi1kenobi/cargo-semver-checks-action@v2
-      with:
-        feature-group: all-features
-        exclude: grpc
+  # semver:
```
LucioFranco: Will re-enable once we get the release out.

tottoto: Can `if: false` condition be used to disable this job so as to maintain it instead of commenting out.

LecrisUT:
> Can you @cijothomas please block these AI chatbots from the project? See above

Pardon me? Are you insinuating that I am an AI chatbot? Look at my github profile and try again.

The reason I am not making a PR is because I am busy investigating the build state of the whole dependency graph for downstream packaging.

PR #2326, opened 2025-07-03
chore(test): Refactor compile test

tottoto:
## Motivation

Reduces the cost of maintaining the multiple test crates.

## Solution

Uses trybuild crate to test compiling.


tests/compile/Cargo.toml
```diff
@@ -1,11 +1,14 @@
 [package]
-authors = ["Lucio Franco <luciofranco14@gmail.com>"]
+name = "compile"
```
LucioFranco: what about calling this `test-compile`? so its a bit more clear in the build graph.

LucioFranco (approved):
LGTM, would just like the one comment addressed.

tests/compile/Cargo.toml
tottoto: Updated.

PR #2327, opened 2025-07-03
chore(grpc): Apply workspace lint config to grpc crate

tottoto:
Applies workspace lint config to grpc crate.


nathanielford:
Hey there! I think I might recommend we close this. We will probably be evaluating linting we want in the `grpc/` directory separately from the root tonic `Cargo.toml`. Let me know if you have any concerns, though!

PR #2330, opened 2025-07-06
feat: Add http proxy support

### LINKED ISSUE #2298: Support for http_proxy environment variables
## Feature Request

Support for http_proxy environment variables by default. This is something gRPC Core supports and I'm unable to replicate manually. https://grpc.github.io/grpc/core/md_doc_core_default_http_proxy_mapper.html. I've had some success with hyper-http-proxy, but unable to proxy mtls certs (separate issue).

### Crates
- https://github.com/metalbear-co/hyper-http-proxy

### Alternatives

- relying on GRPC libraries built on c core

victornguen:
## Motivation

Resolves #2298 

This PR adds support for HTTP proxying to Tonic clients

## Solution

This PR implements HTTP proxy support for Tonic clients by leveraging [`hyper-util`]'s built-in proxy functionality, specifically the `Tunnel` type for HTTP CONNECT proxying. The solution includes:

- Proxy configuration API:
  - `Endpoint::proxy_uri()` allows explicit proxy configuration.
  - `Endpoint::proxy_from_env(true)` enables automatic proxy detection via environment variables (`http_proxy`, `https_proxy`, `no_proxy`).

- Proxy resolution logic:
  - Explicit proxy configuration takes precedence over environment variables.
  - `no_proxy` is respected for bypassing the proxy for specific hosts.

- Tests:
  - Tests verify explicit proxy configuration, environment variable detection, `no_proxy` bypass, and precedence rules.
  - Tests use a mock proxy server to ensure actual proxy connections are made or bypassed as expected.
  - Test isolation is enforced to avoid environment variable leakage between tests.

This change is backward compatible.


victornguen:
@dfawley @LucioFranco please take a look

LeoCBS:
@victornguen Were there any updates on this review? I need to use Tonic over an HTTP proxy

nathanielford:
Heya! Continuing to move through PR cleanups. 

For this:
* It looks like HTTP proxying remains unsupported in Tonic.
* Tonic is not likely to get a new release any time soon, perhaps ever.
* The new grpc-rust implementation does cover HTTP_PROXY functionality.

Given that, @victornguen do you have a sense of how you want to proceed with this? This will likely need to go through Lucio and given that he has not yet attended to it I don't expect that it will happen in the foreseeable future.

nathanielford:
@ganluo960214 Are you able to solve your issue using something like this? (I know this is very belated!)
```
    use hyper_util::client::legacy::connect::{HttpConnector, proxy::Tunnel};
    use tonic::transport::Endpoint;

    async fn connect_via_proxy(endpoint: Endpoint, proxy_url: &str) -> Result<tonic::transport::Channel, Box<dyn std::error::Error>> {
        let proxy_uri = proxy_url.parse()?;

        let mut http = HttpConnector::new();
        http.enforce_http(false);

        let tunnel = Tunnel::new(proxy_uri, http);
        let channel = endpoint.connect_with_connector(tunnel).await?;

        Ok(channel)
    }
```

It looks like rustls received a raw HTTP text payload in your coad instead of a TLS handshake frame. Tunnel should handle the initial plaintext TCP connection to the proxy, then sends the request and verifies the response before handing the socket to Tonic, which allows Tonic to execute the handshake cleanly. I haven't _really_ tested this, though, so ymmv.

ganluo960214:
> @ganluo960214 Are you able to solve your issue using something like this? (I know this is very belated!)
> 
> ```
>     use hyper_util::client::legacy::connect::{HttpConnector, proxy::Tunnel};
>     use tonic::transport::Endpoint;
> 
>     async fn connect_via_proxy(endpoint: Endpoint, proxy_url: &str) -> Result<tonic::transport::Channel, Box<dyn std::error::Error>> {
>         let proxy_uri = proxy_url.parse()?;
> 
>         let mut http = HttpConnector::new();
>         http.enforce_http(false);
> 
>         let tunnel = Tunnel::new(proxy_uri, http);
>         let channel = endpoint.connect_with_connector(tunnel).await?;
> 
>         Ok(channel)
>     }
> ```
> 
> It looks like rustls received a raw HTTP text payload in your coad instead of a TLS handshake frame. Tunnel should handle the initial plaintext TCP connection to the proxy, then sends the request and verifies the response before handing the socket to Tonic, which allows Tonic to execute the handshake cleanly. I haven't _really_ tested this, though, so ymmv.

sorry. i don't have there code anymore. I can't test it for u. u do you a great work.

nathanielford:
Given that we have a workaround and the new grpc library will handle this as well, we're going to close this one.

PR #2331, opened 2025-07-06
feat(codec): Add lz4 support

### LINKED ISSUE #2312: Add lz4 compression
### Motivation

I have a very high bandwidth gRPC stream and zstd is not fast enough. Lz4 is the fastest compression scheme.

victornguen:
## Motivation

Add support for LZ4 compression with lz4_flex. 

Closes #2312 

## Solution

- Added the `lz4_flex` crate as an optional dependency behind a new `lz4` feature flag.
- Updated header parsing, encoding negotiation, and metadata handling to recognize and advertise LZ4.

### DIFF
--- a/tests/compression/Cargo.toml
+++ b/tests/compression/Cargo.toml
@@ -15,7 +15,7 @@ pin-project = "1.0"
 prost = "0.14"
 tokio = {version = "1.0", features = ["macros", "rt-multi-thread", "net"]}
 tokio-stream = "0.1"
-tonic = {path = "../../tonic", features = ["gzip", "deflate", "zstd"]}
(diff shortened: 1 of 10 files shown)


nathanielford:
@victornguen As a heads up, given that we are halting development on Tonic, this is unlikely to be merged. Unfortunately, grpc-rust is also not likely to support this directly in the near term as we prepare for GA. Given that the larger gRPC project does not have lz4 as a standard compression algorithm implemented in any language, we aren't likely to support it in the near term. 

We can possibly consider this to be a pluggable component, but it's not currently on our roadmap. We'll make a decision/discuss it on issue [#2312](https://github.com/grpc/grpc-rust/issues/2312).

nathanielford:
Going to close this, given that we are unlikely to support it in the near future. Feel free to reach out if you have any concerns!

Issue #2301, opened 2025-06-14
Re-export prost dependencies in tonic


Jisu-Woniu:
### Crates

- tonic
- tonic-build
- maybe tonic-types?

### Motivation

An incompatible version of prost will break the build of tonic macros generated code. For example, prost 0.14.x won't work with tonic 0.13.x.

### Proposal

Currently, tonic seems to generate paths like `::prost` in its codegen process. To maintain a working build, I suggest use `pub use prost;` in tonic and `::tonic::prost` in tonic-build generated code instead.

Since tonic and prost are often tightly coupled, it would help reduce the need for users to write `prost = "xxx"` in their Cargo manifests, as they can always use `tonic::prost` or `tonic_build::prost_build`.

### Alternatives

- Keep it as is.
- Make prost hidden pub like:

  ```rust
  #[doc(hidden)]
  // `pub` needed because we need to access it
  // from macro generated code in other crates.
  pub mod __private {
      pub use prost;
  }

  // In generated code:
  ::tonic::__private::prost;
  ```

  I think this is not better than a public re-export, since the user may still need to access prost members in their crates.
- Re-export all prost members like `pub use prost::*;`. (Possible name conflicts)
- ...

LucioFranco:
Totally agree, we've added more complexity but imo its for a good reason. 

1) Removes the need to bump a breaking release of tonic/tonic-build for every bump of prost (removes our need to tie our public version to theirs since we can use the secondary prost crates for that)
2) We are now adding first class support for the official google protobuf library but there are cases where users will still want to pick prost. Moving to this crate structure allows us to support that.
3) In reality this is just the first step we are taking for our 1.0 rework of tonic/grpc-rust. There will be more info soon when we have some more solid plans to share and some progress but you can already see some of that in the `grpc` folders in this repo.

As of now, prost is re-exported now in their new crates, there is still some work to remove the need to also import prost but we are working on that. So for now this issue has been resolved.

PR #2333, opened 2025-07-08
fix: re-export prost crate in tonic

westhide:
Fixes: #2301


LucioFranco:
Fixed in https://github.com/hyperium/tonic/pull/2321

PR #2334, opened 2025-07-08
feat(transport): Add disable user-agent

### LINKED ISSUE #970: Have full control over the User-Agent

i18nsite:
fix https://github.com/hyperium/tonic/issues/970

disable default feature can without any user-agent

in your Cargo.toml
```
tonic = { version = "0.13.1", default-features = false, features = [
  "channel",
  "codegen",
  "prost",
] }
```

if want set, use  with_interceptor like 
```

use bytes::Bytes;
use tonic::{
  service::{Interceptor, interceptor::InterceptedService},
  transport::{Channel, Endpoint, Error},
};

use crate::api_client::ApiClient;

pub async fn conn<F>(
  addr: impl Into<Bytes>,
  interceptor: F,
) -> Result<ApiClient<InterceptedService<Channel, F>>, Error>
where
  F: Interceptor,
{
  let endpoint = Endpoint::from_shared(addr.into())?;
  let channel = endpoint.connect().await?;

  let client = ApiClient::with_interceptor(channel, interceptor);

  Ok(client)
}
    let mut client = conn("12.3.2.3:1111", |mut req: tonic::Request<()>| {
        req.metadata_mut().insert("user-agent", "xxxx".try_into().unwrap());
        Ok(req)
    });


```


nathanielford:
Hello @i18nsite! In the transition to the official `grpc-rust` library we are going through old PRs to try and settle matters as we halt development on Tonic.

There is unlikely to be another Tonic release scheduled any time soon, but if you still want to see this in Tonic and don't mind merging the main branch and resolving any conflicts, we can hand it to Lucio for consideration.  One likely concern that might arise is we probably won't be supporting new feature flags, so if you do want to proceed consider the best way to handle it without a flag. 

Let me know if you have any needs or concerns as we go through this process - we'll do our best to address them!

nathanielford:
@i18nsite - let me know if you'd like to keep this open for any reason. I'll probably close it in a few weeks if I don't hear anything. Thanks!

nathanielford:
Closing this. We may still merge https://github.com/grpc/grpc-rust/pull/2492 but it conflicts with this.

PR #2338, opened 2025-07-14
Initial simple grpc channel implementation

dfawley:
This is a minimum implementation of the grpc channel with a functional, pluggable name resolver, LB policy, and transport implementation.  It includes examples that use an inmemory transport to send and receive messages.  It is still in a very incomplete state, and still needs more review, refining, and tests to be written, and API changes are still expected.  But I would like to merge it into the tonic repo in this approximate state to make ongoing development across the team easier than continuing to add code into my fork.


dfawley:
The MSRV check fails because we are using rust 1.86 features and the tonic repo uses a rust version of 1.75.  Is there a way to work around this?  I tried to set the rust version in the grpc `Cargo.toml`, but it's in the same workspace, so that is not allowed.

LucioFranco:
@dfawley I don't see an easy to scope the msrv check, so I think its fine to bump it to 1.86 for now. I don't expect any msrv issues with the next release of tonic and in around 3 months the msrv will be valid. We just have a 6 month policy for msrv following what tokio does.

grpc/Cargo.toml
```diff
@@ -6,19 +6,33 @@ authors = ["gRPC Authors"]
 license = "MIT"
 
 [dependencies]
```
LucioFranco, line 8: In a follow up we should relax the versions required here unless we need a feature in a specific patch.

grpc/src/client/channel.rs
```diff
@@ -22,19 +22,119 @@
  *
  */
 
-use std::{any::Any, str::FromStr, sync::Arc};
+use core::panic;
+use std::{
```
LucioFranco, line 26: This import style differs from what we do in the rest of tonic, we should think about aligning that.

LucioFranco (approved):
I took a very light look, overall LGTM, I expect we will have more review opportunity going forward when we change code. I also plan to work through this code locally when I get some time. But I think its better to merge this so others can work on top of it.

grpc/src/client/channel.rs
dfawley: I have no strong feelings on this - just let me know how to do it and I can try to fix everything up.

Issue #2341, opened 2025-07-19
Support for customizing trait attributes in tonic-build


ya7on:
### Crates

tonic-build

### Motivation

Currently, tonic-build generates trait without any way to attach additional attributes or proc-macros.
This makes it difficult to integrate with frameworks or tools that rely on macros, such as automocking struct by trait

### Proposal

Add a configuration option in tonic-build to inject custom attributes into generated service traits:

**In build options**
```rust
...
    .trait_atttribute("myservice.MyService", "#[mockall::automock]")
...
```

**In generated code (Before)**
```rust
/// Generated trait containing gRPC methods that should be implemented for use with ProjectsPrivateServer.
#[async_trait]
pub trait MyServce: std::marker::Send + std::marker::Sync + 'static {
  ...
}
```

**In generated code (After)**
```rust
/// Generated trait containing gRPC methods that should be implemented for use with ProjectsPrivateServer.
#[async_trait]
#[mockall::automock]
pub trait MyServce: std::marker::Send + std::marker::Sync + 'static {
  ...
}
```

### Alternatives

I ran into this limitation when trying to use `#[mockall::automock]` for a generated service trait to simplify testing.
In my case:

* I use the trait to describe a client interface for another service.
* Functions in my code accept a generic parameter implementing this trait, so I can pass either:

  * the real gRPC client generated alongside the trait, or
  * a mocked implementation for tests.

Without this feature, mocking is cumbersome because I must **manually implement the entire trait for mocks** instead of simply deriving it via `mockall`.

While my primary need is mocking, I believe there are many other potential use cases for injecting proc-macro attributes, such as logging, tracing, or applying custom derives.

---

If this feature aligns with the vision of the project, I am willing to implement the functionality and open a Pull Request

PR #2348, opened 2025-07-25
feat(build): allow to specify trait attributes

ya7on:
Related to #2341 

## Motivation

Currently, tonic-build generates service traits without any ability to attach additional attributes or proc-macros.

This creates friction when integrating generated traits with tools and frameworks that rely on such attributes.
For example, when using mockall for testing, one cannot simply write:

```rust
#[mockall::automock]
pub trait MyService { ... }
```

because the trait is auto-generated and cannot be easily modified.
As a result, mocking requires duplicating the entire trait or writing manual boilerplate, which is error-prone and difficult to maintain.

## Solution

This PR adds a new `trait_attribute` configuration method to `tonic-build`, allowing custom attributes (including proc-macros) to be injected into generated service traits.

Example usage:

```rust
tonic_build::configure()
    .trait_attribute("myservice.MyService", "#[mockall::automock]")
    .compile_protos(&["proto/myservice.proto"], &["proto"])?;
```

This will place `#[mockall::automock]` directly above the generated `MyService` trait, avoiding manual trait duplication for testing and enabling other use cases (e.g., tracing or custom derives).

I could not find any existing tests in `tonic-build` that cover this code path, so I have not added automated tests yet.
If the maintainers can point me to where such tests should live, I can add them.

### DIFF
--- a/tonic-build/src/lib.rs
+++ b/tonic-build/src/lib.rs
@@ -88,6 +88,8 @@ pub struct Attributes {
     module: Vec<(String, String)>,
     /// `struct` attributes.
     structure: Vec<(String, String)>,
+    /// `trait` attributes.
+    trait_attributes: Vec<(String, String)>,
 }
 
 impl Attributes {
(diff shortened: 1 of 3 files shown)


LucioFranco (approved):
Thanks!

PR #2359, opened 2025-07-30
feat(grpc): server codegen using protobuf rust

arjan-bal:
This change adds server code generation using Protobuf Rust. To test the generated code, a Go client and a Tonic client using Prost are used to run interop tests against the Tonic server implemented with Protobuf Rust.

**Note:** The generated code uses [`generate_default_stubs = true`](https://github.com/hyperium/tonic/blob/a738cabe6b01675cc0bd0c59327beb2080efc5e6/tonic-build/src/code_gen.rs#L73-L77), consistent with other gRPC implementations. This avoids breaking builds when a new method is added to the `.proto` file. In practice, this means that the server trait does **not** include associated types for response streams.


interop/test.sh
```diff
@@ -51,9 +51,16 @@ TLS_KEY="interop/data/server1.key"
 SERVER_PID=$!
 echo ":; started grpc-go test server."
 
+cleanup() {
+  # Temporarily disable 'exit on error' just for this command.
```
dfawley: Can you just do `kill ... || true` instead of toggling this?

Also why is it important?  Would we end up in an infinite loop because of the `trap`? Because this is the last thing the script will do anyway, isn't it?

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -514,10 +518,431 @@ static void GenerateClient(const Service &service, Printer &printer,
...
+    const std::string response_type = method.ResponseName(opts, 1);
+    auto vars = printer.WithVars({
+        {"name", method.Name()},
+        {"request", request_type},
+        {"response", response_type},
+        {"method_doc", ProtoCommentToRustDoc(method.Comment())},
+    });
+    if (!method.IsClientStreaming() && !method.IsServerStreaming()) {
+      printer.Emit(unary_format);
+    } else if (!method.IsClientStreaming() && method.IsServerStreaming()) {
```
dfawley, line 568: Optional: technically you could omit the `IsServerStreaming` since you already know that must be true because of the `if` above.  Same on the line below re: `IsClientStremaing`.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -514,10 +518,431 @@ static void GenerateClient(const Service &service, Printer &printer,
...
+    let max_decoding_message_size = self.max_decoding_message_size;
+    let max_encoding_message_size = self.max_encoding_message_size;
+    let inner = self.inner.clone();
+    let fut = async move {
+        let method = $service_ident$(inner);
+        let codec = $codec_name$::default();
+
+        let mut grpc = tonic::server::Grpc::new(codec)
+            .apply_compression_config(accept_compression_encodings, send_compression_encodings)
+            .apply_max_message_size_config(max_decoding_message_size, max_encoding_message_size);
```
dfawley, line 746: Is there any simple way to refactor and avoid all the duplication?  IIUC this should eventually be going away, as this configuration should be stored within the channel in the future, so I don't think it's too important if it's not simple because of the future/etc.

interop/test.sh
arjan-bal: > Can you just do `kill ... || true` instead of toggling this?

Yes, that's is a better way. I saw the `-e` causing the script to fail and thought of the most straightforward way to fix it. Changed to use `|| true` now.

> Also why is it important? 

In the success case, the server would already have shutdown so calling `kill "${SERVER_PID}"` will fail, causing the the return code of the program to be non-0.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: In my opinion, even though it's more verbose, it makes reasoning about correctness easier.

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
arjan-bal: I was following Tonic's existing codegen while breaking up the code template into pieces. The existing code also [has this duplication](https://github.com/hyperium/tonic/blob/c6e42ebe425f1afe70da41c0b387e2b0807f4672/tonic-build/src/server.rs#L459-L494). I suspect this may be for easier readability.

interop/test.sh
```diff
@@ -63,36 +68,46 @@ sleep 1
...
+    else
+      TLS_ARRAY=()
+    fi
+
+    for CASE in "${TEST_CASES[@]}"; do
+      flags=( "-test_case=${CASE}" )
+      # Avoid unbound variable errors on MacOS with bash version < 4.4.
+      # See: https://stackoverflow.com/a/61551944
+      flags+=( ${TLS_ARRAY[@]+"${TLS_ARRAY[@]}"} )
+      interop/bin/client_"${OS}"_amd64"${EXT}" "${flags[@]}"
```
LucioFranco, line 105: Can we also make it so we can mix and match codecs here with the client?

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -514,10 +518,431 @@ static void GenerateClient(const Service &service, Printer &printer,
...
+      service.Name() + "Server.");
+  printer.Emit(
+      {
+          {"trait_doc", trait_doc},
+          {"methods", [&] { GenerateTraitMethods(printer, service, opts); }},
+      },
+      R"rs(
+    $trait_doc$
+    #[async_trait]
+    pub trait $server_trait$ : std::marker::Send + std::marker::Sync + 'static {
```
LucioFranco, line 595:
```suggestion
    pub trait $server_trait$: std::marker::Send + std::marker::Sync + 'static {
```

Is this space needed? When we use quote we need this for its interpolation but I don't think the c++ stuff needs that?

protoc-gen-rust-grpc/src/grpc_rust_generator.cc
```diff
@@ -514,10 +518,431 @@ static void GenerateClient(const Service &service, Printer &printer,
...
+        $methods$
+    }
+    )rs");
+}
+
+static void GenerateMethods(Printer &printer, const Service &service,
+                            const GrpcOpts &opts) {
+  static const std::string unary_format = R"rs(
+    #[allow(non_camel_case_types)]
```
LucioFranco, line 604: I don't think we have this with the other codegen, why do we need this here?

PR #2375, opened 2025-08-05
add support for parsing X509 certs from DER

### LINKED ISSUE #2343: Add support to tonic::transport::Identity from_der
### Motivation

PEM and DER are both common formats for the certificates and it stands to reason that both should be supported here in this implementation.

### Proposal

Add a trait similar to the from_pem() impl, but from_der()

### LINKED ISSUE #2344: Add support to tonic::transport::Certificate from_der
### Motivation

PEM and DER are both common formats for the certificates and it stands to reason that both should be supported here in this implementation.

### Proposal

Add a trait similar to the from_pem() impl, but from_der()

cristiano-prato:
tls: allow creation of Certificate and Identity from DER

Adds `Certificate::from_der` and `Identity::from_der` constructors to allow creation of Certificate and Identity from DER-encoded data.

Fixes: #2343
Fixes: #2344

## Motivation

PEM and DER are both very common encoding formats for X509 certificates.

## Solution

Convert DER certificates to PEM to prevent changes to the rest of the crate. The conversion is done following RFC 7468, which basically means converting the DER in base64 and adding the proper header and footer.

### DIFF
--- a/tonic/src/transport/tls.rs
+++ b/tonic/src/transport/tls.rs
@@ -28,7 +28,9 @@ impl Certificate {
     pub fn from_der(der: impl AsRef<[u8]>) -> Self {
         let der = der.as_ref();
         let pem = der2pem(der);
-        Self { pem: pem.into_bytes() }
+        Self {
+            pem: pem.into_bytes(),
+        }
     }
 
     /// Get a immutable reference to underlying certificate
@@ -70,7 +72,7 @@ impl Identity {
     }
 
     /// Parse a DER encoded certificate and private key.
-    /// 
+    ///
     /// The provided cert must contain exactly one DER encoded certificate.
     pub fn from_der(cert: impl AsRef<[u8]>, key: impl AsRef<[u8]>) -> Self {
         let cert = Certificate::from_der(cert);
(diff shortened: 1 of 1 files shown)


nathanielford:
Hey there @cristiano-prato!

As you may know we are transitioning this project to the official `grpc-rust` implementation, and are halting development on tonic. For that reason it's unlikely to receive another release in the near future, and PRs that will merge Tonic code are likely to have a much longer timeline for review and merging.

That said, if you're still interested in seeing this change get in, I can work with you to see if that is possible. Some concerns I think are raised here:

* Double Conversion: this looks like it's going to cause redundant heap allocations and encoding/decoding. Maybe it's small enough it's not so bad, but it's a consideration we have.
* Private Key structure: IIUC DER keys come in three different structures (PKCS#1, PKCS#8, SEC1). This sort of wrapping will break PKCS#1 and SEC1. We at the least would want some tests regarding these situations.
* Deferred validation: One awkwardness here is that invalid DER bytes passed to this will silently succeed at tls.rs construction and only fail later during transport initialization. 

I think if this is still a feature you want, either in Tonic or gRPC, we should discuss the design (probably on the linked Issues) to make sure we address those. Of course, this has already lingered for some time, so if it's a non-issue now, feel free to close!

cristiano-prato:
Hello,
thanks for your feedback. I do not need this feature anymore.

PR #2378, opened 2025-08-06
chore: relax msrv for most crates

### LINKED ISSUE #2361: Resolve accidental rust-version bump in tonic crates
Build failure for https://github.com/googleapis/google-cloud-rust/actions/runs/16654782596/job/47136850215?pr=2801#step:8:263 shows that users that pin anything <1.86 will fail to build due to cargo strictly enforcing the cargo version. We should bump all tonic crates to 0.14.1 and relax the rust-version this should be a backwards compatible change. CI MSRV checks will need to be updated.

acrrd:
## Motivation
Closes #2361 

## Solution
Relax version of the workspace to 1.75 ~~1.82~~, specify msrv for `grpc` crate as 1.86.
Exclude `grpc` and `tonic-protobuf*` from msrv check. They require a higher msrv and they do not specify `rust-version = { workspace = true }`.

A small code change is used because we cannot rely on automatic lifetime extension of temporaries.

~~Without the `grpc` crate we would need to bump only to 1.79 because of `protobuf-codegen v4.31.1-release`.~~
~~With `grpc` we need 1.82 because of `icu_*` crates.~~

~~We could move grpc to another workspace to avoid bumping the tonic to 1.82, but I don't think is worth it.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ may be a good resource as it shows examples of many of the gRPC features.
 
 ### Rust Version
 
-`tonic`'s MSRV is `1.75`.
+`tonic`'s MSRV is `1.82`.
 
 ### Dependencies
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ may be a good resource as it shows examples of many of the gRPC features.
 
 ### Rust Version
 
-`tonic`'s MSRV is `1.82`.
+`tonic`'s MSRV is `1.75`.
 
 ### Dependencies
 
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -183,9 +183,9 @@ jobs:
         rust-version: ${{ steps.msrv.outputs.version }}
     - uses: taiki-e/install-action@cargo-no-dev-deps
     - uses: Swatinem/rust-cache@v2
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ may be a good resource as it shows examples of many of the gRPC features.
 
 ### Rust Version
 
-`tonic`'s MSRV is `1.75`.
```
dfawley: Is there any reason not to go back to 1.75?

dfawley: Sorry I didn't see the initial comment in the PR.

Does `protobuf-codegen` need to affect the main tonic crates?

acrrd: No, it doesn't! I assumed all tonic-* crates required the same MSRV, but tonic-protobuf* are even marked as `publish = false`. I excluded them as well from the MSRV check in the CI.

A small code change was needed to work with 1.75 because we cannot rely on the automatic lifetime extension of temporaries.

LucioFranco:
Thanks for getting this solved!

dbolduc:
Thanks!

README.md
dfawley: That's great, thank you for the fix!

PR #2391, opened 2025-08-19
Update protobuf to 4.32.0

arjan-bal:


arjan-bal:
We'll need to wait for the next protobuf release to revert https://github.com/hyperium/tonic/pull/2382. [The fix](https://github.com/protocolbuffers/protobuf/pull/23059) isn't part of v32.0.  

https://github.com/protocolbuffers/protobuf/blob/32.x/rust/map.rs#L47

dfawley (approved):

Issue #978, opened 2022-04-22
pub constructor for tonic::transport::Error


fridge-dev:
### Motivation

We have code that calls `Endpoint::connect()` and handles the returned error. We want to write unit tests that exercise our error handling logic. Currently, we cannot create a `tonic::transport::Error` because all of the constructor methods are `pub(crate)`, and so we can't test our error handling logic.

We have a workaround, where we create a thin wrapper around `Endpoint::connect()` that returns an Err variant like this:

```
enum TonicErrorWrapper {
    Real(tonic::transport::Error),
    Fake,
}
```

...and we only test with `TonicErrorWrapper::Fake`. We would like if `tonic::transport::Error` had a public constructor so we can test our error branches without introducing wrapper types.

### Proposal

Make `Error::new()` or `Error::from_source()` public, instead of `pub(crate)`. Or introduce some new constructors intended for public use.

I would even be fine with `Error::dummy_instance_for_tests()`. I just need an actual instance for my tests.

I am not sure what would be best for public API surface of this crate. I'll leave that up to hyper team.

gabotechs:
I'm also facing this, I'm trying to do some lazy initialization on the first call from a client to a server, and for that I want to use a `MapFuture` tower layer, something like this:

```rust
pub fn build_channel() -> BoxService<> {
    let (channel, changes_tx) = Channel::balance_channel(512);

    let once = Arc::new(OnceCell::new());

    tower::ServiceBuilder::new()
        .map_future(move |f| async move {
            once.get_or_try_init(|| async move {
                // do some lazy initialization stuff.

                Err(tonic::transport::Error::from())
            }).await?;

            f.await
        })
        .service(channel)
        .boxed()

}
```

The issue is that the `Service` implementation of a `tonic::Channel` requires a `tonic::transport::Error`:

https://github.com/hyperium/tonic/blob/dee02feaa7caf1fa8559d0fcbe9a5c8fb5e63437/tonic/src/transport/channel/mod.rs#L208-L211

which means that If I ever want to throw an error from the `.map_future()` layer, it needs to be a `tonic::transport::Error`, but those cannot be built by users.

The workaround I came up with was to build `tonic::transport::Error` using a dirty trick that bypasses the privacy restrictions:

```rust
async fn make_transport_err(msg: String) -> tonic::transport::Error {
    let connector = tower::service_fn(move |_| {
        let msg = msg.clone();
[...]

PR #2394, opened 2025-08-21
feat(tonic): add From<BoxError> implementation for tonic::transport::Error

gabotechs:
Addresses https://github.com/hyperium/tonic/issues/978

This PR keeps the `tonic::transport::Error` constructor methods private while allowing external users to map `BoxError`s into `tonic::transpoert::Error`, which makes it possible to do something like:

```rust
fn test() -> Result<(), crate::transport::Error> {
    let err = Err(crate::BoxError::from("test"));
    err?;
    Ok(())
}
```

More details about why users would want to build `tonic::transport::Error` in the issue.

Also, completely open to alternatives.

### DIFF
--- a/tonic/src/transport/error.rs
+++ b/tonic/src/transport/error.rs
@@ -90,3 +90,9 @@ impl StdError for Error {
             .map(|source| &**source as &(dyn StdError + 'static))
     }
 }
+
+impl From<crate::BoxError> for Error {
+    fn from(value: crate::BoxError) -> Self {
+        Self::from_source(value)
+    }
+}


LucioFranco:
What is the use case here, its not construct-able publicly for a reason.

gabotechs:
> What is the use case here, its not construct-able publicly for a reason.

this comment explains it:
https://github.com/hyperium/tonic/issues/978#issuecomment-3209208871

nathanielford:
I updated [Issue 978](https://github.com/grpc/grpc-rust/issues/978#issuecomment-5109817268) regarding this. Given that Tonic development is winding down, implementing something like this is unlikely to happen, even if we were inclined to - but unless I've missed something (possible!) it doesn't seem like it's actually needed. Feel free to update the issue if you ongoing concerns!

PR #2396, opened 2025-08-21
transport: unify request modifiers and reduce allocations

veecore:
## Motivation

Currently, request modifiers (`AddOrigin`, `UserAgent`) are implemented as separate services, each carrying their own `impl Service`. This leads to duplicated boilerplate, scattered logic, and some unnecessary runtime allocations (e.g. cloning `Uri`, building headers at call boundaries).

Additionally, the derivation of the endpoint’s origin was handled speculatively in `connection.rs` rather than directly on `Endpoint`, making reasoning about the canonical origin less clear.

The goal of this change is to:

* unify modifier logic for maintainability,
* reduce redundant allocations and clones,
* and make origin handling the responsibility of `Endpoint`.

---

## Solution

* Unified `AddOrigin` and `UserAgent` into `request_modifiers.rs`.
* Introduced a generic `Modifier<M, T>` that centralizes the `Service<Request>` impl.
* Future modifiers only need to define an `into_fn()`, avoiding repeated `Service` impls.
* Added `Endpoint::get_origin()` to move origin derivation closer to the data it represents.
* Reduced heap allocations by validating and preparing `UserAgent` once at construction, instead of at every service call.
* Eliminated extra `Uri` clones by borrowing where possible.
* Preserved existing semantics (`try_insert` still respected for user-provided headers).
* Added dedicated unit tests for modifiers; all existing tests continue to pass.

✅ No public API changes.
✅ Cleaner internals, less duplication.
✅ Small but meaningful perf improvements (fewer heap allocations/clones).


veecore:
Totally fair — thanks for the feedback. I don’t want to push an abstraction that feels over-engineered at this stage.
I’ll go ahead and extract the clone/allocation reductions as a smaller, focused PR since those are worthwhile regardless. I’ll ping you for review once it’s up.
If later we get more modifiers, the abstraction here might make sense to revisit.

nathanielford:
@veecore Hey there! Seems like we should close this out? I don't know if you opened a related PR, but the discussion seems to have landed on 'do not merge'. LMK if you have any concerns with that!

PR #2399, opened 2025-08-22
feat(grpc): add gracefulswitch

cjqzhao:
@dfawley @easwars


grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+        GracefulSwitchLbConfig {
+            child_builder,
+            child_config,
+        }
+    }
+}
+
+/**
+Struct for Graceful Switch.
+*/
```
dfawley, line 48: Please write a good docstring for this.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+pub struct GracefulSwitchPolicy {
+    subchannel_to_policy: HashMap<WeakSubchannel, ChildKind>,
```
dfawley, line 50: Can we not use the `ChildManager` for this policy?  I was really hoping we could.

The goal was to make it so that no policy would need to manage its own tracking of children.

Ideally graceful switch could do something with the incoming resolver update so that the update sharder would produce 1 or 2 children with the proper configs.

If this is too much to change at the last minute for you, then we can figure out what TODOs to add.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), Box<dyn Error + Send + Sync>> {
+        if update.service_config.as_ref().is_ok_and(|sc| sc.is_some()) {
+            return Err("can't do service configs yet".into());
+        }
+        let cfg: Arc<GracefulSwitchLbConfig> =
+            match config.unwrap().convert_to::<Arc<GracefulSwitchLbConfig>>() {
+                Ok(cfg) => (*cfg).clone(),
+                Err(e) => panic!("convert_to failed: {e}"),
+            };
```
dfawley, line 70: Why the `match` on the `Result` instead of `.unwrap()`?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+    fn resolver_update(
+        &mut self,
+        update: ResolverUpdate,
+        config: Option<&LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), Box<dyn Error + Send + Sync>> {
+        if update.service_config.as_ref().is_ok_and(|sc| sc.is_some()) {
+            return Err("can't do service configs yet".into());
+        }
+        let cfg: Arc<GracefulSwitchLbConfig> =
```
dfawley, line 66: I think you should be able to elide this type?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+impl LbPolicy for GracefulSwitchPolicy {
+    fn resolver_update(
+        &mut self,
+        update: ResolverUpdate,
+        config: Option<&LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), Box<dyn Error + Send + Sync>> {
+        if update.service_config.as_ref().is_ok_and(|sc| sc.is_some()) {
+            return Err("can't do service configs yet".into());
+        }
```
dfawley, line 65: This seems wrong.  Graceful switch shouldn't need to do anything special with service configs.  It should just forward it to all children.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+            match config.unwrap().convert_to::<Arc<GracefulSwitchLbConfig>>() {
+                Ok(cfg) => (*cfg).clone(),
+                Err(e) => panic!("convert_to failed: {e}"),
+            };
+        let mut wrapped_channel_controller = WrappedController::new(channel_controller);
+        let mut target_child_kind = ChildKind::Pending;
+
+        // Determine if we can switch the new policy in. If there is no children
+        // yet or the new policy isn't the same as the latest policy, then
+        // we can swap.
```
dfawley, line 76:
```suggestion
        // Determine if we can switch to the new policy. If there is no child
        // yet or the new policy isn't the same as the previous policy, then
        // we can swap.
```

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,1223 @@
...
+        let needs_switch = {
+            let mut managing_policy = self.managing_policy.lock().unwrap();
+            managing_policy.no_policy()
+                || managing_policy.latest_policy() != cfg.child_builder.name()
+        };
+
+        if needs_switch {
+            target_child_kind = self.switch_to(config);
+        }
+        {
```
dfawley, line 86: Why add a scope here?

grpc/src/client/load_balancing/graceful_switch.rs
cjqzhao: I found it a lot easier to manage children directly when I looked into using ChildManager but I can look into redesigning. The challenge about this is that the ResolverUpdate from shard_update would not be used and there would have to be some other design. I think that changes will need to be made for the Child Manager due to the Child manager forwarding resolver updates to children even when unnecessary. I will draft an appropriate design and get approval from you and Easwar.

grpc/src/client/load_balancing/graceful_switch.rs
cjqzhao: Due to locking. I wanted to lock the managing_policy just to call resolve_update and then release it for resolve_child_controller.

PR #2401, opened 2025-08-22
grpc(feat): add macro for pregenerated code

cjqzhao:
Add include_generated_proto! macro for Codelabs purposes.

@dfawley @easwars @LucioFranco

### DIFF
--- a/grpc/src/macros.rs
+++ b/grpc/src/macros.rs
@@ -24,8 +24,8 @@
 
 /// Includes generated proto message, client, and server code.
 ///
-/// You must specify the path to the `.proto` file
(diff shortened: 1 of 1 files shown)


grpc/src/macros.rs
```diff
@@ -102,3 +102,29 @@ macro_rules! include_proto {
         ));
     };
 }
+///
```
LucioFranco, line 136: Is this missing more documentation?

cjqzhao: Yes will add!

cjqzhao: Just added documentation!

PR #2406, opened 2025-08-29
Status Formatting and tls-connect-info feature into v0.14.x

xumaple:


LucioFranco:
Okay, I solved this by just ff the v0.14.x branch from master, so we don't need this PR anymore since these commits are already there.

PR #2414, opened 2025-09-18
Fix labels in `Display for Status`

emilk:
* Follow-up to https://github.com/hyperium/tonic/pull/2403

I accidentally used confusing labels. Thanks to @kloakin for pointing it out.

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -748,10 +748,10 @@ impl From<std::io::Error> for Status {
 
 impl fmt::Display for Status {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "status: '{}'", self.code())?;
+        write!(f, "code: '{}'", self.code())?;
 
         if !self.message().is_empty() {
-            write!(f, ", self: {:?}", self.message())?;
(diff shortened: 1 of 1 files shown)


tonic/src/status.rs
```diff
@@ -748,10 +748,10 @@ impl From<std::io::Error> for Status {
 
 impl fmt::Display for Status {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "status: '{}'", self.code())?;
+        write!(f, "code: '{}'", self.code())?;
```
kloakin, line 751: `code` may be more clear
but the old format string was `"status: {:?}, message: {:?}, details: {:?}, metadata: {:?}"`
so I think either one might be ok

tonic/src/status.rs
```diff
@@ -748,10 +748,10 @@ impl From<std::io::Error> for Status {
...
 
         if !self.message().is_empty() {
-            write!(f, ", self: {:?}", self.message())?;
+            write!(f, ", message: {:?}", self.message())?;
```
kloakin, line 754: thanks!

tottoto (approved):

PR #2421, opened 2025-10-08
Allow configuring max_local_error_reset_streams

alexanderrilee:
## Motivation
This makes [max_local_error_reset_streams](https://docs.rs/hyper/latest/hyper/server/conn/http2/struct.Builder.html#method.max_local_error_reset_streams) configurable. We have an isolated environment where we create and close many (valid) short-lived gRPC streams. We very quickly hit the 1024 limit on stream resets, which [initiates a GoAway](https://doc.servo.org/src/h2/proto/streams/streams.rs.html#1583). During connection recreation, the Python gRPC client misbehaves. We would like the option to tune this limit, since this is a safe traffic pattern for us.

## Solution
I've added `http2_max_local_error_reset_streams` to the `Server` API.


tonic/src/transport/server/mod.rs
```diff
@@ -720,6 +704,7 @@ impl<L> Server<L> {
                 .keep_alive_timeout(http2_keepalive_timeout)
                 .adaptive_window(http2_adaptive_window.unwrap_or_default())
                 .max_pending_accept_reset_streams(http2_max_pending_accept_reset_streams)
+                .max_local_error_reset_streams(http2_max_local_error_reset_streams)
```
tottoto, line 738: The required `hyper-util` version seems to be needed to be updated to `0.1.11`, which is the version this API was added.

https://github.com/hyperium/hyper-util/releases/tag/v0.1.11

tottoto (changes requested):
Thanks for working on this. I left a comment.

tonic/src/transport/server/mod.rs
```diff
@@ -340,14 +316,23 @@ impl<L> Server<L> {
         }
     }
 
+    /// Configures the maximum number of local reset streams allowed before a GOAWAY will be sent.
+    ///
+    /// This will default to whatever the default in hyper is. As of v1.7.0, it is 1024.
```
tottoto: I wouldn't think that the `tonic`'s documentation should manage the default value ​​that depend on the specific dependency version.

```suggestion
    /// This will default to whatever the default in hyper is.
```

tonic/src/transport/server/mod.rs
alexanderrilee: Done! It seems the lock doesn't need to be updated, but I updated the `Cargo.toml`.

alexanderrilee:
I'm not sure why the "CI / Interop Tests (macOS-latest)" test failed, but it looks unrelated?

alexanderrilee:
@tottoto, do you know if we can bump the minor version to capture this change?

LucioFranco:
@alexanderrilee yeah, I will work on a patch release 👍

PR #2423, opened 2025-10-09
transport: handle ipv6 endpoints when tls is enabled

### LINKED ISSUE #2422: Configuring TLS doesn't work if IPv6 IP address is passed in
I have some code like the following:

```rust

use tonic::transport::{Channel, Endpoint, ClientTlsConfig, Error};

fn connect_endpoint(url: String) -> Result<Channel, Error> {
    let channel = Endpoint::try_from(url)?
        .tls_config(ClientTlsConfig::new().with_enabled_roots())?
        .connect_lazy();
    Ok(channel)
}
```

This works fine if I pass a URL like `https://example.com` or `http://127.0.0.1:1234` but it fails if I pass `http://[::1]:1234` which is a valid URL.

The error is:

```
Error: 
   0: transport error
   1: invalid dns name
```

Which comes from this line:

https://github.com/hyperium/tonic/blob/ff7b54045c943172de4c65cf949011492db185ee/tonic/src/transport/channel/service/tls.rs#L102

bouk:
Fixes #2422 see that issue for context


nathanielford:
Hey @bouk ! I'm a maintainer on grpc-rust, and am going through old PRs as we halt development on tonic in favor of the new official library. It looks like this was never merged, despite being a reproducible issue. (I had Gemini reproduce it, putting the report and code on this branch: https://github.com/nathanielford/grpc-rust-testbed/tree/repro/ipv6-tls-pr2423

I went ahead and merged all the changes from master and moved the code you wrote to the function (`build_tls_connector()`) that is now called in two places for this work (there was only one when you originally wrote it). If it looks good to you, I can try and get this reviewed and merged. Let me know!

bouk:
Yeah sounds good!

LucioFranco (approved):

PR #2427, opened 2025-10-15
fix(examples): remove now missing tower feature

bobrik:
It was removed in https://github.com/hyperium/tonic/pull/2339

Now clippy is unhappy about us requiring it in examples:

```
warning: invalid feature `tower` in required-features of target `tower-client`: `tower` is not present in [features] section
warning: invalid feature `tower` in required-features of target `tower-server`: `tower` is not present in [features] section
```

### DIFF
--- a/tonic-build/README.md
+++ b/tonic-build/README.md
@@ -110,7 +110,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 }
 ```
 
-Then you can reference the generated Rust like this this in your code:
+Then you can reference the generated Rust like this in your code:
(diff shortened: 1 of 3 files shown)


LucioFranco:
oh oops I accidentally updated this PR with updates from main.

@arjan-bal that is a good question, I honestly don't know the answer. I would say try to re-open that PR that re-introduces the feature and then we can get someone who knows the tool a bit better to help us.

LucioFranco:
We can close this now as I have a fix in https://github.com/hyperium/tonic/pull/2432

PR #2439, opened 2025-11-04
grpc: use pointers for load_balancing::WeakSubchannel hash/eq operations

dfawley:
This is an old change that I found in my client that never got pushed upstream, but is worth having.

Upgrading weak to owned before performing these operations is more expensive.  Just using the pointer is simpler and quicker and equally valid.

### DIFF
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -31,6 +31,7 @@ use std::{
     fmt::{Debug, Display},
     hash::{Hash, Hasher},
     ops::{Add, Sub},
+    ptr::addr_eq,
     sync::{
         atomic::{AtomicI64, Ordering::Relaxed},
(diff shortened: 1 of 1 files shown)


grpc/src/client/load_balancing/mod.rs
```diff
@@ -456,21 +457,13 @@ impl WeakSubchannel {
 
 impl Hash for WeakSubchannel {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        if let Some(strong) = self.upgrade() {
-            return strong.dyn_hash(&mut Box::new(state as &mut dyn Hasher));
-        }
-        panic!("WeakSubchannel is not valid");
+        (self.0.as_ptr() as *const () as usize).hash(state);
```
LucioFranco, line 460: this feels really weird but I guess its okay lol

PR #2442, opened 2025-11-06
feat(grpc): add gracefulswitch LB policy and some other LB policy changes

dfawley:
Sorry in advance for the large and unfocused change.  Some smaller things could be split out if it helps, but since many parts of the design are still in flight and can/will be changing going forward, I thought this would be okay for now.

This mainly supersedes #2399 (and keeps it as a commit), but most of the implementation is different except the tests, which are largely copied verbatim.  And it's a net 200 LoC smaller in graceful switch (100 LoC in total) which hopefully is an indication that the reuse of ChildManager is preferable.

In addition, I'd like to revisit the test implementation later, as I'm not sure the current way they are written is ideal.

--

General:

- Add `Debug` to many traits and derive/impl in structs.
- Pass LB config to LB policies via `Option<LbConfig>` instead of `Option<&LbConfig>`.  It should be rare that policies want to store a config except for the leaf policy.

Child manager:

The original assumption was that all children would be the same type/configuration, but several policies (including gracefulswitch) will not have that property.  So, several changes are made:

- Children are considered unique by both their identifier and their `LbPolicyBuilder`'s name().
- Make it so the sharder also can shard `LbConfig` and provide it via the `ChildUpdate.child_update` field in addition to the `ResolverUpdate`.
- Make `ResolverUpdateSharder` a generic instead of `Box<dyn>`.
- Add booleans so users of child manager can easily easily tell whether any child policies updated themselves, and which ones did.
- Pass `&mut self` for sharder so that it can maintain and update its state if needed.
- Change the sharder's output `ChildUpdate.child_update` field to an `Option`; if `None` then the child will not be called during the resolver update, but will remain in the child manager.
- Change `child_states` into `children` and provide the whole `Child` struct, exposing the fields it contains.
- Provide mutable access to the sharder.
- Minor test cleanups

Graceful switch:

The previous implementation in #2399 contained a lot of logic to manage child policy delegation.  It was intended that only `ChildManager` should need to have this kind of logic.

- Create a new implementation of this policy that delegates to `ChildManager`.
- Uses a Sharder that simply emits the active policy with no update alongside any new policy in the new `LbConfig`.
- `maybe_swap` is called after every call into the `ChildManager` to determine if child updates necessitate a swap.
- This logic is simple: if the active policy is not `Ready`, or if there is a new policy and it is not `Connecting`, then set the new policy as the active policy and call resolver_update on the `ChildManager`.  The sharder will see that no `LbConfig` is provided and just emit the active policy with no config, causing the `ChildManager` to drop the previously active policy.  If no swap is needed, update the picker of the active policy if it had an update.
- Minor test cleanups/fixes vs. #2399.

### DIFF
--- a/grpc/src/client/load_balancing/child_manager.rs
+++ b/grpc/src/client/load_balancing/child_manager.rs
@@ -47,7 +47,7 @@ use super::{Subchannel, SubchannelState};
 // An LbPolicy implementation that manages multiple children.
 #[derive(Debug)]
 pub(crate) struct ChildManager<T: Debug, S: ResolverUpdateSharder<T>> {
-    subchannel_child_map: HashMap<WeakSubchannel, usize>,
+    subchannel_to_child_idx: HashMap<WeakSubchannel, usize>,
     children: Vec<Child<T>>,
     update_sharder: S,
     pending_work: Arc<Mutex<HashSet<usize>>>,
@@ -101,7 +101,7 @@ where
     pub fn new(update_sharder: S, runtime: Arc<dyn Runtime>) -> Self {
         Self {
             update_sharder,
-            subchannel_child_map: Default::default(),
+            subchannel_to_child_idx: Default::default(),
             children: Default::default(),
             pending_work: Default::default(),
             runtime,
@@ -110,7 +110,7 @@ where
     }
 
     /// Returns data for all current children.
-    pub fn children(&mut self) -> impl Iterator<Item = &Child<T>> {
+    pub fn children(&self) -> impl Iterator<Item = &Child<T>> {
         self.children.iter()
     }
 
@@ -163,7 +163,7 @@ where
     ) {
         // Add all created subchannels into the subchannel_child_map.
         for csc in channel_controller.created_subchannels {
-            self.subchannel_child_map.insert(csc.into(), child_idx);
+            self.subchannel_to_child_idx.insert(csc.into(), child_idx);
         }
         // Update the tracked state if the child produced an update.
         if let Some(state) = channel_controller.picker_update {
@@ -211,7 +211,7 @@ where
         let old_children = mem::take(&mut self.children);
 
         // Replace the subchannel map with an empty map.
-        let old_subchannel_child_map = mem::take(&mut self.subchannel_child_map);
+        let old_subchannel_child_map = mem::take(&mut self.subchannel_to_child_idx);
 
         // Reverse the old subchannel map into a vector indexed by the old child ID.
         let mut old_child_subchannels: Vec<Vec<WeakSubchannel>> = Vec::new();
@@ -224,19 +224,22 @@ where
         // Build a map of the old children from their IDs for efficient lookups.
         // This leverages a Child<usize> to hold all the entries where the
         // identifier becomes the index within the old self.children vector.
-        let old_children = old_children.into_iter().enumerate().map(|(old_idx, e)| {
-            (
-                (e.builder.name(), e.identifier),
-                Child {
-                    identifier: old_idx,
-                    policy: e.policy,
-                    builder: e.builder,
-                    state: e.state,
-                    work_scheduler: e.work_scheduler,
-                },
-            )
-        });
-        let mut old_children: HashMap<(&'static str, T), _> = old_children.collect();
+        let mut old_children: HashMap<(&'static str, T), _> = old_children
+            .into_iter()
+            .enumerate()
+            .map(|(old_idx, e)| {
+                (
+                    (e.builder.name(), e.identifier),
+                    Child {
+                        identifier: old_idx,
+                        policy: e.policy,
+                        builder: e.builder,
+                        state: e.state,
+                        work_scheduler: e.work_scheduler,
+                    },
+                )
+            })
+            .collect();
 
         // Split the child updates into the IDs and builders, and the
         // ResolverUpdates/LbConfigs.
@@ -252,7 +255,7 @@ where
             if let Some(old_child) = old_children.remove(&k) {
                 let old_idx = old_child.identifier;
                 for subchannel in mem::take(&mut old_child_subchannels[old_idx]) {
-                    self.subchannel_child_map.insert(subchannel, new_idx);
+                    self.subchannel_to_child_idx.insert(subchannel, new_idx);
                 }
                 if old_pending_work.contains(&old_idx) {
                     pending_work.insert(new_idx);
@@ -324,7 +327,7 @@ where
     ) {
         // Determine which child created this subchannel.
         let child_idx = *self
-            .subchannel_child_map
+            .subchannel_to_child_idx
             .get(&WeakSubchannel::new(&subchannel))
             .unwrap();
         let policy = &mut self.children[child_idx].policy;


grpc/src/client/load_balancing/mod.rs
```diff
@@ -151,13 +153,13 @@ pub(crate) trait LbPolicyBuilder: Send + Sync {
...
 /// RPCs.
-pub trait LbPolicy: Send {
+pub trait LbPolicy: Send + Debug {
     /// Called by the channel when the name resolver produces a new set of
     /// resolved addresses or a new service config.
     fn resolver_update(
         &mut self,
         update: ResolverUpdate,
-        config: Option<&LbConfig>,
+        config: Option<LbConfig>,
```
arjan-bal: We can avoid clones by sending references. For example, pickfirst doesn't seem to be storing the LB config presently. The ChildManager is still cloning the config to pass an owned object to all its children. With a reference, LB policies can choose to clone if they need an owned object.

Is supporting references complicating the implementation?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -64,44 +71,48 @@ pub struct ChildUpdate<T> {
...
-    /// identifier the ChildManager should use for this child.
+    /// Performs the operation of sharding an aggregate ResolverUpdate/LbConfig
+    /// into one or more ChildUpdates.  Called automatically by the ChildManager
+    /// when its resolver_update method is called.
     fn shard_update(
-        &self,
+        &mut self,
         resolver_update: ResolverUpdate,
-    ) -> Result<Box<dyn Iterator<Item = ChildUpdate<T>>>, Box<dyn Error + Send + Sync>>;
+        update: Option<LbConfig>,
```
arjan-bal: nit: `update` seems like a very general name. Maybe we should call this `lb_config` or `config` to be more specific?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -64,44 +71,48 @@ pub struct ChildUpdate<T> {
...
+            updated: false,
         }
     }
 
     /// Returns data for all current children.
-    pub fn child_states(&mut self) -> impl Iterator<Item = (&T, &LbState)> {
-        self.children
-            .iter()
-            .map(|child| (&child.identifier, &child.state))
+    pub fn children(&mut self) -> impl Iterator<Item = &Child<T>> {
```
arjan-bal: Can this method accept an immutable reference instead? Same question regarding the `aggregate_states` method below.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -64,44 +71,48 @@ pub struct ChildUpdate<T> {
     /// The identifier the ChildManager should use for this child.
     pub child_identifier: T,
     /// The builder the ChildManager should use to create this child if it does
-    /// not exist.
+    /// not exist.  The child_policy_builder's name is effectively a part of the
+    /// child_identifier.  If two identifiers are identical but have different
+    /// builder names, they are treated as different children.
```
arjan-bal, line 75: Do you think a simpler design would be to have the type argument for `T` include the builder name if necessary?

arjan-bal (commented):
I haven't reviewed the entire PR, leaving some initial comments.

grpc/src/client/load_balancing/child_manager.rs
dfawley: That's what I started out trying to do, actually.  Somewhere I decided that it was a correctness issue and that the child manager should take care of it for that reason.  Otherwise you could end up sending the wrong config to the wrong type of child.  And there should be no reason to want to switch the builder type for a child while keeping it around - it could only result in bugs.

In terms of code complexity, it makes the child manager itself only a little more complicated, because is really only one place where it matters (in resolver_update when determining which children to create vs. keep).  Anyone using child manager needs to be aware it's happening, e.g. when they're iterating through the children in the child manager, but doesn't have to _do_ anything special to take advantage of it.

grpc/src/client/load_balancing/child_manager.rs
dfawley: Yes, that should be fine.  I am not sure why I made these `&mut self` since they are read-only operations.  I'm expecting all real-world uses will have the ChildManager mutably when calling into it, but there's no reason to make these `&mut`.

grpc/src/client/load_balancing/child_manager.rs
dfawley: These should match `resolver_update` IMO.  For now that's apparently `update: ResolverUpdate, config: Option<LbConfig>`.  I'll go with that and if we want to rename the lb policy API for any reason, then we should come back and rename these too.

grpc/src/client/load_balancing/mod.rs
dfawley: Mainly, it felt like the more natural API to pass by value instead of reference.  Otherwise...why would pass _any_ parameters by value, since passing by reference is theoretically more flexible for the caller?  I.e. why not `&ResolverUpdate` here, too?

I don't think this actually saves us any clones in practice if we require it to be passed by value.  Passing by reference probably results in more clones since the parent is less likely to want to keep it than the child.  The child might keep it since that is supposed to configure its behavior.  The parent is done with it.

I made the change back as a separate commit if you want to compare the two.

dfawley: > For example, pickfirst doesn't seem to be storing the LB config presently

Regarding this, note that PF only uses the lb config for its behavior _during_ `resolver_update` which I believe is much less common than configuring the behavior of its ongoing operation.

dfawley: Thinking more about this, it seems most instances of `LbConfig` will actually be references -- they will come out of parts of the service config.  So perhaps this way is best...

grpc/src/client/load_balancing/child_manager.rs
arjan-bal: Acknowledged. Please feel free to resolve the conversation, I can't.

grpc/src/client/load_balancing/child_manager.rs
arjan-bal: Acknowledged.

grpc/src/client/load_balancing/child_manager.rs
arjan-bal, line 1: Should we have the exported symbols as `pub(crate)` instead of `pub` while they're expected to be changed?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
+#![warn(unused_imports)]
```
arjan-bal: Can this linter rule be enabled in this PR?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+            // A config should always be passed.  When one is not passed, we
+            // delete any children besides active.  This is how the graceful
+            // switch balaner performs a swap of active policies.
+            return Ok(vec![ChildUpdate {
+                child_policy_builder: self.active_child_builder.clone().unwrap(),
+                child_identifier: (),
+                child_update: None,
+            }]
+            .into_iter());
+        }
```
arjan-bal: This feels like a smart way to send a signal to drop the old LB. It took me some time to understand this piece. Do you think either of the following ways would make the code easier to understand?
1. Pass a different type of config inside the LbConfig instead of GracefulSwitchLbConfig. This could be the signal to send a blank update to the active child.
1. Add an extra boolean state in the sharder indicating that a blank update needs to be sent. The state can be reset once sent.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+            }
+        }
+        if could_switch && pending_builder.is_some() {
+            let sharder = self.child_manager.update_sharder();
+            sharder.active_child_builder = pending_builder;
+            self.child_manager.resolver_update(
+                ResolverUpdate::default(),
+                None,
+                channel_controller,
+            )?;
```
arjan-bal: Since this is just dropping a child LB, is this expected to fail? Presently, `maybe_swap` returns a `Result` and all but one of the callers are ignoring the returned error. It may be better to change this to an `expect` and get rid of the `Result` in the return type.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+        fn pick(&self, _req: &Request) -> PickResult {
+            PickResult::Pick(Pick {
+                subchannel: Arc::new(TestSubchannel::new(
+                    Address {
+                        address: self.name.to_string().into(),
+                        ..Default::default()
+                    },
+                    mpsc::unbounded_channel().0,
+                )),
+                // on_complete: None,
```
arjan-bal: Is this comment needed since the same line is present below?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+            // Closure for subchannel_update. Verify that the subchannel that
+            // being updated now is the same one that this child policy created
+            // in resolver_update. It then sends a picker of the same state that
+            // was passed to it.
+            subchannel_update: Some(Arc::new(
+                move |data: &mut StubPolicyData, updated_subchannel, state, channel_controller| {
+                    // Retrieve the specific TestState from the generic test_data field.
+                    // This downcasts the `Any` trait object
+                    if let Some(test_data) = data.test_data.as_mut() {
+                        if let Some(test_state) = test_data.downcast_mut::<TestState>() {
```
arjan-bal: Should we use an `expect()` instead of ignoring the case when the test data is of some other type?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+
+                let pick = update.picker.pick(&req);
+                match &pick {
+                    PickResult::Pick(pick) => {
+                        let received_address = &pick.subchannel.address().address.to_string();
+                        // It's good practice to create the expected value once.
+                        let expected_address = name.to_string();
+
+                        // Check for inequality and panic with a detailed message if they don't match.
+                        if received_address != &expected_address {
```
arjan-bal: nit: Can we use `assert_eq!` here?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+                address: addr.into(),
+                ..Default::default()
+            }],
+            ..Default::default()
+        }
+    }
+
+    // Verifies that the expected number of subchannels is created. Returns the
+    // subchannels created.
+    async fn verify_subchannel_creation_from_policy(
```
arjan-bal: We don't pass the `expected number of subchannels` as a param and it looks like this function only waits for the first event. I think the function name, signature (to return a single subchannel) and doc comment needs to be updated.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+        );
+        reg_stub_policy(
+            "stub-gracefulswitch_switching_to_resolver_update-two",
+            create_funcs_for_gracefulswitch_tests(
+                "stub-gracefulswitch_switching_to_resolver_update-two",
+            ),
+        );
+
+        let service_config = serde_json::json!([
+                { "stub-gracefulswitch_switching_to_resolver_update-one": serde_json::json!({}) }
```
arjan-bal, line 547: nit: Should we use shorter variable names for these `&strs`?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+
+        // Subchannel creation and ready
+        let subchannels = verify_subchannel_creation_from_policy(&mut rx_events).await;
+        let mut subchannels = subchannels.into_iter();
+        move_subchannel_to_state(
+            &mut *graceful_switch,
+            subchannels.next().unwrap(),
+            tcc.as_mut(),
+            ConnectivityState::Ready,
+        );
```
arjan-bal, line 572: To avoid unexpected test failures in the future, I think we should allow legal state changes for subchannels in tests. Here, we should transition the subchannel to connecting and then ready. 

In Go, I had to debug a fix a bunch of tests which were using fake subchannels and making illegal state transitions when adding the new pickfirst for dualstack.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+            ]
+        );
+        let new_parsed_config = GracefulSwitchPolicy::parse_config(&ParsedJsonLbConfig {
+            value: new_service_config,
+        })
+        .unwrap();
+        graceful_switch
+            .resolver_update(update.clone(), Some(&new_parsed_config), &mut *tcc)
+            .unwrap();
+
```
arjan-bal, line 593: Can we assert that a new picker is not produced using a short timeout?
I've done these kind of assertions in the dns test.
https://github.com/hyperium/tonic/blob/8921a2df61163e2fcd884f16756eacb52c256960/grpc/src/client/name_resolution/dns/test.rs#L420-L427

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -0,0 +1,990 @@
...
+                { "stub-gracefulswitch_two_policies_same_type-one": serde_json::json!({}) }
+            ]
+        );
+        let parsed_config2 = GracefulSwitchPolicy::parse_config(&ParsedJsonLbConfig {
+            value: service_config2,
+        })
+        .unwrap();
+        graceful_switch
+            .resolver_update(update.clone(), Some(&parsed_config2), &mut *tcc)
+            .unwrap();
```
arjan-bal, line 670: Similar to the comment above, I think we should wait for a while and assert no new picker is produced.

dfawley:
cc @sauravzg (I can't assign you as a reviewer/assignee)

grpc/src/client/load_balancing/child_manager.rs
dfawley: I _believe_ it's moot since `load_balancing` itself is `pub(crate)`?

But it can't hurt.  I've done a pretty comprehensive change of `pub`s into `pub(crate)`s for `load_balancing`, `name_resolution` and `transport` (all already declared as `pub(crate)` themselves).

There's actually a linter for this (e.g. if you make a field pub but the struct is private), but maybe it doesn't work inter-module?

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yes, thanks for catching this.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yeah, I knew it was a bit magical which is why I added the comment.  I originally started with a boolean in there, too, but it seemed unnecessary.

I have changed the GSB LbConfig to be an enum. It either performs a swap and needs the builder or it performs an update and needs the builder and the config.  I tried a few other options and I think this one feels the best of the things I tried - what do you think?

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yes, this should really not fail.  I feel really uncomfortable every time I use `unwrap` or `expect`, but it does seem better than dropping it for now at least.  Done.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Oops!

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yes, I agree, and it helps with the nesting too.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yes, thanks for catching this.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Done.  Also converted `match`es into `let else`s to reduce nesting.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: I would love to. The issue is that these things get registered globally, so for now they are using the name of the test function.  I want to redo the tests (which mainly Cathy wrote based on some guidance from Easwar) and would prefer to save those for another time when I can focus on just the testing aspect.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: I'm torn on this.  I think it's important that our code handles illegal state transitions "correctly", too, at least where it's possible to have some reasonable behavior.  Because we might change what is a legal state change -- see, e.g. https://github.com/grpc/proposal/blob/54694dd93c333d354f099cc9d1e74a91e27afafa/A105-max_concurrent_streams-connection-scaling.md#backoff-behavior-and-connectivity-state:

> Note that subchannels may exhibit two new state transitions that have not previously been possible. Today, the only possible transition from state READY is to state IDLE, but with this design, the following additional transitions are possible:

I guess more generally, I think transitions should not matter except where they absolutely have to.  I.e. we should try to operate solely based on the new state and not the old state.

> In Go, I had to debug a fix a bunch of tests which were using fake subchannels and making illegal state transitions when adding the new pickfirst for dualstack.

Yes, I remember this being an issue.  Do you remember examples of the illegal transitions that were problematic and why?

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Good idea; done.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Done

dfawley:
https://github.com/hyperium/tonic/pull/2442/commits/96da89301628d5b73c54ce6aa4353bde44d638fb attempts to address @sauravzg's comment https://github.com/hyperium/tonic/pull/2443#discussion_r2510968206

Thoughts?  We can probably do more, but this is at least one possible first step.

grpc/src/client/load_balancing/graceful_switch.rs
arjan-bal: Making the GSB config an enum is nice!

grpc/src/client/load_balancing/graceful_switch.rs
arjan-bal: > I guess more generally, I think transitions should not matter except where they absolutely have to. I.e. we should try to operate solely based on the new state and not the old state.

Acknowledged.

> Do you remember examples of the illegal transitions that were problematic and why?

The problem occured when subchannels went from `Ready` to `TF`. The new PF goes to `Idle` when a subchannel exited `Ready`, irrespective of the new state. The old PF used to enter `TF` in this situation. 

This transition was valid when health updates were sent through the same listener as the connection updates. The tests that failed were not using health checks though.

arjan-bal (approved):
LGTM

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -44,64 +45,74 @@ use crate::rt::Runtime;
 use super::{Subchannel, SubchannelState};
 
 // An LbPolicy implementation that manages multiple children.
-pub struct ChildManager<T> {
+#[derive(Debug)]
+pub(crate) struct ChildManager<T: Debug, S: ResolverUpdateSharder<T>> {
     subchannel_child_map: HashMap<WeakSubchannel, usize>,
```
sauravzg: nit: Since this is not a part of this PR but seems to be used at one of the changes below, should we consider renaming this to reflect what is being mapped to? Since usize is a pretty ambigious type it makes it difficult to understand the purpose of the map. Ex: subchannel_to_id ?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -44,64 +45,74 @@ use crate::rt::Runtime;
...
+            updated: false,
         }
     }
 
     /// Returns data for all current children.
-    pub fn child_states(&mut self) -> impl Iterator<Item = (&T, &LbState)> {
-        self.children
-            .iter()
-            .map(|child| (&child.identifier, &child.state))
+    pub fn children(&self) -> impl Iterator<Item = &Child<T>> {
```
sauravzg: Should we consider encapsulation here instead of returning accessors to non trivial data members(anything that isn't just plain old data type)?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -158,19 +169,37 @@ impl<T> ChildManager<T> {
...
         if let Some(state) = channel_controller.picker_update {
             self.children[child_idx].state = state;
+            self.children[child_idx].updated = true;
+            self.updated = true;
         };
     }
+
+    /// Returns a mutable reference to the update sharder so operations may be
+    /// performed on it for instances in which it needs to retain state.
+    pub fn update_sharder(&mut self) -> &mut S {
```
sauravzg, line 177: Same as above. Should we encapsulate the operations instead of returning a data member to avoid leaking implementations?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -186,25 +215,34 @@ impl<T: PartialEq + Hash + Eq + Send + Sync + 'static> LbPolicy for ChildManager
...
+                    identifier: old_idx,
+                    policy: e.policy,
+                    builder: e.builder,
+                    state: e.state,
+                    updated: e.updated,
+                    work_scheduler: e.work_scheduler,
+                },
+            )
+        });
+        let mut old_children: HashMap<(&'static str, T), _> = old_children.collect();
```
sauravzg: Do we need to collection to happen in a separate statement or can it be combined with the above?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -213,24 +251,23 @@ impl<T: PartialEq + Hash + Eq + Send + Sync + 'static> LbPolicy for ChildManager
         // update, and create new children.  Add entries back into the
         // subchannel map.
         for (new_idx, (identifier, builder)) in ids_builders.into_iter().enumerate() {
```
sauravzg, line 253: This is again probably not related to this PR itself so feel free to ignore and resolve, but should we simplify the state management and the recon process.

We seem to have some form of intertwined states (children, work queue, map) which seem to have some sort of interdependency.


Large cyclomatic complexity blocks which do non trivial state management tend to make it easy to add small bugs. So should we consider abstracting the state into a class which exposes valid operations on the state , separate the recon logic into `diff calculator` and `diff applier` ?  Or is that not enough value and might have a performance overhead over raw state manipulation

sauravzg:
Reviewed non test code. Moved over relevant comments from #2443 to this and resolved comments on the child PR.

grpc/src/client/load_balancing/child_manager.rs
dfawley: Renamed.

grpc/src/client/load_balancing/child_manager.rs
dfawley: Combined.

grpc/src/client/load_balancing/child_manager.rs
dfawley: This section of the code is obviously very complex.  I'd love to simplify it, but I'm worried that encapsulating it is just moving it elsewhere, and not actually simplifying it.

I think we can keep making improvements here over time.    For instance, if we implement either Clone or Default on the identifier T, then we can avoid the need to build a `Child<()>` and just `mem::take` or `.clone` the identifier for the reverse map it builds.  It's unclear to me which, if either, of those options is better, so I just avoided it entirely for now.

grpc/src/client/load_balancing/child_manager.rs
dfawley: I believe that `Child<T>` is essentially POD except for the WorkScheduler and child policy, which are active instances of things. I have moved those from pub to private as I don't think the parent should need to access them.  All the other fields (`identifier`, builder (mainly, for `.name()`), and last-produced-`state`) are static for the life of the child.

Maybe this ChildManager abstraction itself needs some fine-tuning to avoid some of the awkwardness so that the parent is more responsible for dealing with the child states/operations.  E.g. maybe we can inject something into ChildManager to observe the child operations, and then some of these other functions may not be necessary.

grpc/src/client/load_balancing/child_manager.rs
dfawley: This one's fun.  I added it so GSB can ask the sharder what the active policy is so that it can determine whether it needs to change policies.

We wouldn't need it if we're willing to duplicate the state.  I think it's impossible to do it with references because internal references in Rust are basically not a thing.

There might be other approaches as well, e.g. add some other way to perform operations on the ChildManager besides `resolver_update`.  Swapping to a new child also doesn't really want to perform this call, as evidenced by the awkwardness of passing a `ResolverUpdate::default()` and needing to support a sharder output of `Option<(ResolverUpdate, Option<LbConfig>>` to avoid calling the child.

I'll iterate a bit more here tomorrow and see if I can simplify.

sauravzg: Consider this and the above optional. Feel free to add a todo if you want to get the current API merged to have a functional child manager. 
I'll try to gather some thoughts from my own side to see if there's a potential solution here.

PR #2443, opened 2025-11-06
feat(grpc): make ChildManager call the parent work_scheduler

dfawley:
This is based on top of #2442; please review only the final commit in this PR.

I found this pattern for writing tests to be painful.  I will try to spend some time working on another approach.

The main problem I had with the tests was that the test_utils mod creates the `StubPolicy` and `StubPolicyData` meaning the tests have a hard time observing what the children Funcs did except by observing what was called on the child_controller or work_scheduler.

### DIFF
--- a/grpc/src/client/load_balancing/child_manager.rs
+++ b/grpc/src/client/load_balancing/child_manager.rs
@@ -27,7 +27,7 @@
 
 // TODO: This is mainly provided as a fairly complex example of the current LB
 // policy in use.  Complete tests must be written before it can be used in
-// production.  Also, support for the work scheduler is missing.
+// production.
 
 use std::collections::HashSet;
 use std::fmt::Debug;
@@ -53,6 +53,7 @@ pub(crate) struct ChildManager<T: Debug, S: ResolverUpdateSharder<T>> {
     pending_work: Arc<Mutex<HashSet<usize>>>,
     runtime: Arc<dyn Runtime>,
     updated: bool, // Set when any child updates its picker; cleared when accessed.
+    work_scheduler: Arc<dyn WorkScheduler>,
 }
 
 #[non_exhaustive]
@@ -98,13 +99,18 @@ where
 {
     /// Creates a new ChildManager LB policy.  shard_update is called whenever a
     /// resolver_update operation occurs.
-    pub fn new(update_sharder: S, runtime: Arc<dyn Runtime>) -> Self {
+    pub fn new(
+        update_sharder: S,
+        runtime: Arc<dyn Runtime>,
+        work_scheduler: Arc<dyn WorkScheduler>,
+    ) -> Self {
         Self {
             update_sharder,
             subchannel_to_child_idx: Default::default(),
             children: Default::default(),
             pending_work: Default::default(),
             runtime,
+            work_scheduler,
             updated: false,
         }
     }
@@ -272,6 +278,7 @@ where
                 let work_scheduler = Arc::new(ChildWorkScheduler {
                     pending_work: self.pending_work.clone(),
                     idx: Mutex::new(Some(new_idx)),
+                    work_scheduler: self.work_scheduler.clone(),
                 });
                 let policy = builder.build(LbPolicyOptions {
                     work_scheduler: work_scheduler.clone(),
@@ -395,8 +402,9 @@ impl ChannelController for WrappedController<'_> {
 
 #[derive(Debug)]
 struct ChildWorkScheduler {
+    work_scheduler: Arc<dyn WorkScheduler>, // The real work scheduler of the channel.
     pending_work: Arc<Mutex<HashSet<usize>>>, // Must be taken first for correctness
-    idx: Mutex<Option<usize>>,                // None if the child is deleted.
+    idx: Mutex<Option<usize>>,              // None if the child is deleted.
 }
 
 impl WorkScheduler for ChildWorkScheduler {
@@ -405,6 +413,12 @@ impl WorkScheduler for ChildWorkScheduler {
         if let Some(idx) = *self.idx.lock().unwrap() {
             pending_work.insert(idx);
         }
+        // Call the real work scheduler with the lock held to avoid a scenario
+        // where we schedule work and get called before the lock can be taken,
+        // and to avoid the scenario where the child is called before the
+        // schedule_work call is done due to a concurrent call to
+        // ChildManager::work().
+        self.work_scheduler.schedule_work();
     }
 }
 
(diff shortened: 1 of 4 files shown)


grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -30,6 +30,7 @@
 // production.  Also, support for the work scheduler is missing.
```
sauravzg: nit: Probably need to remove todo about the missing work scheduler?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -44,18 +45,25 @@ use crate::rt::Runtime;
 use super::{Subchannel, SubchannelState};
 
 // An LbPolicy implementation that manages multiple children.
-pub struct ChildManager<T> {
+#[derive(Debug)]
+pub struct ChildManager<T: Debug, S: ResolverUpdateSharder<T>> {
     subchannel_child_map: HashMap<WeakSubchannel, usize>,
```
sauravzg: Should we consider renaming this to reflect what is being mapped to? Since `usize` is a pretty ambigious type it makes it difficult to understand the purpose of the map.  Ex: subchannel_to_iindex

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -44,18 +45,25 @@ use crate::rt::Runtime;
...
     children: Vec<Child<T>>,
-    update_sharder: Box<dyn ResolverUpdateSharder<T>>,
+    update_sharder: S,
     pending_work: Arc<Mutex<HashSet<usize>>>,
     runtime: Arc<dyn Runtime>,
+    updated: bool, // Set when any child updates its picker; cleared when accessed.
```
sauravzg, line 55: We seem to have some form of intertwined states (children, work queue, map) which seem to have some sort of interdependency. Would it make sense to encapsulate the dependent states into a something like a `StateHolder` which only exposes potentially valid state manipulations on the underlying data? This'd help us avoid complexities around potentially missing corner cases which result in invalid states.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -64,44 +72,53 @@ pub struct ChildUpdate<T> {
...
+            updated: false,
         }
     }
 
     /// Returns data for all current children.
-    pub fn child_states(&mut self) -> impl Iterator<Item = (&T, &LbState)> {
-        self.children
-            .iter()
-            .map(|child| (&child.identifier, &child.state))
+    pub fn children(&self) -> impl Iterator<Item = &Child<T>> {
```
sauravzg, line 119: Should we consider encapsulation here instead of returning accessors to non trivial data members(anything that isn't just data)?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -158,19 +175,37 @@ impl<T> ChildManager<T> {
...
         if let Some(state) = channel_controller.picker_update {
             self.children[child_idx].state = state;
+            self.children[child_idx].updated = true;
+            self.updated = true;
         };
     }
+
+    /// Returns a mutable reference to the update sharder so operations may be
+    /// performed on it for instances in which it needs to retain state.
+    pub fn update_sharder(&mut self) -> &mut S {
```
sauravzg, line 183: Same as above. Should we encapsulate the operations instead of returning a data member to avoid leaking implementations?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -197,14 +232,16 @@ impl<T: PartialEq + Hash + Eq + Send + Sync + 'static> LbPolicy for ChildManager
         }
 
         // Build a map of the old children from their IDs for efficient lookups.
-        let old_children = old_children
-            .into_iter()
-            .enumerate()
-            .map(|(old_idx, e)| (e.identifier, (e.policy, e.state, old_idx, e.work_scheduler)));
-        let mut old_children: HashMap<T, _> = old_children.collect();
+        let old_children = old_children.into_iter().enumerate().map(|(old_idx, e)| {
```
sauravzg: Should we call collect on the first expression itself?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -197,14 +232,16 @@ impl<T: PartialEq + Hash + Eq + Send + Sync + 'static> LbPolicy for ChildManager
...
+            (
+                (e.builder.name(), e.identifier),
```
sauravzg: Should we abstract these tuples to some reasonable internal structs instead of arbitrary tuples? In general, we seem to have lots of tuples being structured and destructured, so the question would probably apply to it as well.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -213,7 +250,8 @@ impl<T: PartialEq + Hash + Eq + Send + Sync + 'static> LbPolicy for ChildManager
         // update, and create new children.  Add entries back into the
         // subchannel map.
         for (new_idx, (identifier, builder)) in ids_builders.into_iter().enumerate() {
```
sauravzg, line 259: Should we consider simplifying the reconciliation logic ? Along the lines of one of the previous comment about separating state manipulation into it's own encapsulation. Maybe we can split the updation into two steps calculate diff (which returns a set of diff operations) by reading the state and then asking the state manager to apply the set of diffs.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -705,4 +786,153 @@ mod test {
             ConnectivityState::TransientFailure
         );
     }
+
```
sauravzg, line 784: Do we have a test that verifies reconciliation business logic and branches?

grpc/src/client/load_balancing/child_manager.rs
dfawley: +1

grpc/src/client/load_balancing/child_manager.rs
dfawley: I'm not exactly clear on how that would work, but let's save anything on that scale for a separate change?

grpc/src/client/load_balancing/child_manager.rs
dfawley: Do you mean: combine this expression with line 241?  If so, done.

grpc/src/client/load_balancing/child_manager.rs
dfawley: Yes this was very confusing, and got more confusing as more state was added. I've found another way.

PR #2448, opened 2025-11-17
feat(prost-build): Relax bound on compile_protos and compile_with_config

### LINKED ISSUE #2433: Relax bounds on `Builder::compile_protos` and `Builder::compile_with_config`
### Crates

`tonic-prost-build`

### Motivation

Before the split out, `Builder::compile_protos` and `Builder::compile_with_config` used `impl AsRef<Path>` for both arguments (`protos` and `includes`) which allowed you to pass a `String` to one arg and a `PathBuf` to another one, making the usage more ergonomic.

With the split into `tonic-prost-build` both arguments now share a single bound. There wasn't a reason given for this change.

### Proposal

Change to
```rust
 pub fn compile_protos<P, I>(self, protos: &[P], includes: &[I]) -> io::Result<()>
    where
        P: AsRef<Path>,
        I: AsRef<Path>
```

tottoto:
Resolves #2433.

### DIFF
--- a/tonic-prost-build/Cargo.toml
+++ b/tonic-prost-build/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tonic-prost-build"
-version = "0.14.3"
+version = "0.15.0"
 authors = ["Lucio Franco <luciofranco14@gmail.com>"]
 edition = "2021"
 license = "MIT"


tottoto:
Although this is technically a breaking change, this can be treated as a minor change according to [RFC  1105](https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-introducing-a-new-type-parameter).

PR #2449, opened 2025-11-17
tonic: add max connection age grace

garmato:
## Motivation
According to the gRPC ~[documentation](https://grpc.io/docs/guides/server-graceful-stop/)~ [documentation](https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification) when a connection exceeds the graceful shutdown timeout, the server should begin a secondary forceful shutdown timer. If the client still does not close the connection within this period, the server must terminate the connection.

## Problem
Tonic currently does not enforce this second-phase shutdown. As a result, in long-running applications where clients become congested, stuck or simply ignore the `GOAWAY` frame, connections can linger indefinitely. This leads to unbounded growth in memory usage and a continuously increasing number of live Tokio tasks.

## Solution
This PR introduces a `max_connection_age_grace` timeout that starts after `max_connection_age` expires. If the client does not close the connection within this grace period, the server forcefully closes it. This mirrors the behavior of the [gRPC Go implementation](https://github.com/grpc/grpc-go/blob/1ed87ec0fb2a61656d13965a8b3aeb49c98c6e4b/internal/transport/http2_server.go#L1217).

A similar idea was previously mentioned in a [closed PR](https://github.com/hyperium/tonic/issues/1428) but it seems the discussion stalled before the feature was implemented.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -1280,8 +1280,8 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::time::Duration;
     use crate::transport::Server;
+    use std::time::Duration;
 
     #[tokio::test(start_paused = true)]
     async fn test_connection_timeout_no_max_age() {
@@ -1327,7 +1327,7 @@ mod tests {
 
         let action = future.await;
         assert!(matches!(action, TimeoutAction::ForcefulShutdown));
-    }    
+    }
 
     #[test]
     fn server_tcp_defaults() {


arjan-bal (approved):
LGTM. To clarify, this change relates to the MAX_CONNECTION_AGE_GRACE configuration (documented here: https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification).

This is distinct from the `Graceful Shutdown` documentation, which describes shutting down the entire server instance.

This file seems to lack good unit test coverage. @LucioFranco are there existing tests that can be updated adding coverage the new field?

tonic/src/transport/server/mod.rs
```diff
@@ -827,7 +862,8 @@ fn serve_connection<B, IO, S, E>(
 
             let mut conn = pin!(builder.serve_connection(hyper_io, hyper_svc));
 
-            let mut sleep = pin!(sleep_or_pending(max_connection_age));
+            let mut graceful_sleep = pin!(sleep_or_pending(max_connection_age));
+            let mut forceful_sleep = pin!(sleep_or_pending(None));
```
LucioFranco: Am I missing something but we should be able to combine these two and their logic together into a singular future that is `max age wait -> force ful wait -> exit`? From this we can write a simple unit test using tokio `test-utils` with their test timer?

garmato: @LucioFranco Thank you for the feedback. 
>we should be able to combine these two and their logic together into a singular future

Totally! I just made some changes to collapse both futures into a singular one and added some tests.

Please let me know if there's anything else that might require further attention, thank you!

LucioFranco:
@garmato I think if you can fix CI here, we can merge, thanks for the update

LucioFranco:
> Let me know if you prefer to open a brand new PR cherry picking the specific commit or any other resolution strategy you would prefer.

Could you do this and ping me and we can get that merged then rebase this and merge this. You're right I totally forgot about this branch haha. I do think I would like it to be its own commit.

garmato:
@LucioFranco cool, everything's green now, feel free to approve and merge whenever you have the chance, thank you!

LucioFranco (approved):

PR #2450, opened 2025-11-17
grpc: Rework ChildManager some

dfawley:
1. Stop `impl`-ing `LbPolicy` - given the functionality it provided, nothing could use it standalone anyway, since it never updated the picker
2. Replace `resolver_update` + `ResolverUpdateSharder` with update method that acts on `ChildUpdate`s
3. Update `GracefulSwitch` and tests accordingly
4. Add comments for readability

Once we have more policies similar to RoundRobin (which is still pending itself), we can see if it makes sense to have a _separate_ `ChildManagerLbPolicy` with the `ResolverUpdateSharder` and some kind of `PostChildUpdates` that can update the picker if needed.  Or maybe some other abstraction would be better.

### DIFF
--- a/grpc/src/client/load_balancing/child_manager.rs
+++ b/grpc/src/client/load_balancing/child_manager.rs
@@ -828,7 +828,7 @@ mod test {
             tx_events: tx_events.clone(),
         };
 
-        let names = vec![name1, name2];
+        let names = [name1, name2];
         let mut child_manager =
             ChildManager::new(default_runtime(), Arc::new(TestWorkScheduler { tx_events }));
 
--- a/grpc/src/client/load_balancing/graceful_switch.rs
+++ b/grpc/src/client/load_balancing/graceful_switch.rs
@@ -215,9 +215,9 @@ impl GracefulSwitchPolicy {
 
         self.active_child_builder = Some(pending_child_builder.clone());
         self.child_manager
-            .retain_children([((), pending_child_builder)].into_iter());
+            .retain_children([((), pending_child_builder)]);
 
-        return Some(pending_state);
+        Some(pending_state)
     }
 }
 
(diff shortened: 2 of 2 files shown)


grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -296,14 +272,30 @@ where
...
         for (_, old_child) in old_children {
-            *old_child.work_scheduler.idx.lock().unwrap() = None;
+            old_child.work_scheduler.invalidate();
         }
+        // Anything left in old_children will just be Dropped and cleaned up.
+    }
 
-        // Release the pending_work mutex before calling into the children to
-        // allow their work scheduler calls to unblock.
-        drop(pending_work);
```
sauravzg, line 304: Why don't we need to drop pending work anymore? Or should this be moved to the invalidate method below?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -13,109 +11,70 @@ use std::collections::HashMap;
...
+            .ok_or("graceful switch received no config")?
+            .convert_to::<GracefulSwitchLbConfig>()
+            .ok_or_else(|| format!("invalid config: {config:?}"))?;
+        if self.active_child_builder.is_none() {
+            self.active_child_builder = Some(config.child_builder.clone());
         }
 
+        let mut children = Vec::with_capacity(2);
+
         if self.active_child_builder.is_none() {
```
sauravzg, line 46: Isn't this a duplicate of the existing code above and essentially a no-op?

grpc/src/client/load_balancing/child_manager.rs
dfawley: pending_work is a local variable.  Because the function now ends at this point, it is implicitly dropped.  The other work that happened after it was moved into the `update` function, and the lock will not be held during those operations.

grpc/src/client/load_balancing/graceful_switch.rs
dfawley: Yes, I'm not sure how this happened.  Thanks for catching it.

PR #2405, opened 2025-08-27
feat(grpc): add round robin load balancing policy

cjqzhao:
Add round robin load balancing policy

@dfawley @easwars

PR #2451, opened 2025-11-18
feat(grpc): add round_robin LB policy

dfawley:
This change is based heavily on #2405, especially the tests.

### DIFF
--- a/grpc/src/client/load_balancing/round_robin.rs
+++ b/grpc/src/client/load_balancing/round_robin.rs
@@ -238,8 +238,8 @@ mod test {
         self, StubPolicyData, StubPolicyFuncs, TestChannelController, TestEvent, TestWorkScheduler,
     };
     use crate::client::load_balancing::{
-        pick_first, ChannelController, FailingPicker, LbPolicy, LbPolicyBuilder, LbState, Pick,
-        PickResult, Picker, QueuingPicker, Subchannel, SubchannelState, GLOBAL_LB_REGISTRY,
+        pick_first, ChannelController, FailingPicker, LbPolicy, LbState, Pick, PickResult, Picker,
+        QueuingPicker, Subchannel, SubchannelState, GLOBAL_LB_REGISTRY,
     };
     use crate::client::name_resolution::{Address, Endpoint, ResolverUpdate};
     use crate::client::ConnectivityState;


grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1327 @@
...
+            picker: Arc::new(FailingPicker { error }),
+        });
+        channel_controller.request_resolution();
+    }
+
+    // Sends an aggregate picker based on states of children.
+    //
+    // The state is determined according to normal state aggregation rules, and
+    // the picker round-robins between all children in that state.
+    fn resolve_child_updates(&mut self, channel_controller: &mut dyn ChannelController) {
```
sauravzg: I don't particularly see it doing anything related to resolver? It is using `resolve` as an overloaded termm here? It seems closer to `update_picker` to me from the implementation.

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1327 @@
...
+        (rx_events, lb_policy, tcc)
+    }
+
+    struct TestSubchannelList {
+        subchannels: Vec<Arc<dyn Subchannel>>,
+    }
+
+    impl TestSubchannelList {
+        fn new(addresses: &Vec<Address>, channel_controller: &mut dyn ChannelController) -> Self {
+            let mut scl = TestSubchannelList {
```
sauravzg: nit: more descriptive name like `subchannel_list`

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1327 @@
...
+
+    async fn verify_no_activity(rx_events: &mut mpsc::UnboundedReceiver<TestEvent>) {
+        tokio::select! {
+            _ = tokio::time::sleep(DEFAULT_TEST_SHORT_TIMEOUT) => {}
+            event = rx_events.recv() => {
+                panic!("unexpected event {:?}", event.unwrap());
+            }
+        }
+    }
+
```
sauravzg, line 680: This is a lot of boilerplate before the tests even start ~400+ LOC. Is this expected? Will dependency injection and mocking childmanager and pick first help here?

Additionally, I am not sure if it applies here. But go/tott/643 somewhat recommends leaning heavily towards not overabstracting utility functions to keep things dry.

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1327 @@
...
+            }
+        }
+        assert!(
+            picked[0] != picked[1].clone(),
+            "Should alternate between subchannels"
+        );
+        assert_eq!(&picked[0], &picked[2]);
+        assert_eq!(&picked[1], &picked[3]);
+        assert!(picked.contains(&subchannels[0]));
+        assert!(picked.contains(&subchannels[1]));
```
sauravzg, line 849: I assume this is detereministic, so can we assert the entire container of pickers chosen?

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -313,7 +313,25 @@ where
...
+    pub fn resolver_update(
+        &mut self,
+        resolver_update: ResolverUpdate,
+        config: Option<&LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        for child_idx in 0..self.children.len() {
+            let child = &mut self.children[child_idx];
+            let mut channel_controller = WrappedController::new(channel_controller);
+            let _ = child.policy.resolver_update(
```
sauravzg: Do we need this variable?

sauravzg:
Thanks for the amazing test descriptions. They were really amazing at smoothening the test reviews.

grpc/src/client/load_balancing/child_manager.rs
dfawley: This explicitly ignores the `Result` return value. You can't omit the assignment here or else the compiler complains because `Result`s are `#[must_use]`.

Since this is a general-purpose utility, maybe we should return the results of all the calls and let the caller choose whether to ignore?  I've made that change; LMK what you think.  We could also try to coalesce them into just one Result<> that indicates whether any child call errored.  Or return only the errors (`Result<(), Vec<Box<dyn Error>>>`).

At this point, only round_robin uses this method, and it doesn't care about any potential errors.

grpc/src/client/load_balancing/round_robin.rs
dfawley: Correct, this is the generic "resolve" and nothing to do with name resolution.  I assume this was named after:

https://github.com/hyperium/tonic/blob/d0d3f62bd26530d42f4b969d5b037df78cd31818/grpc/src/client/load_balancing/child_manager.rs#L148

Since I called it `update_picker` in graceful switch, I renamed this one.  But note that it doesn't handle all scenarios -- specifically if we got an error from the channel and had no previous addresses, we want to produce a picker that errors with the message we got from the resolver.  That's handled in `handle_resolver_error` instead.

grpc/src/client/load_balancing/round_robin.rs
dfawley: Simplified instead.

grpc/src/client/load_balancing/round_robin.rs
dfawley: I don't disagree with your sentiment. These tests are mainly as-written by Cathy with only small changes that I've made so far, and I didn't want to change anything too big yet, but I do want to do that later.

But I don't really like the idea of mocking the dependencies.  I'd prefer to use the real dependencies as much as possible to avoid as many differences between tests and production as we can.  I think we can have tests that are both system-level and less complex than what's needed here, but I need to spend some time working on it.

grpc/src/client/load_balancing/child_manager.rs
sauravzg: That's probably a classic problem of bulk status errors.  Usually, creating another Error type that represents something similar to `Vec<Box<dyn Error>>` and also provides a mergabe API to coalesce them should work.

Additionally, I am curious about what is the behavior in other languages(especially since they happen out of bounds in a work queue), do they fail silently

sauravzg: Also, feel free to merge , this is a small item which we can discuss offline or resolve later.

dfawley: This component is slightly novel for Rust, but there are at least partial equivalents in all languages.  I want it to be general purpose enough to be used for any parent->child LB policy pairing, because of the complexities around forwarding subchannel updates.  (The other languages have callbacks for those.)

In C++ their version is an [`EndpointList`](https://github.com/grpc/grpc/blob/master/src/core/load_balancing/endpoint_list.h) which only delegates to pick_first.  And it looks like their [`round_robin`](https://github.com/grpc/grpc/blob/247ec42a5f0e0c121f5ea6fe5a2329f45b59f3a6/src/core/load_balancing/round_robin/round_robin.cc#L271-L274) policy combines all the errors to return them back as one.

Java's WIP version of this only uses the most recent error: https://github.com/grpc/grpc-java/compare/master...ejona86:grpc-java:multichild-composition#diff-02b06d6bd9ea79461e364cb4087ed8a81693c578624c8b3d28948c80ac4a81c9R132-R135

And Go's is: https://github.com/grpc/grpc-go/blob/805b1f88c5fb9419e3837c72e1deb9c2ec677ffe/balancer/endpointsharding/endpointsharding.go#L182-L186 which is the same as Java's behavior.

sauravzg: Noted. I personally, would prefer some form of aggregateable status object, but I'd be content with "pick first" for error as well.

dfawley: Yeah that might be nice. For now I've made sure to aggregate into strings everywhere instead of ignoring the errors.  But we can switch to an multi-error type in the future if needed.  I kind of want to change all these `Box<dyn Error + Send + Sync>`s into `String`s instead when we aren't in the context of an RPC.  I don't think anything derives meaning from them programmatically, ever.  They're informational only.  C++ uses grpc::Status but that seems wrong to me since we aren't doing an RPC, and those errors are meant to express RPC failure modes.  But we can see how things go and adjust accordingly.

grpc/src/client/load_balancing/round_robin.rs
dfawley: Sorry, forgot to reply here.  I think this assertion is actually what we want.  The order is shuffled by round_robin's picker [here](https://github.com/hyperium/tonic/pull/2451/files#diff-fe6d7f8a8f9e823bd05d3f800561e9138f16fb55561d26434000677578d818dcR217) (the link probably won't work since the file is too large for it to be rendered by default so it's line 217), but even if it was deterministic, asserting that it both alternates and contains the two endpoints is probably best since that is all that is specifically required.

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1319 @@
...
+            "Received error from name resolver: {}",
+            resolver_update.endpoints.as_ref().unwrap_err()
+        );
+        if self.child_manager.children().next().is_none() {
+            // We had no children so we must produce an erroring picker.
+            self.move_to_transient_failure(err.clone(), channel_controller);
+            return Err(err.into());
+        }
+        // Forward the error to each child, ignoring their responses.
+        let _ = self
```
arjan-bal, line 131: I have a question about why the resolver_update method returns an error. Looking at the Go and C++ interfaces, it seems the only use of the returned error is to trigger re-resolution. However, since LBs can already request re-resolution via the channel control helper, the return value seems redundant. Do you know the reasoning behind this design?

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1319 @@
...
+        tokio::select! {
+            _ = tokio::time::sleep(DEFAULT_TEST_SHORT_TIMEOUT) => {}
+            event = rx_events.recv() => {
+                panic!("unexpected event {:?}", event.unwrap());
+            }
+        }
+    }
+
+    #[test]
+    fn roundrobin_builder_name() -> Result<(), String> {
```
arjan-bal: This test seems to be testing that the name string is "round_robin". I'm not sure if it's very useful.

grpc/src/client/load_balancing/round_robin.rs
```diff
@@ -0,0 +1,1319 @@
...
+
+        assert_eq!(&picked[0], &picked[2]);
+        assert_eq!(&picked[1], &picked[3]);
+        assert!(picked.contains(&subchannels[0]));
+        assert!(!picked.contains(&subchannel_being_removed));
+    }
+
+    // If Round Robin receives a resolver update that removes an endpoint and
+    // adds a new endpoint from a previous update, that endpoint's subchannels
+    // should not be apart of its picks anymore and should be removed. It should
```
arjan-bal: nit: There's a typo here, it should read `a part` instead of `apart`.

grpc/src/client/load_balancing/round_robin.rs
dfawley: Yeah, I thought it was not very useful either.  If anyone were to change it somehow, everything would break anyway, so I just deleted it.

grpc/src/client/load_balancing/round_robin.rs
dfawley: Fixed

grpc/src/client/load_balancing/round_robin.rs
dfawley: These two different paths that do something similar are for slightly different purposes.

Semantically they are different.  If the `resolver_update` method returns an error, it means the resolver update itself it was invalid (e.g. a broken LB policy config or no addresses for a policy that requires them).  Resolvers can forward that error to the control plane for monitoring purposes.  In contrast, the `resolve_now` method is used to indicate that the current state of the system is potentially stale.  This happens any time a subchannel disconnects, since, when that occurs, the addresses we have might be stale.

Another difference is: inside the DNS resolver, we use exponential backoff if there was some illegal state produced, whereas we simply rate limit re-resolution requests in the latter case (valid config, possibly stale addresses).  I'm not sure if this is very significant, however.

PR #2454, opened 2025-11-26
feat(grpc): introduce Status and ServerStatus types

sauravzg:
Introduces the foundational `Status` and `ServerStatus` types for gRPC status handling.

* `Status`: Represents a standard gRPC status with a code and message.
* `ServerStatus`: A wrapper around `Status` designed to prevent accidental leakage of sensitive server errors to the client. This type distinction ensures that handlers must explicitly convert to `Status`, encouraging a conscious decision about what information is exposed.
* `StatusCode`: Enum representing standard gRPC status codes.

**This implementation provides the necessary types to unblock server-side API development without relying on a heavy `tonic::Status`. This is not meant to be an intermediate or final Status implementation but serves as a way to implement functional requirements while we design the final `Status`.**

### DIFF
--- a/grpc/src/status/server_status.rs
+++ b/grpc/src/status/server_status.rs
@@ -22,15 +22,14 @@ impl ServerStatus {
         ServerStatus(Status::new(code, message))
     }
 
-    /// Converts the `ServerStatus` to a `Status` for client responses.
-    pub(crate) fn to_status(self) -> Status {
-        self.0
-    }
-}
-
-impl From<Status> for ServerStatus {
-    fn from(status: Status) -> Self {
+    /// Create a new `ServerStatus` from a `Status`.
+    pub fn from_status(status: Status) -> Self {
         ServerStatus(status)
+    }
+
+    /// Converts the `ServerStatus` to a `Status` for client responses.
+    pub(crate) fn into_status(self) -> Status {
+        self.0
     }
 }
 
@@ -54,14 +53,14 @@ mod tests {
     #[test]
     fn test_server_status_from_status() {
         let status = Status::new(StatusCode::Ok, "ok");
-        let server_status: ServerStatus = status.into();
+        let server_status = ServerStatus::from_status(status);
         assert_eq!(server_status.code(), StatusCode::Ok);
     }
 
     #[test]
-    fn test_server_status_to_status() {
+    fn test_server_status_into_status() {
         let server_status = ServerStatus::new(StatusCode::Ok, "ok");
-        let status = server_status.to_status();
+        let status = server_status.into_status();
         assert_eq!(status.code(), StatusCode::Ok);
     }
 }


grpc/src/status.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_status_new() {
+        let status = Status::new(StatusCode::Ok, "ok");
```
darth-raijin, line 40: Right now the `"ok"` message is duplicated in both the constructor call and the assertion.

To make the test less brittle, I’d suggest assigning the message to a variable and reusing it, e.g.:

```rust
let message = "ok";
let status = Status::new(StatusCode::Ok, message);

assert_eq!(status.code(), StatusCode::Ok);
assert_eq!(status.message(), message);
```

grpc/src/status/server_status.rs
```diff
@@ -0,0 +1,67 @@
...
+}
+
+impl ServerStatus {
+    /// Create a new `ServerStatus` with the given code and message.
+    pub fn new(code: StatusCode, message: impl Into<String>) -> Self {
+        ServerStatus(Status::new(code, message))
+    }
+
+    /// Converts the `ServerStatus` to a `Status` for client responses.
+    pub(crate) fn to_status(self) -> Status {
```
dfawley: Per the clippy warning this probably wants to be named `into_status(self)` instead.

grpc/src/status/server_status.rs
```diff
@@ -0,0 +1,67 @@
...
+        self.0
+    }
+}
+
+impl From<Status> for ServerStatus {
+    fn from(status: Status) -> Self {
+        ServerStatus(status)
+    }
+}
```
dfawley: I believe we want to _not_ implement this type? That's the reason for having a separate `ServerStatus` IIUC.  Otherwise, `?` does an auto-`into()` which would invoke this, meaning you could do:

```rust
fn method_handler(..) -> ServerStatus {
  self.client.call().await?;  // Will convert from Status to ServerStatus and return it
}
```

sauravzg: Damn! I wasn't aware of this magic. I assume an explicit function like ServerStatus::From(status) would be good enough?

dfawley: As long as we don't implement `From<Status>` then I think it is fine.

FYI the magic is documented here: https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.try.restricted-types:
>   * Result::Err(e) returns Result::Err(From::from(e)).

grpc/src/status.rs
sauravzg: Noted.  I personally tend to follow https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html  , so don't mind repeating stuff especially constants if it makes things more readable , simple and explicit.

I'll resolve this keeping things as they are. With that said, I don't particularly have a strong feeling towards either approach. Please reopen the comment if you have so and I'll gladly accomodate your approach.

PR #2456, opened 2025-11-29
Fix well-known Google wrapper types

daft-panda:
## Motivation

Google WKT were resolving incorrectly (e.g. google.protobuf.BoolValue resolved to super::bool).

## Solution

Fixed by using the resolved types from prost, added test cases.

### DIFF
--- a/tests/wellknown/proto/wellknown.proto
+++ b/tests/wellknown/proto/wellknown.proto
@@ -10,4 +10,13 @@ service Admin {
   rpc EmptyCall(google.protobuf.Empty) returns (google.protobuf.Empty);
   rpc StringCall(google.protobuf.StringValue) returns (google.protobuf.Empty);
   rpc AnyCall(google.protobuf.Any) returns (google.protobuf.Empty);
+  // Wrapper types that map to primitives
+  rpc BoolCall(google.protobuf.BoolValue) returns (google.protobuf.BoolValue);
+  rpc Int32Call(google.protobuf.Int32Value) returns (google.protobuf.Int32Value);
+  rpc Int64Call(google.protobuf.Int64Value) returns (google.protobuf.Int64Value);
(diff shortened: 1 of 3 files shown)


tests/wellknown/proto/wellknown.proto
```diff
@@ -10,4 +10,13 @@ service Admin {
   rpc EmptyCall(google.protobuf.Empty) returns (google.protobuf.Empty);
   rpc StringCall(google.protobuf.StringValue) returns (google.protobuf.Empty);
   rpc AnyCall(google.protobuf.Any) returns (google.protobuf.Empty);
+  // Wrapper types that map to primitives
```
LucioFranco, line 13: Are these new? In what version were they added?

daft-panda: They have been around at least since 2015 (https://github.com/protocolbuffers/protobuf/commits/main/src/google/protobuf/wrappers.proto). These wrappers used to work with prost 0.13.

LucioFranco: They used to work? huh I have no memory of seeing them but that legit could just be me. Can you expand on what you mean by it worked and what changed?

daft-panda: Can't expand that much I'm afraid, got pointed in this direction by Claude and considering the handling of other wrapper types this seemed like the place to be for a fix. My best guess would be this change in prost-build from 0.13 to 0.14: https://github.com/tokio-rs/prost/pull/1228 .

daft-panda: @LucioFranco would be great if this got merged, let me know if there is anything still missing.

daft-panda:
Fixed by https://github.com/grpc/grpc-rust/pull/2544

Issue #2444, opened 2025-11-13
Support xDS in Tonic and gRPC


gu0keno0:
## Feature Request
Support xDS in Tonic and gRPC 

### Motivation
In recent years, the official gRPC implementations have been [adding out-of-box service discovery and load-balancing capabilities via the xDS protocol](https://grpc.github.io/grpc/core/md_doc_grpc_xds_features.html). This enables gRPC users to build a service mesh without having to pay the overhead of proxies: by talking to a xDS control plane (e.g. Istio, Google Cloud Traffic Director), gRPC client can automatically discover gRPC servers, route traffic to them and dynamically configure its load-balancing algorithms out of the box.

My team have been using gRPC and xDS massively in our production environment. For Rust, we have recently added xDS support to our internal Tonic-based gRPC framework, and it achieved significant performance improvements over the Proxy-based service mesh implementation. We’d like to contribute our xDS implementation to Tonic as well as the ongoing Rust gRPC effort, so that other Rust users can also benefit from the lightweight proxyless service mesh.

### Proposal

The core components of our current solution are:

1. xDS client for subscribing resources to the xDS control plane, it is used by the Tower layers in `#3`
2. A “Cluster Channel” that leverages the tower::discover for dynamically configuring the endpoints under a xDS EDS cluster and tower::balance for load balancing among them
3. A few Tower layers on top of Cluster Channel for implementing the follows:
  a. Service Discovery & Routing (LDS → RDS → CDS)
  b. Handling retries
  c. Cluster Channel selection and dispatching

On top of this basic work, we are also actively adding other xDS features such as RDS routing rules, outlier detection, sticky routing, etc. For the xDS client, we plan to support delta xDS protocol and LEDS to handle our large fleet.

Per our previous discussions, we could potentially first contribute the xDS client and provide a generic mechanism for subscribing typed resources to xDS control plane. The APIs can be made transport agnostic, potentially similar to the [Go gRPC XDSClient interface](https://github.com/grpc/grpc-go/blob/master/internal/xds/xdsclient/client.go#L51) . The xDS client can be shared by both Tonic and Rust gRPC code.

[...]

PR #2459, opened 2025-12-05
feat(xds): initial crate structure

YutaoMa:
Initialize the crate scaffolding for xds-client.

Refs: #2444

## Motivation

In recent years, the official gRPC implementations have been [adding out-of-box service discovery and load-balancing capabilities via the xDS protocol](https://grpc.github.io/grpc/core/md_doc_grpc_xds_features.html). This enables gRPC users to build a service mesh without having to pay the overhead of proxies.

My team have been using gRPC and xDS massively in our production environment. For Rust, we have recently added xDS support to our internal Tonic-based gRPC framework, and it achieved significant performance improvements over the Proxy-based service mesh implementation. We’d like to contribute our xDS implementation to Tonic as well as the ongoing Rust gRPC effort, so that other Rust users can also benefit from the lightweight proxyless service mesh.

## Solution

This PR includes only the scaffolding of the new crate `xds-client`. While currently unpublished, the plan is to continue developing it along with `tonic` and the upcoming `grpc-rust`, so that eventually we'll have a working xDS implementation in both.

### DIFF
--- a/xds-client/Cargo.toml
+++ b/xds-client/Cargo.toml
@@ -12,13 +12,9 @@ rust-version = { workspace = true }
 workspace = true
 
 [dependencies]
-tonic = { version = "0.14", path = "../tonic", optional = true }
-tokio = { version = "1", optional = true }
-prost = { version = "0.14", optional = true }
-prost-types = { version = "0.14", optional = true }
 
 [features]
 default = ["transport-tonic"]
-transport-tonic = ["dep:tonic", "rt-tokio", "codegen-prost"]
-rt-tokio = ["dep:tokio"]
-codegen-prost = ["dep:prost", "dep:prost-types"]
+transport-tonic = ["rt-tokio", "codegen-prost"]
+rt-tokio = []
+codegen-prost = []


Cargo.toml
```diff
@@ -13,6 +13,7 @@ members = [
   "examples",
   "codegen",
   "grpc",
+  "xds-client",
```
gu0keno0, line 16: (Just to understand) do you plan to create another crate in the future for the Tower layer that leverages xDS?

xds-client/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "xds-client"
+version = "0.1.0-alpha.1"
+edition = "2021"
```
gu0keno0, line 5: Should this be 2025?

xds-client/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+rust-version = { workspace = true }
+
+[lints]
+workspace = true
+
+[dependencies]
+tonic = { version = "0.14", path = "../tonic", optional = true }
+tokio = { version = "1", optional = true }
+prost = { version = "0.14", optional = true }
+prost-types = { version = "0.14", optional = true }
```
gu0keno0: Where will we need these in the xDS implementation?

xds-client/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0-alpha.1"
+edition = "2021"
+license = "MIT"
+rust-version = { workspace = true }
+
+[lints]
+workspace = true
+
+[dependencies]
+tonic = { version = "0.14", path = "../tonic", optional = true }
```
ankurmittal: clippy will complain as these are not used today? remove them and add as needed?

Cargo.toml
YutaoMa: That's correct, based on the discussion with the Google gRPC team, the plan is to keep `xds-client` limited to xDS subscription logic and all gRPC-specific logic will be in a separate crate. That way the `xds-client` can be used in other protocols as well.

xds-client/Cargo.toml
YutaoMa: The latest Rust edition is 2024, and I chose to go with 2021 to align with the current edition `tonic` and `grpc` are using.

xds-client/Cargo.toml
YutaoMa: It would be used in deserialization of xDS resource types (e.g. Listener, RouteConfiguration). Technically the bare bone xDS client don't have to define any concrete xDS resource types, but in our design of this crate, `codegen-prost` would enable the use of `prost` based xDS types for convenience.

xds-client/Cargo.toml
YutaoMa: I included these dependencies to give a sense of what each feature flag will enable, I can remove them from this PR.

Issue #2244, opened 2025-04-04
Update `BadRequest.FieldViolation` richer error type


MathieuTricoire:
### Crates

`tonic-types`

### Motivation

Google has updated the `google.rpc.BadRequest.FieldViolation` message in its error model by adding two new fields: `string reason` and `LocalizedMessage localized_message`. This feature request proposes reflecting this upstream change in the `tonic-types` crate.

Reference:  
https://github.com/googleapis/googleapis/blob/a31d216ae702a184b5b00bdece3775177dfb2cf7/google/rpc/error_details.proto#L216-L226

### Proposal

- Update the `.proto` files to the latest version.
- Extend the `FieldViolation` struct to include two new fields:
  
  - `pub reason: String`
  - `pub localized_message: Option<LocalizedMessage>`

- Update the following functions to accept two additional parameters (`reason: impl Into<String>` and `loc_message: Option<LocalizedMessage>`):

  - `FieldViolation::new`
  - `ErrorDetails::with_bad_request_violation`
  - `ErrorDetails::add_bad_request_violation`
  - `BadRequest::add_violation`

This proposal would involve modifying existing functions and therefore introduces a **breaking change**.

### Alternatives

This is an initial proposal, and I would like to open the discussion to consider a non-breaking alternative.

Instead of modifying existing functions, we could add new functions to support the new fields. In that case, I’d appreciate input on appropriate naming conventions.

I’m happy to contribute the implementation once a direction is agreed upon. I already have a branch prepared for the breaking-change version, if you'd like to review it to get a sense of the scope https://github.com/hyperium/tonic/commit/86d1095be40c90cd64a2e80662d4afd77082bd9f

PR #2461, opened 2025-12-05
fix(types): Serialize `BadRequest::FieldViolation` `reason` and `localized_message`

### LINKED ISSUE #2460: `BadRequest::FieldViolation` `reason` and `localized_message` are not serialized
### Version

`0.14.2`

### Platform

Darwin

### Crates

`tonic-types`

### Description

After the addition of the `reason` and `localized_message` fields to `BadRequest::FieldViolation`, these fields are not actually serialized into the underlying protobuf message and therefore never make it onto the wire.

The problem seems to come from the `From<FieldViolation> for pb::bad_request::FieldViolation` implementation in `tonic-types/src/richer_error/std_messages/bad_request.rs`:

#### Reproduction code:

Create a new rust project, add tonic-types `cargo add tonic-types`.

Update `main.rs`:
```rust
use tonic_types::pb;
use tonic_types::{BadRequest, FieldViolation, LocalizedMessage};

fn main() {
    let field_violations = vec![FieldViolation {
        field: "field".to_string(),
        description: "description".to_string(),
        reason: "REASON".to_string(),
        localized_message: Some(LocalizedMessage::new("en-US", "localized error")),
    }];

    let br_details = BadRequest::new(field_violations);
    let pb_br: pb::BadRequest = br_details.clone().into();

    // Get back from pb type.
    let br_details_from_pb: BadRequest = pb_br.into();

    println!("Original: {br_details:#?}");
    println!("From pb:  {br_details_from_pb:#?}");
    assert_eq!(
        format!("{:?}", br_details),
        format!("{:?}", br_details_from_pb)
    );
}
```

I expected to see this happen, output:
```
Original: BadRequest {
    field_violations: [
        FieldViolation {
            field: "field",
            description: "description",
            reason: "REASON",
            localized_message: Some(
                LocalizedMessage {
                    locale: "en-US",
                    message: "localized error",
                },
            ),
        },
    ],
}
From pb:  BadRequest {
    field_violations: [
        FieldViolation {
            field: "field",
            description: "description",
            reason: "REASON",
            localized_message: Some(
                LocalizedMessage {
                    locale: "en-US",
                    message: "localized error",
                },
            ),
        },
    ],
}
```

Instead, this happened, panic with output:
```
Original: BadRequest {
    field_violations: [
        FieldViolation {
            field: "field",
            description: "description",
            reason: "REASON",
            localized_message: Some(
                LocalizedMessage {
                    locale: "en-US",
                    message: "localized error",
                },
            ),
        },
    ],
}
From pb:  BadRequest {
    field_violations: [
        FieldViolation {
            field: "field",
            description: "description",
            reason: "",
            localized_message: None,
        },
    ],
}
```

#### Fix

Patch `tonic-types` with my fix (PR coming):

```toml
[patch.crates-io]
tonic-types = { git = "https://github.com/MathieuTricoire/tonic", branch = "field-violation-serialization-fix" }
```

MathieuTricoire:
## Motivation

fix #2460

Fields `reason` and `localized_message` from `BadRequest::FieldViolation` are not serialized into the underlying protobuf message and therefore never make it onto the wire.

## Solution

Fix the conversion from `FieldViolation` to the prost-generated `pb::bad_request::FieldViolation` by explicitly forwarding both the `reason` and `localized_message` fields instead of leaving them at their default values.

A new test is added to ensure that behavior.

### DIFF
--- a/tonic-types/src/richer_error/std_messages/bad_request.rs
+++ b/tonic-types/src/richer_error/std_messages/bad_request.rs
@@ -164,8 +164,8 @@ mod tests {
 
     #[test]
     fn gen_bad_request() {
-        let mut br_details = BadRequest::new(Vec::new());
-        let formatted = format!("{br_details:?}");
+        let empty_br_details = BadRequest::new(Vec::new());
+        let formatted = format!("{empty_br_details:?}");
 
         let expected = "BadRequest { field_violations: [] }";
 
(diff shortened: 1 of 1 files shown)


tonic-types/src/richer_error/std_messages/bad_request.rs
```diff
@@ -218,4 +219,51 @@ mod tests {
             "BadRequest from Any differs from expected result"
         );
     }
+
+    #[test]
+    fn gen_bad_request_with_verbose_field_violation() {
```
sauravzg: I'd recommend either merging the business logic into the original test or creating an incremental fine grained test that tests just the new business logic that has been added in this change.

MathieuTricoire: Thanks @sauravzg for the review! I've merged the new coverage into the original test as suggested. Let me know if this looks good to you.

sauravzg (approved):

MathieuTricoire:
Hi @sauravzg 
I rebased onto `master` and retargeted the PR there since `v0.14.x` seems stale, let me know if that's correct.
The diff is unchanged from what you approved: one file (`bad_request.rs`)
Could it be picked up for the next tonic patch release?

MathieuTricoire:
Hi, sorry again but I'm not sure who I need to ping here, is there still any interest in merging this fix?

It's been almost 8 months since I opened the PR, it was approved back in December, and I've rebased it a couple of times since then without much progress.

If there's no interest in merging it, that's genuinely completely fine to me, but I'd rather know so I can close the PR (noise for the project and you) and stop maintaining / tracking it (noise for me).

@dfawley @arjan-bal @ejona86 @YutaoMa @nathanielford @gu0keno0

tonic-types/src/richer_error/std_messages/bad_request.rs
```diff
@@ -53,7 +53,8 @@ impl From<FieldViolation> for pb::bad_request::FieldViolation {
         pb::bad_request::FieldViolation {
             field: value.field,
             description: value.description,
-            ..Default::default()
```
dfawley, line 56: We should keep the `..Default::default()` in case the proto message is extended in the future.

But also if that happens, I'm not sure what that means for the `FieldViolation` struct here, but that's a totally separate issue.

dfawley (commented):
Any idea why these fields weren't getting propagated before?  That would take some spelunking through the commit history.  Were they omitted on purpose?  Were the fields added later?

This does look good to me in theory, but I don't know much about these types.

MathieuTricoire:
No idea why, but I suppose it was just a simple oversight.

If I remember correctly, I had opened https://github.com/grpc/grpc-rust/issues/2244 about these new fields some time ago and proposed an initial solution. It wasn't perfect since it introduced breaking changes, but it was meant as a starting point: https://github.com/MathieuTricoire/tonic/commit/86d1095be40c90cd64a2e80662d4afd77082bd9f

In the meantime a PR about these new fields landed and seems to be where this oversight was introduced: https://github.com/grpc/grpc-rust/pull/2306

I think it's ultimately a project-level design decision regarding `..Default::default()`. The question is whether future fields should be tolerated and defaulted, or deliberately cause a compilation failure until the conversion is updated.

So I can keep it, let me know.

dfawley:
> I think it's ultimately a project-level design decision regarding ..Default::default(). The question is whether future fields should be tolerated and defaulted, or deliberately cause a compilation failure until the conversion is updated.

Adding fields to a proto message should be a backward-compatible operation.  That said, this crate owns its own protobuf code generation, as opposed to an external dependency, so either way is valid here.  I'd rather let @LucioFranco weigh in his opinion on it.

dfawley (approved):

tonic-types/src/richer_error/std_messages/bad_request.rs
nathanielford: FWIW `Default::default()` defaults anything in the receiving struct but not passed explicitly. And in this case, `reason` and `localized_message` were late additions to the `FieldViolation` proto and just never added into the serialization code. It was added to the deserialization code. So it's just been dropping these fields when being serialized.

I did a brief search and I _think_ this is the only place where this oversight happened. 

If `FieldViolation` is extended in the future, extra fields will still be dropped by this serialization if we keep the default call. I think we would need to remove it if we want to use compilation failure to force an update here. I don't think this has any impacts on in-production systems, so I might advocate for that.

PR #2462, opened 2025-12-08
Return a Result type from the PersistentChannel instead of unwrapping.

nathanielford:
Currently the PersistentChannel constructor panics if the URL cannot be parsed. This adds a check for a parse error and returns a corresponding Status::INVALID_ARGUMENT if the parse failed.  Using the Status struct should be conformant with the C++ implementation and the current tonic implementation.

## Motivation

Standardize error handling, handle a TODO.

## Solution

Using a Result object to return. The Status code is to conform with the [gRPC spec ](https://github.com/ejona86/grpc/blob/call-semantics/doc/PROTOCOL-SEMANTICS.md)

### DIFF
--- a/grpc/examples/inmemory.rs
+++ b/grpc/examples/inmemory.rs
@@ -46,7 +46,8 @@ async fn main() {
 
     println!("Creating channel for {}", lis.target());
     let chan_opts = ChannelOptions::default();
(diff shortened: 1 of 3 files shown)


nathanielford:
After discussion, opting for a different path.

PR #2464, opened 2025-12-10
feat(xds): add public APIs for xds-client

YutaoMa:
## Motivation

As explained previously #2459 , we are upstreaming an implementation of xDS client in Rust, in order to support xDS based routing and load balancing in tonic / gRPC Rust eventually. This PR is a continuation of the effort, by exposing all the expected public APIs (traits, types, function signatures) first to lay out the design.

## Solution

Here is an overview of all the public APIs we are adding in this PR:

| File | Public Types |
|------|--------------|
| `error.rs` | `Error`, `Result` |
| `resource.rs` | `Resource` trait |
| `transport.rs` | `Transport`, `TransportStream`, `DiscoveryRequest`, `DiscoveryResponse` |
| `runtime.rs` | `Runtime` trait |
| `config.rs` | `ClientConfig` |
| `watcher.rs` | `ResourceWatcher<T>`, `ResourceEvent<T>` |
| `client.rs` | `XdsClient`, `XdsClientBuilder` |

In case the reviewer is curious if the provided APIs could support the expected functionality of an xDS client, I have a working proof of concept that implements these APIs in my fork: https://github.com/YutaoMa/tonic/tree/yutaoma/xds-client-poc, it showcases the APIs working together with an implementation based on the `tonic` stack (`tonic`, `tokio`, `prost`).

### DIFF
--- a/xds-client/Cargo.toml
+++ b/xds-client/Cargo.toml
@@ -15,6 +15,7 @@ workspace = true
 [dependencies]
 bytes = "1.11.0"
 thiserror = "2"
+futures-channel = "0.3"
 
 [features]
 default = ["transport-tonic"]
--- a/xds-client/src/client/watch.rs
+++ b/xds-client/src/client/watch.rs
@@ -2,19 +2,97 @@
 
 use crate::error::Error;
 use crate::resource::Resource;
+use futures_channel::oneshot;
+
+/// A signal to indicate that processing of a resource event is complete.
+///
+/// The xDS client waits for this signal before sending ACK/NACK to the server.
+/// This allows watchers to add cascading subscriptions (e.g. LDS -> RDS -> CDS -> EDS)
+/// that will be included in the same ACK.
+///
+/// # Automatic Signaling
+///
+/// Signals automatically when dropped, so you don't need to call [`.complete()`](Self::complete)
+/// explicitly if you have no cascading watches to add.
+///
+/// # Example
+///
+/// ```ignore
+/// match event {
+///     ResourceEvent::ResourceChanged { resource, done } => {
+///         // Process the resource, possibly add cascading watches.
+///         client.watch::<RouteConfiguration>(&resource.route_name());
+///         done.complete();
+///     }
+///     ResourceEvent::ResourceError { error, done } => {
+///         eprintln!("Error: {}", error);
+///         done.complete();
+///     }
+///     ResourceEvent::AmbientError { error, .. } => {
+///         // Can also rely on auto-signal on drop
+///         eprintln!("Ambient error: {}", error);
+///     }
+/// }
+/// ```
+#[derive(Debug)]
+pub struct ProcessingDone(Option<oneshot::Sender<()>>);
+
+impl ProcessingDone {
+    /// Create a channel pair for signaling.
+    ///
+    /// Returns the `ProcessingDone` sender and a receiver future that resolves
+    /// when `complete()` is called or the sender is dropped.
+    #[allow(dead_code)] // TODO: remove once used by XdsClient worker
+    pub(crate) fn channel() -> (Self, oneshot::Receiver<()>) {
+        let (tx, rx) = oneshot::channel();
+        (Self(Some(tx)), rx)
+    }
+
+    /// Signal that processing is complete.
+    ///
+    /// This is equivalent to dropping the `ProcessingDone`, but more explicit.
+    pub fn complete(&mut self) {
+        if let Some(tx) = self.0.take() {
+            let _ = tx.send(());
+        }
+    }
+}
+
+impl Drop for ProcessingDone {
+    fn drop(&mut self) {
+        // Auto-signal on drop to prevent deadlocks if the caller forgets
+        // or doesn't need to explicitly signal.
+        self.complete();
+    }
+}
 
 /// Events delivered to resource watchers.
 #[derive(Debug)]
 pub enum ResourceEvent<T> {
     /// Indicates a new version of the resource is available.
-    ResourceChanged(T),
+    ResourceChanged {
+        /// The updated resource.
+        resource: T,
+        /// Signal when processing is complete.
+        done: ProcessingDone,
+    },
     /// Indicates an error occurred while trying to fetch or decode the resource.
-    ResourceError(Error),
+    ResourceError {
+        /// The error that occurred.
+        error: Error,
+        /// Signal when processing is complete.
+        done: ProcessingDone,
+    },
     /// Indicates an error occurred after a resource has been received that should
     /// not modify the use of that resource but may provide useful information
     /// about the state of the XdsClient. The previous version of the resource
     /// should still be considered valid.
-    AmbientError(Error),
+    AmbientError {
+        /// The error that occurred.
+        error: Error,
+        /// Signal when processing is complete.
+        done: ProcessingDone,
+    },
 }
 
 /// A watcher for resources of type `T`.
@@ -37,9 +115,19 @@ impl<T: Resource> ResourceWatcher<T> {
     /// ```ignore
     /// while let Some(event) = watcher.next().await {
     ///     match event {
-    ///         ResourceEvent::ResourceChanged(resource) => { /* handle */}
-    ///         ResourceEvent::ResourceError(error) => { /* handle */}
-    ///         ResourceEvent::AmbientError(error) => { /* handle */}
+    ///         ResourceEvent::ResourceChanged { resource, done } => {
+    ///             // Process the resource, possibly add cascading watches.
+    ///             client.watch::<RouteConfiguration>(&resource.route_name());
+    ///             done.complete();
+    ///         }
+    ///         ResourceEvent::ResourceError { error, done } => {
+    ///             eprintln!("Error: {}", error);
+    ///             done.complete();
+    ///         }
+    ///         ResourceEvent::AmbientError { error, .. } => {
+    ///             // Can also rely on auto-signal on drop
+    ///             eprintln!("Ambient error: {}", error);
+    ///         }
     ///     }
     /// }
     /// ```
--- a/xds-client/src/transport/mod.rs
+++ b/xds-client/src/transport/mod.rs
@@ -32,7 +32,7 @@ pub trait Transport: Send + Sync + 'static {
     /// Creates a new bidirectional ADS stream to the xDS server.
     ///
     /// This may be called multiple times for reconnection.
-    fn connect(&self) -> impl Future<Output = Result<Self::Stream>> + Send;
+    fn new_stream(&self) -> impl Future<Output = Result<Self::Stream>> + Send;
 }
 
 /// A bidirectional stream for xDS ADS communication.
--- a/xds-client/src/lib.rs
+++ b/xds-client/src/lib.rs
@@ -25,8 +25,19 @@
 //! let mut watcher = client.watch::<Listener>("my-listener");
 //! while let Some(event) = watcher.next().await {
 //!     match event {
-//!         ResourceEvent::ResourceChanged(resource) => { /* handle resource change */ }
-//!         // ... handle other events ...
+//!         ResourceEvent::ResourceChanged { resource, done } => {
+//!             // Process the resource, possibly add cascading watches.
+//!             client.watch::<RouteConfiguration>(&resource.route_name());
+//!             done.complete();
+//!         }
+//!         ResourceEvent::ResourceError { error, done } => {
+//!             eprintln!("Error: {}", error);
+//!             done.complete();
+//!         }
+//!         ResourceEvent::AmbientError { error, .. } => {
+//!             // Can also rely on auto-signal on drop
+//!             eprintln!("Ambient error: {}", error);
+//!         }
 //!     }
 //! }
 //! ```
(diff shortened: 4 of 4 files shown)


xds-client/src/client/mod.rs
```diff
@@ -1 +1,72 @@
...
+use crate::error::Result;
+use crate::resource::Resource;
+
+pub mod config;
+pub mod watch;
+
+/// Builder for [`XdsClient`].
+#[derive(Debug)]
+pub struct XdsClientBuilder {
+    _config: ClientConfig,
```
gu0keno0, line 15: We can use builder to get configuration knobs from caller, so is it still necessary to make ClientConfig public, or maybe simply remove ClientConfig?

xds-client/src/client/mod.rs
```diff
@@ -1 +1,72 @@
...
+impl XdsClientBuilder {
+    /// Create a new builder with the given configuration.
+    pub fn new(config: ClientConfig) -> Self {
+        Self { _config: config }
+    }
+
+    /// Build the client with the given transport and runtime.
+    ///
+    /// This starts the background worker that manages the ADS stream.
+    pub async fn build<T, R>(self, _transport: T, _runtime: R) -> Result<XdsClient> {
```
gu0keno0: Nit: make build() simple without parameter, and add methods for overriding transport and runtime?

xds-client/src/client/mod.rs
```diff
@@ -1 +1,72 @@
...
+/// This is a handle to the background worker that manages the ADS stream.
+/// Cloning this handle creates a new reference to the same worker.
+#[derive(Clone, Debug)]
+pub struct XdsClient {
+    // TODO: add fields as needed
+}
+
+impl XdsClient {
+    /// Create a new builder.
+    pub fn builder(config: ClientConfig) -> XdsClientBuilder {
```
gu0keno0, line 43: This does not seem necessary?

xds-client/src/client/watch.rs
```diff
@@ -0,0 +1,39 @@
...
+use crate::resource::Resource;
+use futures::Stream;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// Events delivered to resource watchers.
+#[derive(Debug)]
+pub enum ResourceEvent<T> {
+    /// Resource was added or updated.
+    Upsert(T),
```
gu0keno0: Maybe rename to "Changed"

Also, probably we can just follow Golang's existing API https://pkg.go.dev/google.golang.org/grpc@v1.75.1/xds/internal/xdsclient/xdsresource#ResourceWatcher

xds-client/src/resource/mod.rs
```diff
@@ -1,4 +1,52 @@
...
+///     fn name(&self) -> &str {
+///         &self.name
+///     }
+/// }
+/// ```
+pub trait Resource: Send + Sync + Clone + std::fmt::Debug + 'static {
+    /// The xDS type URL for this resource type.
+    ///
+    /// Example: `"type.googleapis.com/envoy.config.listener.v3.Listener"`
+    const TYPE_URL: &'static str;
```
gu0keno0: Shall we crate a new type wrapper TypeUrl ?

xds-client/src/client/mod.rs
YutaoMa: This is so the user don't have to import the builder type, more of a convenience helper. The commonly used `bon` provides a similar ux. I'd prefer to keep this.

xds-client/src/client/mod.rs
YutaoMa: Makes sense, I can modify it to default to `tonic` stack, given we already made that the default feature flag anyway.

xds-client/src/client/mod.rs
YutaoMa: imo the `ClientConfig` makes discovery of all available knobs easier, and more info about those knobs `required/optional`, `default values` can be expressed more easily.

xds-client/src/client/watch.rs
YutaoMa: Good idea, I'll change to be aligned with the `grpc-go` API.

xds-client/src/resource/mod.rs
YutaoMa: Good point, using new type pattern is a good approach here, let me apply that.

xds-client/Cargo.toml
LucioFranco, line 1: should add `publish = false`

xds-client/src/client/watch.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// ```ignore
+    /// while let Some(event) = watcher.next().await {
+    ///     match event {
+    ///         ResourceEvent::Upsert(resource) => { /* handle */}
+    ///         ResourceEvent::Removed { name } => { /* handle */}
+    ///         ResourceEvent::Error(error) => { /* handle */}
+    ///     }
+    /// }
+    /// ```
+    pub fn next(&mut self) -> impl Future<Output = Option<ResourceEvent<T>>> + '_ {
```
LucioFranco: take a look at some of the tokio::sync apis for what this signature should look like in regards to async fn/return types etc in general tokio is a great reference point

xds-client/src/resource/mod.rs
LucioFranco: might be good to capture a list of things like this to see if there is types we can share across the ecosystem

xds-client/src/resource/mod.rs
```diff
@@ -1,4 +1,52 @@
...
+///         // Validate fields...
+///         Ok(Self { name: proto.name, /* ... */ })
+///     }
+///
+///     fn name(&self) -> &str {
+///         &self.name
+///     }
+/// }
+/// ```
+pub trait Resource: Send + Sync + Clone + std::fmt::Debug + 'static {
```
LucioFranco: Debug should not be enforced here it should be enforced at the lowest level call site where debug is called or else you will need to include it on all the bounds. Clone could be thought of in the same way though if all listeners implement clone they it could be fine to keep up to you, send/sync/static are basically necessary for anything sendable futures so that is fine to keep.

xds-client/src/runtime/mod.rs
```diff
@@ -1,4 +1,20 @@
...
+use std::future::Future;
+use std::time::Duration;
+
 #[cfg(feature = "rt-tokio")]
 pub mod tokio;
+
+/// Trait for async runtime operations.
+///
+/// This abstraction allows the xDS client to be runtime-agnostic.
+pub trait Runtime: Send + Sync + Clone + 'static {
```
LucioFranco, line 13: We need to think how this might interact with the `grpc-rust` version cc @dfawley @arjan-bal

xds-client/src/transport/mod.rs
```diff
@@ -1,4 +1,50 @@
...
+    /// Creates a new bidirectional ADS stream to the xDS server.
+    ///
+    /// This may be called multiple times for reconnection.
+    fn connect(&self) -> impl Future<Output = Result<Self::Stream>> + Send;
+}
+
+/// A bidirectional stream for xDS ADS communication.
+pub trait TransportStream: Send + 'static {
+    /// Send a discovery request to the server.
+    fn send(&mut self, request: DiscoveryRequest) -> impl Future<Output = Result<()>> + Send;
```
LucioFranco, line 41: there might be a crate that provides macro sugar for the sendable async trait fn

xds-client/Cargo.toml
```diff
@@ -12,9 +12,17 @@ rust-version = { workspace = true }
 workspace = true
 
 [dependencies]
+bytes = "1.11.0"
+thiserror = "2.0.17"
```
LucioFranco: just pin these to `thiserror = "2"`

xds-client/src/transport/mod.rs
sauravzg: https://crates.io/crates/trait-variant

xds-client/src/runtime/mod.rs
sauravzg: We can probably adapt the grpc-rust rt api to xds RT API and pass it along. I believe have an hyper rt wrapper which does something similar for hyper

YutaoMa: Added `TODO` for now. Export and re-use `grpc-rust` rt API should work.

xds-client/src/resource/mod.rs
YutaoMa: Makes sense. I don't think `Debug` or `Clone` is needed universally so I removed both bounds.

xds-client/src/resource/mod.rs
YutaoMa: Added `TypeUrl` wrapper type. Will make notes of the ecosystem-worthy APIs.

xds-client/src/client/mod.rs
YutaoMa: I now have a setup in mind with default generic type set to tonic stack. But since I don't want to create `TonicTransport` and `TokioRuntime` in this PR, I'm going to do that change in the next one.

YutaoMa: Added a TODO for now.

gu0keno0 (approved):

xds-client/src/transport/mod.rs
```diff
@@ -1,4 +1,50 @@
...
+#[derive(Debug, Clone)]
+pub struct DiscoveryRequest {
+    // TODO: Add fields as needed
+}
+
+/// A discovery response from the xDS server.
+#[derive(Debug, Clone)]
+pub struct DiscoveryResponse {
+    // TODO: Add fields as needed
+}
```
dfawley, line 19: Is it possible to leave the transport as a byte transport, or does that not work with tonic so well?

The main design idea is that it's able to send generic messages (but not have to say "these are 'proto messages'" since "what a proto message is" will likely change from prost to google-protobuf).  So we serialize inside the xdsclient and send the serialized bytes to the transport.  It's still transmitted as proto, it just doesn't need to do the serialization in tonic.

xds-client/src/transport/mod.rs
```diff
@@ -1,4 +1,50 @@
...
+/// - Mock transport for testing
+/// - Other custom transports
+pub trait Transport: Send + Sync + 'static {
+    /// The stream type produced by this transport.
+    type Stream: TransportStream;
+
+    /// Creates a new bidirectional ADS stream to the xDS server.
+    ///
+    /// This may be called multiple times for reconnection.
+    fn connect(&self) -> impl Future<Output = Result<Self::Stream>> + Send;
```
dfawley: Could you please rename this?  Maybe to `new_stream` or `create_stream`?  "Connect" implies there's a new TCP connection being made, but this just creates a stream.

xds-client/src/client/watch.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// ```ignore
+    /// while let Some(event) = watcher.next().await {
+    ///     match event {
+    ///         ResourceEvent::ResourceChanged(resource) => { /* handle */}
+    ///         ResourceEvent::ResourceError(error) => { /* handle */}
+    ///         ResourceEvent::AmbientError(error) => { /* handle */}
+    ///     }
+    /// }
+    /// ```
+    pub async fn next(&mut self) -> Option<ResourceEvent<T>> {
```
dfawley, line 134: The other gRPC versions of this have a callback that is used to indicate the returned result has "finished being processed".  Generally, what that means is: it's added a watch for all the related resources it think it will need.  IIRC this was useful to batch these subscription requests as we block all ADS stream operations while calling the watchers.  In Go we struggled for awhile with whether we should just say "the function blocks until processing is done" or "it's useful to pass this closure elsewhere to allow all watchers to be doing some async processing at the same time" and we eventually decided on the latter.  In Rust this could be a oneshot channel or some equivalent that is just a signal that something is finished.

https://pkg.go.dev/google.golang.org/grpc@v1.77.0/internal/xds/clients/xdsclient#ResourceWatcher

xds-client/src/client/mod.rs
```diff
@@ -1 +1,73 @@
...
+    ///         ResourceEvent::AmbientError(error) => {
+    ///             println!("Ambient error: {}", error);
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    pub fn watch<T: Resource>(&self, _name: impl Into<String>) -> ResourceWatcher<T> {
+        todo!()
+    }
+}
```
dfawley, line 73: We will also need the `dump_resources` method eventually in order to feed CSDS for debugging (https://pkg.go.dev/google.golang.org/grpc@v1.77.0/internal/xds/clients/xdsclient#XDSClient.DumpResources).

xds-client/src/transport/mod.rs
YutaoMa: Definitely doable and a good idea, I can use a pattern similar to the `BytesCodec` in `grpc/codec.rs`.
I'll make these Request/Response types wrapper for bytes.

xds-client/src/transport/mod.rs
YutaoMa: Makes sense, updated

xds-client/src/client/mod.rs
YutaoMa: Noted, will add in following PRs

xds-client/src/client/watch.rs
YutaoMa: Good point, I added a oneshot channel based signaling mechanism accordingly, let me know if that matches your expectation

xds-client/src/transport/mod.rs
YutaoMa: Noted for potential future simplification

PR #2465, opened 2025-12-15
Separate channel creation and state probe, and error handling thereof.

nathanielford:
## Motivation

Currently the `Channel` object is calling into the `PersistentChannel` internals to do things like provision an active channel and check state. The division of where work and error handling is managed becomes muddied in this situation, which complicates handling errors in ways that aren't `unwrap`. 

This was also motivated by work to implement the always-failing PersistentChannel PR, but it seemed like it could be broken off into it's own PR.

## Solution

Moving the actual mechanics into the `PersistentChannel` object, which owns channel config, and allows `Channel` to be a lightweight handles for users to interact with. Specifically state probing and channel creation are now in `PersistentChannel`.

`PersistentChannel` calls are wrapped in `Result` objects that are translated at the `Channel` level to the appropriate public-facing API objects.  `get_or_create_active_channel()` just becomes `get_active_channel()` and passes the 'create' flag, which controls whether a new channel is created if it does not already exist.

The external semantics should stay the same as they have been.

This does not yet address the error that can surface in `call()`, because it's not yet clear what the error reporting story is.

### DIFF
--- a/grpc/src/client/channel.rs
+++ b/grpc/src/client/channel.rs
@@ -188,8 +188,8 @@ struct PersistentChannel {
 }
 
 impl PersistentChannel {
-    // Channels begin idle so new does not automatically connect.
-    // ChannelOptions are only non-required parameters.
+    // Channels begin idle so `new()` does not automatically connect.
+    // ChannelOption contain only optional parameters.
     fn new(
         target: &str,
         _credentials: Option<Box<dyn Credentials>>,


grpc/src/client/channel.rs
```diff
@@ -154,23 +154,14 @@ impl Channel {
...
-            }
-            ac.as_ref().unwrap().clone()
-        } else {
-            // Otherwise, get or create the active channel.
-            self.get_or_create_active_channel()
-        };
-        if let Some(s) = ac.connectivity_state.cur() {
-            return s;
+        let state = self.inner.state(connect);
+        match state {
```
dfawley: Wouldn't we want the two state() methods to have the same return value so this could simply delegate?

grpc/src/client/channel.rs
```diff
@@ -228,6 +207,47 @@ impl PersistentChannel {
...
+    /// Gets the underlying active channel. If `connect` is true, will create a new channel iff
+    /// there is no active channel.
+    fn get_active_channel(
+        &self,
+        connect: bool,
+    ) -> Result<Arc<ActiveChannel>, Box<dyn std::error::Error + Sync + Send>> {
+        let mut s = self
+            .active_channel
+            .lock()
+            .map_err(|_| "Could not get channel lock.".to_string())?;
```
dfawley: I *think* this can stay as unwrap().  IIUC locks can only fail in pretty spectacular ways (someone holding the lock panicked).  In that case we might as well panic too.

grpc/src/client/channel.rs
```diff
@@ -228,6 +207,47 @@ impl PersistentChannel {
...
+                    self.runtime.clone(),
+                ));
+            } else {
+                return Err("No active channel.".into());
+            }
+        }
+
+        s.as_ref()
+            .cloned()
+            .ok_or_else(|| "Could not clone channel".into())
```
dfawley: By here you know s is `Some`, so you can do `s.unwrap().clone()` I think.

Or if you do a `if let Some(ac) = s else {}` instead of `if s.is_none()` then it will already be unwrapped.

grpc/src/client/channel.rs
```diff
@@ -228,6 +207,47 @@ impl PersistentChannel {
...
+        return Ok(ConnectivityState::Idle);
+    }
+
+    /// Gets the underlying active channel. If `connect` is true, will create a new channel iff
+    /// there is no active channel.
+    fn get_active_channel(
+        &self,
+        connect: bool,
+    ) -> Result<Arc<ActiveChannel>, Box<dyn std::error::Error + Sync + Send>> {
+        let mut s = self
```
dfawley: Maybe name this `ac` or `active_channel`?  I'm not sure what `s` represents.

grpc/src/client/channel.rs
```diff
@@ -228,6 +207,47 @@ impl PersistentChannel {
...
+
+        if s.is_none() {
+            if connect {
+                *s = Some(ActiveChannel::new(
+                    self.target.clone(),
+                    &self.options,
+                    self.runtime.clone(),
+                ));
+            } else {
+                return Err("No active channel.".into());
```
dfawley: This function can probably return just an Option?

Also maybe we shouldn't combine these two functionalities into one function.  They're a little disjoint.

- Anyone asking for the active channel itself will _always_ want connect=true.
- Only when requesting the current state, exiting idle mode is optional.

grpc/src/client/channel.rs
nathanielford: Done!

grpc/src/client/channel.rs
nathanielford: I went back and forth on this, but ultimately decided to leave it as `unwrap()` on the theory that we really do want lock poisoning to cause an all-stop, and fix whatever causes the poisoning rather than allow apps to recover from it. I'm only _vaguely_ uncomfortable with leaving the unwrap in there.

grpc/src/client/channel.rs
nathanielford: Without needing to be clear that the error is a lock poisoning, the option makes the most sense, so I changed it to this.

nathanielford: I also removed the bool flag after going back and forth with various options. After being satisfied that I could arrange the deck chairs any which way settled on the current version. LMK what you think!

grpc/src/client/channel.rs
```diff
@@ -213,8 +188,8 @@ struct PersistentChannel {
 }
 
 impl PersistentChannel {
-    // Channels begin idle so new is a simple constructor.  Required parameters
-    // are not in ChannelOptions.
+    // Channels begin idle so new does not automatically connect.
+    // ChannelOptions are only non-required parameters.
```
dfawley: Nit:

```suggestion
    // ChannelOptions contains only optional parameters.
```

grpc/src/client/channel.rs
```diff
@@ -228,6 +203,49 @@ impl PersistentChannel {
...
+    }
+
+    /// Returns the current state of the channel. If there is no underlying active channel,
+    /// returns Idle. If `connect` is true, will create a new active channel iff none exists.
+    fn state(&self, connect: bool) -> ConnectivityState {
+        // Done this away to avoid potentially locking twice.
+        let active_channel = if connect {
+            self.get_active_channel()
+        } else {
+            match self.locked_active_channel().as_ref().cloned() {
```
dfawley: `as_ref().cloned()` -> `clone()` ?  (And below)

grpc/src/client/channel.rs
```diff
@@ -228,6 +203,49 @@ impl PersistentChannel {
             runtime,
         }
     }
+
+    // Internal use only to get the locked active channel. If this panics it means the lock is
+    // poisoned and we should also panic.
+    fn locked_active_channel(&self) -> std::sync::MutexGuard<'_, Option<Arc<ActiveChannel>>> {
+        self.active_channel.lock().unwrap()
```
dfawley: I am not sure having a method for this one liner is helpful, and in fact makes the call sites more opaque which is probably a bad thing.

grpc/src/client/channel.rs
```diff
@@ -154,23 +154,10 @@ impl Channel {
 
     // TODO: enter_idle(&self) and graceful_stop()?
 
-    /// Returns the current state of the channel.
+    /// Returns the current state of the channel. Any errors translate into a
+    /// TransientFailure state.
```
dfawley: I'm not sure what the "translate" comment is about.  Does that mean there could be an error while simply retrieving the state?  I don't _think_ this is possible or appropriate...this method should just tell me the current state of the channel, period.

grpc/src/client/channel.rs
nathanielford: The only reason I thought it was worth it was to try and put the unsafe unwrap in one place. But I'll revert it, since I think it's probably not that big a deal.

grpc/src/client/channel.rs
nathanielford: This should be done as well - apologies if it got clobbered while I was fixing my git flow.

grpc/src/client/channel.rs
nathanielford: Fixed!

grpc/src/client/channel.rs
```diff
@@ -154,23 +154,10 @@ impl Channel {
 
     // TODO: enter_idle(&self) and graceful_stop()?
 
-    /// Returns the current state of the channel.
+    /// Returns the current state of the channel. If there is no underlying active channel,
+    /// returns Idle. If `connect` is true, will create a new active channel.
```
dfawley, line 158: This leaks implementation details into the documentation.  Users shouldn't know about our active channel at all.  How about something like:

```suggestion
    /// Returns the current state of the channel. If `connect` is true, will begin connecting.
```

PR #2467, opened 2025-12-19
feat(tonic-build): rpit / async_trait configuration

tomDev5:
## Motivation

#2283 

## Solution

Essentially used the modifications to `server.rs` from the original PR, but under a config flag (the default is still async_trait).

Tests will be added shortly, creating this to get some feedback on the code (I currently plan on asserting that the generated content from some proto file equals exactly what I expect, but other ideas are welcome).


nathanielford:
Thanks for working on this! Active development on legacy Tonic is winding down, and the new official `grpc-rust` library (`protoc-gen-rust-grpc`) natively supports modern async traits without `#[async_trait]` boxing. Because maintaining a dual code generation pipeline in `tonic-build` adds substantial complexity during this transition, we won't be able to land this. We really appreciate the effort and look forward to your feedback on the new library. Let us know if you have any outstanding concerns. Thanks again!

Issue #2388, opened 2025-08-18
`super::bool` not found in `super`


pronebird:
### Version

```
│   │   ├── tonic v0.14.1
│   ├── tonic v0.14.1 (*)
│   ├── tonic-prost v0.14.1
│   │   └── tonic v0.14.1 (*)
│   └── tonic-prost-build v0.14.1
│       └── tonic-build v0.14.1
├── tonic v0.14.1 (*)
```

### Platform

```
Darwin foxmulder.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Dec 19 20:47:53 PST 2024; root:xnu-10063.141.1.703.2~1/RELEASE_ARM64_T6030 arm64
```

### Description

```rs
pub mod proto { 
  tonic::include_proto!("myproto") 
}
``` 

Lotta compilation errors related to referencing boolean types via `super::bool` which errors until I add `use std::primitive::bool` into the `mod`

pronebird:
Maybe pull https://github.com/nymtech/nym-vpn-client/commit/289acb83f61a48d3777da37ab8026fda3b846fa6 and try `cargo build -p nym-vpn-proto`. See what `tonic::include_proto!` expands to.

zhaohaiyun9:
@LucioFranco Hello! I encounter this problem too. It's simple to reproduce it. In the examples/proto/helloworld/helloworld.proto, add a method using google well known wrapper types:

```
import "google/protobuf/empty.proto";

// The greeting service definition.
service Greeter {
  // Sends a greeting
  rpc SayHello (HelloRequest) returns (HelloReply) {}

  rpc Echo (google.protobuf.BoolValue) returns (google.protobuf.BoolValue) {}
}

```

In the generated rust code, google.protobuf.BoolValue is converted to "super::bool".

I think the fix is in tonic-prost-build/src/lib.rs, as follows:

```
diff --git a/tonic-prost-build/src/lib.rs b/tonic-prost-build/src/lib.rs
index 3fcc9ac..7374502 100644
--- a/tonic-prost-build/src/lib.rs
+++ b/tonic-prost-build/src/lib.rs
@@ -108,7 +108,7 @@ pub fn compile_fds(fds: prost_types::FileDescriptorSet) -> io::Result<()> {
 }

 /// Non-path Rust types allowed for request/response types.
-const NON_PATH_TYPE_ALLOWLIST: &[&str] = &["()"];
[...]

PR #2471, opened 2025-12-30
fix(prost-build): Map wrapper types to primitives

jen20:
## Motivation

This commit address the issue raised in #2388, where well-known types such as `google.protobuf.BoolValue` would appear as `super::bool` rather than the correct primitives.

The issue was reproduced [here](https://github.com/jen20/tonic-repros/blob/main/tonic-2388/proto/helloworld.proto#L9), and produces the following output when `cargo build` is run (only the first occurrence is shown):

```
error[E0412]: cannot find type `bool` in module `super`
  --> /Users/James/Code/personal/tonic-repros/tonic-2388/target/debug/build/tonic-2388-685c45971c3ed332/out/helloworld.rs:95:53
   |
95 |             request: impl tonic::IntoRequest<super::bool>,
   |                                                     ^^^^ not found in `super`
   |
help: consider importing this builtin type
   |
11 +     use std::primitive::bool;
   |
help: if you import `bool`, refer to it directly
   |
95 -             request: impl tonic::IntoRequest<super::bool>,
95 +             request: impl tonic::IntoRequest<bool>,
   |
```

## Solution

We address this by adding new cases to the existing enum for the various wrapper types contained in `google/protobuf/wrappers.proto`, when not compiling well-known types as proto structs directly. This is broadly in line with the patch suggested in the issue by @zhaohaiyun9, but is a little less broad and adds test coverage.

### DIFF
--- a/tonic-prost-build/src/lib.rs
+++ b/tonic-prost-build/src/lib.rs
@@ -196,9 +196,17 @@ impl tonic_build::Method for TonicBuildMethod {
         {
             // For well-known types, map to absolute paths that will work with super::
             match self.prost_method.input_type.as_str() {
-                ".google.protobuf.Empty" => quote!(()),
                 ".google.protobuf.Any" => quote!(::prost_types::Any),
+                ".google.protobuf.BoolValue" => quote!(bool),
+                ".google.protobuf.BytesValue" => quote!(::prost::bytes::Bytes),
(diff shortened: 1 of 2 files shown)


nathanielford:
It looks like this PR was superceded by #2544. Going to close it, but let me know if there is an outstanding concern! Thank you!

PR #2472, opened 2025-12-31
tonic: expose tcp_keepalive_interval and tcp_keepalive_retries options

kevinWangSheng:
Add two new configuration methods to the Server builder:
  - `tcp_keepalive_interval()`: Sets TCP_KEEPINTVL socket option
  - `tcp_keepalive_retries()`: Sets TCP_KEEPCNT socket option

  These options were already implemented in TcpIncoming but not exposed through the Server builder API. This change allows users to fine-tune TCP keepalive behavior for scenarios like:
  - Cloud load balancer idle timeout matching
  - NAT traversal requirements
  - Faster dead connection detection

  The new options default to None (system default) and only take effect when tcp_keepalive() is also configured.

  ## Motivation

  `TcpIncoming` already supports `with_keepalive_interval()` and `with_keepalive_retries()`, but these are not exposed through the `Server` builder. Users need fine-grained control over TCP keepalive for cloud LB timeout matching and faster dead connection detection.

  ## Solution

  Add `tcp_keepalive_interval()` and `tcp_keepalive_retries()` methods to `Server` builder, passing them through to `TcpIncoming` in `bind_incoming()`. Both default to `None` and only take effect when `tcp_keepalive()` is also set. Tests updated accordingly.

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -98,6 +98,8 @@ pub struct Server<L = Identity> {
     init_connection_window_size: Option<u32>,
     max_concurrent_streams: Option<u32>,
     tcp_keepalive: Option<Duration>,
+    tcp_keepalive_interval: Option<Duration>,
+    tcp_keepalive_retries: Option<u32>,
     tcp_nodelay: bool,
     http2_keepalive_interval: Option<Duration>,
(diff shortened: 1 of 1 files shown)


tottoto (approved):
Thanks!

PR #2475, opened 2026-01-06
feat(xds): implement tonic-based transport and prost-based codec

YutaoMa:
## Motivation

Ref: #2444 

The transport layer in xDS client is responsible for sending `DiscoveryRequest` to xDS management server, and receiving `DiscoveryResponse` back.

## Solution

This PR implements tonic-based transport and prost-based codec. As discussed in #2464, the transport layer sends and receives bytes and leaves (de)serialization to the worker with the help of codegen-specific codecs. Specifically,

tonic-based transport is responsible for:
1. connecting to xDS management server over tonic transport channel.
2. send/receive gRPC requests/responses as serialized bytes.

prost-based codec is responsible for:
1. (de)serialize between bytes and prost-generated xDS binding types.
2. convert between prost-generated xDS binding types to `xds-client`'s codegen-agnostic message types. This allows the worker logic we are going to write later be codegen-agnostic.

Note regarding the codec: String fields in `DiscoveryRequest` are mostly cloned around instead of borrowed, these are intentional to reduce complexity of lifetime. xDS subscription requests are infrequent, small, and typically on non-critical path, so the allocation cost here should be negligible compared to network latency.

## Next Steps

Once this PR is merged, the next step to complete `xds-client` would be to wire the transport, codec and client interface together by implementing workers, i.e. event loops that manage ADS connection and xDS subscriptions. I've added a skeleton for worker in this PR to demonstrate how that will look like at a high level.

Note regarding ADS stream connection: we'll add retry with backoff and jitter similar to `grpc-go`'s xDS client implementation in subsequent PR. That is handled at worker layer, so it shouldn't impact the interface we have here in the transport layer.

### DIFF
--- a/xds-client/src/transport/mod.rs
+++ b/xds-client/src/transport/mod.rs
@@ -7,6 +7,10 @@ use std::future::Future;
 #[cfg(feature = "transport-tonic")]
 pub mod tonic;
 
+mod sealed {
+    pub trait Sealed {}
+}
+
 /// Factory for creating xDS transport streams.
 ///
 /// This abstraction allows for different transport implementations:
@@ -28,7 +32,8 @@ pub trait Transport: Send + Sync + 'static {
 ///
 /// Raw byte transport where the bytes are serialized DiscoveryRequest/DiscoveryResponse
 /// (de)serialization is handled at the xDS client worker layer.
-pub trait TransportStream: Send + 'static {
+// Sealed for now to limit API surface.
+pub trait TransportStream: sealed::Sealed + Send + 'static {
     /// Send serialized DiscoveryRequest bytes to the server.
     fn send(&mut self, request: Bytes) -> impl Future<Output = Result<()>> + Send;
 
@@ -40,3 +45,6 @@ pub trait TransportStream: Send + 'static {
     /// - `Err(_)` - Stream error (connection dropped, etc.)
     fn recv(&mut self) -> impl Future<Output = Result<Option<Bytes>>> + Send;
 }
+
+#[cfg(feature = "transport-tonic")]
+impl sealed::Sealed for tonic::TonicAdsStream {}
--- a/xds-client/src/transport/tonic.rs
+++ b/xds-client/src/transport/tonic.rs
@@ -18,6 +18,8 @@ use tonic::{Status, Streaming};
 const ADS_PATH: &str =
     "/envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources";
 
+const ADS_CHANNEL_BUFFER_SIZE: usize = 16;
+
 /// A codec that passes bytes through without serialization.
 ///
 /// This allows us to handle serialization in the xDS client layer
@@ -101,7 +103,7 @@ impl Transport for TonicTransport {
             .await
             .map_err(|e| Error::Connection(e.to_string()))?;
 
-        let (tx, rx) = mpsc::channel::<Bytes>(16);
+        let (tx, rx) = mpsc::channel::<Bytes>(ADS_CHANNEL_BUFFER_SIZE);
         let request_stream = tokio_stream::wrappers::ReceiverStream::new(rx);
 
         let path = PathAndQuery::from_static(ADS_PATH);


xds-client/src/codec/prost.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::error::{Error, Result};
+use crate::message::{DiscoveryRequest, DiscoveryResponse, ResourceAny};
+use bytes::Bytes;
+use prost::Message;
+
+/// A codec that uses prost/envoy-types for serialization.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct ProstCodec;
+
+impl XdsCodec for ProstCodec {
```
LucioFranco, line 13: I am forgetting but can you remind me what this codec is for? And why we can't use the tonic one? Or is this just an unfortunate name collision and this codec has nothing to do with the grpc encoding? From what I can tell this is related to the stuff later in transport, it would be really nice if yall could do a very short form write up about the usage of the low level grpc client since ive not yet actually seen someone do that yet

xds-client/src/transport/mod.rs
```diff
@@ -35,16 +24,19 @@ pub trait Transport: Send + Sync + 'static {
     fn new_stream(&self) -> impl Future<Output = Result<Self::Stream>> + Send;
 }
 
-/// A bidirectional stream for xDS ADS communication.
+/// A bidirectional byte stream for xDS ADS communication.
+///
+/// Raw byte transport where the bytes are serialized DiscoveryRequest/DiscoveryResponse
+/// (de)serialization is handled at the xDS client worker layer.
 pub trait TransportStream: Send + 'static {
```
LucioFranco: I think this should be sealed for now just for api surface area saftey

xds-client/src/error.rs
```diff
@@ -5,7 +5,23 @@ use thiserror::Error;
...
 pub enum Error {
-    // TODO: Add error variants as needed
+    /// Failed to connect to the xDS server.
+    #[error("failed to connect: {0}")]
+    Connection(String),
+
+    /// Error on the ADS stream.
+    #[cfg(feature = "transport-tonic")]
+    #[error("stream error: {0}")]
+    Stream(#[from] tonic::Status),
```
LucioFranco, line 15: my gut feeling is that this isnt always a stream error and this scoped down error could be confusing, not a big blocker at all right now just a gut check

xds-client/src/transport/tonic.rs
```diff
@@ -1 +1,250 @@
...
+
+/// Factory for creating ADS streams using tonic.
+#[derive(Clone, Debug)]
+pub struct TonicTransport {
+    channel: Channel,
+}
+
+impl TonicTransport {
+    /// Connect to an xDS server.
+    pub async fn connect(uri: impl Into<String>) -> Result<Self> {
```
gu0keno0, line 85: For large-scale use cases, the application can attempt to connect to many xds:// targets. According to the gRPC StoW xDS protocol, each target should own its own xDS stream; however, I think it would be preferrable for these streams to share one / one pool of Tonic channels, this will reduce the connection overhead on the xDS management server.

Can we maybe make ChannelProvider a parameter? I think for the initial implementation, we only need a single Tonic channel for all the streams, the gRPC multiplexing should be able to handle the xDS subscription load .

xds-client/src/transport/tonic.rs
```diff
@@ -1 +1,250 @@
...
+    type Stream = TonicAdsStream;
+
+    async fn new_stream(&self) -> Result<Self::Stream> {
+        let mut grpc = Grpc::new(self.channel.clone());
+
+        grpc.ready()
+            .await
+            .map_err(|e| Error::Connection(e.to_string()))?;
+
+        let (tx, rx) = mpsc::channel::<Bytes>(16);
```
gu0keno0: Nit: make 16 a const?

xds-client/src/client/worker.rs
```diff
@@ -0,0 +1,116 @@
...
+        todo!("implement subscription tracking")
+    }
+
+    /// Run the worker event loop.
+    pub async fn run(&mut self) -> Result<()> {
+        // TODO: Implement the main event loop that:
+        // - Receives responses from the stream
+        // - Dispatches resources to watchers
+        // - Sends ACK/NACK
+        // - Handles reconnection
```
gu0keno0, line 113: How do you plan to handle reconnections since stream is already passed-in from new()?

xds-client/src/client/worker.rs
```diff
@@ -0,0 +1,116 @@
...
+            version_info,
+            response_nonce,
+            error_detail: None,
+        };
+
+        self.send_request(request).await
+    }
+
+    /// Send ACK for a received response.
+    pub async fn ack(&mut self, response: &DiscoveryResponse) -> Result<()> {
```
gu0keno0, line 78: Need nack as well

xds-client/src/codec/prost.rs
gu0keno0: Good question, @LucioFranco !

I had a discussion about this with @YutaoMa yesterday , IIUC the intention is to let callers of xds-client to use xds_client::DiscoveryRequest and xds_client::DiscoveryResponse, which are 1:1 mapped xDS' DiscoveryRequest and DiscoveryResponse but are created as native Rust structs to allow using different protobuf implementations.

So then here we are first converting between native Rust structs of DiscoveryRequest and DiscoveryResponse and the prost codegen types of Envoy xDS equivalences, then use prost to do the codec.

Please advise if you have a better path.

xds-client/src/client/worker.rs
YutaoMa: Noted, will add when I get to worker

xds-client/src/client/worker.rs
YutaoMa: Now you mention it, I'll need to change the worker to take in the factory ("TonicTransport") instead of the stream.

xds-client/src/transport/tonic.rs
YutaoMa: 👍

xds-client/src/transport/tonic.rs
YutaoMa: Makes sense, I'll implement this this channel sharing

xds-client/src/transport/mod.rs
YutaoMa: Got it, yeah that makes sense for all the transport related traits here, I can seal those

xds-client/src/message.rs
```diff
@@ -0,0 +1,79 @@
...
+    pub type_url: String,
+    /// Serialized resource bytes.
+    pub value: Bytes,
+}
+
+/// Node identification for the client.
+#[derive(Debug, Clone, Default)]
+pub struct Node {
+    /// An opaque node identifier.
+    pub id: String,
```
ankurmittal, line 54: what is this id used for? who will create this type? maybe id can be a simple uuid? or a static &str

ankurmittal: oh this comes from proto, ignore this comment

xds-client/src/codec/prost.rs
YutaoMa: @LucioFranco this is indeed an unfortunate name collision, this is for converting between codegen-specific wire format and codegen-agnostic concrete types so that later when we write worker logic to create these Request/Response, we don't have to worry about codegen at all. I can see now how calling it codec could cause confusion with tonic transport codec, perhaps I'd better rename these?

xds-client/src/error.rs
YutaoMa: The current only use site is the stream error in TonicTransport's ADS stream. Is the concern that because this wraps a generic gRPC status that it can be misused later to wrap other gRPC errors?

PR #2478, opened 2026-01-12
feat(xds): implement xDS subscription worker

YutaoMa:
## Motivation

Ref: #2444 

With #2475 transport and codec change merged, the remaining change required to get the xDS workflow work end to end is to wire them together with `XdsClient` through a worker loop. This PR implements that.

## Solution

1. Implement `AdsWorker`, a transport/runtime/codegen-agnostic event loop for managing xDS subscriptions and ADS stream.
  - The worker conceptually manages a pair of mpsc channel, where the sender is used by `XdsClient` to send subscription requests and the receiver is used by `TransportStream` to send `DiscoveryRequest` to xDS servers.
  - When the underlying ADS stream closes, retry with exponential backoff is supported. Configurable via `ClientConfig`.
3. Implement `ResourceWatcher` and wire it with `XdsClient` so user can now subscribes to xDS resources.

Some design choice highlights:

1. Created a `DecodedResource` that is a type-erased representation of xDS resource with its decoding function carried in a closure. AdsWorker sends and receives this type on channels so it can stay transport and codec generic.
2. The ADS stream connection waits for the first subscription from the user. This is because `tonic`'s gRPC stream `::connect()` awaits for the response headers. Depending on the xDS server implementation, it may not respond back with headers until the first subscription, creating a deadlock if we await for the stream creation before sending any requests. (Btw `grpc-go` works around this by having send/recv in different go routines, here I kept both in the same worker loop to reduce the shared state complexity)

## Testing

Created a `basic.rs` example to showcase the user experience. I've used it to test against a local xDS management server and successfully subscribed to multiple Listener resources.

```
=== xds-client Example ===

Connecting to xDS server: https://[redacted-private-server]
Connected!

Enter listener names to watch (one per line, Ctrl+C to exit):
(Use empty string for wildcard subscription)

[redacted-listener-1]
→ Watching for Listener: '[redacted-listener-1]'
✓ Listener received:
  name:        [redacted-listener-1]
  rds_config:  [redacted-route-1]

[redacted-listener-2]
→ Watching for Listener: '[redacted-listener-2]'
✓ Listener received:
  name:       [redacted-listener-2]
  rds_config:  [redacted-route-2]

✓ Listener received:
  name:        [redacted-listener-1]
  rds_config:  [redacted-route-1]
```

## Next Steps

The current implementation completes the basic functionality end to end, but have these improvements opportunities:

1. `xds-client` currently do not cache received resources. This means new watchers to a subscribed resource need to wait for the next response from xDS server. If we add in resource caching, new watchers get instant response. 
4. If we bring in a proper xDS server implementation we can run integration test in CI.
5. Observability. We'll bring in tracing, logging and metrics support.
6. Delta xDS. The current version implements State of the World variant of xDS only, but in some use cases delta variant is more efficient and scalable.

Additionally, these features are xDS-related but are also gRPC-specific so we have plan to implement them in a separate `tonic-xds` crate, using this `xds-client` crate:

1. Cascadingly subscribes to RDS, CDS, and EDS resources from a top-level Listener, for gRPC routing and load balancing.
2. xDS server connection bootstrapping, this includes TLS configuration, connecting pooling, multi-server fallback, etc.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,13 +9,13 @@ members = [
   "tonic-reflection",
   "tonic-prost",
   "tonic-prost-build",
-  "tonic-web", # Non-published crates
+  "tonic-web",
   "examples",
   "codegen",
   "grpc",
   "xds-client",
   "tonic-xds",
-  "interop", # Tests
+  "interop",
   "tests/disable_comments",
   "tests/wellknown",
   "tests/wellknown-compiled",


xds-client/src/client/watch.rs
```diff
@@ -101,14 +109,34 @@ pub enum ResourceEvent<T> {
 /// Dropping the watcher unsubscribes from the resource.
 #[derive(Debug)]
 pub struct ResourceWatcher<T: Resource> {
-    // TODO: replace with proper implementation
-    _marker: std::marker::PhantomData<T>,
+    /// Channel to receive events from the worker.
+    event_rx: mpsc::UnboundedReceiver<ResourceEvent<DecodedResource>>,
```
ankurmittal: does this need to be unbounded? can we bound these senders?

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,849 @@
...
+use crate::message::{DiscoveryRequest, DiscoveryResponse, ErrorDetail, Node};
+use crate::resource::{DecodedResource, DecoderFn};
+use crate::runtime::Runtime;
+use crate::transport::{Transport, TransportStream};
+
+/// Unique identifier for a watcher.
+///
+/// Generated by an atomic counter in `XdsClient`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct WatcherId(pub(crate) u64);
```
ankurmittal: why does it need to be pub(crate), can anyone else in the crate change this value? if not we can create a internal raw_id method to read it?

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,180 @@
...
+use prost::Message;
+use tokio::io::{AsyncBufReadExt, BufReader};
+
+use xds_client::resource::TypeUrl;
+use xds_client::{
+    ClientConfig, ProstCodec, Resource, ResourceEvent, TokioRuntime, TonicTransport, XdsClient,
+};
+
+/// URI of your xDS management server.
+const XDS_SERVER_URI: &str = "http://localhost:18000";
```
gu0keno0: maybe use clap with env variable support?

Otherwise users of this example will have to change the const.

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// The RDS route config name (from HttpConnectionManager).
+    pub rds_route_config_name: Option<String>,
+}
+
+impl Resource for Listener {
+    const TYPE_URL: TypeUrl = TypeUrl::new("type.googleapis.com/envoy.config.listener.v3.Listener");
+
+    fn decode(bytes: Bytes) -> xds_client::Result<Self> {
+        use envoy_types::pb::envoy::config::listener::v3::Listener as ListenerProto;
+        use envoy_types::pb::envoy::extensions::filters::network::http_connection_manager::v3::{
```
gu0keno0: Nit: move them to the top?

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        while let Ok(Some(line)) = lines.next_line().await {
+            let name = line.trim().to_string();
+            if name.is_empty() {
+                continue;
+            }
+
+            println!("→ Watching for Listener: '{name}'");
+
+            let mut watcher = client_clone.watch::<Listener>(&name);
```
gu0keno0: The basic example is probably too specific about one resource (Listener).

As a more generic idea, maybe get inputs such as Listener::ListenerName, Cluster::ClusterName, etc, and just do generic subscriptions?

xds-client/src/client/mod.rs
```diff
@@ -1,74 +1,160 @@
...
+            command_rx,
+        );
+
+        self.runtime.spawn(async move {
+            worker.run().await;
+        });
+
+        XdsClient {
+            command_tx,
+            next_watcher_id: Arc::new(AtomicU64::new(0)),
```
gu0keno0: How relevant is watcher_id being an increasing integer? Maybe just use uuid instead? (Stateless)

xds-client/src/client/watch.rs
```diff
@@ -132,6 +160,32 @@ impl<T: Resource> ResourceWatcher<T> {
...
     /// ```
     pub async fn next(&mut self) -> Option<ResourceEvent<T>> {
-        todo!()
+        let event = self.event_rx.next().await?;
+
+        Some(match event {
+            ResourceEvent::ResourceChanged { resource, done } => {
+                let typed_resource = resource
+                    .downcast::<T>()
+                    .expect("resource type mismatch - this is a bug in xds-client");
```
gu0keno0: Just return an error?

The scenario I'm considering is that management server's bug should not crash clients.

xds-client/src/error.rs
```diff
@@ -22,6 +22,19 @@ pub enum Error {
...
     #[error("decode error: {0}")]
     Decode(#[from] prost::DecodeError),
+
+    /// Resource validation failed.
+    #[error("resource validation failed: {0}")]
+    Validation(String),
+
+    /// Resource does not exist on the server.
+    #[error("resource does not exist: {type_url}/{name}")]
+    ResourceDoesNotExist {
```
gu0keno0: Should we rename this to ResourceTimeout instead?

It might be confusing to return & log "not exist" if it's caused by xds management server being overloaded, because technically the resource exists, just that xds server cannot return it properly.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,849 @@
...
+        }
+    }
+}
+
+/// Per-type_url state tracking.
+struct TypeState {
+    /// Decoder function for this resource type.
+    decoder: DecoderFn,
+    /// Version from last successful response.
+    version_info: String,
```
gu0keno0, line 196: this is true for LDS and CDS;

for RDS and EDS I believe the resource version is tied to the resource itself: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#four-variants

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,849 @@
...
-    // TODO: Add version/nonce tracking
-    // versions: HashMap<String, String>,  // type_url -> version_info
-    // nonces: HashMap<String, String>,    // type_url -> response_nonce
+    /// Worker configuration.
+    config: WorkerConfig,
+
+    /// Receiver for commands from XdsClient.
+    command_rx: mpsc::UnboundedReceiver<WorkerCommand>,
+    /// Per-type_url state.
+    type_states: HashMap<String, TypeState>,
```
gu0keno0, line 341: Seem my previous comments about resource versions for StoW.

I tend to think that it might be better to just tie version to each resource name directly. For LDS and CDS, we can just duplicate the versions for different resources.

It will make it easier to support Delta in the future.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,849 @@
...
+    /// Sender for internal timer events.
+    timer_tx: mpsc::UnboundedSender<TimerEvent>,
+    /// Receiver for internal timer events.
+    timer_rx: mpsc::UnboundedReceiver<TimerEvent>,
```
gu0keno0: I'm wondering if we can simplify the client by not dealing with timeouts for now.

Reason is that caller to xds client (tonic-xds) can deal with subscription timeouts and log them.

xds-client/src/error.rs
YutaoMa: Makes sense, I went thru [gRFC A88](https://github.com/grpc/proposal/blob/master/A88-xds-data-error-handling.md) and learned about the change from previous Not Exist Error to a transient Timeout status, so this proposed change would align with later version of gRPC. I'll make this change.

xds-client/src/client/mod.rs
YutaoMa: uuid also works. I went with atomic int for code simplicity, but agree that being stateless also has its benefit.

xds-client/src/client/watch.rs
YutaoMa: This would not trigger under server bug, but would if `xds-client` worker impl stored wrong decoder function, or when two different Resource with the same type_url are present. Now I think of it, given the Resource conflict case can be caused by user implementation, it would be better to not panic here. I'll change to an error variant instead.

xds-client/src/client/watch.rs
YutaoMa: Doesn't need to, went with unbounded to not worry about handling buffer size. I can change to bounded one if that's a concern.

xds-client/src/client/worker.rs
YutaoMa: sure, I can change to an immutable approach.

xds-client/examples/basic.rs
YutaoMa: 👍

xds-client/examples/basic.rs
YutaoMa: Adding a new type of subscription requires adding the `Resource` impl, for now I'd like to keep the example as basic as possible.

xds-client/src/client/worker.rs
YutaoMa: That should also work, let me remove the timer related logic for now and try implement them in `tonic-xds` gRPC layer.

xds-client/examples/basic.rs
YutaoMa: 👍 simplified using clap

xds-client/src/client/worker.rs
YutaoMa: Summary of investigation: gRPC does this type-level version_info tracking for State of the World. It can cause inefficiency in setup where endpoints number is high. The plan is to add delta xDS support to address that kind of use case, and we'll keep the stow implementation aligned with other open source gRPC implementations.

gu0keno0: SG!

Since STOW is scoped under listener, it should not be super crazy.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,659 @@
...
+        requests
     }
 
-    /// Subscribe to resources of a given type.
-    pub async fn subscribe(&mut self, type_url: &str, resource_names: Vec<String>) -> Result<()> {
-        // TODO: Track subscription in self.subscriptions
-        // TODO: Get current version/nonce from tracking state
-        let (version_info, response_nonce) = self.get_version_nonce(type_url);
+    /// Handle a command while disconnected (just update state, can't send requests).
+    fn handle_command_disconnected(&mut self, cmd: WorkerCommand) {
```
gu0keno0: Is it possible to unify this with the handle_command?

YutaoMa: Good point, updated to unify them.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,659 @@
...
+
+        for resource in resources {
+            let resource_name = resource.name().to_string();
+            let resource = Arc::new(resource);
+
+            for (_watcher_id, mut event_tx, watcher_name) in watcher_info.clone() {
+                // Watcher is interested if:
+                // - It's a wildcard watcher (empty name), or
+                // - Its name matches the resource name
+                if watcher_name.is_empty() || watcher_name == resource_name {
```
gu0keno0: Under gRPC usage of StoW, would it be necessary to have multiple workers for each resource?

I'm thinking about this: each channel will have a scoped ADS stream under a LDS name, and each time some type of resource change, the channel anyways need to deal with the change, so it feels that there's no need for supporting multiple workers?

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,621 @@
...
+
+        if !decode_errors.is_empty() {
+            self.send_nack(stream, &response, decode_errors.join("; "))
+                .await?;
+            self.notify_watchers_error(&type_url, Error::Validation(decode_errors.join("; ")))
+                .await;
+            return Ok(());
+        }
+
+        let processing_done_futures = self.dispatch_resources(&type_url, decoded_resources).await;
```
gu0keno0: Are we using ProcessingDone struct?

xds-client/src/client/mod.rs
```diff
@@ -1,74 +1,156 @@
...
 pub struct XdsClient {
-    // TODO: add fields as needed
+    /// Channel to send commands to the worker.
+    command_tx: mpsc::UnboundedSender<WorkerCommand>,
 }
 
+/// Default buffer size for watcher event channels.
+///
+/// This provides backpressure when watchers are slow to process events.
+const WATCHER_CHANNEL_BUFFER_SIZE: usize = 16;
```
gu0keno0, line 89: It will be simpler if we just use callbacks require that watcher callbacks are always non-blocking?

No need to block this PR on this, we can continue the discussion and iterations in the follow-ups.

xds-client/src/client/worker.rs
gu0keno0: Oh, here is the answer from Golang https://github.com/grpc/grpc-go/blob/v1.38.1/xds/internal/client/singleton.go#L38

xds-client/src/client/worker.rs
YutaoMa: Yes, it's internally used by dispatch_resources and this future returned is the receiving side of the oneshot channel

xds-client/src/client/mod.rs
YutaoMa: The choice of channel over callback was intentional, given how async callbacks in Rust generally adds tons of complexity: move/Pin/Send/etc. I suspect even with non-blocking callbacks it will still involve many of those complexities, therefore I propose to continue with this channel approach for now.

gu0keno0: what if we just make the callback a regular function?

In all the use cases, the handler will just copy the xDS somewhere and return. The callback handler should never block anyways.

ankurmittal: if callback is not implemented correctly (by the user) it can block your thread, async channel will not block the thread, so channels should be ideally better?

gu0keno0: Fair point. Then shall we use tokio::sync::watch instead? When receiver is slow, worst case is that it misses a few old values but will be guaranteed to receive newest value -- that's the most important thing for service discovery.

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,191 @@
...
+/// Example demonstrating xds-client usage.
+#[derive(Parser, Debug)]
+#[command(name = "basic")]
+#[command(about = "xds-client example - watch Listener resources")]
+struct Args {
+    /// URI of the xDS management server.
+    #[arg(short, long, default_value = "http://localhost:18000")]
+    server: String,
+
+    /// Path to CA certificate for TLS (enables TLS when set).
```
arjan-bal: Nit: It might be worth mentioning that the cert and key are expected to be PEM-encoded.

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,191 @@
...
+    println!("Connecting to xDS server: {}", args.server);
+
+    let config = ClientConfig::with_node_id("example-node").user_agent("grpc");
+
+    let transport = match &args.ca_cert {
+        Some(ca_path) => {
+            let ca_cert = std::fs::read_to_string(ca_path)?;
+            let mut tls = ClientTlsConfig::new().ca_certificate(Certificate::from_pem(&ca_cert));
+
+            if let (Some(cert_path), Some(key_path)) = (&args.client_cert, &args.client_key) {
```
arjan-bal, line 175: To assist with debgging, we should validate that the `cert` and `key` are provided as a pair; if only one is present, the code should log an error.

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,191 @@
...
+    // Start watchers for each listener from args
+    for name in &args.listeners {
+        println!("→ Watching for Listener: '{name}'");
+
+        let mut watcher = client.watch::<Listener>(name);
+        let tx = event_tx.clone();
+
+        tokio::spawn(async move {
+            while let Some(event) = watcher.next().await {
+                if tx.send(event).is_err() {
```
arjan-bal, line 207: I think it will be useful to log the error here before breaking.

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,110 @@
...
+        if let Some(ref mut node) = self.node {
+            node.user_agent_name = name.into();
+        } else {
+            self.node = Some(Node {
+                user_agent_name: name.into(),
+                ..Default::default()
+            });
+        }
+        self
+    }
```
arjan-bal: In my opinion, the `Node` struct should have its own builder, and `ClientConfig` should accept a fully initialized `Node`. This decouples the construction of the two structs. If a new field is added to `Node` in the future, `ClientConfig` won't need to be updated. This also avoids the need for repetitive `if let Some` checks within `ClientConfig` methods. Let me know what you think.

xds-client/src/message.rs
```diff
@@ -56,6 +56,8 @@ pub struct Node {
     pub cluster: String,
     /// Locality specifying where the node is running.
     pub locality: Option<Locality>,
+    /// Free-form string identifying the client type (e.g., "envoy", "grpc").
+    pub user_agent_name: String,
```
arjan-bal, line 63: Should we add a builder for `Node` to avoid exposing the fields directly? Since half of the fields are optional, it may also make construction easier for callers.

xds-client/src/message.rs
```diff
@@ -56,6 +56,8 @@ pub struct Node {
     pub cluster: String,
```
arjan-bal: The `id` and `cluster` fields are also optional, see https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#config-core-v3-node

xds-client/src/message.rs
arjan-bal, line 63: There is also a `user_agent_version` field in the Node proto that other gRPC implementations set; we should add that here as well.

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,110 @@
...
+    /// Default: 30 seconds.
+    pub max_backoff: Duration,
+
+    /// Multiplier for exponential backoff.
+    ///
+    /// After each failed connection attempt, the backoff duration is multiplied
+    /// by this value, up to `max_backoff`.
+    ///
+    /// Default: 2.0.
+    pub backoff_multiplier: f64,
```
arjan-bal: There is a gRFC that defines the retry parameters and algorithm: [A6-client-retries.md](https://github.com/grpc/proposal/blob/d926ac47de91eec7527f6b8658ec35f4363ed4cc/A6-client-retries.md#exponential-backoff).

I suggest encapsulating these fields into a `RetryPolicy` struct so they can be propagated together across different objects.

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,110 @@
...
+use crate::message::Node;
+
 /// Configuration for the xDS client.
 #[derive(Debug, Clone)]
 pub struct ClientConfig {
-    // TODO: Add fields as needed
+    /// Node identification sent to the xDS server.
+    ///
+    /// Default: None.
+    pub node: Option<Node>,
```
arjan-bal: Can we make the `node` field required, while keeping some of the fields inside `Node` optional? However, I believe `user_agent` and `user_agent_version` should be mandatory in the final struct, as they are typically provided by the gRPC implementation rather than the user.

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,191 @@
...
+                println!("✓ Listener received:");
+                println!("  name:        {}", resource.name());
+                if let Some(ref rds) = resource.rds_route_config_name {
+                    println!("  rds_config:  {rds}");
+                }
+                println!();
+
+                // In gRPC xDS, you would cascadingly subscribe to RDS, CDS, EDS, etc.
+                // before completing the done signal.
+                done.complete();
```
arjan-bal: I haven't checked the previous PRs, so apologies if this was already discussed: Could we remove the `complete` method and rely only on the `Drop` trait to signal completion? Since `complete` is synchronous and doesn't return a value, moving this logic to `drop` seems like a good fit for RAII.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,621 @@
...
+use crate::transport::{Transport, TransportStream};
+
+/// Unique identifier for a watcher.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct WatcherId(Uuid);
+
+impl WatcherId {
+    /// Create a new unique watcher ID.
+    pub fn new() -> Self {
+        Self(Uuid::new_v4())
```
arjan-bal: We could use an atomic to have an auto incrementing ID to avoid taking a dependency on uuild.
```rust
static NEXT_ID: AtomicU64 = AtomicU64::new(1);

impl WatcherId {
    fn new() -> Self {
        Self(NEXT_ID.fetch_add(1, Ordering::Relaxed))
    }
}
```

xds-client/src/client/mod.rs
YutaoMa: The `done` signal is necessary for this ADS stream flow-control: https://github.com/grpc/grpc-go/pull/7458. `tokio::sync::watch` channel doesn't work well with this "synchronous processing semantic", I'd prefer to keep the current mpsc channel design which acts as a natural back pressure for ADS flow that prevents server from overwhelming the client.

xds-client/src/client/config.rs
YutaoMa: Makes sense, will extract to a Node builder

xds-client/src/client/worker.rs
YutaoMa: I actually switched from atomic ID to UUID implementation based on previous PR discussion on how this keeps Watcher creation stateless. I think both approach has its pros and cons. One potential advantage of UUID I can think of: it offers more debuggability if we choose to log it.

YutaoMa: On the other hand I agree with the value of reducing any dependency we can. What do you think?

xds-client/examples/basic.rs
YutaoMa: Currently both Drop and explicit `complete` are offered, the latter is just an ergonomic API for `drop(done)` in case the user needs explicit drop, like async processing:

```rust
  ResourceEvent::ResourceChanged { resource, done } => {
      tokio::spawn(async move {
          async_process(&resource).await;
          drop(done);  // = done.complete()
      });
  }
```

I feel an explicit `complete()` API offers a clearer semantics, but I'm also good with removing it to keep API minimal. wdyt?

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,191 @@
...
+        });
+    }
+
+    // Drop the original sender so the loop exits when all watchers complete
+    drop(event_tx);
+
+    while let Some(event) = event_rx.recv().await {
+        match event {
+            ResourceEvent::ResourceChanged { resource, mut done } => {
+                println!("✓ Listener received:");
```
dfawley: I'm not sure how others feel about this, but I'd prefer to avoid unicode in our source files unless it's strictly necessary.  Maybe I shouldn't be concerned about editors not being able to handle it, but it does seem like a possibility, and I don't think it adds sufficient value to outweigh the cost/risk.

xds-client/examples/basic.rs
dfawley: +1 - I think it's unnecessary to have both options for signalling.  Maybe this type changes to `XdsUpdateLease` or something and it's just dropped to signify the update is done being processed.

It might also be interesting to include this as a feature of the update itself, but that might be bundling functionality in an unexpected way.

EDIT: sorry, I did not see @YutaoMa 's reply before writing this.  I believe I'd still prefer to remove the explicit function and make sure the name of the type/parameter convey that it's a token to be dropped when update processing is done, but I'm not strongly opposed to leaving it as-is and deciding to change it later based on feedback/ergonomics when actually using it.  (I'm also not opposed to removing it now and re-adding it later if we think ergonomics/usability are a problem when it's not present.)

xds-client/src/client/config.rs
YutaoMa: Makes sense 👍

xds-client/src/client/config.rs
YutaoMa: 👍 was thinking of encapsulating these, good to know there is gRFC on it. One question I have is whether the gRFC policy is grpc request flow specific, as here it's specifically tied to ADS stream, not the RPC. Let me go over that doc.

xds-client/examples/basic.rs
YutaoMa: 👍

xds-client/examples/basic.rs
YutaoMa: 👍

xds-client/examples/basic.rs
YutaoMa: > include this as a feature of the update itself

ah that's a good point, could potentially use a guard pattern like `ResourceGuard<T>` for this. It will at least make it hard to miss the async case above.

For now I'll remove the explicit `complete` method, and based on `tonic-xds`'s use pattern I'll consider implement the guard pattern

xds-client/examples/basic.rs
YutaoMa: valid point, not that I liked how AI adds in emoji and symbols to all log lines anyway 😆 , just didn't bother removing them

dfawley: :laughing: let's make sure we're very closely reviewing anything that comes from it please!  And maybe it can be given an instruction to never use unicode to hopefully make it less likely to do so.

xds-client/src/client/worker.rs
dfawley: +1 to minimizing dependencies.  Also I generally dislike UUIDs if there are alternatives.  Atomics are probably fine here, or pointer values are probably OK, too, if it's just there for logging.

xds-client/examples/basic.rs
gu0keno0: > "let's make sure we're very closely reviewing anything that comes from it"
+1000

Removing unicode is a good point! I read this example line by line and thought that check mark is fine, yet I think it's a very valid point that this might display weirdly depending on the terminal setting.

xds-client/src/resource/mod.rs
```diff
@@ -67,3 +69,65 @@ pub trait Resource: Sized + Send + Sync + 'static {
     /// The resource name combined with the type URL uniquely identifies a resource.
     fn name(&self) -> &str;
```
arjan-bal: There is another piece of configuration required to indicate whether a resource missing in the ADS response should be interpreted as a removal. gRPC Go stores this as a [boolean](https://github.com/grpc/grpc-go/blob/c7ec4d9ae3281bc57a8adce59b572e56965fb728/internal/xds/clients/xdsclient/resource_type.go#L39-L43).

Since RDS resource names are derived from LDS resources and EDS resource names from CDS resources, a missing RDS or EDS resource in the ADS response is not considered a removal. Consequently, the watcher will not receive a `ResourceError`.

There's some more information in the envoy docs: https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#knowing-when-a-requested-resource-does-not-exist

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,60 @@
...
 pub struct ClientConfig {
-    // TODO: Add fields as needed
+    /// Node identification sent to the xDS server.
+    pub node: Node,
+
+    /// Retry policy for connection attempts.
+    ///
+    /// Controls the backoff behavior when reconnecting to the xDS server.
+    pub retry_policy: RetryPolicy,
+}
```
arjan-bal, line 63: An additional configuration is required for the "watch expiry timeout". This is used to detect if a requested resource does not exist when it has not been received from the xDS server after subscription. See the [Envoy docs](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#knowing-when-a-requested-resource-does-not-exist).

Implementation of this timeout logic can be deferred to a follow-up to limit the scope of this PR. However, I am calling it out now to ensure the existing code can be easily extended to support it.

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,60 @@
...
 #[derive(Debug, Clone)]
 pub struct ClientConfig {
-    // TODO: Add fields as needed
+    /// Node identification sent to the xDS server.
+    pub node: Node,
+
+    /// Retry policy for connection attempts.
+    ///
+    /// Controls the backoff behavior when reconnecting to the xDS server.
+    pub retry_policy: RetryPolicy,
```
arjan-bal: I think we should either add the `non_exhaustive` macro to this struct of hide these fields to ensure we can easily add more fields in the future without breaking users.

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,60 @@
 //! Configuration for the xDS client.
 
+use crate::client::retry::RetryPolicy;
+use crate::message::Node;
+
 /// Configuration for the xDS client.
 #[derive(Debug, Clone)]
 pub struct ClientConfig {
```
arjan-bal, line 37: After comparing this struct to the [Go implementation](https://github.com/grpc/grpc-go/blob/c7ec4d9ae3281bc57a8adce59b572e56965fb728/internal/xds/clients/xdsclient/xdsconfig.go#L41-L86), a few changes are likely needed to ensure future extensibility.

Instead of accepting a pre-built `Transport` in the client's constructor, the configuration should include a priority-ordered list of management servers, and the client should have access to a `Transport` builder. This structure is essential for implementing [xDS fallback](https://github.com/grpc/proposal/blob/master/A71-xds-fallback.md) and [xDS federation](https://github.com/grpc/proposal/blob/master/A47-xds-federation.md), which require a single xDS client to communicate with multiple xDS servers.

xds-client/src/client/config.rs
YutaoMa: The timeout behavior is discussed in previous commits of this PR, we decided to remove it for now and will likely implement it at the gRPC xDS layer since that keeps the main worker loop clean.

xds-client/src/client/config.rs
YutaoMa: Good point, will do that

xds-client/src/client/config.rs
YutaoMa: Makes sense, I was debating between implementing the servers lists abstraction at `tonic-xds` level or at `xds-client` level. I'll go through those gRFCs to see which layer to implement them is the best

xds-client/src/resource/mod.rs
YutaoMa: Noted, adding that right now

xds-client/src/client/config.rs
YutaoMa: After reviewing both gRFCs, it seems keeping multi-authority and multi-servers internal to `xds-client` would be a better design. I'll add in the `TransportFactory` trait and servers config interface for now.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
-/// - Receiving discovery responses
-/// - Version/nonce tracking for ACK/NACK
+/// This enum captures the mutually exclusive subscription states:
+/// - Wildcard: receive all resources of this type
+/// - Named: receive only specific resources by name
+#[derive(Debug, Clone, PartialEq, Eq)]
+enum SubscriptionMode {
+    /// Wildcard subscription - receive all resources of this type.
+    /// In xDS protocol, this is represented by an empty resource_names list.
+    Wildcard,
```
arjan-bal, line 87: Out of curiosity, what is the specific use case for requesting all resources? We haven't encountered a need for this in our other gRPC implementations yet.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
+
+/// Per-watcher state.
 #[derive(Debug)]
-pub struct AdsWorker<S, C> {
-    stream: S,
+struct WatcherEntry {
+    /// Channel to send events to this watcher.
+    event_tx: mpsc::Sender<ResourceEvent<DecodedResource>>,
+    /// Resource name this watcher subscribed to (empty = wildcard).
+    name: String,
```
arjan-bal: nit: I think it's better to use an Option or a custom enum to represent the wildcard state to avoid relying on "magic values" like empty strings/lists.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
 
-    /// Subscribe to resources of a given type.
-    pub async fn subscribe(&mut self, type_url: &str, resource_names: Vec<String>) -> Result<()> {
-        // TODO: Track subscription in self.subscriptions
-        // TODO: Get current version/nonce from tracking state
-        let (version_info, response_nonce) = self.get_version_nonce(type_url);
+    /// Run the main event loop while connected.
+    ///
+    /// Returns `true` if the worker should shut down, `false` to reconnect.
+    async fn run_connected<S: TransportStream>(&mut self, mut stream: S) -> bool {
```
arjan-bal: nit: Would it be more readable to returns a `Result<(), SomeError>` from this function instead of a boolean?

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
+        let type_state = self
+            .type_states
+            .entry(type_url_string)
+            .or_insert_with(|| TypeState::new(decoder));
+
+        let old_subscription = type_state.subscription.clone();
+
+        type_state
+            .watchers
+            .insert(watcher_id, WatcherEntry { event_tx, name });
```
arjan-bal: I believe the client needs to cache the latest state of each resource to bootstrap new watchers. If a new watcher registers for a resource that is already subscribed to, the management server might not re-send it unless there’s a change. Without a local cache, that watcher could potentially wait indefinitely for an update that never comes.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
+    }
+
+    /// Send a DiscoveryRequest for a type.
+    async fn send_request<S: TransportStream>(&self, stream: &mut S, type_url: &str) -> Result<()> {
+        let type_state = match self.type_states.get(type_url) {
+            Some(s) => s,
+            None => return Ok(()),
+        };
 
         let request = DiscoveryRequest {
```
arjan-bal, line 653: nit: Since all current usages pass DiscoveryRequest to the codec by reference, we could use references for some fields (e.g., type_url: &str) and add a lifetime parameter to the struct to avoid clones. Feel free to ignore this if it adds too much complexity—especially since xDS requests are relatively infrequent compared to standard data-plane RPCs.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,600 @@
...
+            type_state.nonce = response.nonce.clone();
+        }
+
+        if !decode_errors.is_empty() {
+            self.send_nack(stream, &response, decode_errors.join("; "))
+                .await?;
+            self.notify_watchers_error(&type_url, Error::Validation(decode_errors.join("; ")))
+                .await;
+            return Ok(());
+        }
```
arjan-bal: If I understand correctly, this code is sending all resource validation errors to all watchers. The expected behavior is that each resource error should only be sent to the watcher for that particular resource. However, since wildcard watchers exist, perhaps they should continue to receive all errors?

xds-client/examples/basic.rs
```diff
@@ -0,0 +1,198 @@
...
+    /// The listener name.
+    pub name: String,
+    /// The RDS route config name (from HttpConnectionManager).
+    pub rds_route_config_name: Option<String>,
+}
+
+impl Resource for Listener {
+    const TYPE_URL: TypeUrl = TypeUrl::new("type.googleapis.com/envoy.config.listener.v3.Listener");
+
+    fn decode(bytes: Bytes) -> xds_client::Result<Self> {
```
arjan-bal: I see a potential challenge with the Resource trait. In grpc-go, we assume a single protobuf library, allowing the decoder to handle both deserialization and field validation (e.g., range checks or cross-field relationships).

In Rust, we have multiple libraries (prost and protobuf-rust). If we follow the same abstraction, we might end up duplicating validation logic across different implementations. This might not require changing the `Resource` trait itself, but we should ensure implementations support pluggable codecs. Alternatively, we could separate decode and validate into distinct methods to make this explicit in the API and encourage separation of concerns?

What are your thoughts, @dfawley?

xds-client/src/error.rs
arjan-bal: I believe the timeout should be managed by the client, as it's a requirement of the SotW protocol. The timeout may be the only way to detect when resources don't exist. When multiple watchers exist for a single resource, the client must cache the latest versions because the management server may only respond when a resource actually changes. Consequently, individual watchers have no visibility into how long the underlying client has been waiting for a response from the server.

I’ve added a related comment here: https://github.com/hyperium/tonic/pull/2478#discussion_r2715799914

xds-client/src/client/worker.rs
YutaoMa: What I know of is

> [Envoy will always use wildcard subscriptions for Listener and Cluster resources](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol)

For gRPC I haven't seen anyone done it.

xds-client/src/client/worker.rs
YutaoMa: Makes sense. We already do that for managing subscription in worker, should definitely use the same approach here in watcher.

xds-client/src/client/worker.rs
YutaoMa: yeah I explained in previous comments the excessive use of clones here are done for simplicity given the infrequent requests. Let me try implement references, if it gets too complicated I'll defer them to later optimizations.

xds-client/examples/basic.rs
YutaoMa: my current assumption though is the validation logic will be coupled to the codec-specific types, it could be hard to avoid duplication there

xds-client/src/client/worker.rs
YutaoMa: That's valid, I'll change it to resource-specific error dispatch. In fact I think this will come naturally as I'm adding in the resource cache, so most of the ack/nack logic will be resource rather than type level.

YutaoMa:
@arjan-bal re-requested your review, I've implemented the following changes in recent commits according to your comments:

	1. `Resource` API change
		1. Added `ALL_RESOURCES_REQUIRED_IN_SOTW` flag
		2. Split up `decode` into `deserialize` and `validate`, specifically allow for name extraction even on failed validation
	2. Added resource cache
	3. Use references in `DiscoveryRequest`
	4. Send resource validation error to specific watcher
	5. `Config` API change
		1. Added `non_exhaustive`
		2. Added servers list config and `TransportFactory`
	6. Style changes
		1. enum for Wildcard case in WatcherEntry
		2. Return `Result<(), Error>` for `run_connected`

xds-client/src/client/worker.rs
arjan-bal: I see. gRPC likely won't need this since it only subscribes to specific listener resources, whereas Envoy must intercept traffic for all services. If there are no known users of this feature, it might make sense to skip supporting it to keep the implementation simpler.

xds-client/src/client/config.rs
```diff
@@ -51,10 +120,39 @@ impl ClientConfig {
...
+    ///
+    /// // Use a custom timeout
+    /// let config = ClientConfig::new(node.clone(), "https://xds.example.com:443")
+    ///     .with_resource_initial_timeout(Some(Duration::from_secs(60)));
+    ///
+    /// // Disable the timeout
+    /// let config = ClientConfig::new(node, "https://xds.example.com:443")
+    ///     .with_resource_initial_timeout(None);
+    /// ```
+    pub fn with_resource_initial_timeout(mut self, timeout: Option<Duration>) -> Self {
```
arjan-bal, line 158: nit: Should with_resource_initial_timeout avoid taking an Option? If a caller doesn't want to set a timeout, they can just forgo calling the method entirely.

xds-client/src/client/worker.rs
```diff
@@ -597,4 +939,56 @@ where
...
+        let command_tx = self.command_tx.clone();
+        let runtime = self.runtime.clone();
+
+        self.runtime.spawn(async move {
+            runtime.sleep(timeout).await;
+            let _ = command_tx.unbounded_send(WorkerCommand::ResourceTimerExpired {
+                type_url: type_url_owned,
+                name,
+            });
+        });
```
arjan-bal, line 991: We should ensure the spawned task is cancelled upon resource receipt so we don't leak any futures.

xds-client/src/client/config.rs
YutaoMa: I suppose it could still be useful to offer a way to clear timeout if set.

xds-client/src/client/worker.rs
YutaoMa: Added a oneshot channel based cancellation mechanism.

xds-client/Cargo.toml
```diff
@@ -40,11 +40,21 @@ rt-tokio = ["dep:tokio"]
...
 
 [dev-dependencies]
-tokio = { version = "1", features = ["rt-multi-thread", "macros", "net"] }
+tokio = { version = "1", features = [
+    "rt-multi-thread",
+    "macros",
+    "net",
+] }
+tonic = { version = "0.14", features = ["tls-ring"] }
+clap = { version = "4", features = ["derive"] }
```
LucioFranco: clap is a really heavy dep we should try to keep it out of the tree and just do the env var parsing manually?

xds-client/src/client/config.rs
```diff
@@ -1,7 +1,158 @@
...
+
+use crate::client::retry::RetryPolicy;
+use crate::message::Node;
+
+/// Configuration for an xDS management server.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub struct ServerConfig {
+    /// URI of the management server (e.g., "https://xds.example.com:443").
+    pub uri: String,
```
LucioFranco: Even though we have non_exhaustive set changing the field type (which we may want to do in the future) will be breaking. I suggest that we keep String as an impl detail and expose it as a &str which allows us to move to say an Arc<str> in the future if we want. I would apply this same rule to all public facing types (if its not public facing prob best to make it pub(crate)).

xds-client/src/client/mod.rs
```diff
@@ -1,74 +1,160 @@
 //! Client interface through which the user can watch and receive updates for xDS resources.
 
+use futures::channel::mpsc;
```
LucioFranco: We need to use tokio here rather than the futures channel

xds-client/src/client/mod.rs
```diff
@@ -1,74 +1,160 @@
...
-    /// Build the client with the given transport and runtime.
+    /// Build the client and start the background worker.
     ///
-    /// This starts the background worker that manages the ADS stream.
-    pub async fn build(self) -> Result<XdsClient> {
-        todo!()
+    /// This spawns a background task that manages the ADS stream.
+    /// The task runs until all `XdsClient` handles are dropped.
+    pub fn build(self) -> XdsClient {
+        let (command_tx, command_rx) = mpsc::unbounded();
```
LucioFranco: I think we want to always avoid an unbounded queue? I know feedback loops can get hard here but I think there is always a way to set an upper bound.

xds-client/src/client/mod.rs
```diff
@@ -1,74 +1,160 @@
...
     /// }
     /// ```
-    pub fn watch<T: Resource>(&self, _name: impl Into<String>) -> ResourceWatcher<T> {
-        todo!()
+    pub fn watch<T: Resource>(&self, name: impl Into<String>) -> ResourceWatcher<T> {
+        let name = name.into();
+        let watcher_id = WatcherId::new();
+        let (event_tx, event_rx) = mpsc::channel(WATCHER_CHANNEL_BUFFER_SIZE);
+
+        let decoder: DecoderFn = Box::new(|bytes| match crate::resource::decode::<T>(bytes) {
```
LucioFranco, line 141: Whats the advantage of doing a boxdn Fn over an actual trait and trait object?

xds-client/src/client/watch.rs
```diff
@@ -1,8 +1,14 @@
 //! Resource watcher types.
 
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+use futures::channel::{mpsc, oneshot};
+use futures::StreamExt;
```
LucioFranco: need to use `tokio_stream`

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,1031 @@
...
+//! - ACK/NACK protocol
 
+use std::collections::{HashMap, HashSet};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+
+use bytes::Bytes;
+use futures::channel::{mpsc, oneshot};
+use futures::{FutureExt, SinkExt, StreamExt};
```
LucioFranco: @dfawley and I have discussed this but I would like us to also avoid `Sink` at all costs right now. Its not a good abstraction for many reasons. We can discuss aternative approaches on tuesday or I can break out some time to meet with yall on it.

xds-client/src/client/worker.rs
```diff
@@ -1,116 +1,1031 @@
...
+        requests
+    }
+
+    /// Run the main event loop while connected.
+    ///
+    /// Returns `Ok(())` if the worker should shut down (command channel closed).
+    /// Returns `Err` if an error occurred and the worker should reconnect.
+    async fn run_connected<S: TransportStream>(&mut self, mut stream: S) -> Result<()> {
+        loop {
+            futures::select! {
```
LucioFranco: This needs to use tokio::select! instead of the futures crates

xds-client/src/client/config.rs
YutaoMa: Makes sense, I've updated all config structs to hide the internal types and exposed `&str` accessor

xds-client/src/client/worker.rs
YutaoMa: np, it was needed only because futures mpsc Sender need that trait for the ergonomic `send()` function. No need for something like that with the change to tokio channel.

xds-client/src/client/mod.rs
YutaoMa: The main reason was to make `watch` API sync because C++/Go gRPC does it synchronously by blocking in place. I evaluated this use case and agree it should be ok to expose `watch` function as async in Rust, which should be ok as the xds-client should be used in async context.

xds-client/src/client/mod.rs
YutaoMa: Both should work. I went with closure so I don't have to create a new trait, is there any particular reason you prefer trait?

YutaoMa:
@LucioFranco I've completed the change to use tokio channels, ptal. The only one I'm not sure about is the DecodeFn closure vs. trait discussion, let me know that you think.

xds-client/src/client/mod.rs
LucioFranco: I find a trait a more clear way to express the abstractness of this sort of thing over closures that tend to hid away capture complexity and data requirements. Not blocking though, can always be changed later.

PR #2482, opened 2026-01-15
feat(xds): initial structures for tonic-xds

gu0keno0:
Initialize the tonic-xds crate for Tower-based xDS implementations.
Refs: https://github.com/hyperium/tonic/issues/2444

## Motivation
We are structuring the xDS contribution as follows:
1. [tonic/xds-client](https://github.com/hyperium/tonic/tree/master/xds-client) provides an runtime-independent, codec-independent and dataplane-independent xDS client implementation
2. **tonic/tonic-xds (this PR creates it)** contains the Tower-based implementation of xDS routing and load-balancing. It can be used directly to enable xDS on Tonic gRPC clients.
3. (Future) [tonic/grpc](https://github.com/hyperium/tonic/tree/master/grpc) will host the xDS routing and load-balancing implementation for Google Rust gRPC.

This PR creates the tonic-xds crate and lays out the initial Tower services for routing and load balancing.

## Solution
The xDS functionality can be implemented as a stack of Tower services. In this PR, we introduce:
1. XdsRoutingService for routing a xDS service name (LDS) to xDS cluster, it will implement xDS-based routing
2. XdsLbService for load-balancing among endpoints under a xDS cluster

The top level APIs are client::channel::{XdsChannel, XdsChannelBuilder, XdsChannelConfig, XdsChannelGrpc}. XdsChannel is transport agnostic, and can be used for adding xDS support to generic Http clients as well. XdsChannelGrpc uses Tonic Channel as the transport and can therefore be used to route gRPC traffic via xDS.

The current PR does not provide implementation for the actual xDS subscription and routing logic. These are abstracted out by the XdsManager trait. In future PRs, we will implement XdsManager trait to add the logics. By separating XDS logics from the Tower services, we will be able to evolve both parts cleanly in parallel.

### DIFF
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -6,6 +6,7 @@ rust-version.workspace = true
 homepage = "https://github.com/hyperium/tonic"
 repository = "https://github.com/hyperium/tonic"
 authors = [
+    "Ankur Mittal <ankurmit2006@gmail.com>",
     "Jeff Jiang <guokeno0@gmail.com>",
     "Lucio Franco <luciofranco14@gmail.com>",
     "Yutao Ma <ytma98@gmail.com>"


tonic-xds/proto/test/helloworld.proto
```diff
@@ -0,0 +1,37 @@
+// Copyright 2015 gRPC authors.
```
ankurmittal, line 1: Copyright 2026?

tonic-xds/src/client/channel.rs
```diff
@@ -0,0 +1,389 @@
...
+    E: std::hash::Hash + Eq + Clone + Send + 'static,
+    S: Service<Request<B>> + Load + Send + 'static,
+    S::Response: Send + 'static,
+    S::Error: Into<BoxError>,
+    S::Future: Send,
+    <S as tower::load::Load>::Metric: std::fmt::Debug,
+{
+    type Response = S::Response;
+    type Error = BoxError;
+    type Future = BoxFuture<Result<Self::Response, Self::Error>>;
```
ankurmittal, line 100: type Future = S::Future?

tonic-xds/src/client/channel.rs
```diff
@@ -0,0 +1,389 @@
...
+    Request<B>: Send + 'static,
+    E: std::hash::Hash + Eq + Clone + Send + 'static,
+    S: Service<Request<B>> + Load + Send + 'static,
+    S::Response: Send + 'static,
+    S::Error: Into<BoxError>,
+    S::Future: Send,
+    <S as tower::load::Load>::Metric: std::fmt::Debug,
+{
+    type Response = S::Response;
+    type Error = BoxError;
```
ankurmittal, line 99: type Error = S:Error??

tonic-xds/src/client/lb.rs
```diff
@@ -0,0 +1,108 @@
...
+{
+    fn clone(&self) -> Self {
+        Self {
+            cluster_registry: self.cluster_registry.clone(),
+            cluster_discovery: self.cluster_discovery.clone(),
+        }
+    }
+}
+
+impl<B, E, S> Service<Request<B>> for XdsLbService<Request<B>, E, S>
```
ankurmittal: Can we rename generic E to Endpoint, as E is associated with Error, it will be easier to read, same thing above

tonic-xds/src/client/channel.rs
```diff
@@ -0,0 +1,389 @@
...
+type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send>>;
+
+impl<B, E, S> Service<http::Request<B>> for XdsChannel<Request<B>, E, S>
+where
+    B: Send + 'static,
+    Request<B>: Send + 'static,
+    E: std::hash::Hash + Eq + Clone + Send + 'static,
+    S: Service<Request<B>> + Load + Send + 'static,
+    S::Response: Send + 'static,
+    S::Error: Into<BoxError>,
```
ankurmittal, line 94: do we need to convert Error?

tonic-xds/src/xds/uri.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// use tonic_xds::XdsUri;
+    ///
+    /// let uri = XdsUri::parse("xds:///my-service").expect("Failed to parse valid xDS URI");
+    /// assert_eq!(uri.target, "my-service");
+    /// ```
+    pub fn parse(uri: &str) -> Result<Self, XdsUriError> {
+        let uri = Url::parse(uri)?;
+
+        if uri.scheme() != XDS_SCHEME {
+            return Err(XdsUriError::InvalidScheme);
```
YutaoMa: may be helpful to log the actual scheme in error as well

tonic-xds/proto/test/helloworld.proto
gu0keno0: This is copied directly from Google gRPC repo, I noticed other copies also just keep the Copyright as is.

tonic-xds/src/client/channel.rs
gu0keno0: Good question! This fundamentally goes down to tower::balance::p2c::Balance implementation. It has a bound of BoxError: https://docs.rs/tower/latest/tower/balance/p2c/struct.Balance.html#impl-Balance%3CD,+Req%3E . We layer on top of Balance, and that's why BoxError is also here.

Soon I'll work on a more generic balancer, so we can potentially make E a type parameter.

tonic-xds/src/client/channel.rs
gu0keno0: Another good question! This is mainly because XdsRouteService need to await on xds_manager.route() -> Future in the call() method of Service trait implementation, call() returns a Future, so here cannot use S::Future as is.

tonic-xds/src/client/channel.rs
gu0keno0: Same reason as the previous one.

tonic-xds/src/xds/uri.rs
gu0keno0: Done.

tonic-xds/src/client/lb.rs
gu0keno0: Done. Good suggestion!

tonic-xds/Cargo.toml
```diff
@@ -0,0 +1,39 @@
+[package]
```
LucioFranco, line 1: Probably should add an authors section here you can include what we have for the others crates and then add yourselves as well.

tonic-xds/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[dependencies]
+tonic = "0.14"
+http = "1"
+tower = { version = "0.5", default-features = false, features = ["discover"] }
+dashmap = "6.1"
+thiserror = "2.0.17"
+futures = "0.3.31"
+url = "2.5.8"
+tonic-prost-build = { version = "0.14", optional = true }
```
LucioFranco: I think this needs to be in dev deps?

tonic-xds/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+
+[lints]
+workspace = true
+
+[dev-dependencies]
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "net"] }
+tonic = { version = "0.14", features = [ "server", "channel", "tls-ring" ] }
+prost = "0.14"
+tonic-prost = "0.14"
+anyhow = "1"
```
LucioFranco: Don't think we need anyhow here can just use box<dyn error>

tonic-xds/src/client/route.rs
```diff
@@ -0,0 +1,75 @@
+use crate::xds::route::RouteInput;
+use crate::xds::xds_manager::XdsRouter;
+use futures::future::BoxFuture;
```
LucioFranco: I dont remember but I think in tonic we defer to not use this crate but instead just define the types we need, futures is an end users consumable crate but I think libs should try to not depend on it.

LucioFranco (approved):
LGTM

tonic-xds/Cargo.toml
gu0keno0: Moved the codegen for test protos to examples/

tonic-xds/Cargo.toml
gu0keno0: Done.

tonic-xds/Cargo.toml
gu0keno0: Done.

tonic-xds/src/client/route.rs
gu0keno0: Sounds good. I did a search in the other crates and understandd the intention now.

I removed most of the futures crate usage, the only thing left is for the BoxDiscovery , which needs to implement tower::Discover. The tower:Discover is sealed, and is implemented for Stream<>, so I changed the code to use futures-core for now.

In our discord chat, I'm also proposing to replace tower::Balance with an internal implementation so that we can iterate on the load balancing implementation. If we do that we can also get rid of the dep to Stream as well.

PR #2484, opened 2026-01-18
fix(tonic): respect max_message_size when decompressing a message

bmwill:
This patch fixes a bug where the configured max_message_size is not respected when a payload is compressed, leading the possibility of over allocating and exhausting memory over the configured maximum message size when decompressing the message.

## Motivation

When compression is enabled, and a compressed message is sent or received, the configured max_message_size (for both clients and servers) is only checked against the compressed message size and that limit is not respected for the resulting uncompressed message, which can lead to resource exhaustion.

## Solution

Respect the configured, or default, max_message_size limit while decompressing a message, returning an error if the resultant decompressed message would exceed the limit.

### DIFF
--- a/tests/compression/src/compressing_response.rs
+++ b/tests/compression/src/compressing_response.rs
@@ -1,5 +1,4 @@
 use super::*;
-use prost::Message;
 use tonic::codec::CompressionEncoding;
 
 util::parametrized_tests! {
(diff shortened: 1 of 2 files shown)


tests/compression/src/compressing_response.rs
```diff
@@ -488,3 +489,76 @@ async fn disabling_compression_on_response_from_client_stream(encoding: Compress
...
 }
+
+util::parametrized_tests! {
+    limit_decoded_message_size,
+    zstd: CompressionEncoding::Zstd,
+    gzip: CompressionEncoding::Gzip,
+    deflate: CompressionEncoding::Deflate,
+}
+
+#[allow(dead_code)]
```
LucioFranco: What combo of features flags makes us add this? can we reduce it to that list so its easier to grok why we have this

bmwill: just checked again and it seems like this is how all of the existing tests were written so i was just copying what was already there. I think its probably due to these needing to just be `#[cfg(test)]` though.

LucioFranco (approved):

PR #2491, opened 2026-01-30
feat: Introduce gRPC over Windows named pipes for platform parity

hackeramitkumar:
### Motivation

Tonic is widely adopted across many open-source projects for gRPC, and AGNTCY/slim also relies on Tonic for its gRPC implementation. Initially, SLIM supported gRPC communication only over TCP sockets. More recently, we added support for gRPC over Unix Domain Sockets (UDS) on Unix-based systems.

However, on Windows platforms, Tonic currently does not support gRPC over named pipes, which limits parity with Unix-based transports. This PR introduces an implementation to enable gRPC over Windows named pipes in Tonic.

Once this functionality is merged into Tonic, AGNTCY/slim will be able to add native support for gRPC over Windows named pipes as well, bringing feature parity across platforms and improving the Windows developer experience.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -22,6 +22,7 @@ jobs:
     - uses: actions/checkout@v6
     - uses: hecrj/setup-rust-action@v2
       with:
+        rust-version: stable
         components: rustfmt
     - run: cargo fmt --all --check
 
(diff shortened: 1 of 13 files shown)


nathanielford:
Hey there! I'm closing this for a few reasons, but the primary one being that we are halting tonic development in favor of the grpc-rust library. I looked into what it would take to merge this, but there is enough work that would have to go into it as to be prohibitive. We appreciate the contribution, though! Thank you!

PR #2492, opened 2026-02-02
Allow full control over User-Agent header

teodordelibasic-db:
## Motivation

Currently, tonic appends its user-agent string (`tonic/x.x.x`) to any existing `User-Agent` header on a request. This means middleware and applications built on top of tonic cannot have full control over the `User-Agent` header.

This addresses the original request in #970 for full control over the User-Agent header.

## Solution

Skip setting the `User-Agent` header if it has already been set on the request, giving users complete control when needed.

## Testing

Updated existing unit tests to verify the new behavior preserves the request's user-agent unchanged.


nathanielford (approved):
Looks good! If you sign the CLA we can get this merged. Small caveat; it's unclear if there will be another tonic release, so this may only ever exist in the repo.

nathanielford:
@teodordelibasic-db Bumping this! If you accept the CLA authorization we can get this merged. Thanks again!

dfawley:
This isn't a common grpc feature -- applications don't usually get to control it at all.  It also constitutes a behavior change, which we may or may not want.

Can't an application tower-wrap the http layer before wrapping it with tonic to control the headers as desired?

LucioFranco:
As doug said I believe the solution here is to overwrite the tonic header on the outbound, since headervalues are cheap (via bytes) I don't see this as an issue. Thanks for the contribution!

PR #2493, opened 2026-02-02
Introduce generic RPC abstractions

dfawley:
*See [PR comment below](https://github.com/hyperium/tonic/pull/2493#issuecomment-3843005140) for a more detailed design of these changes.*

In this change:

- Add `Invoke` and `Handle` and related types.
- Add `client::RecvStreamValidator` which will be used by the codegen to enforce proper RPC response stream behavior for unary and non-unary streams.
- Update `Status` to `Result<(), StatusError>` so that a status in an Err() cannot be OK.

**Some very large, structural questions about these APIs remain to be answered:**

- Is it better to use generics for message types and streams instead of dynamic types?  (Generics may offer better flexibility for the future -- e.g. interceptors may one day be able to determine "does my message T implement trait Foo?", at the cost of larger binaries.)  It's unlikely generics would be significantly faster as there should be very few places where we make these calls.

- Do we need to support dyn compatibility on these traits?

- Should we use types that match Stream / Sink?

- Are we sure this is the optimal API?  There are three different models we have been investigating:

1. call(request_stream, response_sink)
2. call(request_stream) -> response_stream
3. call() -> (request_sink, response_stream)

This change proposes (3) for clients and (1) for servers, but we need more time working with this before deciding for sure.

**In spite of the unanswered questions**

I believe this proposal is ready for review so we can start tying the pieces together and to enable people to work on different things that are dependent on these APIs at the same time (e.g. proto codegen, channel implementation, server implementation, contents of headers/trailers, etc).  Moderate changes to these APIs are likely and large changes are possible, but we can refactor the implementation as the types evolve.

### DIFF
--- a/tonic/Cargo.toml
+++ b/tonic/Cargo.toml
@@ -99,7 +99,7 @@ bencher = "0.1.5"
 quickcheck = "1.0"
 quickcheck_macros = "1.0"
 static_assertions = "1.0"
-tokio = {version = "1.0", features = ["rt-multi-thread", "macros"]}
+tokio = {version = "1.0", features = ["rt-multi-thread", "macros", "test-util"]}
 tower = {version = "0.5", features = ["load-shed", "timeout"]}
 
 [lints]
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -111,6 +111,7 @@ pub struct Server<L = Identity> {
     accept_http1: bool,
     service_builder: ServiceBuilder<L>,
     max_connection_age: Option<Duration>,
+    max_connection_age_grace: Option<Duration>,
 }
 
 impl Default for Server<Identity> {
@@ -139,13 +140,14 @@ impl Default for Server<Identity> {
             accept_http1: false,
             service_builder: Default::default(),
             max_connection_age: None,
+            max_connection_age_grace: None,
         }
     }
 }
 
 /// A stack based [`Service`] router.
 #[cfg(feature = "router")]
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Router<L = Identity> {
     server: Server<L>,
     routes: Routes,
@@ -282,6 +284,36 @@ impl<L> Server<L> {
     pub fn max_connection_age(self, max_connection_age: Duration) -> Self {
         Server {
             max_connection_age: Some(max_connection_age),
+            ..self
+        }
+    }
+
+    /// Sets the maximum duration that a connection may continue to exist
+    /// **after** the graceful shutdown period (`max_connection_age`) has elapsed.
+    ///
+    /// This timeout only takes effect *after* a connection has exceeded its
+    /// configured `max_connection_age`. Once that happens, the server will begin
+    /// graceful shutdown for the connection. If the connection does not close
+    /// gracefully within the `max_connection_age_grace` duration, the server will then
+    /// forcefully terminate it.
+    ///
+    /// If no `max_connection_age` is configured, this forced shutdown timeout will
+    /// **never trigger**, because the server will not know when to begin the
+    /// graceful shutdown phase.
+    ///
+    /// Default is no limit (`None`).
+    ///
+    /// ```
+    /// # use tonic::transport::Server;
+    /// # use tower_service::Service;
+    /// # use std::time::Duration;
+    /// # let builder = Server::builder();
+    /// builder.max_connection_age_grace(Duration::from_secs(60));
+    /// ```
+    #[must_use]
+    pub fn max_connection_age_grace(self, max_connection_age_grace: Duration) -> Self {
+        Server {
+            max_connection_age_grace: Some(max_connection_age_grace),
             ..self
         }
     }
@@ -614,6 +646,7 @@ impl<L> Server<L> {
             max_frame_size: self.max_frame_size,
             accept_http1: self.accept_http1,
             max_connection_age: self.max_connection_age,
+            max_connection_age_grace: self.max_connection_age_grace,
         }
     }
 
@@ -746,6 +779,7 @@ impl<L> Server<L> {
         let http2_max_pending_accept_reset_streams = self.http2_max_pending_accept_reset_streams;
         let http2_max_local_error_reset_streams = self.http2_max_local_error_reset_streams;
         let max_connection_age = self.max_connection_age;
+        let max_connection_age_grace = self.max_connection_age_grace;
 
         let svc = self.service_builder.service(svc);
 
@@ -825,7 +859,7 @@ impl<L> Server<L> {
                     let hyper_io = TokioIo::new(io);
                     let hyper_svc = TowerToHyperService::new(req_svc.map_request(|req: Request<Incoming>| req.map(Body::new)));
 
-                    serve_connection(hyper_io, hyper_svc, server.clone(), graceful.then(|| signal_rx.clone()), max_connection_age);
+                    serve_connection(hyper_io, hyper_svc, server.clone(), graceful.then(|| signal_rx.clone()), max_connection_age, max_connection_age_grace);
                 }
             }
         }
@@ -846,6 +880,29 @@ impl<L> Server<L> {
     }
 }
 
+enum TimeoutAction {
+    GracefulShutdown,
+    ForcefulShutdown,
+}
+
+async fn connection_timeout_future(
+    max_connection_age: Option<Duration>,
+    max_connection_age_grace: Option<Duration>,
+) -> TimeoutAction {
+    if let Some(age) = max_connection_age {
+        tokio::time::sleep(age).await;
+
+        if let Some(grace) = max_connection_age_grace {
+            tokio::time::sleep(grace).await;
+            TimeoutAction::ForcefulShutdown
+        } else {
+            TimeoutAction::GracefulShutdown
+        }
+    } else {
+        future::pending().await
+    }
+}
+
 // This is moved to its own function as a way to get around
 // https://github.com/rust-lang/rust/issues/102211
 fn serve_connection<B, IO, S, E>(
@@ -854,6 +911,7 @@ fn serve_connection<B, IO, S, E>(
     builder: ConnectionBuilder<E>,
     mut watcher: Option<tokio::sync::watch::Receiver<()>>,
     max_connection_age: Option<Duration>,
+    max_connection_age_grace: Option<Duration>,
 ) where
     B: http_body::Body + Send + 'static,
     B::Data: Send,
@@ -872,7 +930,10 @@ fn serve_connection<B, IO, S, E>(
 
             let mut conn = pin!(builder.serve_connection(hyper_io, hyper_svc));
 
-            let mut sleep = pin!(sleep_or_pending(max_connection_age));
+            let mut connection_timeout = pin!(connection_timeout_future(
+                max_connection_age,
+                max_connection_age_grace,
+            ));
 
             loop {
                 tokio::select! {
@@ -882,13 +943,20 @@ fn serve_connection<B, IO, S, E>(
                         }
                         break;
                     },
-                    _ = &mut sleep  => {
-                        conn.as_mut().graceful_shutdown();
-                        sleep.set(sleep_or_pending(None));
+                    timeout_action = &mut connection_timeout => {
+                        match timeout_action {
+                            TimeoutAction::GracefulShutdown => {
+                                conn.as_mut().graceful_shutdown();
+                            },
+                            TimeoutAction::ForcefulShutdown => {
+                                debug!("forcefully closed connection");
+                                break;
+                            }
+                        }
                     },
                     _ = &mut sig => {
                         conn.as_mut().graceful_shutdown();
-                    }
+                    },
                 }
             }
         }
@@ -896,13 +964,6 @@ fn serve_connection<B, IO, S, E>(
         drop(watcher);
         trace!("connection closed");
     });
-}
-
-async fn sleep_or_pending(wait_for: Option<Duration>) {
-    match wait_for {
-        Some(wait) => tokio::time::sleep(wait).await,
-        None => future::pending().await,
-    };
 }
 
 #[cfg(feature = "router")]
@@ -1218,9 +1279,55 @@ where
 
 #[cfg(test)]
 mod tests {
+    use super::*;
+    use crate::transport::Server;
     use std::time::Duration;
 
-    use crate::transport::Server;
+    #[tokio::test(start_paused = true)]
+    async fn test_connection_timeout_no_max_age() {
+        let future = connection_timeout_future(None, None);
+
+        tokio::select! {
+            _ = future => {
+                panic!("timeout future should never complete when max_connection_age is None");
+            }
+            _ = tokio::time::sleep(Duration::from_secs(1000)) => {
+            }
+        }
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_connection_timeout_with_max_connection_age() {
+        let future = connection_timeout_future(Some(Duration::from_secs(10)), None);
+
+        let action = future.await;
+        assert!(matches!(action, TimeoutAction::GracefulShutdown));
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_connection_timeout_with_max_connection_age_grace() {
+        let mut future = pin!(connection_timeout_future(
+            Some(Duration::from_secs(10)),
+            Some(Duration::from_secs(5)),
+        ));
+
+        tokio::select! {
+            _ = &mut future => {
+                panic!("should not complete before max_connection_age");
+            }
+            _ = tokio::time::sleep(Duration::from_secs(9)) => {}
+        }
+
+        tokio::select! {
+            _ = &mut future => {
+                panic!("should not complete before max_connection_age_grace");
+            }
+            _ = tokio::time::sleep(Duration::from_secs(4)) => {}
+        }
+
+        let action = future.await;
+        assert!(matches!(action, TimeoutAction::ForcefulShutdown));
+    }
 
     #[test]
     fn server_tcp_defaults() {


grpc/src/client/interceptor.rs
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+use crate::client::CallOptions;
+use crate::client::Invoke;
+use crate::client::InvokeOnce;
+use crate::client::RecvStream;
+use crate::client::SendStream;
+
+/// A trait which allows intercepting an RPC invoke operation.
+pub trait Intercept: Send + Sync {
```
sauravzg: Maybe we want to highlight the difference between Intercept and InterceptOnce ?

grpc/src/client/mod.rs
```diff
@@ -63,3 +71,116 @@ impl Display for ConnectivityState {
...
+
+/// Contains settings to configure an RPC.
+///
+/// Most applications will not need this type, and will set options via the
+/// generated (e.g. protobuf) APIs instead.
+#[derive(Default)]
+#[non_exhaustive]
+pub struct CallOptions {
+    /// The deadline for the call.  If unset, the call may run indefinitely.
+    deadline: Option<Instant>,
```
sauravzg, line 85: isn't `deadline` usually a part of the metadata ?

grpc/src/client/stream_util.rs
```diff
@@ -0,0 +1,404 @@
...
+    AwaitingTrailers,
+    Done,
+}
+
+/// RecvStreamValidator wraps a client's RecvStream and enforces proper
+/// RecvStream semantics on it so that protocol validation does not need to be
+/// handled by the consumer.
+pub struct RecvStreamValidator<R> {
+    recv_stream: R,
+    state: RecvStreamState,
```
sauravzg, line 74: Instead of making the Validator stateful and verifying state at runtime, what are your thoughts on making this a compile time verifiable state machine, which performs struct/trait transitions?


```
RecvStreamHeaderGetter.get(self) -> (Header , RecvStreamMessageRetreiver.getMessage(&self) -> message,
RecvStreamMessageReteriver.getTrailers(self) -> Trailers
```
All APIs are optional to represent trailer only responses

grpc/src/core/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+        None
+    }
+}
+
+#[allow(unused)]
+pub trait RecvMessage: Send + Sync {
+    fn decode(&mut self, data: Vec<Vec<u8>>);
+
+    #[doc(hidden)]
+    fn _ptr_for(&mut self, id: TypeId) -> Option<*mut ()> {
```
sauravzg: Should this be in the public API?  Shouldn't the trait itself be unsafe maybe

grpc/src/status.rs
```diff
@@ -5,22 +5,24 @@ pub use server_status::ServerStatus;
...
-pub struct Status {
+#[derive(Debug, Clone, PartialEq)]
+pub struct StatusError {
     code: StatusCode,
     message: String,
 }
 
-impl Status {
+pub type Status = Result<(), StatusError>;
+
```
sauravzg: While we are at it , should we consider also adding a try macro to make our lives easier in future by ensuring we use the macro in our code from get go?

dfawley:
Here is the relevant section of a larger client-side design that explains these types, or at least the client-side versions of them:

# The `grpc` Crate

The main `grpc` crate will contain traits and types to support the generic RPC API.  At this layer, all RPCs are bidi streams.

## Core Traits

### Invoke

The entry point to the system is the `Invoke` trait, which acts as a factory for RPC streams. The `Invoke` trait starts an RPC. It takes a method name and arguments (for any grpc-specific call options like timeouts, call credentials, etc), and synchronously returns a pair of streams for sending and receiving items on the stream.  `invoke` must return a valid `SendStream` and `RecvStream` even in error situations, e.g. the channel is not connected.  In this case `RecvStream` must synthesize a failing status code and message.

```rust
/// A trait which may be implemented by types to perform RPCs (Remote Procedure
/// Calls, often shortened to "call").
pub trait Invoke: Send + Sync {
    /// Starts an RPC, returning the send and receive streams to interact with
    /// it.
    ///
    /// Note that invoke is synchronous, which implies no pushback may be
    /// enforced via execution flow.  If a call cannot be started or queued
    /// locally, the returned SendStream and RecvStream may represent a
    /// locally-erroring stream immediately instead.  However, SendStream and
    /// RecvStream are asynchronous, and may block their first operations until
    /// quota is available, a connection is ready, etc.
    fn invoke(
        &self,
        method: impl Into<String>,
        options: CallOptions,
    ) -> (impl SendStream, impl RecvStream);
}
```

#### 

### Streams (`SendStream` & `RecvStream`)

Once a call is created, all stream interaction happens via these two traits. They follow gRPC's stream semantics.

```rust
/// Represents the sending side of a client stream.  When a `SendStream` is
/// dropped, the send side of the stream is closed.  Clients may continue to
/// read from the RecvStream
#[trait_variant::make(Send)]
pub trait SendStream: Send {
    /// Sends T on the stream.  If Err(()) is returned, the message could not be
    /// delivered because the stream was closed.  Future calls to SendStream
    /// will do nothing.
    ///
    /// # Cancel safety
    ///
    /// This method is not intended to be cancellation safe.  If the returned
    /// future is not polled to completion, the behavior of any subsequent calls
    /// to the SendStream are undefined and data may be lost.
    async fn send(&mut self, item: &dyn SendMessage, options: SendOptions) -> Result<(), ()>;
}

/// Contains settings to configure a send operation on a SendStream.
///
/// Most applications will not need this type directly, and will use the
/// generated (e.g.  protobuf) APIs to configure RPCs instead.
#[derive(Default)]
#[non_exhaustive]
pub struct SendOptions {
    /// Closes the stream immediately after sending this message.
    pub final_msg: bool,
    /// If set, compression will be disabled for this message.
    pub disable_compression: bool,
}
```

and

```rust
/// Represents the receiving side of a client stream.  When a `RecvStream` is
/// dropped, the associated call is cancelled if the server has not already
/// terminated the stream.
#[trait_variant::make(Send)]
pub trait RecvStream: Send {
    /// Returns the next item on the stream.  If that item represents a message,
    /// `msg` has been updated directly to contain the received message.
    ///
    /// # Cancel safety
    ///
    /// This method is not intended to be cancellation safe.  If the returned
    /// future is not polled to completion, the behavior of any subsequent calls
    /// to the RecvStream are undefined and data may be lost.
    async fn next(&mut self, msg: &mut dyn RecvMessage) -> ClientResponseStreamItem;
}

/// The client's view of a ResponseStream in a RecvStream: the message type is
/// void as the received message is passed in via the `next` method.
pub type ClientResponseStreamItem = ResponseStreamItem<()>;

/// ResponseStreamItem represents an item in a response stream (either server
/// sending or client receiving).
///
/// A response stream must always contain items exactly as follows:
///
/// [Headers *Message] Trailers *StreamClosed
///
/// That is: optionaly, a Headers value and any number of Message values
/// (including zero), followed by a required Trailers value.  A response stream
/// should not be used after Trailers, but reads should return StreamClosed if
/// it is.
#[derive(Debug, Clone)]
pub enum ResponseStreamItem<M> {
    /// Indicates the headers for the stream.
    Headers(ResponseHeaders),
    /// Indicates a message on the stream.
    Message(M),
    /// Indicates trailers were received on the stream and includes the trailers.
    Trailers(Trailers),
    /// Indicates the response stream was closed.  Trailers must have been
    /// provided before this value may be used.
    StreamClosed,
}
```

Notably, `RecvStream`'s `next` function receives a mutable reference to the message to be received into. This enables zero-copy deserialization directly from the network buffer into the application struct, and allows for the message to be allocated from an arena if desired.

### Messages (`SendMessage` and `RecvMessage`)

Messages are traits that self-encode and decode and also provide a way to downcast to a concrete type, described later.

```rust
pub trait SendMessage: Send + Sync {
    fn encode(&self) -> Arc<dyn Buf>;
}

pub trait RecvMessage: Send + Sync {
    fn decode(&mut self, data: Arc<dyn Buf>);
}
```

### Interceptors

Interceptors allow middleware to wrap a `Invoke` impl and its resulting streams. The trait is exactly like `Invoke`, but its method receives an `Invoke` for the channel it is intercepting, which may be used to start the RPC.

```rust
/// A trait which allows intercepting an RPC invoke operation.
pub trait Intercept: Send + Sync {
    /// Intercepts the start of a call.  Implementations should generally use
    /// next to create and start a call whose streams are optionally wrapped
    /// before being returned.
    fn intercept(
        &self,
        method: impl Into<String>,
        args: CallOptions,
        next: &impl Invoke,
    ) -> (impl SendStream, impl RecvStream);
}
```

The below is an example interceptor for illustration purposes.

```rust
pub struct FailStatusInterceptor {}

impl Intercept for FailStatusInterceptor {
    fn intercept(...) -> (impl SendStream, impl RecvStream) {
        let (tx, rx) = next.call(method, args).await;
        (tx, FailingRecvStreamInterceptor { delegate: rx })
    }
}

struct FailingRecvStreamInterceptor<Delegate> {
    delegate: Delegate,
}

impl<Delegate: RecvStream> RecvStream for FailingRecvStreamInterceptor<Delegate> {
    async fn next(...) -> ClientResponseStreamItem {
        let i = self.delegate.next(msg).await;
        if let ClientResponseStreamItem::Trailers(mut t) = i {
            t.status.code = 3;
            return ClientResponseStreamItem::Trailers(t);
        }
        i
    }
}
```

We then provide a simple facility to wrap a `Invoke` impl with an `Intercept` impl to produce a new `Invoke`:

```rust
/// Wraps `Invoke` and an `Intercept` impls and implements `Invoke` for the
/// combination.
#[derive(Clone)]
pub struct Interceptor<Inv, Int> {
    invoke: Inv,
    intercept: Int,
}

impl<Inv: Invoke, Int: Intercept> Invoke for Interceptor<Inv, Int> {
    fn invoke(&self, method: impl Into<String>, options: CallOptions)
           -> (impl SendStream, impl RecvStream) {
        self.intercept.intercept(method, options, &self.invoke)
    }
}
```

Which can then be implemented for every `Invoke` via an extension trait, if desired:

```rust
pub trait InvokeExt: Invoke + Sized {
    fn with_interceptor(self, intercept: impl Intercept) -> impl Invoke;
}

impl<I: Invoke> InvokeExt for I {
    fn with_interceptor<I: Intercept>(self, interceptor: I) -> impl Invoke {
        Interceptor::new(self, interceptor)
    }
}
```

### Downcasting Messages

Interceptors often need to access the concrete message data (e.g., to log or validate a specific field), but the generic API only provides `dyn` views of the messages. Standard `Any` downcasting in Rust is insufficient, because `Any` requires `'static` types, and our messages require view types with lifetimes.

This is solved with a custom `_ptr_for` method on the message traits and a helper `MessageType` trait to identify the types.

```rust
pub trait SendMessage: Send + Sync {
    #[doc(hidden)]
    fn _ptr_for(&self, id: TypeId) -> Option<*const ()> { None }
}

/// A MessageType describes what underlying message is inside a SendMessage or
/// RecvMessage so that it can be downcast, e.g. by interceptors.  It allows for
/// safe downcasting to views containing a lifetime.
pub trait MessageType {
    /// The message view's type, which may have a lifetime.
    type Target<'a>;
    /// The 'static TypeId of the message view.
    fn type_id() -> TypeId;
}

// impl for all SendMessage traits.
impl dyn SendMessage + '_ {
    /// Downcasts the SendMessage to T::Target if the SendMessage contains a T.
    pub fn downcast_ref<T: MessageType>(&self) -> Option<&T::Target<'_>> {
        if let Some(ptr) = self._ptr_for(T::type_id()) {
            unsafe { Some(&*(ptr as *const T::Target<'_>)) }
        } else {
            None
        }
    }
}

pub trait RecvMessage: Send + Sync {
    #[doc(hidden)]
    fn _ptr_for(&mut self, id: TypeId) -> Option<*mut ()> { None }
}

impl dyn RecvMessage + '_ {
    // As above, but with downcast_mut and mut reference types.
}
```

This allows an interceptor to safely inspect the payload as follows:

```rust
// In interceptor code...
async fn send(...) -> Result<(), ()> {
    if let Some(req) = msg.downcast_ref::<ProtoSendMessage<MyRequest>>() {
        // See below for ProtoSendMessage details.
        // req: MyRequest::View<'_> (lifetime matches msg)
        println!("request.query: {}", req.query);
    }
}
```

Typically the `MessageType` trait would be implemented by the same type implementing the `SendMessage` or `RecvMessage` for simplicity.

### "Once" Interceptors

Some interceptors are stateful and can only be executed a single time. One example is header extraction.  If a user wants to read response headers from a specific call, the most ergonomic way to provide the header data back to the application is via a `oneshot` channel.  Because a `oneshot::Sender` is consumed when sending, an interceptor wanting to use this must be able to consume its own state to prevent reuse.

Because `Intercept` takes `&self`, it is reusable.  Instead, we will create `InvokeOnce` and `InterceptOnce` traits that *consume* `self` instead.

```rust
/// The `InvokeOnce` trait is implemented by types that can perform a single
/// RPC.  It is automatically implemented on references of types that implement
/// `Invoke`.
///
/// Most applications will not use this type directly, and will instead use the
/// generated APIs (e.g.  protobuf) to perform RPCs instead.
pub trait InvokeOnce: Send + Sync {
    /// Provides the same semantics as `Invoke.invoke`.
    fn invoke_once(
        self,
        method: impl Into<String>,
        options: CallOptions,
    ) -> (impl SendStream, impl RecvStream);
}

/// A trait which allows intercepting a call one time only.
pub trait InterceptOnce: Send + Sync {
    /// Starts a call.  Implementations should generally use next to create and
    /// start a call whose streams are optionally wrapped before being returned.
    fn intercept_once(
        self,
        method: impl Into<String>,
        args: CallOptions,
        next: impl InvokeOnce,
    ) -> (impl SendStream, impl RecvStream);
}
```

#### Example: HeaderReader Interceptor

This interceptor captures headers and sends them over a `oneshot` channel. It must be a `CallInterceptorOnce` because `tx.send()` consumes `tx`.

```rust
pub struct HeaderReader {
    tx: oneshot::Sender<Headers>,
}

impl HeaderReader {
    pub fn new() -> (Self, oneshot::Receiver<Headers>) {
        let (tx, rx) = oneshot::channel();
        (Self { tx }, rx)
    }
}

impl InterceptOnce for HeaderReader {
    fn intercept_once(
        self,
        descriptor: MethodDescriptor,
        args: Args,
        next: impl InvokeOnce,
    ) -> (impl SendStream, impl RecvStream) {
        let (tx, delegate) = next.invoke(descriptor, args).await;
        (
            tx,
            // Wrap the stream to intercept the headers
            HeaderReaderRecvStream {
                tx: Some(self.tx), // Move the sender into the stream wrapper
                delegate,
            },
        )
    }
}

impl<Delegate: RecvStream> RecvStream for HeaderReaderRecvStream<Delegate> {
    async fn next(&mut self, msg: &mut dyn RecvMessage) -> ClientResponseStreamItem {
        let i = self.delegate.next(msg).await;
        if let ClientResponseStreamItem::Headers(h) = &i
            && let Some(tx) = self.tx.take()
        {
            let _ = tx.send(h.clone());
        }
        i
    }
}
```

This design ensures compile-time safety: you cannot accidentally attach a "one-shot" interceptor to a reusable `ClientStub` in a way that would cause it to panic or not function properly on a second call.

As an example of how the application would use this interceptor:

```rust
// Create the HeaderReader.
let (header_reader_interceptor, headers_rx) = HeaderReader::new();

// Perform the call using the interceptor.
let res = client
    .unary_call(proto!(MyRequest { query: 1 }))
    .with_interceptor(header_reader_interceptor)
    .await;

// After the call has completed, the headers_rx Reader should already have the value --
// or the Sender was dropped first, because the call response did not have any headers.
match headers_rx.await {
    Ok(v) => println!("saw headers: {:?}", v),
    Err(_) => println!("RPC finished as trailers-only"),
}
```

grpc/src/client/interceptor.rs
dfawley: Agreed; done

EDIT: oops - and then I deleted one :)

grpc/src/client/mod.rs
dfawley: Yes and no.  It ends up as an HTTP/2 header, but it's handled differently by gRPC itself.  We don't want to expose it via the application-visible headers, which should ideally only have entries directly set by users.

(Also we'll have a deadline in a task local eventually that will be 2nd priority after this before we assume `infinity`.)

grpc/src/client/stream_util.rs
dfawley: That's really interesting.

It depends on who the consumer is for this.  I was planning to add it in the codegen for validation since there might be interceptors between our channel (which we need to guarantee is not breaking any of the non-unary-specific requirements).  But it's also possible an intermediate interceptor wants to use this to validate if it is making assumptions that are important to ensure.

In the latter case, it's more helpful if it is essentially just an interceptor.  And, considering that, maybe I should just implement it that way?  I have done that now.

---

In the process, I completely had to revamp the traits. Otherwise we were going to need four interception traits:

1. Takes Invoke (requires multiple uses) and returns Invoke
2. Takes Invoke OR InvokeOnce and returns whatever it got
3. Takes Invoke (requires multiple uses) and returns InvokeOnce (consuming)
4. Takes Invoke OR InvokeOnce and returns InvokeOnce (consuming)

And then I think FIVE(??) wrapper types somehow.

This way we have only one trait for Invoke and one trait for Intercept, and they both consume.  There's some confusing stuff here, though, because you have to add + Clone to make it reusable and Clone _can be_ expensiveish if you don't do it via references.  So I'm not sure about this, still. I could probably use some help here.

---

There's still the bigger question about whether the API itself should be a state machine instead of technically allowing any type of item on the response stream at any time.  I found it to be harder to use (and wrap) with that kind of approach, but I'm still open to ideas.

grpc/src/core/mod.rs
dfawley: It needs to be in the public API.  But I was thinking to make this method `unsafe`.  The trait shouldn't be `unsafe` though AFAICT.

grpc/src/status.rs
dfawley: Probably so...let's save that for another PR though.  I did this mainly for testing the validator.

grpc/src/core/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+//! servers.  Note that must gRPC applications should not need these types
+//! unless they are implementing custom interceptors.
+
+use std::any::TypeId;
+
+use crate::Status;
+
+#[allow(unused)]
+pub trait SendMessage: Send + Sync {
+    fn encode(&self) -> Vec<Vec<u8>>;
```
YutaoMa: My understanding is that this encode/decode interface should match whatever the transport layer will use? e.g. `Bytes`, streams, etc. Otherwise these owned vectors would force allocation in implementation?

grpc/src/status.rs
```diff
@@ -5,22 +5,24 @@ pub use server_status::ServerStatus;
 pub use status_code::StatusCode;
 
 /// Represents a gRPC status.
-#[derive(Debug, Clone)]
-pub struct Status {
+#[derive(Debug, Clone, PartialEq)]
```
gu0keno0: Nit: is it a deliberate decision that we don't derive Eq? (allow error_a == error_a) Might be useful that way

grpc/src/status/server_status.rs
```diff
@@ -1,66 +1,59 @@
...
-    pub fn new(code: StatusCode, message: impl Into<String>) -> Self {
-        ServerStatus(Status::new(code, message))
-    }
-
-    /// Create a new `ServerStatus` from a `Status`.
-    pub fn from_status(status: Status) -> Self {
-        ServerStatus(status)
+impl ServerStatusError {
+    /// Create a new `ServerStatus` from a `StatusError`.
+    pub fn from_status_error(status_error: StatusError) -> Self {
```
gu0keno0: Nit: maybe better to implement From\<StatusError\> ?

grpc/src/status/server_status.rs
```diff
@@ -1,66 +1,59 @@
...
+        ServerStatusError(status_error)
     }
 
-    /// Converts the `ServerStatus` to a `Status` for client responses.
-    pub(crate) fn into_status(self) -> Status {
+    /// Converts the `ServerStatusError` to a `StatusError` for client responses.
+    pub(crate) fn into_status_error(self) -> StatusError {
```
gu0keno0: Same here

grpc/src/status/server_status.rs
```diff
@@ -1,66 +1,59 @@
...
 /// leaking of sensitive information from the server to the client.
 #[derive(Debug, Clone)]
-pub struct ServerStatus(Status);
+pub struct ServerStatusError(StatusError);
 
-impl std::ops::Deref for ServerStatus {
-    type Target = Status;
+pub type ServerStatus = Result<(), ServerStatusError>;
+
+impl std::ops::Deref for ServerStatusError {
```
gu0keno0: IMO, per https://rust-lang.github.io/api-guidelines/predictability.html#c-deref we don't have to do Deref here if we have From<> implementations.

grpc/src/core/mod.rs
gu0keno0: I share the same question. Is Vec of another Vec because of upb (micro protobuf) or scatter-gather I/O? I guess we are missing some context here.

grpc/src/core/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+//! Types used to implement core gRPC functionality common to clients and
+//! servers.  Note that must gRPC applications should not need these types
```
gu0keno0: most gRPC applications

grpc/src/core/mod.rs
sauravzg: The transport layer(at least hyper's) is generic on B:Buf, so as long as we return a Buf we should be fine. 

So maybe something like
 `fn encode(&self, &dyn BufMut) -> Result<(),Status>` would be good.

grpc/src/status/server_status.rs
sauravzg: This was intentional.
We current seem to want serverError to be an entity that can't be trivially converted using `From`. The idea is to avoid issues like

```
asyn fn handle() -> Result<(), ServerStatus> {
   // With From this may leak the client status which may contain sensitive information , 
   // so we want it to be an explicit decision instead of an inoculours `?` 
   return dep_service.dep_method()?
}
```

grpc/src/core/mod.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+/// A MessageType describes what underlying message is inside a SendMessage or
+/// RecvMessage so that it can be downcast, e.g. by interceptors.  It allows for
+/// safe downcasting to views containing a lifetime.
+pub trait MessageType {
+    /// The message view's type, which may have a lifetime.
+    type Target<'a>;
+    /// The 'static TypeId of the message view.
+    fn type_id() -> TypeId;
```
gu0keno0: Could we derive type_id from Target? E.g.

```
pub trait MessageType {
    /// The message view's type, which may have a lifetime.
    type Target<'a>;
}

pub fn msg_type_id<M: MessageType>() -> TypeId
where
    M::Target<'static>: 'static
{
    TypeId::of::<M::Target<'static>>()
}
```

So then msg_type_id is strictly tied to M::Target

grpc/src/status/server_status.rs
gu0keno0: Got it, makes lots of sense. Maybe add to the rustdoc.

grpc/src/core/mod.rs
dfawley: Yes, this works.  I think originally I was planning to use a different type ID (e.g. the owned Message type instead of the View type), but that would probably be a bad idea, or at least not necessary.  Thanks!

grpc/src/core/mod.rs
dfawley: Fixed; thanks

grpc/src/status.rs
dfawley: No, just an oversight.  Added.

grpc/src/status/server_status.rs
dfawley: Per below, we don't want From<StatusError> -- but LMK if you had something else in mind.

grpc/src/status/server_status.rs
dfawley: This is documented on the type:
```rust
/// This is a separate type from `Status` to prevent accidental conversion and
/// leaking of sensitive information from the server to the client.
```

grpc/src/status/server_status.rs
dfawley: As above.

grpc/src/client/stream_util.rs
dfawley: I have fully revamped everything, and it seems to be working well.  (Finally!)

- Invoke and Intercept<T> are the only traits users might implement, and they consume.
- InvokeFactory is introduced as a major simplification of trait bounds for users.  This is auto-impl'd for all types that impl Invoke on their reference - i.e. are reusable.
- Users should impl Intercept<T:Invoke> or Intercept<T:InvokeFactory> depending on their needs.
- This is impl'd on a value for a non-reusable interceptor or on a reference for a reusable interceptor.
- These automagically can be merged with existing Invoke impls through two extension traits with identical methods:

  - `InvokeExt` for non-reusable Invoke
  - `InvokeFactoryExt` for reusable Invokes
  - Both have `with_interceptor` and `with_interceptor_once` methods.  Reusable interceptors would be passed to `with_interceptor` and non-resuable to `with_interceptor_once`.
  - `InvokeFactoryExt.with_interceptor` returns another `InvokeFactory`; the other three return non-reusable `Invoke` impls.

- I added a test that includes every valid combination of invoke/intercept.
- I implemented a full retry interceptor and a test for it.  The tests so far are fairly basic, but stress the functionality.  This should be moved out of `interceptor.rs` and inside the channel when we implement proper retry support.

Also if you don't want to review the retry interceptor, I'm happy to remove it from this PR; I wanted to include it in this commit at least to give an idea of what a very complicated interceptor might look like.

grpc/src/core/mod.rs
dfawley: I don't think a status is appropriate here.  Probably just a String makes the most sense.  Our library will determine what code is appropriate.

I changed this to `VecDeque<Bytes>` for now.  This allows the data to be noncontiguous and also allows the application to access the data in the chunks it is already in.  In theory a future proto deserializer could want to store the bytes buffers as-is and access them lazily.

PTAL.  I'm happy to debate this further but we should not dwell on it _too_ long for now -- we should pick something reasonable and more forward with implementing on top of it.  I suspect this will be an easy thing to change in the future.

dfawley (commented):
Thank you everyone for the review feedback. This PR is now ready to review again after some substantial changes.  I will update the mini-design in the PR comments accordingly tomorrow.

grpc/src/status.rs
sauravzg, line 1: Can we move all status related changes to a different PR and submit it earlier?  This isn't particularly relevant to this PR and might help avoid merge conflicts if submitted early

grpc/src/core/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+/// A response stream must always contain items exactly as follows:
+///
+/// [Headers *Message] Trailers *StreamClosed
+///
+/// That is: optionaly, a Headers value and any number of Message values
+/// (including zero), followed by a required Trailers value.  A response stream
+/// should not be used after Trailers, but reads should return StreamClosed if
+/// it is.
+#[derive(Debug, Clone)]
+pub enum ResponseStreamItem<M> {
```
sauravzg, line 114: Is enum a correct abstraction here? 
Wouldn't it suffer from all problems we discussed with `Stream<Result<T,Err>>` for RecvStream ? 

- we leave it to the implementation of the trait to  ensure that the order is maintained correctly i.e. state machine is verified at runtime instead of build time  which in this case is provided by `StreamValidator`. 
-  there's an ever so slight performance implication of using enums which will always require a branch to identify what type is being returned.

grpc/src/status.rs
dfawley: Done. I'm not sure why I thought this was going to depend on it more originally. Really it will only be important when we do the proto APIs.

grpc/src/core/mod.rs
dfawley: > Is enum a correct abstraction here?

It's a **valid** abstraction, anyway.  Is it the most optimal given all the tradeoffs?  Hard to say.  The other options I explored had downsides, too:

- A full-on state machine is harder to wrap and use:

```rust
impl ClientResponseStream {
  async fn headers(self) -> Either<(Headers, MessageReader), Trailers>;
}
impl MessageReader {
  async fn next(self, &mut dyn RecvMessage) -> Either<MessageReader, Trailers>;
}
```

- A struct that contains all functionality has questions about the behavior:
```rust
impl ClientResponseStream {
  async fn headers(&mut self) -> Option<Headers>; // what is returned on a second call?
  async fn next(&mut self, &mut dyn RecvMessage) -> Result<(), ()>; // what happens if you don't read headers first?
  async fn trailers(self) -> Trailers;
}
```
One specific problem I ran into with the latter approach is: the application (including our proto wrapper) generally doesn't care about the headers.  So how would the header-reader interceptor work?  The implementation I had needed to wait until the first operation, no matter what kind, and then read the headers first, which was pretty awkward.

Do you have other ideas?

Note that some concerns were raised about the cost of this approach for interceptors that only want to do something simple like validate the outgoing headers and fail the RPC in certain conditions.  In this case, because there can only be one impl behind the returned RecvStream, that impl will need to intercept every stream operation:

```rust
impl RecvStream for Whatever {
   async fn next(&mut self, msg: &mut dyn RecvMessage) -> ClientResponseStreamItem {
      if let Some(self.error) {
         return ClientResponseStreamItem::Trailers{status:self.error.clone()} // e.g.
      }
      self.delegate.next(msg)
   }
}
```

If that turns out to be a performance problem then we may need to adjust this approach or use something different internally where performance is most important.

sauravzg: `Do you have other ideas?` - No. These 3 are essentially all potential solutions i.e. 
* compile time state machine ( performant , idiot proof, poor UX) 
* Runtime state machine (the current RecvStream) (slightly less performant, maybe idiot proof, good ux)
* Move to application layer( similar to what H2 in rust does). 

I tend to lean towards compile time state machine + a extension traits to simplify the wrapping . (Similar to StreamExt). 

Re : performance issues . I am not sure about it, we'll need to benchmark. Isn't it just an additional branch on the enum at each interceptor for headers/trailers and the stream items can essentially be delegate ? The compiler (or us hopefully via profile driven or compiler hints) be able to use branch prediction to minimise this.

dfawley: > we'll need to benchmark

Exactly.  It _is_ as simple as a very-predictable branch and an extra trampoline through a function call (which may be inlined since it's all generics, so, not even?).

The state machine was <span>#</span>2 for me. The usability is quite bad and all the same couner-arguments about performance that you said above apply to this approach vs. that one as well.

grpc/src/core/mod.rs
dfawley: As discussed offline, my current plan will be to change this to:

```rust
fn encode(&self) -> Result<Box<dyn Buf + Send>, String>;
fn decode(&self, &mut dyn Buf) -> Result<(), String>;
```

For `encode`, I don't think we want to pass a `&mut BufMut` since that would prevent zero-copy.

For `decode`, this should be fine, because of `copy_to_bytes`'s behavior of being zero-copy.

LucioFranco (approved):

grpc/src/core/mod.rs
dfawley: I think `encode` is going to need to return a reference [to allow arena-allocated messages and zero-copy support].  But let's go with this for now and revisit in follow-ups?

sauravzg: `For encode, I don't think we want to pass a &mut BufMut since that would prevent zero-copy.` - Can you explain this slightly ? I think it should be trivial to support zero copy with dyn&BufMut , depending on the implementation.  I think we have a different notion of zero copy.  

`For decode, this should be fine, because of copy_to_bytes's behavior of being zero-copy.` - This is not correct. It depends on the implementation of Buf by default. The default implementation is actually copying the data. Bytes overrides this method to perform a cheap clone. We reallistically will almost never be using `Bytes` directly since that requires a fully contiguous memory block. 

So, let's set a few requirements in place
- We need to ensure GRPC codec has a way to ser/deser which can perform zero copies from the http frames to the proto/prost API. What the proto and prost API does is beyond our control.
- GCS use case: We need to find a way to share raw bytes between proto and Buffer to support https://protobuf.dev/reference/go/opaque-faq/#lazydecoding . 
- Arena: I think we might want to mark it out of scope. Arenas are unary only and the GCS use cases for sharing bytes are usually streaming. Additionally, hyper is currently not Arena aware, so it'll never create network frame bytes on arena.  LMK if you think this additional complexity is worth tackling, IMO no until we have a valid use case.


```
pub trait Deserialize {
    fn deserialize(&mut self, buf: &mut dyn Buf) -> Result<(), crate::Status>;
}
pub trait Serialize {
    fn serialize(&self, buf: &mut dyn BufMut) -> Result<(), crate::Status>;
}
```

So, this is essentially `Prost`s API for ser/deser, So it's drop in compatible with zero copy for prost. 

I have been able to get zero copy working(subject to the limitations of protobuf  API ) with the above API.  Here's my current implementation for prost and protobuf: https://github.com/sauravzg/tonic/tree/f8c0c449b3259d6f980a0cbf6069298b72dcf5d8/grpc/src/codec/serialization . I don't recall why `encode` is generic, but I think I can make it work with dyn as well. 

My `Buf` implementation is essentially currently a wrapper over `VecDeque<Bytes>` on both ser/deser side. On , deser it allows chaining http frame buffers to represent a contiguous single message, on ser side it allows direct serialization from into the Buffer and for protobuf it allows protobuf -> [u8] -> Byest -> append workflow for zero copy.

The https://protobuf.dev/reference/go/opaque-faq/#lazydecoding was a new use case we discussed yesterday, and that's slightly difficult to model since it requires owned cheaply clonable buffers (c++ uses absl::Cord which is implemented as a rope data structure) , but technically we can use something like 

```
    fn deserialize(&mut self, buf: &mut dyn Buf) -> Result<(), crate::Status>{
      vec<Bytes> buffer_chain;
      while(buf.remaining()>0){
        // We have to override `copy_to_bytes` to allow this cheap clone. The default impl is a copy by default.
         buffer_chain.append(buf.copy_to_bytes(buf.chunk().len()));
         buf.advance(buf.chunk().len());
      }
       protobuf::fancy_lazy_decode_serialize(self,buffer_chain.into_buf());
    }
```

This API is dyn compatible, avoids boxing and supports lazy manipulation, but cannot support arena based buffers.

sauravzg: I stand corrected. I realized why my `Serialize` trait isn't dyn but generic, because it requires https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html#method.put to be able to append bytes to allow zero copy which is not dyn compatible. So, maybe we need a different dyn compatible API for encoding.

This is solvable but would need a custom trait which adds a method with the signature `put(Bytes ...)` to the current `BufMut` API.

dfawley: > Arena: I think we might want to mark it out of scope

I'm not convinced actually convinced arenas are possible to get to zero-copy on the encode/send path, as the lifetimes there would imply we have to block in send until the data is written out to the wire, which is counter to other optimizations, like batching syscalls for multiple messages.  Unless we have some concept of an arena attached to the invoke operation, which might be possible?

> it requires https://docs.rs/bytes/latest/bytes/buf/trait.BufMut.html#method.put to be able to append bytes to allow zero copy which is not dyn compatible. So, maybe we need a different dyn compatible API for encoding.

What is your concern with `encode() -> Box<dyn Buf + Send>`?

At least short-term, maybe we can merge this approach and then work on improving it as follow-ups?  These APIs are foundational for implementing most everything in the channel, so they need to be merged in some mostly-reasonable state (which I believe we're ~at) to unblock folks.

sauravzg: `like batching syscalls for multiple messages.` Does this really happen? That seems like a consistency issue. So, in theory I could get an ack for a message being sent, but then my box crashes and the message was buffered , died before being sent. So the system lied to me?  As long as we don't do out of bounds stuff it shouldn't happen. Additionally, arenas only apply to unary APIs, so it might be doable maybe? 

`What is your concern with encode() -> Box<dyn Buf + Send>?` - This is problematic because this means that the message is responsbile for allocating buffers for itself, which is an issue because our API demands an allocation per message, which is not good since most messages are very small. 
Providing a reference unlocks both the ways, 
* the message could serialize to an new buffer and append new buffer to existing buffer
* the message could append directly to the last buffer in the Buf
This unblocks using things like bufferpool for optimising allocations

I am okay with adding a todo to fix this later, but I do think it's slightly problematic, because this is a problem that we are currently facing with the rust protobuf API

sauravzg: I think I have enough information to conlude the discussion about arenas and zero copy deser based on my discussions with craig.

I'll take c++ implementation as the benchmark/north star for us.

Zero copy deser/ser is largely unsupported in the protobuf API for all types of data except `Cord` . https://protobuf.dev/reference/cpp/cpp-generated/#cord . 

Arenas usually tend to become irrelevant for Cords . Cords are essentially a rope. https://en.wikipedia.org/wiki/Rope_(data_structure) where the individual `buffers` are refcounted and are located on the heap and not on the arena(roughly a SortedTree<bytes::Bytes>). Arenas, simply contain a refcounted ptr to the data on the heap. 

The implications being, if we stick to supporting the same notion as c++ i.e. true zero copy deser for cord fields, we need not worry about arena at all. 

So, let's proceed to go ahead with "true zero copy will supported for cord type fields" , which scopes out all arena requirements from us.

dfawley: > Does this really happen

Yes, this is something we go out of our way to optimize, probably in all languages.

> in theory I could get an ack for a message being sent, but then my box crashes and the message was buffered , died before being sent. So the system lied to me?

We provide no ack for messages sent.  Just because the send call completes doesn't mean the message was written to the wire.  And even if it did, there would be no guarantee from there that it made it to the destination or that the application on the other side saw it, given how networking works.

> Additionally, arenas only apply to unary APIs, so it might be doable maybe?

That's true but we need one generic API.  So if we need to support it, then it impacts the API for streaming, too.

> So, let's proceed to go ahead with "true zero copy will supported for cord type fields" , which scopes out all arena requirements from us.

I'm not entirely sure what you're proposing.  Are you suggesting that we should only need owned data buffers (which _can_ be ref-countented, in-place owned buffers) and not worry about the output from `encode` keeping a reference to the `SendMessage` from which it came?

sauravzg: I'll close the discussion on supporting generic APIs using arenas, since even if it may be possible even grpc c++ isn't close to being able to trivially achieve this.

Coming to zero copy for GCS.

TLDR: whatever api that we design for heap based zero copy for GCS will trivially work with arenas as well because cords are not allocated on arena.



Assuming, we want to meet the goal of how c++ currently supports true zero copy just for the corded bytes field, we don't have to worry about designing our APIs with arenas in mind i.e. returning refa from encode. Since corded data is allowed to exist beyond arena and hence grpc can own this data. 

Owned Bytes should in general be enough to support true zero copy needs for cords.

dfawley: That sounds right to me, although it's something we should keep thinking about before we release a 1.0.

I think the current change here is OK then?  It's using `Buf`.  And we can keep iterating as we implement things.

PR #2495, opened 2026-02-04
feat(grpc) Add channel credentials API + Insecure credentials

arjan-bal:
This change introduces a channel credentials API to abstract over TLS/Insecure/Local credentials. Implementations for TLS and local credentials will follow in future PRs.

Key Changes
* Renamed `rt::TcpStream` to `GrpcEndpoint` to reflect a more generic connection interface.
* `GrpcEndpoint` is now sealed. This allows us to change the API in the future (e.g., to align with gRPC C++) without breaking changes. This restricts implementations of `rt::Runtime` and channel credentials to the `grpc` crate.

Other changes
* Most credential components are `pub(crate)` as users can't to implement them due to `GrpcEndpoint` trait being sealed.
* Restricted the `rt` module to `pub(crate)` and exposed `Runtime` as `pub` to resolve `private_bounds` lints when using runtime types in `pub` APIs.


grpc/src/rt/hyper_wrapper.rs
```diff
@@ -66,13 +66,13 @@ pin_project! {
     /// Tokio's async IO traits.
     pub(crate) struct HyperStream {
         #[pin]
-        inner: Box<dyn TcpStream>,
+        inner: Box<dyn GrpcEndpoint>,
     }
 }
 
 impl HyperStream {
     /// Creates a new `HyperStream` from a type implementing `TcpStream`.
```
sauravzg, line 74: nit: update docs

grpc/src/credentials/client.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// This is primarily used for HTTP/2 connection reuse (coalescing). If the
+    /// underlying security handshake (e.g., a TLS certificate) covers the provided
+    /// `authority`, the existing connection may be reused for that host.
+    ///
+    /// # Returns
+    ///
+    /// * `true` - The connection is valid for this authority.
+    /// * `false` - The connection cannot be reused; a new connection must be created.
+    fn validate_authority(&self, authority: &Authority) -> bool {
+        false
```
sauravzg, line 74: Why do we have a default implementation?

grpc/src/credentials/client.rs
```diff
@@ -0,0 +1,105 @@
...
+
+use std::sync::Arc;
+
+use tonic::async_trait;
+
+use crate::attributes::Attributes;
+use crate::credentials::common::{Authority, SecurityLevel};
+use crate::rt::{GrpcEndpoint, Runtime};
+
+#[async_trait]
```
sauravzg: Maybe we want to move out of `async_trait` .  Especially because I see you made a lot of effort in writing dyn_wrapper

For send, we should probably use https://docs.rs/trait-variant/0.1.2/trait_variant/ 
For dyn, we should probably use https://docs.rs/dynosaur/latest/dynosaur/ , but this doesn't make the existing trait dyn compatible. It creates a new `DynFoo` which implements `Foo` and is dyn compatible.

grpc/src/client/name_resolution/dns/test.rs
```diff
@@ -296,7 +296,7 @@ impl rt::Runtime for FakeRuntime {
         &self,
         target: std::net::SocketAddr,
         opts: rt::TcpOptions,
-    ) -> Pin<Box<dyn Future<Output = Result<Box<dyn rt::TcpStream>, String>> + Send>> {
+    ) -> Pin<Box<dyn Future<Output = Result<Box<dyn rt::GrpcEndpoint>, String>> + Send>> {
```
sauravzg, line 299: nit: Outside the scope of this CL, but   maybe we need to modularize our runtime traits similar to hyper  into separate components. This will probably fall into IO ?

grpc/src/credentials/client.rs
arjan-bal: I've replaced `async_trait` with `trait_variant::make` here, but I’m having trouble getting the type-erased wrapper code to compile. I'm pushing the code now despite the broken build for visibility.

grpc/src/credentials/client.rs
arjan-bal: This is a safe-by-default implementation for credentials. If a specific implementation needs to support authority overrides, it must explicitly add the necessary check.

grpc/src/rt/hyper_wrapper.rs
arjan-bal: Fixed.

grpc/src/client/name_resolution/dns/test.rs
arjan-bal: We can review the organization once we start making the APIs public. For now, the runtime trait is still work in progress.

grpc/src/credentials/client.rs
arjan-bal: Following offline discussions and troubleshooting with Gemini, we identified that the build issues stem from known Rust compiler limitations (specifically around Send analysis). I've implemented a workaround based on the approach used in the `send_future` crate.

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,132 @@
+/*
```
sauravzg, line 1: Does this file need unit test coverage to verify the blanket implementation?

grpc/src/credentials/insecure.rs
```diff
@@ -0,0 +1,134 @@
+/*
```
sauravzg, line 1: Unit tests

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,132 @@
...
+        runtime: Arc<dyn Runtime>,
+    ) -> Result<
+        (
+            BoxEndpoint,
+            ClientConnectionSecurityInfo<Box<dyn ClientConnectionSecurityContext>>,
+        ),
+        String,
+    > {
+        let (stream, sec_info) =
+            SendFuture::send(self.connect(authority, source, info, runtime)).await?;
```
sauravzg: nit: Shouldn't we be able to do `self.connect(...).send().await?;`

arjan-bal: Yes, that works. Changed.

grpc/src/credentials/insecure.rs
arjan-bal: Added tests for the client side. I realized that the `Runtime` trait lacks methods to create TCP listeners, which are required to test the server-side logic. I will add these methods in a follow-up PR, along with tests for insecure and type-erased server credentials.

grpc/src/credentials/dyn_wrapper.rs
arjan-bal: Added tests for the client creds wrapper, will add tests for the server side after making the required additions in the Runtime trait.

grpc/src/credentials/client.rs
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+use std::sync::Arc;
+
+use crate::attributes::Attributes;
+use crate::credentials::common::{Authority, SecurityLevel};
+use crate::rt::{GrpcEndpoint, Runtime};
+
+#[trait_variant::make(Send)]
+pub trait Sealed {
```
ankurmittal: why name this trait Sealed?

grpc/src/credentials/insecure.rs
```diff
@@ -0,0 +1,200 @@
...
+/// An implementation of [`ClientChannelCredential`] for insecure connections.
+///
+/// This credential type does not perform any encryption or authentication. It
+/// simply passes the raw underlying transport as the output.
+#[derive(Debug, Clone, Default)]
+pub struct InsecureClientChannelCredentials;
+
+impl InsecureClientChannelCredentials {
+    /// Creates a new instance of `InsecureChannelCredentials`.
+    pub fn new() -> Self {
```
ankurmittal, line 46: are we planning to add fields to this struct? then maybe we should add private phantom data to it if user should not create object directly and always use new

grpc/src/credentials/insecure.rs
```diff
@@ -0,0 +1,200 @@
...
+        &INFO
+    }
+}
+
+/// An implementation of [`ServerChannelCredentials`] for insecure connections.
+#[derive(Debug, Clone, Default)]
+pub struct InsecureServerChannelCredentials;
+
+impl InsecureServerChannelCredentials {
+    pub fn new() -> Self {
```
ankurmittal, line 98: same

grpc/src/credentials/insecure.rs
```diff
@@ -0,0 +1,200 @@
...
+        Self
+    }
+}
+
+impl server::Sealed for InsecureServerChannelCredentials {
```
ankurmittal: the trait should be not maned Sealed. It does not tell anything about underlying API

grpc/src/send_future.rs
```diff
@@ -0,0 +1,110 @@
...
+/// [#102211]: https://github.com/rust-lang/rust/issues/102211
+/// [#96865]: https://github.com/rust-lang/rust/issues/96865
+/// [`Future`]: core::future::Future
+/// [`Send`]: core::marker::Send
+pub trait SendFuture: core::future::Future {
+    /// Consumes the future and returns it as an opaque type that is guaranteed to be [`Send`].
+    ///
+    /// This is a zero-cost abstraction (it simply returns `self`) used primarily
+    /// to help the compiler resolve auto-traits or to produce better error diagnostics.
+    fn send(self) -> impl core::future::Future<Output = Self::Output> + Send
```
ankurmittal: maybe rename it to make_send?

grpc/src/credentials/insecure.rs
arjan-bal: We may not need to add fields, but I agree that we should disallow bypassing the constructor. I've added a zero-sized private field to prevent this.

grpc/src/credentials/insecure.rs
arjan-bal: Added a zero sized private field.

grpc/src/credentials/client.rs
arjan-bal: I believe this is known as the "Sealed trait pattern" in rust, where a private supertrait is used to prevent external crates from implementing a public trait. Defining methods on the private supertrait also prevents external crates from calling those methods. I think it is idiomatic to name the private supertrait `Sealed`.

Tonic does something similar in `MetadataMap`:
https://github.com/hyperium/tonic/blob/3efc5f9ce817614d0444f67c02b94bf7f767029d/tonic/src/metadata/map.rs#L2114-L2131

grpc/src/credentials/insecure.rs
arjan-bal: Responded to the comment above, the main trait is `ServerChannelCredentials`, the `Sealed` trait is used to limit the visibility of methods that users can't implement and only gRPC needs to call.

grpc/src/send_future.rs
arjan-bal: Renamed.

grpc/src/credentials/client.rs
ankurmittal: But here it is public not private, that is why I was confused. Maybe wrap it in non pub mod like tonic code? Or make this pub crate?

arjan-bal: The trait itself must be `pub`; otherwise, the compiler will issue a warning regarding a private supertrait within a public trait. To restrict external usage, this `pub` trait should be placed inside a private module. In `credentials/mod.rs`, the `client` module is exported as `pub(crate)`, ensuring that only first-party credential implementations within the crate can access it.

grpc/src/send_future.rs
```diff
@@ -0,0 +1,111 @@
...
+///   analysis incorrectly believes it is held across the await.
+/// * Complex control flow confuses the auto-trait deduction for `Send`.
+///
+/// These scenarios often result in obscure error messages when trying to spawn
+/// the future on an executor (like `tokio::spawn`), claiming the future is not
+/// `Send`.
+///
+/// # The Solution
+///
+/// The `send()` method acts as an identity function (a no-op at runtime) but
```
dfawley: The docs should be updated to match the name change to `make_send`

grpc/src/send_future.rs
```diff
@@ -0,0 +1,111 @@
...
+/// pub trait SendFuture: Future {
+///     fn send(self) -> impl Future<Output = Self::Output> + Send
+///     where
+///         Self: Sized + Send,
+///     {
+///         self
+///     }
+/// }
+///
+/// impl<T: Future> SendFuture for T {}
```
dfawley: Do we need to repeat this here? It's the implementation.

grpc/src/send_future.rs
arjan-bal: The method was renamed during the reviews, missed updating the comments. Fixed now.

grpc/src/send_future.rs
arjan-bal: Yes, it's redundant. Replaced this with a `use` statement that brings the trait in scope.

arjan-bal: I removed the example code because the `SendFuture` trait is not accessible outside the grpc crate. Since cargo executes these examples as an external crate, they effectively serve as integration tests and cannot access private traits. The `doc` CI check was failing due to this.

grpc/src/credentials/client.rs
```diff
@@ -0,0 +1,103 @@
...
+    type ContextType: ClientConnectionSecurityContext;
+    type Output<I>;
+    /// Performs the client-side authentication handshake on a raw endpoint.
```
LucioFranco, line 33: dont think you need the duplicate docs on the sealed trait?

arjan-bal: The trait-level documentation for the externally reachable sub-trait is quite abstract and doesn't cover the specific methods. Even though this method isn't exposed externally, I think keeping the docs is valuable for internal maintainability and for developers using IDE features like hover-to-preview. What do you think?

LucioFranco: iirc the trait you actually interact with is the root one not the sealed one so I think just having docs on the main trait is enough and keeping the sealed one a mirror of the type interface makes more sense.

grpc/src/credentials/client.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// Represents the security state of an established client-side connection.
+#[non_exhaustive]
+pub struct ClientConnectionSecurityInfo<C> {
+    pub(crate) security_protocol: &'static str,
+    pub(crate) security_level: SecurityLevel,
+    pub(crate) security_context: C,
+    /// Stores extra data derived from the underlying protocol.
+    pub(crate) attributes: Attributes,
+}
```
LucioFranco: not a big fan of tons of pub(crates) makes refactoring things later much harder. Something we can maybe discuss as a group but I tend to want to keep a lot of things function based rather than field based.

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,199 @@
...
+        &self,
+        authority: &Authority,
+        source: BoxEndpoint,
+        info: ClientHandshakeInfo,
+        runtime: Arc<dyn Runtime>,
+    ) -> Result<
+        (
+            Box<dyn GrpcEndpoint>,
+            ClientConnectionSecurityInfo<Box<dyn ClientConnectionSecurityContext>>,
+        ),
```
LucioFranco: Think this should just return a struct this is really hard to read due to being a tuple

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,199 @@
...
+    >;
+
+    fn info(&self) -> &ProtocolInfo;
+}
+
+#[async_trait]
+impl<T> DynClientChannelCredential for T
+where
+    T: ClientChannelCredential,
+    T::Output<BoxEndpoint>: GrpcEndpoint + 'static,
```
LucioFranco: if grpcEndpoint is almost always static just promote the `'static` to the trait bounds `trait Foo: 'static`

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,199 @@
...
+            ClientConnectionSecurityInfo<Box<dyn ClientConnectionSecurityContext>>,
+        ),
+        String,
+    > {
+        let (stream, sec_info) = self
+            .connect(authority, source, info, runtime)
+            .make_send()
+            .await?;
+
+        let boxed_stream: BoxEndpoint = Box::new(stream);
```
LucioFranco: should be able to drop this type hint here, its funneled right into a fixed return type.

grpc/src/credentials/dyn_wrapper.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+}
+
+// Bridge trait for type erasure.
+#[async_trait]
+pub(crate) trait DynServerChannelCredentials: Send + Sync {
+    async fn accept(
+        &self,
+        source: BoxEndpoint,
+        runtime: Arc<dyn Runtime>,
```
LucioFranco: nit: probably should just make this `GrpcRuntime` or just `grpc::Runtime` that wraps the arc internally

grpc/src/credentials/insecure.rs
LucioFranco: good call!

grpc/src/credentials/insecure.rs
```diff
@@ -0,0 +1,204 @@
...
+    _private: (),
+}
+
+impl InsecureServerChannelCredentials {
+    pub fn new() -> Self {
+        Self { _private: () }
+    }
+}
+
+impl server::Sealed for InsecureServerChannelCredentials {
```
LucioFranco: seems like the patterns of sealed traits have changed  bit since I last looked at them, I would just say double check the patterns here https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/ I trust predrag and we use semver checks. If you think this is following that I am good on that I trust you to make the right choice here.

grpc/src/credentials/mod.rs
```diff
@@ -1 +1,74 @@
...
+pub(crate) mod client;
+pub(crate) mod dyn_wrapper;
+mod insecure;
+pub(crate) mod server;
+
+pub use insecure::{InsecureClientChannelCredentials, InsecureServerChannelCredentials};
+
+/// Defines the common interface for all live gRPC wire protocols and supported
+/// transport security protocols (e.g., TLS, ALTS).
+pub trait ClientChannelCredential: client::Sealed + Sync {
```
LucioFranco: Why do these need Sync? I assume they would also get send and 'static?

grpc/src/send_future.rs
```diff
@@ -0,0 +1,80 @@
...
+///     creates an opaque type boundary. This can sometimes help the compiler's
+///     trait solver "lock in" the `Send` guarantee and disregard phantom lifetime
+///     issues that might otherwise propagate.
+///
+/// [#64552]: https://github.com/rust-lang/rust/issues/64552
+/// [#102211]: https://github.com/rust-lang/rust/issues/102211
+/// [#96865]: https://github.com/rust-lang/rust/issues/96865
+/// [`Future`]: core::future::Future
+/// [`Send`]: core::marker::Send
+pub trait SendFuture: core::future::Future {
```
LucioFranco, line 65: This is interesting, and makes a ton of sense. Assuming this is totally fine since the compiler will noop this out.

grpc/examples/inmemory.rs
```diff
@@ -1,5 +1,6 @@
 use std::any::Any;
 
+use grpc::credentials::InsecureClientChannelCredentials;
```
dfawley: Bikeshedding time (sorry):

Maybe "channel" credentials are good enough to imply "client", so the two are redundant?

`InsecureChannelCredentials`?

grpc/src/client/channel.rs
```diff
@@ -136,16 +136,16 @@ impl Channel {
...
     /// will fail all RPCs.
     // TODO: should this return a Result instead?
-    pub fn new(
-        target: &str,
-        credentials: Option<Box<dyn Credentials>>,
-        options: ChannelOptions,
-    ) -> Self {
+    pub fn new<C>(target: &str, credentials: C, options: ChannelOptions) -> Self
+    where
+        C: ClientChannelCredential + 'static,
```
dfawley: Likewise `ChannelCredential`(s?)  Why no s?

grpc/src/client/channel.rs
```diff
@@ -192,7 +192,7 @@ impl PersistentChannel {
     // ChannelOption contain only optional parameters.
     fn new(
         target: &str,
-        _credentials: Option<Box<dyn Credentials>>,
+        _credentials: Box<dyn DynClientChannelCredential>,
```
dfawley: `DynChannelCredentials`?

grpc/src/credentials/mod.rs
dfawley: I'm fairly sure Java's `ChannelCredentials` supports call credentials inside it: https://grpc.github.io/grpc-java/javadoc/io/grpc/ChannelCredentials.html

I'm not sure about C++ but it seems like they do, too.  We might need to do the same later.

grpc/src/credentials/mod.rs
```diff
@@ -1 +1,74 @@
...
+    //// Provides the ProtocolInfo of this ClientChannelCredential.
+    fn info(&self) -> &ProtocolInfo;
+}
+
+pub trait ServerChannelCredentials: server::Sealed + Sync {
```
dfawley: Just `ServerCredentials` (also matching Java/C++ naming).  https://grpc.github.io/grpc-java/javadoc/io/grpc/ServerCredentials.html

grpc/src/credentials/mod.rs
```diff
@@ -1 +1,74 @@
...
+        /// The connection guarantees both privacy (confidentiality) and data
+        /// integrity.
+        ///
+        /// This is the standard level for secure transports like TLS.
+        PrivacyAndIntegrity,
+    }
+    /// Represents the value passed as the `:authority` pseudo-header, typically
+    /// in the form `host:port`.
+    pub struct Authority<'a> {
+        pub(crate) host: &'a str,
```
dfawley: Why are we attaching a lifetime here?  Should this be an owned `String` instead?

grpc/examples/inmemory.rs
arjan-bal: Renamed to `InsecureChannelCredentials`.

grpc/src/client/channel.rs
arjan-bal: Renamed to `ChannelCredentials`.

grpc/src/client/channel.rs
arjan-bal: Renamed to `DynChannelCredentials`.

grpc/src/credentials/client.rs
arjan-bal: Removed the use of exported fields throughout.

grpc/src/credentials/mod.rs
arjan-bal: Accepting a reference allows implementations to avoid unnecessary copies in most cases. The channel likely maintains a url::Url type and uses the [host_str](https://github.com/hyperium/tonic/pull/2495/changes) method to retrieve the authority for the credentials. Requiring an owned string would force the channel to perform a copy, which would then be consumed by the credentials during the handshake.

grpc/src/credentials/dyn_wrapper.rs
arjan-bal: Introduced a `HandshakeOutput` struct to replace the tupple.

grpc/src/credentials/dyn_wrapper.rs
arjan-bal: Moved the `'static` bound to `GrpcEndpoint` trait.

grpc/src/credentials/dyn_wrapper.rs
arjan-bal: You're correct, removed.

grpc/src/credentials/mod.rs
arjan-bal: Renamed.

grpc/src/credentials/mod.rs
arjan-bal: When using a load balancing policy that manages multiple TCP connections (e.g., round_robin, ringhash, or least_request), multiple concurrent connections may be established. Consequently, the same credentials object must be wrapped in an Arc to be shared during the handshaking process.

arjan-bal: Yes, it should be `'static`. The static bound was being applied in the channel constructor. I've moved it to the trait now.

grpc/src/credentials/dyn_wrapper.rs
arjan-bal: Added a `GrpcRuntime` type as suggested and removed Arc<dyn Runtime> from all method signatures.

grpc/src/credentials/mod.rs
arjan-bal: Acknowledged. I'll add this capability in a later PR, when I'm working on the CallCredentials API.

grpc/src/credentials/insecure.rs
arjan-bal: After reading a few different sources, I realized the pattern I'm using isn't exactly the standard "sealed trait" pattern. 

In my case, I want to ensure all the methods on the trait are only callable from within the crate. Most guides on the sealed pattern don't actually add methods to the sealed trait itself.

1. The [official Rust guidelines](https://rust-lang.github.io/api-guidelines/future-proofing.html) suggest adding a `#[doc(hidden)]` macro to private methods. However, this is just a gentleman's agreement. External crates can still call the method, though we reserve the right to break their code if they do.
2. [Predrag's blog](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/) suggests adding an unnameable type parameter to each private method. However, this pollutes the call sites.
3. [Jack Wrenn's blog](https://jack.wrenn.fyi/blog/private-trait-methods/#approach-1-private-super-trait) mentions the technique I'm currently following: adding the private methods to a "pub-in-priv" trait. The only downside mentioned is that this pattern can't be applied to an existing public trait without causing breaking changes for downstream users.

In my opinion, Option 3 is the best solution for us. It guarantees that private methods are not callable externally, and the call sites remain clean. I've renamed these private traits to `XXXInternal` to avoid confusion with the regular `Sealed` pattern. 

I would love to hear your thoughts on this approach.

grpc/src/credentials/client.rs
arjan-bal: I'm not quite following this. Could you provide a rough sketch of how the supertrait and subtrait would look? Also, please take a look at [my comment below](https://github.com/hyperium/tonic/pull/2495#discussion_r2799433664) describing the existing approach.

grpc/src/credentials/mod.rs
dfawley: That's once per connection? It seems like avoiding having lifetimes on structs would be worth the cost of a clone per connection.  And even if it was performance-critical, could the [sub?]channel save away an `Authority` instance that it passes (by reference already) to all the handshakers?

arjan-bal: The `connect` and `accept` methods are already accepting `&Authority`, so it should be possible for the channel to re-use the same `Authority` object. I've replaced the `&str` with a `String` in the `host` field.

PR #2497, opened 2026-02-06
feat(grpc): Add attributes API

arjan-bal:
## Attributes Implementation

Implements `Attributes`, an immutable map keyed by `TypeId` (similar to [http::Extensions](https://docs.rs/http/latest/http/struct.Extensions.html)) for passing arbitrary data between control plane components like resolvers, LB policies, and channels.

Each `Endpoint` and `Address` produced by the resolver receives its own `Attribute` objects, which often share many common values. As these endpoints and addresses are passed down the Load Balancing policy tree, they may be copied to include new metadata—such as endpoint weights or xDS metadata—resulting in the creation of numerous objects with significant data duplication. To maximize memory efficiency, the `Attributes` struct is immutable and supports low-cost copying and value updates.

### Implementation Details
* Backing: A persistent list optimized for memory efficiency and low-cost cloning/extending.
* Performance Trade-off: Linear lookup time is acceptable given the small expected size (10-20 elements).
* Usage Pattern: Optimized for scenarios where attributes are passed down the LB policy or Subchannel trees, requiring frequent creation of modified copies that differ slightly from their parent.

### DIFF
--- a/grpc/src/attributes/linked_list.rs
+++ b/grpc/src/attributes/linked_list.rs
@@ -49,7 +49,7 @@ struct Node<K, V> {
 /// This list is intended to store a small number of values (few hundreds) and
 /// is optimized for memory usage. It is **not** optimized for query speed.
 #[derive(Debug)]
-pub struct LinkedList<K, V> {
+pub(crate) struct LinkedList<K, V> {
     head: Option<Arc<Node<K, V>>>,
 }
 
@@ -69,7 +69,7 @@ impl<K, V> Default for LinkedList<K, V> {
 
 impl<K, V> LinkedList<K, V> {
     /// Creates a new, empty list.
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
         Self::default()
     }
 
@@ -77,7 +77,7 @@ impl<K, V> LinkedList<K, V> {
     ///
     /// If the key already exists in the list, this new entry will shadow the
     /// old one, effectively updating the value.
-    pub fn add(&self, key: K, value: V) -> Self {
+    pub(crate) fn add(&self, key: K, value: V) -> Self {
         LinkedList {
             head: Some(Arc::new(Node {
                 key,
@@ -103,7 +103,7 @@ impl<K: Eq, V> LinkedList<K, V> {
     ///
     /// The value associated with the key, or `None` if the key is not present or
     /// has been removed.
-    pub fn get(&self, key: &K) -> Option<&V> {
+    pub(crate) fn get(&self, key: &K) -> Option<&V> {
         let mut current = self.head.as_ref();
         while let Some(node) = current {
             if &node.key == key {
@@ -121,7 +121,7 @@ impl<K: Ord, V> LinkedList<K, V> {
     /// The iterator yields unique keys. If a key has been added multiple times,
     /// only the most recent value is returned. Keys that have been removed are
     /// skipped.
-    pub fn iter(&self) -> Iter<'_, K, V> {
+    pub(crate) fn iter(&self) -> Iter<'_, K, V> {
         Iter {
             current: self.head.as_ref(),
             seen: BTreeSet::new(),
--- a/grpc/src/attributes/mod.rs
+++ b/grpc/src/attributes/mod.rs
@@ -65,11 +65,13 @@ impl<T: Any + Send + Sync + Eq + Ord + Debug> AttributeTrait for T {
 }
 
 #[derive(Debug)]
-struct AttributeValue(Box<dyn AttributeTrait>);
+struct AttributeValue {
+    inner: Box<dyn AttributeTrait>,
+}
 
 impl PartialEq for AttributeValue {
     fn eq(&self, other: &Self) -> bool {
-        self.0.dyn_eq(other.0.as_ref())
+        self.inner.dyn_eq(other.inner.as_ref())
     }
 }
 
@@ -83,7 +85,7 @@ impl PartialOrd for AttributeValue {
 
 impl Ord for AttributeValue {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.0.dyn_cmp(other.0.as_ref())
+        self.inner.dyn_cmp(other.inner.as_ref())
     }
 }
 
@@ -103,7 +105,7 @@ impl Ord for AttributeValue {
 /// and is optimized for memory usage. It is **not** optimized for query speed.
 #[derive(Clone, Default, Debug)]
 pub struct Attributes {
-    map: LinkedList<TypeId, AttributeValue>,
+    elements: LinkedList<TypeId, AttributeValue>,
 }
 
 impl Attributes {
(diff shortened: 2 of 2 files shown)


grpc/src/attributes/avl.rs
sauravzg, line 1: Why doesn't something like the im crate satisfy our needs ?  AVL trees seem relatively low level for us to bother implementing if an implementation already exists.

grpc/src/attributes/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+/// `Attributes` provides a map-like interface where values are keyed by their
+/// TypeId.
+///
+/// Equality and ordering of `Attributes` are structural.
+/// This means two `Attributes` maps are equal if they contain the same set of
+/// values, compared by value (via `Eq` trait).
+/// Stored types must implement `Any + Send + Sync + Eq + Ord + Debug`.
+#[derive(Clone, Default, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Attributes {
+    map: Avl<TypeId, AttributeValue>,
```
sauravzg: Okay for now.
But maybe we don't want tie ourselves to an implementation. 
Can we abstract operations to a trait instead of concrete AVL to decouple things?

grpc/src/attributes/avl.rs
sauravzg, line 1: Maybe we should benchmark this implementation against existing rust implementations before we make a call.
Additionally, I am not sure but maybe AVL (+ a trait if we decide to make one) should be a separate PR?

grpc/src/attributes/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::cmp::Ordering;
+use std::fmt::Debug;
+use std::sync::Arc;
+
+use crate::attributes::avl::Avl;
+
+mod avl;
+
+trait AttributeTrait: Any + Send + Sync + Debug {
+    fn as_any(&self) -> &dyn Any;
```
sauravzg: any_ref maybe?

grpc/src/attributes/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::any::{Any, TypeId};
+use std::cmp::Ordering;
+use std::fmt::Debug;
+use std::sync::Arc;
+
+use crate::attributes::avl::Avl;
+
+mod avl;
+
+trait AttributeTrait: Any + Send + Sync + Debug {
```
sauravzg, line 37: comments for the trait and the methods? It's particularly difficult to understand the need and requirements for ordering ?

sauravzg: NVM. The implementation makes it clear but still worth documenting

grpc/src/attributes/mod.rs
arjan-bal: I don't believe any public methods on the Attributes type expose its internal data structure. Could you clarify the benefit of introducing a trait here? 

Introducing an interface for a single implementation feels like unnecessary indirection.

grpc/src/attributes/avl.rs
arjan-bal: The decision to implement an AVL tree was made to avoid pulling in a new dependency. Since gRPC is used across many projects, any library we add increases the binary size for all downstream users. While the im crate might be more performant due to its use of B-trees, this struct is only used for control plane operations. In my opinion, the performance gain doesn't justify the additional dependency.

I’m open to changing this if we decide that taking on the dependency is acceptable.

arjan-bal: The decision to implement an AVL tree wasn't driven by performance; I’m confident the `im` crate would outperform anything we write ourselves. The primary question is whether we're willing to accept the additional dependency.

grpc/src/attributes/mod.rs
arjan-bal: Renamed.

grpc/src/attributes/mod.rs
arjan-bal: Added a rustdoc mentioning that it's required to support value based equality checks instead of pointer comparisions.

grpc/src/attributes/avl.rs
sauravzg: It is acceptable to me 1.6k . Although, I'd probably prefer rpds : https://crates.io/crates/rpds , since it's more actively maintained with simiar-ish star count.

I'll defer the decision to @dfawley , if depending on a create for data structures is fine.

grpc/src/attributes/mod.rs
sauravzg: Which is why I mention that it's okay for now.

It's going to be a classic tradeoff between YAGNI and ISP.

But a concrete dependendency on something(even if it's an implementation detail) as specific as AVL sounds odd. We might be better off abstracting to a standard `Map` like interface with AVL as its implementation. That allows the flexibility of moving to a different data structure(especially given that we are discussing it in the review) without changing the current attributes code. 

With that said, I am okay with adding a todo to revisit this later.

grpc/src/attributes/avl.rs
arjan-bal: After reviewing the performance requirements, I realized the data structure won't hold many elements. Since we're using TypeId as the key, we're looking at only ~20–30 unique entries. The main reason for using a persistent data structure is structural sharing—specifically, to minimize memory usage when creating multiple copies of the struct that differ by only one element. After discussing this with a C++ maintainer, it seems a persistent linked list would maximize space efficiency and be simpler to implement. I'm going to give that approach a try.

LucioFranco: Remind me again why we dont want to follow the internal structure of HeadersMap? I have not looked in a while but the properties should be the same?

grpc/src/attributes/mod.rs
LucioFranco: Id rather pick something and not turn this into a bunch of traits as well, that tends to make the code quite hard to read.

grpc/src/attributes/avl.rs
sauravzg: Attributes are different from HeadersMap. They are a side channel for propagating information in the control plane and very rarely for letting the data plane propagate some information to the control plane.
They will likely not be serialized on to the wire, so might not be particularly related to headersmap.

This is  closer to http extensions , but the way GRPC's tends to use it is "make a bunch of copies and update them slightly" which somewhat leads to the need for a persistent structure instead of the standard http extensions like approach.

LucioFranco: Sorry you're right, got things mixed up. And why dont we want to reuse extensions, I don't see a reason in the PR desc.

sauravzg: Extensions is essentially a wrapper over hashmap of any. 
So, it provides good read semantics  but poor copy semantics since you need to clone entire map.
Cloning current attributes and slightly modifying is an important requirement, so we need a "persistent  variations of data structures" that can allow cheap copies + updates.

arjan-bal: I've replaced the AVL tree with a linked list now.

sauravzg (approved):

grpc/src/attributes/linked_list.rs
```diff
@@ -0,0 +1,194 @@
...
+///
+/// Deletions are handled by adding a "deletion marker" node to the front of the
+/// list, which hides the key during iteration.
+///
+/// # Warning
+///
+/// This list is intended to store a small number of values (few hundreds) and
+/// is optimized for memory usage. It is **not** optimized for query speed.
+#[derive(Debug)]
+pub struct LinkedList<K, V> {
```
dfawley: `pub(crate)`? Isn't the list an implementation detail that should be hidden from users?

arjan-bal: I was relying on the entire module being imported as private. Changed all `pub` symbols in this module to `pub(crate)` to make it explicit.

PR #2498, opened 2026-02-09
feat: allow get Grpc inner

Sherlock-Holo:
## Motivation

when user use custom channel implement, for example a dynamic channel, which can connect multi endpoints, and call `ban` method to ban the bad endpoints when rpc call is failed, with this PR user now can access the inner field

## Solution

add the getter methods to allow user access the inner field


nathanielford:
Hey there! This seems to be trying to modify generated code, which isn't going to work. I'm not wholly certain what your use case is, but I'm pretty sure we can address it via other means.

I'm going to close this for now - if you still have a concern do you mind opening an Issue and we can attempt to address it? Thanks!

PR #2499, opened 2026-02-10
Update LICENSE

xiaobei0715:
Update the copyright year to 2026.


nathanielford:
We keep the first year that it was copyrighted. It's not necessary to update with each year. Thank you, though!

PR #2502, opened 2026-02-10
Fix permission of a series of files

r0x0d:
## Motivation

The files present in this commit have chmod 0755 (rwxr-xr-x) without need, making linux packaging process a bit difficult. 

## Solution

This patch fixes them to chmod 0644 (rw-r--r--).


tottoto (approved):
Thanks!

PR #2507, opened 2026-02-16
feat(grpc): Add TCP listener API in the Runtime trait + tests for server credentials

arjan-bal:
This change adds methods to the `Runtime` trait for creating TCP listeners and accepting remote connections. These methods are required to test the insecure server credentials and the upcoming TLS server credentials, which authenticate streams on the server side.

### DIFF
--- a/grpc/src/rt/mod.rs
+++ b/grpc/src/rt/mod.rs
@@ -31,6 +31,8 @@ pub(crate) mod tokio;
 
 pub type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send>>;
 pub type BoxedTaskHandle = Box<dyn TaskHandle>;
+pub type BoxEndpoint = Box<dyn GrpcEndpoint>;
+pub type ScopedBoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
 
 /// An abstraction over an asynchronous runtime.
 ///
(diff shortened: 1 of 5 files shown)


grpc/src/rt/mod.rs
```diff
@@ -58,6 +58,13 @@ pub trait Runtime: Send + Sync + Debug {
         target: SocketAddr,
         opts: TcpOptions,
     ) -> BoxFuture<Result<Box<dyn GrpcEndpoint>, String>>;
+
+    /// Create a new listener for the given address.
+    fn listen_tcp(
```
sauravzg, line 65: Are `tcp_listener` (or `tcp_stream`) the correct abstraction ?
I am thinking about something like a LocalChannelServer/Client . I don't think they'll be using TCP, but may need these abstractions for creating the IO streams etc ?

sauravzg (approved):

grpc/src/rt/mod.rs
arjan-bal: The abstraction for the I/O stream is `GrpcEndpoint`. Currently, `listen_tcp` and `tcp_stream` are just one way to obtain `GrpcEndpoints` through the runtime. We can add factory methods for other transports, such as UDP or Unix sockets, as the need arises.

We could use a single constructor that accepts the transport type, but I suspect the configuration for each transport would differ significantly, causing the abstraction to become leaky.

PR #2518, opened 2026-02-19
grpc: update generic API to use RequestHeaders instead of method string

dfawley:
Also:

- Move metadata from CallOptions to RequestHeaders - this allows the client and server to both use the same RequestHeaders type.
- Add Dyn versions of several traits which will be necessary for type erasure in upcoming implementation work.
- Update 1yr old nightly rust version used by udeps to work around errors from code that compiles cleanly on stable

This PR is part 1 of probably 3 or 4 of introducing the new generic API to the channel/server implementations.  The work is mostly complete but I'm breaking it up into more easily reviewable chunks.

### DIFF
--- a/grpc/src/client/stream_util.rs
+++ b/grpc/src/client/stream_util.rs
@@ -437,10 +437,12 @@ mod test {
             // Assert the codes match.
             assert_eq!(got_t.status().code(), expect_t.status().code());
             // Assert the status received contains the expected status error message.
-            assert!(got_t
-                .status()
-                .message()
-                .contains(expect_t.status().message()));
+            assert!(
+                got_t
+                    .status()
+                    .message()
+                    .contains(expect_t.status().message())
+            );
         }
     }
 


grpc/src/core/mod.rs
```diff
@@ -133,13 +135,19 @@ pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
...
 #[non_exhaustive]
-#[derive(Debug, Clone)]
-pub struct ResponseHeaders {}
+#[derive(Debug, Clone, Default)]
+pub struct ResponseHeaders {
+    pub metadata: MetadataMap,
+}
 
 /// Contains all information transmitted in the request headers of an RPC.
-#[non_exhaustive]
```
arjan-bal, line 140: Is removing `non_exhaustive` a mistake? It's still there on the `ResponseHeaders`.

grpc/src/core/mod.rs
```diff
@@ -133,13 +135,19 @@ pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
...
-#[derive(Debug, Clone)]
-pub struct RequestHeaders {}
+#[derive(Debug, Clone, Default)]
+pub struct RequestHeaders {
+    /// The full (e.g. "/Service/Method") method name specified for the call.
+    pub method_name: String,
```
arjan-bal: Exposing this as a `pub` field would allow external crates to modify it. Should we restrict that?

grpc/src/server/mod.rs
```diff
@@ -97,13 +118,53 @@ pub trait SendStream {
...
+    async fn send<'a>(
         &mut self,
-        item: ServerResponseStreamItem,
+        item: ServerResponseStreamItem<'a>,
         options: SendOptions,
     ) -> Result<(), ()>;
 }
 
+#[async_trait]
+pub trait DynSendStream: Send {
```
arjan-bal: Do the dyn traits need to be exposed to external crates?

grpc/src/core/mod.rs
```diff
@@ -133,13 +135,19 @@ pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
...
+pub struct ResponseHeaders {
+    pub metadata: MetadataMap,
+}
 
 /// Contains all information transmitted in the request headers of an RPC.
-#[non_exhaustive]
-#[derive(Debug, Clone)]
-pub struct RequestHeaders {}
+#[derive(Debug, Clone, Default)]
+pub struct RequestHeaders {
```
arjan-bal, line 167: In previous PR reviews, Lucio had mentioned he prefers using methods instead of exposing fields: https://github.com/hyperium/tonic/pull/2495#discussion_r2799406905

grpc/src/server/mod.rs
dfawley: I thought it was a good idea in case others needed them, but let's go with "no" for now and see if the need actually arises.

grpc/src/core/mod.rs
dfawley: Yes, sorry.

grpc/src/core/mod.rs
dfawley: That comment says something about "pub(crate)s make it harder to refactor later", whereas this would be an external API, so seems different?

Using a struct with public fields does limit our flexibility to change the underlying data, but also makes it harder to use.  The basic idea is this is arguments for the RPC.  It could instead be multiple parameters to `invoke` / `handle`, but grouping them together allows for extensibility and more readable code.

grpc/src/core/mod.rs
dfawley: What is the concern? If someone owns this or a mutable reference to it, then they _should_ be able to modify it, no?  Otherwise they need to allocate and copy to do any mutations?  Mutating these parameters is something interceptors will definitely want to do often.

arjan-bal: The concern is whether the method name should be modifiable by the user, or set exclusively by gRPC.

grpc/src/core/mod.rs
arjan-bal: My understanding was that exposing the fields as `pub` fixes the internal representation of the struct and prevents us from making changes, such as deriving fields from other fields on the fly and validating the input. It may be acceptable for simple data objects.

This isn't explicitly mentioned in the comment though.

grpc/src/core/mod.rs
dfawley: I see.  On the client side, I could see an interceptor wanting to change it. On the server side, it's pretty dubious to want to do it.  But it's impossible to stop an interceptor from doing it.  They could easily just create a new `RequestHeaders` with a different `method_name` when calling into the handler.

grpc/src/core/mod.rs
dfawley: I am happy to update the comment, but let's make sure the basic approach is acceptable first.  @LucioFranco  do you have any opinions here?  I see this as effectively a bunch of RPC-level parameters that apply on both the client and server (so it wouldn't include optional configuration things like "disable retry" or "disable compression", which would be part of `CallOptions` instead since they are client-side-only).

We could instead combine the two concepts into one single thing, but have different structs on client & server.

dfawley: Hmm, I think the problem here is related to the reason `#[non_exhaustive]` was removed -- you can't actually do what you want: have a struct that users build by specifying all the fields they care about and adding `..Default::default()` if it is non-exhaustive.  Since we want it to be extensible, it would need to be opaque.  It works fine in the same crate, but the examples fail (after I update them to use this type).

grpc/src/core/mod.rs
```diff
@@ -133,13 +135,20 @@ pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
 
 /// Contains all information transmitted in the response headers of an RPC.
 #[non_exhaustive]
-#[derive(Debug, Clone)]
-pub struct ResponseHeaders {}
+#[derive(Debug, Clone, Default)]
+pub struct ResponseHeaders {
```
sauravzg, line 138: nit: Headers are `http` specific . Maybe we should use Grpc specific terminology like `Client/ServerInitialMetadata`, `ServerTrailingMetadata`.

sauravzg (approved):

grpc/src/core/mod.rs
dfawley: I'm not sure whether there are "official" gRPC terms for these concepts.  C++ calls them "initial" and "trailing" metadata, but Go and Java use headers and trailers since that's so much more concise.

- https://pkg.go.dev/google.golang.org/grpc#ClientStream
- https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.html#start(io.grpc.ClientCall.Listener,io.grpc.Metadata)
- https://grpc.github.io/grpc-java/javadoc/io/grpc/ClientCall.Listener.html#onClose(io.grpc.Status,io.grpc.Metadata)

sauravzg: Got it. Resolving in that case.

grpc/src/core/mod.rs
dfawley: I have updated these three basic RPC types to be opaque structs instead of having `pub` fields.  PTAL & thank you for catching this sooner than later.

Issue #1476, opened 2023-08-25
Ability to pass customized connector for channel with multiple endpoints


wenym1:
### Crates

`tonic`

https://github.com/hyperium/tonic/tree/master/tonic/src/transport/channel

### Motivation

For channel with a single endpoint, we can create the channel with customized connector by calling `Endpoint::connect_with_connector`. 

However, we are not able to pass a customized connector for channel with multiple endpoints created from `balance_xxx` methods. Reasons are as followed. 

Currently we have several public methods to create a balanced channel: `balance_list`, `balance_channel` and `balance_channel_with_executor`, where `balance_list` is based on `balance_channel` and `balance_channel` is based on `balance_channel_with_executor`. The `balance_channel_with_executor` is based on the `balance` method. The `balance` method accepts a parameter that implements `Discover<Service = Connection>`, which provides the ability to pass some customized logic. However, the `balance` method is private to crate, and cannot be used externally. When `balance_channel_with_executor` calls `balance`, it pass the `DynamicServiceStream` as the parameter that implements `Discover<Service = Connection>`. But in the `poll_next` of `DynamicServiceStream`, it creates all the connection from `hyper::client::connect::HttpConnector`, which is hard-coded and cannot be customized. 

### Proposal

Make the following changes:

1. Turn `DynamicServiceStream` from `pub(crate)` to `pub`. Meanwhile, add new generic parameter `M: MakeConnection<Uri> + From<Endpoint>` to DynamicServiceStream. Implements `Into<hyper::client::connect::HttpConnector>` for `Endpoint` so that `hyper::client::connect::HttpConnector` satisfies the trait bound of `M`.
2. Turn the `balance` method from `pub(crate)` to `pub`.

wenym1:
I am willing to make such refactor if needed.

tillrohrmann:
I would be interested in such a feature as we want to talk to TCP sockets as well as Unix domain sockets.

wenym1:
> I would be interested in such a feature as we want to talk to TCP sockets as well as Unix domain sockets.
> 
> @wenym1 did you solve the problem differently in the meantime?

The feature is not needed in our project for the time being, so no implementation for it yet.

nathanielford:
Because we are winding down Tonic development and this is a significant API change we are going to close this issue. The new `grpc-rust` library should address this - if you have any concerns we encourage you to open a new issue against that library!

PR #2521, opened 2026-02-23
tonic: connector support for balance channels

Logan-Pageler:
This commit adds the ability for connectors to be
used with balance channels. This commit only adds
new functions but does not change any existing
public functions, so it should be fully backwards
compatible. This commit also modifies the
internal implementation details to support
connectors.

Fixes: #1476

## Motivation

Previously balance channels only supported http connectors. However, it is very important to be able
to use https and other connectors with the balance channels. This pr adds the extensibility to use any
connector type with balance channels.

## Solution

The pr contains two main parts. Firstly it adds generic parameters `DynamicServiceStream` to work with both just `Endpoint` objects as it did previously and now `(Connector, Endpoint)` pairs so that the connectors can be configured. Secondly the pr adds new functions for `Channel` for creating balancing channels with specific connectors

### DIFF
--- a/tonic/src/transport/channel/mod.rs
+++ b/tonic/src/transport/channel/mod.rs
@@ -117,6 +117,42 @@ impl Channel {
         channel
     }
 
+    /// Balance a list of [`Endpoint`] with specific connectors.
+    ///
+    /// This creates a [`Channel`] that will load balance across all the
+    /// provided endpoints.
+    ///
+    /// This allows you to build a balance [`Channel`] that that uses a non-HTTP transport to connect to the endpoints.
+    ///
+    /// # Example
+    /// ```rust
+    /// use hyper_util::client::legacy::connect::HttpConnector;
+    /// use tonic::transport::{Endpoint, Channel};
+    ///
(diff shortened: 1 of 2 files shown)


nathanielford:
Hey there, @Logan-Pageler! Thanks for the contribution. Unfortunately, for a couple of reasons we are not going to merge this. The primary one is that we are winding down Tonic development in favor of the `grpc-rust` library, and avoiding any large changes to the current API of the Tonic library. 

To merge this we'd have to address a number of key concerns, which I'll briefly outline:
* I think this PR will silently bypass TLS and timeouts. `endpoint.connector()` is skipped here, which means any TLS or timeout settings will be ignored.
* I think this heavily impacts how service discovery mechanisms (i.e. DNS watchers), forcing them to yield transport connectors on every endpoint change.
* It also looks like this breaks connection pooling. 

I do think the underlying desire to have channel balancing is being addressed in the new library - I'd encourage you to take a look! We are very open to feedback and are eager to hear how people are using grpc and pain points they meet in the Rust library in particular. Thanks again for all your work here!

PR #2524, opened 2026-02-24
feat: add lz4 and snappy compression

CEmocca:
## Motivation
We use `GZIP` compression as the default compression for the server. But profiling shows `GZIP` compression takes a portion of CPU under high load, causing the high latency from our server

We look into another type of compression method supported by Tonic (0.12.3). We tested `ZSTD` compared to `GZIP`; `ZSTD` performs better than `GZIP`. 
However, my colleagues want more comparisons between `LZ4` and `SNAPPY`. So we run another round of tests

#### Setup
We have two setups, which are
1. Always return random **String** depends on setting size
    - Accept request and just return a random **String** without querying the database
image

2. Replay the production traffic to the server, query, and return the actual response from the database with a different compression method
image

Both setups will receive **1,000** requests per second constantly

#### How do we measure the E2E latency?
We measure it from the client side, `start` and `end` in the client request-response interceptor

### Terminology
1. `acm-identity`: expect no compression response back
2. `acm-gzip`: expect `GZIP` compression response back
3. `acm-zstd`: expect `ZSTD` compression response back
4. `acm-lz4`: expect `LZ4` compression response back
5. `acm-snappy`: expect `SNAPPY` compression response back

## Result
For the random string case
- The bigger the message size is, the more `GZIP` CPU usage
  - For example, with a 100KB message. `GZIP` server CPU usage is ~70% while other compression methods is still ~13%
  - The P99 latency for `GZIP` is up to ~13.4ms, while others are still at ~9ms
- `SNAPPY` and `LZ4` compression ratios is almost 1:1 in this case (Comparing from `identity` vs `SNAPPY` and `LZ4`)
- Memory usage is almost the same for all compressions

For replaying the production traffic case
- `SNAPPY` outperforms other compression methods in our use case. CPU usage is ~10.5% while the P99 is ~6.68ms
  - The server SLA is 10ms. So 6.68ms vs 7.60ms (`SNAPPY` vs `ZSTD`) is significant
  - We will deep dive to investigate why `SNAPPY` has low P99 compared to other

#### Result deck
[GRPC Compression evaluation for github PR.pdf](https://github.com/user-attachments/files/25521415/GRPC.Compression.evaluation.for.github.PR.pdf)

### DIFF
--- a/tests/compression/Cargo.toml
+++ b/tests/compression/Cargo.toml
@@ -15,7 +15,7 @@ pin-project = "1.0"
 prost = "0.14"
 tokio = {version = "1.0", features = ["macros", "rt-multi-thread", "net"]}
 tokio-stream = "0.1"
-tonic = {path = "../../tonic", features = ["gzip", "deflate", "zstd"]}
+tonic = {path = "../../tonic", features = ["gzip", "deflate", "zstd", "lz4", "snappy"]}
 tonic-prost = {path = "../../tonic-prost"}
 tower = "0.5"
 tower-http = {version = "0.6", features = ["map-response-body", "map-request-body"]}
--- a/tests/compression/src/bidirectional_stream.rs
+++ b/tests/compression/src/bidirectional_stream.rs
@@ -7,6 +7,8 @@ util::parametrized_tests! {
     zstd: CompressionEncoding::Zstd,
     gzip: CompressionEncoding::Gzip,
     deflate: CompressionEncoding::Deflate,
+    lz4: CompressionEncoding::Lz4,
+    snappy: CompressionEncoding::Snappy,
 }
 
 #[allow(dead_code)]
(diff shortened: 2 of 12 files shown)


CEmocca:
> Hi @CEmocca,
> 
> Thank you for pushing this. By any chance, do you have benchmarks on Protobuf payloads with different kinds of types? Correct me if I’m wrong, but it seems the benchmarks were only done on strings with varying sizes.

The replay production traffic contains multiple types of responses. Unfortunately, I didn't distinguish them into their respective types.
In the production scenario
- The key is a `String` type
- The value will be returned as `Map<String, Value>`
  - `Value` can be any of these
```
message Value {
  oneof val {
    bytes bytes_val = 1;
    string string_val = 2;
    int32 int32_val = 3;
    int64 int64_val = 4;
    double double_val = 5;
    float float_val = 6;
    bool bool_val = 7;
    BytesList bytes_list_val = 11;
    StringList string_list_val = 12;
    Int32List int32_list_val = 13;
    Int64List int64_list_val = 14;
    DoubleList double_list_val = 15;
    FloatList float_list_val = 16;
    BoolList bool_list_val = 17;
  }
}
```

Most likely, the production traffic should cover all types. I just don't have number for each specific type

nathanielford:
Hey there! Thanks for the work on this contribution. Unfortunately, we're going to close this because there are enough outstanding issues that would have to be resolved that it's probably not going to happen given that Tonic development is winding down. 

Briefly looking at it, it seems like snappy has corruption with an empty payload, there are interop issues with Java and the error swallowing on flush would probably all prevent us from merging this. Given that lz4 is not a standard gRPC compression, we don't want to ask you to spend a lot of time working on this for a library we are unlikely to have another release for. 

All that said, we do really appreciate the contributions, and if the underlying concerns you had around compression options remain we would love to hear more about it so that we can see what we can do in the official grpc-rust implementation. Thanks again!

PR #2528, opened 2026-02-28
grpc: implement Invoke in Channel, Subchannel, and Tonic transport

dfawley:
- Adds a Channel->Tonic Server interop test similar to the existing tonic transport test.
- Convert Picker API to use the new `RequestHeaders` type instead of the outgoing `Request`.
- Leave `inmemory` transport's `Invoke` impl as TODO (it's complicated).

Includes #2527 as the first commit; please ignore those changes for this PR.

### DIFF
--- a/grpc/src/client/subchannel.rs
+++ b/grpc/src/client/subchannel.rs
@@ -252,7 +252,7 @@ impl DynInvoke for InternalSubchannel {
     ) -> (Box<dyn DynSendStream>, Box<dyn DynRecvStream>) {
         let svc = match &self.inner.lock().unwrap().state {
             InternalSubchannelState::Ready(s) => s.svc.clone(),
-            _ => panic!("invoke called on non-READY subchannel"),
+            _ => todo!("handle non-READY subchannel"),
         };
         svc.dyn_invoke(headers, options).await
     }
--- a/grpc/src/client/transport/tonic/mod.rs
+++ b/grpc/src/client/transport/tonic/mod.rs
@@ -156,9 +156,10 @@ impl Invoke for TonicTransport {
         let (req_tx, req_rx) = mpsc::channel(1);
         let request_stream = ReceiverStream::new(req_rx);
         let mut request = TonicRequest::new(Box::pin(request_stream));
-        *request.metadata_mut() = headers.metadata().clone();
+        let (method, metadata) = headers.into_parts();
+        *request.metadata_mut() = metadata;
 
-        let Ok(path) = PathAndQuery::from_maybe_shared(headers.method_name().clone()) else {
+        let Ok(path) = PathAndQuery::from_maybe_shared(method) else {
             return err_streams(Status::new(StatusCode::Internal, "invalid path"));
         };
 
@@ -171,11 +172,11 @@ impl Invoke for TonicTransport {
         }
 
         // Note that Tonic's streaming call blocks until the server's headers
-        // are received.  We must return a working send (and, consequently,
-        // recv) stream before this to allow the application to write its
-        // request(s), so we need to spawn a task for this period of time, and
-        // then we send the response (headers, stream) to the TonicRecvStream
-        // when it is available.
+        // are received.  The client needs a SendStream to provide the request
+        // message(s), which the server may be awaiting before sending its
+        // headers.  So, we spawn a task for this period of time, and then we
+        // send the response (headers, stream) to the TonicRecvStream when it is
+        // available.
         let (resp_tx, resp_rx) = oneshot::channel();
         self.runtime.spawn(Box::pin(async move {
             let response = grpc.streaming(request, path, BufCodec {}).await;
@@ -185,17 +186,23 @@ impl Invoke for TonicTransport {
         (
             TonicSendStream { sender: Ok(req_tx) },
             TonicRecvStream {
-                receiver: None,
-                error: None,
-                response_rx: Some(resp_rx),
+                state: StreamState::AwaitingHeaders(resp_rx),
             },
         )
     }
 }
 
-// Converts from a tonic status to a grpc-crate status.
-fn from_tonic_status(status: TonicStatus) -> Status {
-    Status::new(StatusCode::from(status.code() as i32), status.message())
+// Converts from a tonic status to a trailers stream item.
+fn trailers_from_tonic_status(status: TonicStatus) -> ClientResponseStreamItem {
+    ClientResponseStreamItem::Trailers(Trailers::new(Status::new(
+        StatusCode::from(status.code() as i32),
+        status.message(),
+    )))
+}
+
+// Builds a trailers with a status
+fn trailers_from_status(code: StatusCode, msg: impl Into<String>) -> ClientResponseStreamItem {
+    ClientResponseStreamItem::Trailers(Trailers::new(Status::new(code, msg)))
 }
 
 struct TonicSendStream {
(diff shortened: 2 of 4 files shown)


grpc/src/client/subchannel.rs
```diff
@@ -232,6 +243,21 @@ impl Service for InternalSubchannel {
...
+#[async_trait]
+impl DynInvoke for InternalSubchannel {
+    async fn dyn_invoke(
+        &self,
+        headers: RequestHeaders,
+        options: CallOptions,
+    ) -> (Box<dyn DynSendStream>, Box<dyn DynRecvStream>) {
+        let svc = match &self.inner.lock().unwrap().state {
+            InternalSubchannelState::Ready(s) => s.svc.clone(),
+            _ => panic!("invoke called on non-READY subchannel"),
```
arjan-bal: nit: It would be good to add a comment here describing what runtime invariant makes this situation impossible

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -122,8 +144,160 @@ impl Service for TonicTransport {
...
+
+    async fn invoke(
+        &self,
+        headers: RequestHeaders,
+        options: CallOptions,
+    ) -> (Self::SendStream, Self::RecvStream) {
+        let (req_tx, req_rx) = mpsc::channel(1);
+        let request_stream = ReceiverStream::new(req_rx);
+        let mut request = TonicRequest::new(Box::pin(request_stream));
+        *request.metadata_mut() = headers.metadata().clone();
```
arjan-bal: To avoid cloning here, we can add an `into_parts()` method to `RequestHeaders` that consumes self and returns the owned fields.

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -122,8 +144,160 @@ impl Service for TonicTransport {
...
+        // are received.  We must return a working send (and, consequently,
+        // recv) stream before this to allow the application to write its
+        // request(s), so we need to spawn a task for this period of time, and
+        // then we send the response (headers, stream) to the TonicRecvStream
+        // when it is available.
+        let (resp_tx, resp_rx) = oneshot::channel();
+        self.runtime.spawn(Box::pin(async move {
+            let response = grpc.streaming(request, path, BufCodec {}).await;
+            let _ = resp_tx.send(response);
+        }));
```
arjan-bal, line 184: Without the spawn, what causes the deadlock?
* The client waiting for request data before sending its request headers?
* The server waiting for request data before sending its response headers?
* Or something else entirely?

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -122,8 +144,160 @@ impl Service for TonicTransport {
...
+            return Ok(());
+        }
+        Err(())
+    }
+}
+
+struct TonicRecvStream {
+    error: Option<Status>,
+    response_rx: Option<oneshot::Receiver<Result<tonic::Response<Streaming<Bytes>>, TonicStatus>>>,
+    receiver: Option<Streaming<Bytes>>,
```
arjan-bal: The implementation of RecvStream is a little complex as the reader needs to keep track of when each of these fields are present. In Rust, enums make is easy to represent the state of a state machine. Here, we can have single field for the state:

```rust
enum StreamState {
    Error(tonic::Status),
    WaitingForHeaders(oneshot::Receiver<Result<tonic::Response<...>, ...>>),
    Streaming(tonic::codec::Streaming<...>),
    Closed,
}
```

This would also make impossible states unrepresentable. This is the implementation with the state enum from Gemini, I haven't reviewed it.
```rust
impl TonicRecvStream {
    async fn next(&mut self, msg: &mut dyn RecvMessage) -> ClientResponseStreamItem {
        // Take the current state, leaving `Closed` in its place temporarily
        let state = std::mem::replace(&mut self.state, StreamState::Closed);

        match state {
            StreamState::Error(err) => {
                // We leave the state as Closed
                ClientResponseStreamItem::Trailers(Trailers::new(err))
            }

            StreamState::WaitingForHeaders(rx) => {
                match rx.await {
                    Ok(Ok(response)) => {
                        let (metadata, stream, _) = response.into_parts();
                        // Transition to Streaming
                        self.state = StreamState::Streaming(stream);
                        ClientResponseStreamItem::Headers(ResponseHeaders::new().with_metadata(metadata))
                    }
                    Ok(Err(status)) => ClientResponseStreamItem::Trailers(Trailers::new(from_tonic_status(status))),
                    Err(_) => /* Handle cancelled */,
                }
            }

            StreamState::Streaming(mut stream) => {
                match stream.next().await {
                    Some(Ok(mut buf)) => {
                        match msg.decode(&mut buf) {
                            Ok(()) => {
                                // Put the stream back to keep streaming
                                self.state = StreamState::Streaming(stream);
                                ClientResponseStreamItem::Message(())
                            }
                            Err(_) => {
                                // State remains Closed
                                ClientResponseStreamItem::Trailers(/* Internal Error */)
                            }
                        }
                    }
                    Some(Err(status)) => ClientResponseStreamItem::Trailers(Trailers::new(from_tonic_status(status))),
                    None => ClientResponseStreamItem::Trailers(Trailers::new(Status::new(StatusCode::Ok, ""))),
                }
            }

            StreamState::Closed => ClientResponseStreamItem::StreamClosed,
        }
    }
}
```

grpc/src/client/transport/tonic/test.rs
```diff
@@ -146,16 +158,122 @@ pub(crate) async fn tonic_transport_rpc() {
...
+        },
+    )
+    .await
+    .unwrap();
+
+    // Response should be Headers, Message, Trailers (OK).
+    let mut resp = WrappedEchoResponse(EchoResponse::default());
+    let mut received = false;
+    let mut headers = false;
+    loop {
```
arjan-bal: We can simplify this by replacing the loop and the state variables with a series of `if/else` blocks:
```rust
// Response should be Headers, Message, Trailers (OK).
let mut resp = WrappedEchoResponse(EchoResponse::default());

// 1. Expect Headers first
let ClientResponseStreamItem::Headers(_) = rx.next(&mut resp).await else {
    panic!("Expected Headers first");
};

// 2. Expect the Message next
let ClientResponseStreamItem::Message(()) = rx.next(&mut resp).await else {
    panic!("Expected Message after Headers");
};
assert_eq!(resp.0.message, "hello interop");

// 3. Expect Trailers at the end
let ClientResponseStreamItem::Trailers(t) = rx.next(&mut resp).await else {
    panic!("Expected Trailers, got StreamClosed or other item");
};
assert_eq!(t.status().code(), crate::StatusCode::Ok, "RPC failed: {:?}", t.status());
```

grpc/src/client/channel.rs
```diff
@@ -324,7 +344,11 @@ impl ActiveChannel {
         let mut i = self.picker.iter();
         loop {
             if let Some(p) = i.next().await {
-                let result = &p.pick(&request);
+                let result = &p.pick(
+                    &RequestHeaders::new()
+                        .with_method_name(&method)
+                        .with_metadata(request.metadata().clone()),
```
arjan-bal, line 350: I think we should avoid cloning the metadata as it will allocate a new vector for the key/value pairs and increment a ref-count on all the underlying Bytes. A better solution could be to send an `Arc<RequestHeaders>` (or a wrapper struct) to the picker.

arjan-bal: I realized this method is getting removed in https://github.com/hyperium/tonic/pull/2536. Resolving.

grpc/src/client/subchannel.rs
dfawley: It's not. It should have been a TODO.  This situation is very possible if there's a race between picking a subchannel and it disconnecting.  (The `Service` impl has a todo.)  Updated to todo.

grpc/src/client/transport/tonic/mod.rs
dfawley: The client cannot send a request if we don't return the SendStream to it.  And the server may not return headers until a request is received (this is guaranteed to happen for any non-bidi RPC).  I tried to make the comment clearer; PTAL.

grpc/src/client/transport/tonic/mod.rs
dfawley: Done.  It has to be private for now because we haven't figured out everything `RequestHeaders` will hold yet.

grpc/src/client/transport/tonic/test.rs
dfawley: Good call.  Sadly `assert_matches!()` is unstable or it could have been even tidier.

grpc/src/client/transport/tonic/mod.rs
dfawley: Good idea. I've cleaned it up a bit and I think this is better.

PR #2529, opened 2026-03-02
feat(grpc): Call credentials API

arjan-bal:
This change introduces a [Call Credentials](https://grpc.io/docs/guides/auth/#credential-types) API for gRPC, allowing users to attach authentication metadata (such as OAuth or JWT tokens) to individual RPCs.

## API Design

* `CallCredentials`: The base trait that specific credential implementations must implement.
* Ergonomics: Public-facing APIs accept `Arc<dyn CallCredentials>`. 

## Why Reference Counting?
Call credentials must be reference-counted (`Arc`) to support sharing across out-of-band gRPC channels, such as those used by Load Balancers for load reporting.

## Functional Changes

* Added `fn get_call_credentials(&self) -> Option<&CallCredentials>` to the `ChannelCredentials` trait. This enables attaching call credentials to channel-level credentials that apply to every RPC.
* `CompositeCallCredentials`: A new utility for combining multiple call credentials.
* `CompositeChannelCredentials`: A new utility for attaching call credentials to an existing channel credentials object.


grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+            method_name,
+        }
+    }
+
+    /// Returns the base URL of the service for this RPC call.
+    pub fn service_url(&self) -> &str {
+        &self.service_url
+    }
+
+    /// The method name suffix (e.g., `Method` or `package.Service/Method`).
```
dfawley: This example lost me.  Why would the method suffix include the service and a slash?  Or did you typo `of->or`?

grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+use std::sync::Arc;
+
+use tonic::Status;
+use tonic::async_trait;
+use tonic::metadata::MetadataMap;
+
+use crate::attributes::Attributes;
+use crate::credentials::common::SecurityLevel;
+
+/// Details regarding the RPC call.
```
dfawley: Nit: "RPC call" -> "RPC" or just "call".

(x3)

grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// Generates the authentication metadata for a specific RPC call.
+    ///
+    /// This method is called by the transport layer on each request.
+    /// Implementations should populate the provided `metadata` map with the
+    /// necessary authorization headers (e.g., `authorization: Bearer <token>`).
+    ///
+    /// If this returns an `Err`, the RPC will fail immediately with a status
+    /// derived from the error if the status code is in the range defined in
+    /// gRFC A54. Otherwise, the RPC is failed with an internal status.
+    async fn get_metadata(
```
dfawley, line 114: Are we going to call this in a `select` with a timer?

If so this should have a section on cancel safety stating that all implementations must be cancel safe.

grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+        &self,
+        call_details: &CallDetails,
+        auth_info: &ChannelSecurityInfo,
+        metadata: &mut MetadataMap,
+    ) -> Result<(), Status>;
+
+    /// Indicates the minimum transport security level required to send
+    /// these credentials.
```
dfawley, line 122: Does the docstring need to call out the default value here?

grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+impl CallCredentialsProvider for CompositeCallCredentials {
+    async fn get_metadata(
+        &self,
+        call_details: &CallDetails,
+        auth_info: &ChannelSecurityInfo,
+        metadata: &mut MetadataMap,
+    ) -> Result<(), Status> {
+        for cred in &self.creds {
+            cred.get_metadata(call_details, auth_info, metadata).await?;
+        }
```
dfawley, line 163: Do we want to `futures::future::join_all` these so they are run (semi-)concurrently instead of serially?  (Or spawn tasks and join them as another option?)  Or document that they are run serially at least.

grpc/src/credentials/call.rs
```diff
@@ -0,0 +1,280 @@
...
+    fn minimum_channel_security_level(&self) -> SecurityLevel {
+        self.creds
+            .iter()
+            .map(|c| c.minimum_channel_security_level())
+            .max_by_key(|&l| match l {
+                SecurityLevel::NoSecurity => 0,
+                SecurityLevel::IntegrityOnly => 1,
+                SecurityLevel::PrivacyAndIntegrity => 2,
+            })
+            .unwrap_or(SecurityLevel::NoSecurity)
```
dfawley: `PrivacyAndIntegrity`?

Although, why would `unwrap` fail here?  Only if there were zero elements in `self.creds` (which is impossible)?

grpc/src/credentials/call.rs
arjan-bal: Fixed.

grpc/src/credentials/call.rs
arjan-bal: The "or" should have been "in". Fixed that now.

While gRPC Go [only sends the service URL](https://github.com/grpc/grpc-go/blob/1d4fa8a7b772553e82137b059ad4a8f632a1c522/internal/transport/http2_client.go#L671),  [after stripping the method ](https://github.com/grpc/grpc-go/blob/1d4fa8a7b772553e82137b059ad4a8f632a1c522/internal/transport/http2_client.go#L650-L663) to the call credentials, gRPC C++ also [sends the method](https://github.com/grpc/grpc/blob/060a21762295ba058c599a4e879996363bc8ca46/include/grpcpp/security/credentials.h#L272-L280) name separately. I've gone with the C++ approach here.

grpc/src/credentials/call.rs
arjan-bal: I hadn't thought of this. Added a section to the rustdoc stating that the future may be dropped due to RPC timeouts.

grpc/src/credentials/call.rs
arjan-bal: To run `get_metadata`, we would need to allocate a new metadata map for each credential and merge the results into the main map passed to `CompositeCallCredentials`. I suspect most implementations will have the headers cached, so calling them concurrently may not be worth the extra allocations. Java and Go both call the method sequentially.

I've mentioned the sequential execution in the struct's rustdoc,

grpc/src/credentials/call.rs
arjan-bal: Because the `CompositeCallCredentials` constructor requires two call credentials, the vector is guaranteed to be non-empty. I've replaced this with an `expect()` call and a clear message.

arjan-bal:
> I'm not sure whether I like the `impl Into<CallCredentials>`. It is nice in some ways, but:
> 
> 1. It adds an extra type (`CallCredentials` vs. `CallCredentialsProvider`) to the API surface.
> 2. It requires the user to know that there is a blanket `impl<T: Provider> From<T> for CC`.
> 3. It only saves the user needing to do an `Arc::new()` (or `.clone()`) when configuring the channel.
> 4. It means if the user already `Arcs` their `Provider`, e.g. if they're going to use the same call creds for multiple channels, then this arcs their arc, which is wasteful in the same ways you're trying to avoid.

I don't have strong opinions on this, and I see the benefit of having a simpler API surface. The double Arcing can be avoided if the user creates the wrapper object (CallCredentials) once, and passed it to the channel constructor.

I've replaced the wrapper type with `Arc<dyn CallCredentials>` directly.

grpc/src/credentials/call.rs
dfawley: Ping - I'm not sure what the documentation norms are for this kind of situation (default impls in traits) off the top of my head.

grpc/src/credentials/client.rs
```diff
@@ -214,8 +216,8 @@ impl<T: ChannelCredentials> ChannelCredsInternal for CompositeChannelCredentials
             .await
     }
 
-    fn get_call_credentials(&self) -> Option<&CallCredentials> {
-        Some(&self.call_creds)
+    fn get_call_credentials(&self) -> Option<Arc<dyn CallCredentials>> {
+        Some(self.call_creds.clone())
     }
```
dfawley: I'm not sure if it's worth it but this could still return `Option<&Arc<dyn CallCredentials>>` to delay the clone as much as possible.  (That's effectively what you were doing before, too.)

grpc/src/credentials/call.rs
arjan-bal: Added a section in the rustdoc.

grpc/src/credentials/client.rs
arjan-bal: Done.

PR #2530, opened 2026-03-02
feat(grpc): TLS Server Credentials

arjan-bal:
This change implements the ServerCredentials trait using ruslts, allowing users to configure (m)TLS on the server side.

### DIFF
--- a/grpc/src/client/channel.rs
+++ b/grpc/src/client/channel.rs
@@ -261,7 +261,6 @@ impl PersistentChannel {
 }
 
 struct ActiveChannel {
-    cur_state: Mutex<ConnectivityState>,
     abort_handle: Box<dyn rt::TaskHandle>,
     picker: Arc<Watcher<Arc<dyn Picker>>>,
     connectivity_state: Arc<Watcher<ConnectivityState>>,
@@ -284,8 +283,6 @@ impl ActiveChannel {
             connectivity_state.clone(),
             runtime.clone(),
         );
-
-        let resolver_helper = Box::new(tx.clone());
 
         // TODO(arjan-bal): Return error here instead of panicking.
         let rb = global_registry().get(target.scheme()).unwrap();
@@ -315,7 +312,6 @@ impl ActiveChannel {
         }));
 
         Arc::new(Self {
-            cur_state: Mutex::new(ConnectivityState::Connecting),
             abort_handle: jh,
             picker: picker.clone(),
             connectivity_state: connectivity_state.clone(),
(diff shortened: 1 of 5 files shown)


examples/data/tls/create.sh
```diff
@@ -24,7 +24,7 @@ openssl req -x509 \
   -extensions test_ca \
   -sha256
 
-# Generate one server cert.
+# Generate two server cert.
```
dfawley:
```suggestion
# Generate two server certs.
```

grpc/src/credentials/rustls/server/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+#[non_exhaustive]
+pub enum TlsClientCertificateRequestType<R = StaticRootCertificatesProvider> {
+    /// Server does not request client certificate.
+    ///
+    /// This is the default behavior.
+    ///
+    /// The certificate presented by the client is not checked by the server at
+    /// all. (A client may present a self-signed or signed certificate or not
+    /// present a certificate at all and any of those option would be accepted).
+    DontRequestClientCertificate,
```
dfawley: Are these names from elsewhere or can we bikeshed?  I think we could remove `ClientCertificate` from all the enum values:

`TlsClientCertificateRequestType::DontRequest`
`TlsClientCertificateRequestType::RequestAndVerify`
`TlsClientCertificateRequestType::RequireAndVerify`

Or something along those lines?

grpc/src/credentials/rustls/server/test.rs
```diff
@@ -0,0 +1,719 @@
...
+        result.err()
+    );
+
+    let mut tls_stream = result.unwrap();
+    tls_stream.write_all(b"ping!").await.unwrap();
+    let mut buf = [0u8; 5];
+    tls_stream.read_exact(&mut buf).await.unwrap();
+    assert_eq!(&buf, b"pong!");
+
+    let _ = server_task.await;
```
dfawley: I'm pretty sure this is making the test ignore an assertion failure in the spawned task -- i.e. this should be `server_task.await.unwrap();` instead.  Same in the tests below.

grpc/src/credentials/rustls/server/mod.rs
arjan-bal: These names are based on the [C++ API](https://github.com/grpc/grpc/blob/9c9db8a349c1383b2ae9f0725dc8f02354e576da/include/grpc/grpc_security_constants.h#L79-L126). However, I agree the shorter names look better, so I've updated them.

grpc/src/credentials/rustls/server/test.rs
arjan-bal: Oh, I didn't know that. Fixed all the usages in the client and server TLS tests.

PR #2531, opened 2026-03-02
Add missing license files for the public crates

LecrisUT:
As noted in https://github.com/hyperium/tonic/pull/2321#issuecomment-3960424068, these license symlinks were missed when the crate split happened.

I am not sure if I identified the public crates correctly, but I went by if the crate had a `description` or not, (`grpc`, `codegen` and `interop` were hence omitted).

I need to point out though that the license file is not technically correct for `tonic-protobuf` because it seems the copyright authors do not match (`gRPC Authors` vs `Lucio Franco`). It can be fine if it implies that the contributors have yielded their copyright, but `gRPC Authors` being quite the external entity, I don't think it applies here? There is even a header with the relevant copyright indicated. That's why I added a new license file in there.

Alternatively, you can make the top-level `LICENSE` stripped of copyright notice and instead add the copyright notice in the source files themselves. [Reuse](https://reuse.software/tutorial/) is also an interesting project to consider to make sure the appropriate license/copyright headers are included.


nathanielford:
Thanks for putting this up - it was superceded by recent PRs, so I'm going to close it, but we appreciate the contribution!

PR #2533, opened 2026-03-02
feat(tonic-xds): implement xds resources and route matchers

YutaoMa:
## Motivation

Ref: #2444 

### xDS Resource Types and Route Matching

`xds-client` (PR #2478 ) handles the ADS protocol but knows nothing about gRPC. `tonic-xds` has its Tower service stack scaffolded but lacks the middle layer bridging xDS updates to routing decisions. This PR builds that foundation: validated resource types and a route matching engine that the upcoming resource manager will use to drive the LDS→RDS→CDS→EDS cascade and feed the existing Tower services.

This is the 1st PR of a series to complete `tonic-xds`. The next steps include:
- xDS Resource cache
- Endpoint management: EDS diffing into `tower::discover::Change` streams
- Resource manager: cascade task managing LDS→RDS→CDS→EDS watches, implementing `XdsRouter` + `XdsClusterDiscovery`
- Bootstrap config: JSON parsing from env vars → `xds_client::ClientConfig`
- Builder wiring: wire full stack, E2E test
- Advanced routing (gRFC A28+A63): Weighted cluster selection
- Least request LB (gRFC A48)

## Solution

Implements the xds_client::Resource trait for the four gRPC xDS resource types and adds a route compilation/matching engine, bridging xds-client watches to validated, efficient routing structures:

- **Resource types** `xds/resource/`: `ListenerResource` (LDS), `RouteConfigResource` (RDS), `ClusterResource` (CDS), `EndpointsResource` (EDS) — each implements `xds_client::Resource` with two-phase decode (deserialize + validate).
- **Route matching** `xds/route/`: `CompiledRouteConfig` pre-compiles a `RouteConfigResource` at RDS update time (regex compilation, domain indexing), allows for allocation-free per-request matching. 
- **Refactors**: `RouteInput`/`RouteDecision` moved to `client/route` (since it's used for Tower layer specifically, rather than the xDS subscription side of route logic this PR deals with).

### DIFF
--- a/tonic-xds/src/xds/resource/cluster.rs
+++ b/tonic-xds/src/xds/resource/cluster.rs
@@ -20,8 +20,8 @@ pub(crate) struct ClusterResource {
 /// Load balancing policies.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum LbPolicy {
-    RoundRobin {},
-    LeastRequest {},
+    RoundRobin,
+    LeastRequest,
 }
 
 impl Resource for ClusterResource {
@@ -51,8 +51,8 @@ impl Resource for ClusterResource {
             .filter(|s| !s.is_empty());
 
         let lb_policy = match cluster::LbPolicy::try_from(message.lb_policy) {
-            Ok(cluster::LbPolicy::RoundRobin) => LbPolicy::RoundRobin {},
-            Ok(cluster::LbPolicy::LeastRequest) => LbPolicy::LeastRequest {},
+            Ok(cluster::LbPolicy::RoundRobin) => LbPolicy::RoundRobin,
+            Ok(cluster::LbPolicy::LeastRequest) => LbPolicy::LeastRequest,
             _ => {
                 return Err(Error::Validation(format!(
                     "unsupported load balancing policy: {}",
@@ -95,7 +95,7 @@ mod tests {
         let cluster = make_cluster("my-cluster");
         let validated = ClusterResource::validate(cluster).expect("should validate");
         assert_eq!(validated.name, "my-cluster");
-        assert_eq!(validated.lb_policy, LbPolicy::RoundRobin {});
+        assert_eq!(validated.lb_policy, LbPolicy::RoundRobin);
         assert!(validated.eds_service_name.is_none());
     }
 
@@ -130,7 +130,7 @@ mod tests {
             ..Default::default()
         };
         let validated = ClusterResource::validate(cluster).unwrap();
-        assert_eq!(validated.lb_policy, LbPolicy::LeastRequest {});
+        assert_eq!(validated.lb_policy, LbPolicy::LeastRequest);
     }
 
     #[test]


tonic-xds/Cargo.toml
```diff
@@ -28,14 +28,18 @@ dashmap = "6.1"
 thiserror = "2.0.17"
 url = "2.5.8"
 futures-core = "0.3.31"
+bytes = "1"
+xds-client = { path = "../xds-client" }
```
gu0keno0: Better to pin to a version?

gu0keno0: Should use a published version?

tonic-xds/src/xds/resource/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+
+use bytes::Bytes;
+use envoy_types::pb::envoy::config::cluster::v3::{Cluster, cluster};
+use prost::Message;
+use xds_client::resource::TypeUrl;
+use xds_client::{Error, Resource};
+
+/// Validated Cluster resource.
+#[derive(Debug, Clone)]
+pub(crate) struct ClusterResource {
```
gu0keno0, line 11: TLS config will be useful. In the future we need TLS to (re)connect.

tonic-xds/src/xds/resource/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+            eds_service_name,
+            lb_policy,
+        })
+    }
+}
+
+impl ClusterResource {
+    /// Returns the EDS service name for cascading EDS subscriptions.
+    /// Falls back to the cluster name if no EDS service name is set.
+    pub(crate) fn cascade_eds_service_name(&self) -> &str {
```
gu0keno0: nit
```suggestion
    pub(crate) fn eds_service_name(&self) -> &str {
```

tonic-xds/src/xds/resource/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub eds_service_name: Option<String>,
+    /// The load balancing policy for this cluster.
+    pub lb_policy: LbPolicy,
+}
+
+/// Load balancing policies.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum LbPolicy {
+    RoundRobin,
+    LeastRequest,
```
gu0keno0, line 24: Probably not critical for now, yet these should be struct variant.

tonic-xds/src/xds/resource/listener.rs
```diff
@@ -0,0 +1,244 @@
...
+            RouteSpecifier::ScopedRoutes(_) => Err(Error::Validation(
+                "scoped_routes not supported for gRPC".into(),
+            )),
+        }
+    }
+}
+
+impl ListenerResource {
+    /// Returns the RDS route config name for cascading subscriptions.
+    pub(crate) fn cascade_route_config_name(&self) -> Option<&str> {
```
gu0keno0: Nit: cascade

tonic-xds/src/xds/resource/endpoints.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Validated ClusterLoadAssignment (EDS resource).
+#[derive(Debug, Clone)]
+pub(crate) struct EndpointsResource {
+    pub cluster_name: String,
+    pub localities: Vec<LocalityEndpoints>,
+}
+
+/// Endpoints within a locality.
+#[derive(Debug, Clone)]
+pub(crate) struct LocalityEndpoints {
```
gu0keno0, line 25: Add Locality? 

https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/base.proto#envoy-v3-api-msg-config-core-v3-locality

tonic-xds/src/xds/resource/endpoints.rs
```diff
@@ -0,0 +1,260 @@
...
+            cluster_name: String::new(),
+            ..Default::default()
+        };
+        let err = EndpointsResource::validate(cla).unwrap_err();
+        assert!(err.to_string().contains("cluster_name"));
+    }
+
+    #[test]
+    fn test_not_all_resources_required() {
+        assert!(!EndpointsResource::ALL_RESOURCES_REQUIRED_IN_SOTW);
```
gu0keno0, line 262: what's this?

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub headers: Vec<HeaderMatcherConfig>,
+    pub case_sensitive: bool,
+}
+
+/// Path matching specifier.
+#[derive(Debug, Clone)]
+pub(crate) enum PathSpecifierConfig {
+    Prefix(String),
+    Path(String),
+    SafeRegex(String),
```
gu0keno0: Shall we just use regex here?

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    fn name(message: &Self::Message) -> &str {
+        &message.name
+    }
+
+    fn validate(message: Self::Message) -> xds_client::Result<Self> {
+        let name = message.name;
+        let mut virtual_hosts = Vec::with_capacity(message.virtual_hosts.len());
+
+        for vh in message.virtual_hosts {
```
gu0keno0, line 113: do we want to allow 0-length VH array?

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -0,0 +1,473 @@
...
+        Some(HeaderMatchSpecifier::SafeRegexMatch(r)) => {
+            HeaderMatchSpecifierConfig::SafeRegex(r.regex)
+        }
+        Some(HeaderMatchSpecifier::PresentMatch(_)) => HeaderMatchSpecifierConfig::Present,
+        Some(HeaderMatchSpecifier::StringMatch(sm)) => match sm.match_pattern {
+            Some(MatchPattern::Exact(v)) => HeaderMatchSpecifierConfig::Exact(v),
+            Some(MatchPattern::Prefix(v)) => HeaderMatchSpecifierConfig::Prefix(v),
+            Some(MatchPattern::Suffix(v)) => HeaderMatchSpecifierConfig::Suffix(v),
+            Some(MatchPattern::Contains(v)) => HeaderMatchSpecifierConfig::Contains(v),
+            Some(MatchPattern::SafeRegex(r)) => HeaderMatchSpecifierConfig::SafeRegex(r.regex),
```
gu0keno0: if we use regex directly, here we can just construct regex and validate it naturally

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -0,0 +1,473 @@
...
+        )),
+        None => Err(Error::Validation(
+            "route action missing cluster specifier".into(),
+        )),
+    }
+}
+
+impl RouteConfigResource {
+    /// Returns cluster names referenced by this route configuration for cascading CDS subscriptions.
+    pub(crate) fn cascade_cluster_names(&self) -> HashSet<String> {
```
gu0keno0: Nit: cascade

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -0,0 +1,473 @@
...
+                ..Default::default()
+            }],
+            ..Default::default()
+        };
+        let validated = RouteConfigResource::validate(rc).unwrap();
+        assert!(matches!(
+            validated.virtual_hosts[0].routes[0]
+                .match_criteria
+                .path_specifier,
+            PathSpecifierConfig::Path(_)
```
gu0keno0: Just do a exact match?

tonic-xds/src/xds/route/compiler.rs
```diff
@@ -0,0 +1,462 @@
+//! Compiles a [`RouteConfigResource`] into an efficient per-request matching structure.
+//!
+//! Compilation happens once at RDS update time — regex patterns are pre-compiled
```
gu0keno0: What are other optimizations? Seems like most of the structures are duplicated

tonic-xds/Cargo.toml
ankurmittal: it needs to be versioned, else cargo publish will fail

tonic-xds/src/xds/resource/cluster.rs
YutaoMa: Good point, changed to empty struct for now

tonic-xds/src/xds/resource/endpoints.rs
YutaoMa: This is related to [gRFC A88](https://github.com/grpc/proposal/blob/master/A88-xds-data-error-handling.md)'s behavior on how xds client should react to partial response, specifically LDS and CDS enforces all resources to be sent in SotW (and missing will be treated as error) while RDS and EDS do not enforces this (and missing will be ignored and continue to use previously cached value). I've updated the code comment to clarify this

tonic-xds/src/xds/resource/route_config.rs
YutaoMa: Makes sense, this also ties to the duplicate code comment on the compiled config below. I moved regex parsing into this resource level so we no longer need the additional compiled config types. Instead I refactored the routing logic to a `routing` module that handles specifically matching and routing logic.

tonic-xds/src/xds/resource/cluster.rs
YutaoMa: Left unchanged for now, will design TLS config API and add accordingly in future PRs

tonic-xds/src/xds/resource/route_config.rs
YutaoMa: Updated in the new refactoring

tonic-xds/src/xds/resource/route_config.rs
YutaoMa: A27 or later gRFCs didn't specifically clarify the behavior at resource validation level (they did however explain how the RPCs should error when VH can't be find which is fine with the current code), it does make sense to reject the config early to help catch misconfiguration so I added the 0-length check validation as suggested.

tonic-xds/src/xds/route/compiler.rs
YutaoMa: it was mainly regex parsing, therefore I removed this compiled config level after moving the regex into resource validation.

tonic-xds/src/xds/resource/route_config.rs
YutaoMa: updated to use conditional match branch. comparing the config struct is a little tricky given I couldn't derive `PartialEq` directly because of `Regex`, so settling for that approach for now.

tonic-xds/src/xds/resource/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+    /// If not set, the cluster name is used.
+    pub eds_service_name: Option<String>,
+    /// The load balancing policy for this cluster.
+    pub lb_policy: LbPolicy,
+}
+
+/// Load balancing policies.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum LbPolicy {
+    RoundRobin {},
```
ankurmittal: why `{}` ?

tonic-xds/src/xds/resource/endpoints.rs
```diff
@@ -0,0 +1,278 @@
...
+use prost::Message;
+use xds_client::resource::TypeUrl;
+use xds_client::{Error, Resource};
+
+use crate::client::endpoint::EndpointAddress;
+
+/// Validated ClusterLoadAssignment (EDS resource).
+#[derive(Debug, Clone)]
+pub(crate) struct EndpointsResource {
+    pub cluster_name: String,
```
ankurmittal, line 19: do we need Clone here? how many times will this be cloned? as we are already allocating for String, might be better to switch to `Arc<str>` if this is going to be cloned a lot

tonic-xds/src/xds/resource/endpoints.rs
```diff
@@ -0,0 +1,278 @@
...
+pub(crate) struct LocalityEndpoints {
+    pub locality: Option<Locality>,
+    pub endpoints: Vec<ResolvedEndpoint>,
+    pub load_balancing_weight: u32,
+    pub priority: u32,
+}
+
+/// Locality information for a set of endpoints.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct Locality {
```
ankurmittal, line 34: Same question here

tonic-xds/src/xds/resource/listener.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Route configuration embedded inline in the listener.
+    Inline(RouteConfigResource),
+}
+
+/// Validated Listener resource.
+///
+/// Extracts the route source from the
+/// `ApiListener` -> `HttpConnectionManager` -> `route_specifier` chain.
+#[derive(Debug, Clone)]
+pub(crate) struct ListenerResource {
```
ankurmittal, line 28: Same

tonic-xds/src/xds/resource/cluster.rs
```diff
@@ -0,0 +1,162 @@
...
+        let eds_service_name = message
+            .eds_cluster_config
+            .map(|eds| eds.service_name)
+            .filter(|s| !s.is_empty());
+
+        let lb_policy = match cluster::LbPolicy::try_from(message.lb_policy) {
+            Ok(cluster::LbPolicy::RoundRobin) => LbPolicy::RoundRobin {},
+            Ok(cluster::LbPolicy::LeastRequest) => LbPolicy::LeastRequest {},
+            // Default to round-robin for unsupported policies.
+            _ => LbPolicy::RoundRobin {},
```
mingley: per https://grpc.github.io/grpc/core/md_doc_grpc_xds_features.html

```
A gRPC client will NACK the configuration that contains unsupported cluster load balancing policy.
```

Should we handle this differently than silently defaulting to round-robin then?

tonic-xds/src/xds/resource/endpoints.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    let host_identifier = match lb_ep.host_identifier {
+        Some(lb_endpoint::HostIdentifier::Endpoint(ep)) => ep,
+        // Named endpoints not supported.
+        _ => return Ok(None),
+    };
+
+    let addr = host_identifier
+        .address
+        .ok_or_else(|| Error::Validation("endpoint missing address".into()))?;
```
mingley, line 143: - host_identifier = `None`:
    returns `Ok(None)`, no validation error
- host_identifier = `Some(Endpoint(ep)) but ep.address = None`:
    returns `Err("endpoint missing address")`
 - host_identifier = `Some(...)` with an unsupported variant like named endpoint:
    returns `Ok(None)`, no validation error

Does this matter?

should we reject an unsupported or missing host_identifier?

tonic-xds/src/xds/resource/cluster.rs
YutaoMa: Per above PR comment, since we expect LbPolicy to be struct variants (so they could include specific LB configs), I'm pre-structuring these to be struct variants

tonic-xds/src/xds/resource/endpoints.rs
YutaoMa: `Clone` should not be needed given all resource types will be wrapped with `Arc` in `xds-client` to be shared with `tonic-xds` subscribing logic. I'll remove the `Clone` derive while keeping the `String`s simple.

tonic-xds/src/xds/resource/endpoints.rs
YutaoMa: See above

tonic-xds/src/xds/resource/listener.rs
YutaoMa: See above

tonic-xds/src/xds/resource/cluster.rs
YutaoMa: good point, updated to fail the validation to surface the error

tonic-xds/src/xds/resource/endpoints.rs
YutaoMa: The current behavior is intentional, I've updated the comment to explain

- host_identifier = None or unsupported variant → skip `Ok(None)`: Could be alternative endpoint type that is not supported in gRPC but used by other client types like Envoy. Since EDS uses the "partial failure doesn't break the entire config" semantic, we skip those.
- Endpoint variant with missing/invalid address → `error (Err)`: "I understand this type but it's malformed." This is a control plane bug worth surfacing via NACK.

This matches grpc-go's behavior in `unmarshal_eds.go`.

tonic-xds/Cargo.toml
```diff
@@ -28,14 +28,18 @@ dashmap = "6.1"
 thiserror = "2.0.17"
 url = "2.5.8"
 futures-core = "0.3.31"
+bytes = "1"
+xds-client = { version = "0.1.0-alpha.1", path = "../xds-client" }
+envoy-types = "0.7"
```
LucioFranco, line 35: we probably want to not depend on this external crate?

LucioFranco (approved):

tonic-xds/Cargo.toml
YutaoMa: I could vendor in the relevant proto files and run `prost-build` in `build.rs`. Would that be your recommendation?

tonic-xds/src/xds/resource/cluster.rs
ankurmittal: so just include a empty struct? or is the struct definition going to be inline in enum?

YutaoMa: yeah a tuple variant wrapping a struct might be better for organizing the code, let me just remove the `{}` for now and add the tuple variants later

tonic-xds/Cargo.toml
YutaoMa: Discussed with the team, will move to in-library binding generation in later PR.

PR #2536, opened 2026-03-04
grpc: implement Handle for servers and remove old Service abstraction

dfawley:
Note: This PR is based on top of #2528, so review can wait until after that is approved (or close to ready).

Also:
- Update `inmemory` example to use new abstractions.
- Remove `multiaddr` example; update `inmemory` example to handle multiple addresses instead (there was a lot of code duplication).

The `inmemory` transport implementation is doing some fishy things right now.  But "it works".  It's actually covering up bugs in the subchannel implementation that are, I believe, holding circular references between internal/external subchannels and preventing proper shutdown from occurring.  This is a pretty major issue, but it needs to be solved before the inmemory transport can be fixed.  I will start working on that problem next, before dealing with the protobuf codegen for clients.

### DIFF
--- a/grpc/src/lib.rs
+++ b/grpc/src/lib.rs
@@ -44,10 +44,10 @@ pub use status::ServerStatus;
 pub use status::Status;
 pub use status::StatusCode;
 
-pub mod attributes;
-pub mod byte_str;
-pub mod rt;
-pub mod send_future;
+mod attributes;
+mod byte_str;
+mod rt;
+mod send_future;
 
 #[cfg(test)]
 mod echo_pb {


grpc/src/client/transport/tonic/mod.rs
```diff
@@ -488,6 +415,27 @@ impl Future for ResponseFuture {
...
+
+    fn encoder(&mut self) -> Self::Encoder {
+        BytesEncoder {}
+    }
+
+    fn decoder(&mut self) -> Self::Decoder {
+        BytesDecoder {}
+    }
+}
+
```
arjan-bal: I believe this type can be removed. The `BufCodec` servers the same purpose as this now.

grpc/src/lib.rs
```diff
@@ -35,18 +35,20 @@ pub mod client;
...
-mod codec;
-mod rt;
-mod send_future;
-pub mod server;
-pub mod service;
+
+pub(crate) mod attributes;
+pub(crate) mod byte_str;
+pub(crate) mod rt;
+pub(crate) mod send_future;
```
arjan-bal: Using `pub(crate)` in the root lib.rs file is essentially the same as making the module private. In both cases, only descendants of lib.rs can access it. I removed the `pub(crate)` modifiers in a previous PR.

grpc/src/server/mod.rs
```diff
@@ -178,7 +187,7 @@ pub struct SendOptions {
 
 /// Represents the receiving side of a server stream.
 #[trait_variant::make(Send)]
-pub trait RecvStream {
+pub trait RecvStream: Send {
```
arjan-bal: This : Send seems redundant. The trait_variant::make(Send) macro should already be adding this bound. Same for `SendStream` above.

See https://docs.rs/trait-variant/latest/trait_variant/attr.make.html

grpc/src/rt/mod.rs
```diff
@@ -37,6 +39,7 @@ pub type BoxFuture<T> = Pin<Box<dyn Future<Output = T> + Send>>;
 pub type BoxedTaskHandle = Box<dyn TaskHandle>;
 pub type BoxEndpoint = Box<dyn GrpcEndpoint>;
 pub type ScopedBoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + Send + 'a>>;
+pub type SharedService = Arc<dyn DynInvoke>;
```
arjan-bal: The runtime module doesn't seem to be the right place to define this type alias. I think we should move it the module containing `DynInvoke` trait.

grpc/src/client/subchannel.rs
```diff
@@ -412,10 +388,10 @@ impl InternalSubchannel {
                 _ = runtime.sleep(min_connect_timeout) => {
                     let _ = state_machine_tx.send(SubchannelStateMachineEvent::ConnectionTimedOut);
                 }
-                result = transport.connect(address.to_string().clone(), runtime, &transport_opts) => {
+                result = transport.dyn_connect(address.to_string().clone(), runtime, &transport_opts) => {
                     match result {
                         Ok(s) => {
-                            let _ = state_machine_tx.send(SubchannelStateMachineEvent::ConnectionSucceeded(Arc::from(s.service), s.disconnection_listener));
+                            let _ = state_machine_tx.send(SubchannelStateMachineEvent::ConnectionSucceeded(Arc::from(s.0), s.1));
```
arjan-bal: nit: I prefer a struct with named fields instead of a tuple struct with `.0` and `.1` for better readability.

grpc/src/rt/mod.rs
arjan-bal: This also seems to be the same type as `SharedInvoke` in `grpc/src/client/subchannel.rs`.

grpc/src/client/stream_util.rs
```diff
@@ -179,6 +179,7 @@ mod test {
     use crate::client::interceptor::InvokeOnceExt as _;
     use crate::core::ClientResponseStreamItem;
     use crate::core::RecvMessage;
+    use crate::core::RequestHeaders;
```
arjan-bal: This seems like an unused import which can be reverted.

grpc/src/client/service_config.rs
```diff
@@ -27,8 +27,18 @@ use std::sync::Arc;
...
+pub struct ServiceConfig {
+    pub load_balancing_policy: Option<LbPolicyType>,
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub enum LbPolicyType {
+    #[default]
+    PickFirst,
+    RoundRobin,
+}
```
arjan-bal, line 41: Does the presence of this enum mean that we won't support third party load balancing policies in the future?

grpc/src/client/service_config.rs
```diff
@@ -27,8 +27,18 @@ use std::sync::Arc;
 
 /// An in-memory representation of a service config, usually provided to gRPC as
 /// a JSON object.
+// TODO: this shouldn't be public; users should set with JSON instead.
 #[derive(Debug, Default, Clone)]
-pub(crate) struct ServiceConfig;
+pub struct ServiceConfig {
```
arjan-bal: Why do we need to make it `pub`? The build seems to work when it's `pub(crate)`.

grpc/src/client/channel.rs
```diff
@@ -587,10 +545,15 @@ impl LbController {
...
         controller: &mut InternalChannelController,
     ) -> Result<(), Box<dyn Error + Send + Sync>> {
-        if update.service_config.as_ref().is_ok_and(|sc| sc.is_some()) {
-            return Err("can't do service configs yet".into());
+        let mut policy_name = pick_first::POLICY_NAME;
+        if let Ok(Some(service_config)) = update.service_config.as_ref()
+            && service_config
+                .load_balancing_policy
+                .as_ref()
+                .is_some_and(|p| *p == LbPolicyType::RoundRobin)
```
arjan-bal, line 553: Depending on whether we want to allow custom LBs, I can think of two better ways:
1. Use the LbPolicyType enum as the key to the `GLOBAL_LB_REGISTRY`, or
2. Have a function that maps `LbPolicyType` to an `&'static str`

The seems to be easier to extend when we add more LBs.

grpc/src/client/service_config.rs
dfawley: I was going to use this in the example to get RR, but then I ended up just making the inmemory transport use it automatically.  I'll revert it.

grpc/src/client/channel.rs
dfawley: I was "implementing support for the `loadBalancingPolicy` field" with this:

https://github.com/grpc/grpc-proto/blob/dafe8c0aa9c80e28da7808ffade0f33638a16e3d/grpc/service_config/service_config.proto#L600

Apparently Java supports this being any string, so @nathanielford will update this later to work similarly.

But this is all here just so I can do RR in the example.

grpc/src/client/service_config.rs
dfawley: (See other comment below.)

grpc/src/client/stream_util.rs
dfawley: Oh... I didn't realize `use super::*` was pulling in super's `use`s too.

This is going to happen a _lot_, then:

1. Add some code
2. Add a test module
3. `use` stuff in the test module not in the parent
4. Add more code
5. Add more `use`s in the outer module that overlaps a `use` in the test module

I've removed the redundant `use`s here, but I kind of think we shouldn't rely on this and think it's fine or even better to have the redundant `use`.

Because otherwise, if we then continue:

6. Remove `use` in test module
7. Realize new code belongs elsewhere or can be refactored
8. Remove `use` in outer module

Then now the tests break.

grpc/src/client/subchannel.rs
dfawley: It's not defined as a tuple struct; it's a compound returned value.  I'll destructure in the match arm instead, but I agree this is too vague otherwise.

grpc/src/client/transport/tonic/mod.rs
dfawley: Correct; thanks!

grpc/src/lib.rs
dfawley: Private mods are the same as pub(crate) mods?  TIL. Changed.

grpc/src/server/mod.rs
dfawley: Neat; thanks!  I deleted all the redundant `Send` bounds.

grpc/src/rt/mod.rs
dfawley: Sorry I missed cleaning up this from some long-ago temporary state (this PR has been through a lot!).

grpc/src/lib.rs
```diff
@@ -35,18 +35,20 @@ pub mod client;
...
-mod codec;
-mod rt;
-mod send_future;
-pub mod server;
-pub mod service;
+
+pub mod attributes;
+pub mod byte_str;
+pub mod rt;
+pub mod send_future;
```
arjan-bal: These should not be `pub`, simply `mod`.

dfawley: Fixed; thanks!

PR #2537, opened 2026-03-05
grpc: fix shutdown ref cycles in Channel and Subchannel

dfawley:
- Channel ref cycle:
  - LbController impl'd WorkScheduler which was given to child LB policy
  - LbController owns child LB policy directly

- Subchannel ref cycle:
  - Spawned task holds a ref to the InternalSubchannel
  - Task relies on receiver erroring to exit
  - Sender is held by the InternalSubchannel, creating a direct reference cycle

---

- Fixed Channel by creating a separate type to impl WorkScheduler

- Fixed Subchannel:
  - Removed the "one task" entirely
  - Restructured so that an Arc<Mutex<Inner>> handles all the state transitions
  - Any spawned tasks await a signal that the outer InternalSubchannel was dropped

### DIFF
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -538,7 +538,7 @@ impl Subchannel for ExternalSubchannel {
 
     fn connect(&self) {
         println!("connect called for subchannel: {self}");
-        self.isc.as_ref().unwrap().connect(false);
+        self.isc.as_ref().unwrap().connect();
     }
 }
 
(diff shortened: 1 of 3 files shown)


nathanielford:
(Notes from in-person discussion)
* The InternalSubchannel/InnerSubchannel distinction is confusing and has the scent of being too complicated.
* It's probable that what we want is that the task managing the FSM should own the InnerSubchannel, allowing us to skip the data Arc. 
* This would mean the InternalSubchannel should just have send ability to the InnerSubchannel. If we have a watcher that can have the transport filled in, that should be sufficient. This gets us out of both of having a loop that prevents the task from shutting down and also the awkward ownership/packaging the Internal/Inner subchannels has at present.

This refactor should happen on it's own, after this PR. Given that this removes the cyclical references and unblocks us, we're going to push it through.

nathanielford (approved):
(See previous notes - there is work to simplify this area of code, but the core pain point is handled for now!)

PR #2543, opened 2026-03-10
fix: map no trailers ok status to unknown

rumenov:
## Motivation

The change updates gRPC status handling so that HTTP 200 responses missing grpc-status trailers are treated as internal errors instead of silently completing. A new integration test simulates a truncated response to ensure clients surface the issue. This prevents data loss scenarios from being masked as successful completions.
Clients will now see explicit internal errors when a streaming response ends without grpc-status trailers, preventing silent data loss. The new integration test guards this behavior against future regressions.

## Solution

1. gRPC Status Mapping
Adjusted tonic’s infer_grpc_status logic to treat missing trailers as protocol violations.
http 200 responses without grpc-status trailers now return an internal Status error instead of appearing as clean end-of-stream.
Expanded comments explaining protocol expectations and rationale for mapping to Internal.
Clarified that h2::Reason::NO_ERROR on RST_STREAM signals missing grpc-status trailers.

2. Integration Test Coverage
Added an integration test that simulates a truncated server response to verify the new error mapping.
Introduced TruncatedBody and tower layer to cut off the response before trailers.
Created missing_grpc_status_trailer_is_internal_error test ensuring client surfaces tonic::Code::Internal.
Verified error message contains 'missing grpc-status trailer' to guard against regressions.

### DIFF
--- a/tests/integration_tests/tests/status.rs
+++ b/tests/integration_tests/tests/status.rs
@@ -398,7 +398,11 @@ async fn missing_grpc_status_trailer_is_unknown_error() {
     //
     //      tonic's decoder sees None with an empty buffer, calls
     //      `infer_grpc_status(trailers=None, status=200)`, and returns
-    //      `Code::Unknown` — the expected behaviour.
+    //      `Code::Unknown` because it is not able to observe the
+    //      RST_STREAM(NO_ERROR) at this time and only sees a stream end
+    //      successfully but without trailers containing grpc-status.
+    //      TODO: this should expect Code::Internal instead.
+
     let (headers_acked_tx, headers_acked_rx) = tokio::sync::oneshot::channel::<()>();
 
     tokio::spawn(async move {


tonic/src/status.rs
```diff
@@ -796,13 +798,27 @@ pub(crate) fn infer_grpc_status(
...
-        // https://github.com/hyperium/tonic/issues/681
-        http::StatusCode::OK => return Err(None),
+        // https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#responses
+        //
+        // Note: the old behaviour (returning Err(None) == "stream ended cleanly") was needed
+        // to handle the edge case in https://github.com/hyperium/tonic/issues/681 where a
+        // misbehaving server sends two Status frames. That case is already handled by the
+        // code above that parses grpc-status from whatever trailers *are* present, so
+        // changing this branch to a hard error is safe.
+        http::StatusCode::OK => return Err(Some(Status::internal(
```
dfawley: As mentioned in the issue, this should be `unknown` instead of `internal`.

(See https://github.com/grpc/grpc-go/pull/8702 for Go's conscious decision to change it to `Unknown` after it had inadvertently been changed to `Internal`.)

tests/integration_tests/tests/status.rs
```diff
@@ -351,3 +352,153 @@ async fn message_and_then_status_from_server_stream() {
     assert_eq!(stream.message().await.unwrap_err().message(), "foo");
     assert_eq!(stream.message().await.unwrap(), None);
 }
+
+// ---------------------------------------------------------------------------
+// Bug fix: HTTP 200 response with no grpc-status trailer must surface as
+// Internal, not silently treated as a clean end-of-stream.
```
dfawley: (Unknown, as below)

tests/integration_tests/tests/status.rs
```diff
@@ -351,3 +352,153 @@ async fn message_and_then_status_from_server_stream() {
...
+//
+// We simulate this by interposing a tower layer that replaces the real
+// server response body with one that immediately ends (no frames, no
+// trailers) while keeping the HTTP 200 status code.
```
dfawley: Is this how is looks when a RST_STREAM(NO_ERROR) occurs?  Because otherwise I'm a little concerned that this doesn't accurately represent the scenario you described in the bug.  This test is still useful -- I just want to check to make sure this fixes the issue for you, since it seems tonic _attempts_ to map RST_STREAM(NO_ERROR) to Internal (as it should), and that is the scenario you indicated you are encountering.

rumenov: I changed the test so the h2 traces match exactly what we observed

tests/integration_tests/tests/status.rs
```diff
@@ -351,3 +351,104 @@ async fn message_and_then_status_from_server_stream() {
...
+
+    // Oneshot channel: client signals the server once `stream_call().await`
+    // has returned, i.e. the response HEADERS have been received and the
+    // client is about to enter the body-reading loop.  The server delays
+    // `send_reset` until it receives this signal so that:
+    //
+    //   1. RST_STREAM is queued only AFTER the HEADERS frame has already
+    //      been flushed over the wire (h2's `send_reset` calls `clear_queue`
+    //      which would otherwise discard any still-pending DATA frames, but
+    //      here there are none — only the already-flushed HEADERS matter).
```
dfawley, line 386: > h2's `send_reset` calls `clear_queue`

Side note on this particular behavior of h2: gRPC will not want this, as gRPC wants to send trailers with END_STREAM and then immediately RST_STREAM(NO_ERROR) after it if the client has not sent an END_STREAM first.

tests/integration_tests/tests/status.rs
```diff
@@ -351,3 +351,104 @@ async fn message_and_then_status_from_server_stream() {
...
+    //
+    //   2. The client is inside `Incoming::poll_data` (or about to enter it)
+    //      when RST_STREAM arrives.  That is exactly the branch where hyper
+    //      converts `Reset(NO_ERROR)` to `Poll::Ready(None)` (a clean
```
dfawley, line 390: I think ultimately we'll want to move away from hyper for gRPC, because of things like this, but first I want to focus on adding all the things tonic is missing, and not reimplementing the transport until necessary.

tests/integration_tests/tests/status.rs
```diff
@@ -351,3 +351,104 @@ async fn message_and_then_status_from_server_stream() {
...
+    //
+    //        Some(Err(e)) => match e.reason() {
+    //            Some(h2::Reason::NO_ERROR) | Some(h2::Reason::CANCEL)
+    //                => Poll::Ready(None),   // ← taken here
+    //            _   => Poll::Ready(Some(Err(...))),
+    //        }
+    //
+    //      tonic's decoder sees None with an empty buffer, calls
+    //      `infer_grpc_status(trailers=None, status=200)`, and returns
+    //      `Code::Unknown` — the expected behaviour.
```
dfawley: One nit-pick: this is not the expected behavior in this case, actually.  It should be INTERNAL.

I would say something like "...returns Code::Unknown because it is not able to observe the RST_STREAM(NO_ERROR) at this time and only sees a stream end successfully but without trailers containing `grpc-status`.  TODO(i/2547): this should expect Code::Internal instead."  (Or don't re-use 2547 and we file a new issue for this small, lower-priority refinement instead; I'm OK either way.)

tests/integration_tests/tests/status.rs
rumenov: I guess this is a self note ☺️

Issue #2526, opened 2026-02-27
prost-types resolves some well known types to their rust counterparts


hermeGarcia:
### Version

`v0.14.5`

### Platform

`Darwin 192.168.1.131 24.6.0 Darwin Kernel Version 24.6.0: Mon Aug 11 21:16:05 PDT 2025; root:xnu-11417.140.69.701.11~1/RELEASE_X86_64 x86_64`

### Crates

`tonic-prost-build`

### Description

When using the following well known types as input/output types in service definitions: 

- DoubleValue
- FloatValue
- Int32Value
- Int64Value
- UInt32Value
- UInt64Value

`cargo build` will prefix the primitive type with `super::`, for example `google.protobuf.UInt32Value` translates into `super::u32`.

Reproducing this error is quite easy, just recreate the [helloworld example](https://github.com/hyperium/tonic/blob/master/examples/helloworld-tutorial.md) and add the following method:

```
rpc SomeMethod (HelloRequest) returns (google.protobuf.UInt32Value) {}
```

I think the problem is that `prost-build` is already [resolving such types into their primitive counterparts](https://github.com/tokio-rs/prost/blob/822b19b95588b194ed5c9f0e43dd95e5f256b984/prost-build/src/code_generator.rs#L925-L926), hence `tonic-prost-build` receives `u32` instead of the well known type.

I implemented [this solution](https://github.com/hermeGarcia/tonic/commit/706ef38b94938d77eb161d4a580a821eb09fa43f) to take such primitive types into account, using my fork seems to compile fine the helloworld example. The approach is quite simple, I just added the primitive types that can be mapped from a well known type to `NON_PATH_TYPE_ALLOWLIST`.

PR #2544, opened 2026-03-11
Support well known types resolved by prost to their rust counterparts

hermeGarcia:
## Motivation

Solves issue #2526.

## Solution

This PR adds a new constant, `EXTENDED_NON_PATH_TYPE_ALLOWLIST`, that supports every rust primitive that can be returned by prost when a well known is found. This constant should be used instead of `DEFAULT_NON_PATH_TYPE_ALLOWLIST` to successfully compile protos that use well-known types in requests and responses.

To make this behavior optional a new flag, `with_extended_rust_types`, is introduced. Due to interface limitations this flag can not threaded into `prost-build`, instead a thread local is used.


LucioFranco (approved):

LucioFranco:
not sure why CI isn't running 🤔

PR #2570, opened 2026-03-25
feat(grpc): Implement PickFirst load balancer

nathanielford:
## Motivation

Full implementation of the pick first load balancer, including 'Happy eyeballs' features.

## Solution

Load balancing implementation to pick the first available endpoint to connect to, maintaining stickiness across endpoint updates if configured. Handles accepting new LB configuration and subchannel reconstruction. 

Prototype is at https://github.com/nathanielford/grpc-rust-testbed/tree/main/pick_first_lib

## Notes

* Ended up including all happy eyeball features because it wasn't clear where best to slice the line. Considering this a full implementation, and it should be reviewed as such.
* This does use tokio::spawn and tokio::time, which may need to be replaced to make things runtime agnostic. Please comment in the PR whether this is the case.


grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
 #[derive(Debug)]
 pub(crate) struct PickFirstPolicy {
     work_scheduler: Arc<dyn WorkScheduler>,
-    subchannel: Option<Arc<dyn Subchannel>>,
-    next_addresses: Vec<Address>,
     runtime: GrpcRuntime,
```
arjan-bal, line 105: This `runtime` field appears to be currently unused, so we can remove it. We will need it later for the Happy Eyeballs implementation (which requires a timer), but we can add it back when those changes are ready.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
-        self.runtime.spawn(Box::pin(async move {
-            runtime.sleep(Duration::from_millis(200)).await;
-            work_scheduler.schedule_work();
-        }));
-        // TODO: return a picker that queues RPCs.
+        self.update_config(config)?;
+        let new_addresses = self.compile_address(
+            update.endpoints.map_err(|e| e.to_string())?,
+            channel_controller,
+        )?;
```
arjan-bal: It appears the only reference to the LB config is within `compile_address`. We can likely remove the struct field and pass the config directly as a function parameter.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
+
+        // Stickiness: Check if currently selected subchannel is in the new list.
+        if let Some(ref selected) = self.selected {
+            if new_addresses.contains(&selected.address()) {
+                self.rebuild_subchannels(new_addresses, channel_controller);
+                return Ok(());
+            }
+        }
```
arjan-bal: We can combine the nested `if` blocks and replace the `ref` with a borrow on the rhs to make the code more readable.

```suggestion
        if let Some(selected) = &self.selected
            && new_addresses.contains(&selected.address())
        {
            self.rebuild_subchannels(new_addresses, channel_controller);
            return Ok(());
        }
````

File: grpc/src/client/load_balancing/pick_first.rs [hunk 4]
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
+    // If the configuration is set, this list will be shuffled.
+    fn compile_address(
+        &mut self,
+        endpoints: Vec<Endpoint>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<Vec<Address>, String> {
+        let mut addresses = endpoints
+            .into_iter()
+            .flat_map(|e| e.addresses)
+            .collect::<Vec<_>>();
```
arjan-bal: For random shuffling, the expected behaviour is to shuffle the endpoints before flattening the address list. See point 1 here: https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#address-list-handling-in-pick_first

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: In addition to returning errors here, the balancer may need to transition the channel to TRANSIENT_FAILURE. Furthermore, if the resolver provides an empty endpoint list, the balancer must respect the update and drop any existing subchannels. [Here](https://github.com/grpc/grpc-go/blob/8de729c03f732f5d27528cf07632ed5b0b805b88/balancer/pickfirst/pickfirst.go#L242-L249) is how Go handles this.

arjan-bal (commented):
Leaving initial comments while I review the remaining changes.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
+        }
+    }
+
+    fn rebuild_subchannels(
+        &mut self,
+        new_addresses: Vec<Address>,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        // Map existing subchannels by address.
+        let mut existing_map: std::collections::HashMap<Address, Arc<dyn Subchannel>> = self
```
arjan-bal: nit: Unless there are symbols with conflicting names, we usually import the symbol directly, i.e. `use std::collections::HashMap`.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
+    current_index: usize,
+    connectivity_state: ConnectivityState,
+    config: Option<PickFirstConfig>,
+}
+
+impl PickFirstPolicy {
+    fn start_connection_pass(&mut self, channel_controller: &mut dyn ChannelController) {
+        self.current_index = 0;
+        self.selected = None;
+        if let Some(sc) = self.subchannels.get(0) {
```
arjan-bal: It is preferable to let the code panic if runtime behavior guarantees the vector will not be empty.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -74,43 +92,141 @@ pub(crate) fn reg() {
...
+            self.connectivity_state = ConnectivityState::Connecting;
+            sc.connect();
+            channel_controller.update_picker(LbState {
+                connectivity_state: ConnectivityState::Connecting,
+                picker: Arc::new(QueuingPicker {}),
+            });
```
arjan-bal: To avoid picker churn, A62 describes a "sticky TRANSIENT_FAILURE" behaviour [here](https://github.com/grpc/proposal/blob/master/A62-pick-first.md#sticky-transient-failure) where `pick_first` stays in TF until a subchannel becomes READY. We should implement this here.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: Since subchannels are reused when reconciling address lists during a resolver update, the first subchannel may be in backoff (reporting `TRANSIENT_FAILURE`). Instead of unconditionally calling `connect()` on the first subchannel, `pick_first` should ideally find the first `IDLE` or `CONNECTING` subchannel.

[Here](https://github.com/grpc/grpc-go/blob/2de5b1c09879a51b545069a29cd4611702f533ae/balancer/pickfirst/pickfirst.go#L541-L583) is Go's implementation.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal, line 532: The iteration over the subchannel list needs to be split into two passes, as [described in A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#happy-eyeballs-in-the-pick_first-lb-policy). During the initial pass, `pick_first` will sequentially iterate over the subchannel list, ensuring it sees a `TRANSIENT_FAILURE` for each subchannel before proceeding to the next. Once all subchannels have failed, the first phase is complete. Now, `pick_first` will reconnect each subchannel as it reports `IDLE`. Ignoring the Happy Eyeballs timer, this corresponds to the circled portion of the gRFC:

Image

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -120,29 +236,72 @@ impl LbPolicy for PickFirstPolicy {
...
+                }
+                return;
+            }
+        }
+
+        // 2. Otherwise, check if this is from the subchannel we are currently attempting.
+        if let Some(attempting) = self.subchannels.get(self.current_index) {
+            if attempting.address() == subchannel.address() {
+                match state.connectivity_state {
+                    ConnectivityState::Ready => {
```
arjan-bal: Since we have subchannel sharing in Rust, we would need to consider the [this section of A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#implications-of-subchannel-sharing-in-c-core) also:
* pick_first needs to be prepared for any subchannel to report READY at any time
* When we choose a subchannel that has become successfully connected, we will unref all of the other subchannels

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -155,9 +314,322 @@ impl Picker for OneSubchannelPicker {
...
+
+impl Picker for IdlePicker {
+    fn pick(&self, _: &RequestHeaders) -> PickResult {
+        self.work_scheduler.schedule_work();
+        PickResult::Queue
+    }
+}
+
+#[cfg(test)]
+mod test {
```
arjan-bal, line 734: There are some tricky edge cases found by Go's e2e-style tests that we discovered when implementing a new dual-stack `pick_first` LB in Go. I added unit tests for them to ensure the balancer behaved as expected. These tests are mainly in two files:
* [pickfirst_ext_test.go](https://github.com/grpc/grpc-go/blob/master/balancer/pickfirst/pickfirst_ext_test.go): These are e2e-style tests that create a gRPC channel to a fake server.
* [pickfirst_test.go](https://github.com/grpc/grpc-go/blob/master/balancer/pickfirst/pickfirst_test.go): These use a mock channel, similar to the tests in Rust.

It may not be possible to write e2e-style tests in Rust right now since we don't have the same test utilities as Go. However, we can still test the same scenarios using a mock channel.

I would recommend seeing if Gemini can convert these Go tests into tests for Rust, skipping the Happy Eyeballs ones. This would act as a conformance test. I did something similar to generate tests for credentials in the following way:
1. Clone gRPC Go.
2. Point the Gemini CLI to the test files for gRPC Go and gRPC Rust.
3. Ask Gemini to create similar test cases for gRPC Rust.

If it gives decent results without much effort, that's great; otherwise, we can improve test coverage later.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Is this as simple as setting the connection state to transient failure and then attempting to reconnect? Or is it safe to assume the picker is already in transient failure and skip this update altogether?

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Fixed!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Fixed!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Good catch - fixed this!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Done!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Done.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I've made this change. I'm not sure if with ServiceConfig we'll have to do something different going forward but this reads cleaner now, and reduces state surface.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Done!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: To be clear, this is a happy eyeballs requirement, correct? I think we were going to try and put in a pick_first implementation _without_ Happy eyeballs first. If you think it's important, I can include that but it was initially out of scope.

arjan-bal: Technically, Happy Eyeballs involves only the interleaving of IPv4/IPv6 addresses and running a timer to resume iteration if subchannels are slow to fail. The first-pass/second-pass logic is independent of that. Given this, the Happy Eyeballs implementation probably accounts for only around 20% of the dev effort. If we don't add the two-pass logic, we would end up re-implementing a significant portion of the LB policy later. After discussing this with @dfawley, I suggest we increase the scope of this change to implement the two-pass logic without Happy Eyeballs. This would involve increasing the estimates for this work and reducing the estimates for the Happy Eyeballs work, which should be fine.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,351 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        // Start 250ms timer
+        let timer_expired = self.timer_expired.clone();
+        let work_scheduler = self.work_scheduler.clone();
+
+        let handle = tokio::spawn(async move {
+            tokio::time::sleep(std::time::Duration::from_millis(250)).await;
+            timer_expired.store(true, Ordering::SeqCst);
+            work_scheduler.schedule_work();
+        });
```
arjan-bal: To do this without depending on `tokio` directly, you would need to use similar methods on the gRPC `Runtime` type:
https://github.com/hyperium/tonic/blob/602af31a8f3d84b61ee6ccf21e0dee35cc36fbaa/grpc/src/rt/mod.rs#L344-L380

Presently, there's one [one implementation](https://github.com/hyperium/tonic/blob/602af31a8f3d84b61ee6ccf21e0dee35cc36fbaa/grpc/src/rt/tokio/mod.rs#L90) based on `tokio`. The abstraction exists to allow using different runtimes in the future.

arjan-bal: The runtime is passed to the LB through the `LbPolicyOptions` passed to the `build` method.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,351 @@ use crate::client::load_balancing::LbPolicy;
...
+type ShufflerFn = dyn Fn(&mut [Endpoint]) + Send + Sync + 'static;
+
+#[derive(Debug, serde::Deserialize, Clone)]
+pub struct PickFirstConfig {
+    #[serde(rename = "shuffleAddressList")]
+    pub shuffle_address_list: bool,
+}
+
+#[derive(Debug)]
+pub(crate) struct PickFirstBuilder {}
```
arjan-bal: Keeping the builder private should work because it's always referenced using the `LbPolicyBuilder` trait.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,351 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        self.subchannels = new_addresses
+            .into_iter()
+            .map(|addr| {
+                let subchannel = if let Some(sc) = existing_subchannels.remove(&addr) {
+                    sc // Reuse existing subchannel.
+                } else {
+                    // New subchannel for new address.
+                    let (sc, state) = channel_controller.new_subchannel(&addr);
+                    self.subchannel_states.insert(addr.clone(), state);
```
arjan-bal: In functional programming, mutating external state inside an iterator is considered an anti-pattern. I would recommend either change this into a `for` loop or use unzip to generate both `subchannels` and `subchannel_state` together:
```rust
let (new_subchannels, new_states_iter): (Vec<_>, Vec<_>) = new_addresses
    .into_iter()
    .map(|addr| {
        if let Some(sc) = existing_subchannels.remove(&addr) {
            // Retrieve existing state
            let state = self.subchannel_states.get(&addr).unwrap().clone();
            (sc, (addr, state))
        } else {
            // Create new subchannel and state
            let (sc, state) = channel_controller.new_subchannel(&addr);
            (sc, (addr, state))
        }
    })
    .unzip();
```

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Ah, yes. I've fixed this now. Thanks!

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
...
+        // Failover on Failure: Only if the subchannel at the exact frontier_index fails.
+        if let Some(attempting) = self.subchannels.get(self.frontier_index) {
+            if attempting.address() == subchannel.address() {
+                if state.connectivity_state == ConnectivityState::TransientFailure {
+                    // Move to next address via advance_frontier
+                    if let Some(next_sc) = self.advance_frontier(false) {
+                        let next_sc = next_sc.clone(); // Clone to avoid borrow issues
+                        self.trigger_subchannel_connection(next_sc, channel_controller);
+                    } else {
+                        // Exhausted: TRANSIENT_FAILURE and request re-resolution.
```
arjan-bal: Some of the subchannels in the list may have already completed their backoff. Therefore, pick_first must iterate over all the subchannels and connect the ones in IDLE. Otherwise, they'll be stuck in IDLE forever.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,353 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        while self.frontier_index < self.subchannels.len() {
+            let sc = &self.subchannels[self.frontier_index];
+            let addr = sc.address();
+            let state = self
+                .subchannel_states
+                .get(&addr)
+                .map(|s| s.connectivity_state);
+
+            if state == Some(ConnectivityState::TransientFailure) {
```
arjan-bal: Is it possible for the state to be `None`? If not, we should probably call `.expect()` on it; that way, we'll know if an invariant has been broken and can fix the bug.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,353 @@ use crate::client::load_balancing::LbPolicy;
...
+        let unique_addresses: Vec<Address> = endpoints
+            .into_iter()
+            .flat_map(|ep| ep.addresses)
+            .filter(|addr| seen.insert(addr.clone()))
+            .collect();
+
+        // Partition by family (Basic IPv6 detection via colon).
+        let (ipv6, ipv4): (Vec<Address>, Vec<Address>) = unique_addresses
+            .into_iter()
+            .partition(|addr| addr.address.contains(':'));
```
arjan-bal: Both Go and C++ partition the address into 3 groups: IPv4, IPv6, and unknown. We should also have a similar grouping here to be consistent. The simplest way would be to verify that  `Address.network_type`  is "tcp" before applying the  `:`  heuristic.

arjan-bal (commented):
I'm still reviewing the PR, leaving the first set of comments.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,353 @@ use crate::client::load_balancing::LbPolicy;
...
+        // Map existing subchannels by address.
+        let mut existing_subchannels: HashMap<Address, Arc<dyn Subchannel>> = self
+            .subchannels
+            .drain(..)
+            .map(|sc| (sc.address(), sc))
+            .collect();
+
+        let mut new_states = HashMap::new();
+
+        self.subchannels = new_addresses
```
arjan-bal: Since Rust has subchannel pooling like C++, I think it's simpler implementation-wise to create a completely new subchannel map and drop the old one.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,353 @@ use crate::client::load_balancing::LbPolicy;
...
+            channel_controller.update_picker(LbState {
+                connectivity_state: ConnectivityState::Connecting,
+                picker: Arc::new(QueuingPicker {}),
+            });
+        } else {
+            // Otherwise all addresses are in transient failure: update picker and request re-resolution.
+            let error = self
+                .last_connection_error
+                .clone()
+                .unwrap_or_else(|| "all addresses in transient failure".to_string());
```
arjan-bal: I think if we update `self.last_connection_error` with an error message in the following two events, it is guaranteed to be present:
1. When a subchannel starts in `TRANSIENT_FAILURE`.
2. When a subchannel enters `TRANSIENT_FAILURE`.

Then we can call `.expect()` and avoid using a custom error message generated by the LB.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,353 @@ use crate::client::load_balancing::LbPolicy;
...
+        config: Option<&Self::LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), String> {
+        // Reset steady state on new update
+        self.steady_state = None;
+
+        match update.endpoints {
+            Ok(endpoints) => {
+                let new_addresses = self.compile_address(endpoints, config, channel_controller)?;
+                let sticky = self.sticky(&new_addresses);
```
arjan-bal: Based on [A61](https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#implications-of-subchannel-sharing-in-c-core), `pickfirst` should iterate over all the subchannels and choose one that is already ready. If we follow that logic, the explicit check for the previously connected subchannel should be unnecessary.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
         state: &SubchannelState,
         channel_controller: &mut dyn ChannelController,
     ) {
-        match state.connectivity_state {
-            // Assume the update is for our subchannel.
-            ConnectivityState::Ready => {
+        // Update the cache for all updates.
+        self.subchannel_states
+            .insert(subchannel.address(), state.clone());
```
arjan-bal, line 550: Before inserting, we should first ensure that the address is part of the current list of subchannels. If not, the function should return early. This can happen when a subchannel update has been queued by the channel while a resolver updated arrives, removing the address.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
...
+        // If any subchannel in the current pass reports READY, select it.
+        if state.connectivity_state == ConnectivityState::Ready {
+            // Check if this subchannel is part of the current pass (index <= frontier_index)
+            let is_in_pass = self
+                .subchannels
+                .iter()
+                .take(self.frontier_index + 1)
+                .any(|sc| sc.address() == subchannel.address());
+
+            if is_in_pass {
```
arjan-bal: I believe  pickfirst  can choose a subchannel that is across the frontier too, as long as it's in the current list of addresses from the resolver. C++ seems to be doing the same.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
...
+                            work_scheduler: self.work_scheduler.clone(),
+                        }),
+                    });
+                }
+                return;
+            }
+        }
+
+        // Steady State Retries: Automatically connect when a subchannel goes IDLE.
+        if self.steady_state.is_some() && state.connectivity_state == ConnectivityState::Idle {
```
arjan-bal: nit: It would be good to group the handling of  steady_state  and the first pass into an  `if/else`  block, and handle common transitions (e.g., READY or READY->IDLE) before this block. This would allow readers to understand the first pass and steady-state behavior separately.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
...
+            subchannel.connect();
+        }
+
+        // Steady State Failures: Count failures across all subchannels.
+        if let Some(ref mut ss) = self.steady_state {
+            if state.connectivity_state == ConnectivityState::TransientFailure {
+                if ss.record_failure(self.subchannels.len()) {
```
arjan-bal: nit: As a minor simplification, we can pass the channel controller to `record_failure` and let it handle requesting re-resolution itself.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -115,31 +394,838 @@ impl LbPolicy for PickFirstPolicy {
...
+
+        // Failover on Failure: Only if the subchannel at the exact frontier_index fails.
+        if let Some(attempting) = self.subchannels.get(self.frontier_index) {
+            if attempting.address() == subchannel.address() {
+                if state.connectivity_state == ConnectivityState::TransientFailure {
+                    // Move to next address via advance_frontier
+                    if let Some(next_sc) = self.advance_frontier(false) {
+                        let next_sc = next_sc.clone(); // Clone to avoid borrow issues
+                        self.trigger_subchannel_connection(next_sc, channel_controller);
+                    } else {
```
arjan-bal: We need to cancel the timer in the `else` case also.It can probably happen before the `if/else` block.

Comparing the references of `timer_handle` with the Go implementation, we also need to abort the timer when the LB policy is dropped and when a new resolver update arrives.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Ah, this is great advice! I had to modify it a bit but I've made this fix.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I don't *think* so, so it does make sense to panic here if we somehow get a None subchannel. I've made the fix.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I ran into a problem with this in the `impl LbPolicy for PickFirstPolicy` section. Specifically I got a compile error around `type LbConfig = PickFirstConfig;` about private type information leaking.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I've made this change, and on the assumption the `unknown` values should be included in the interleaving updated that as well. This seems to be what C++ is doing, but please do check that I did it right!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Sorry, I think I got confused here. I've updated the builder to be private and the config to be pub(crate).

grpc/src/client/load_balancing/pick_first.rs
nathanielford: This has been fixed. I had to move things around a fair bit and put a higher level match case to handle the routing more clearly. Hopefully that covers it!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I think I was able to do this in the tests, particularly the last two tests in pick_first.rs. LMK if you think something is misaligned with these!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: This should be fixed now!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: This should be handled now!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Ah, yes. I've fixed this.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I moved all the steady-state stuff into the steady state handler. I think that it makes it a bit cleaner, to keep all the state tracking there. LMK if you think something is leaky though!

grpc/src/client/load_balancing/pick_first.rs
nathanielford: This should all be handled - though do please audit all the possible cases!

nathanielford:
Note that this now includes https://github.com/hyperium/tonic/pull/2631, which will merge first so as to keep a clearer record. However, I needed the fix to get the tests ported from Go to work.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+        }
+
+        // If we have no addresses, clear subchannels and set TRANSIENT_FAILURE.
+        if interleaved.is_empty() {
+            self.subchannels.clear();
+            self.selected = None;
+            let error = self
+                .last_resolver_error
+                .clone()
+                .unwrap_or_else(|| "empty address list".to_string());
```
arjan-bal: Is there a reason to use the older error here? I think we should use the new `empty address list` error here.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        Ok(interleaved)
+    }
+
+    // Sets state to TRANSIENT_FAILURE and updates picker with error. Triggers a re-resolution request.
+    fn set_transient_failure(
+        &mut self,
+        channel_controller: &mut dyn ChannelController,
+        error: String,
+    ) -> Result<(), String> {
```
arjan-bal, line 465: Why does this method unconditionally return an error? It seems cleaner to have the caller use the error string it already has.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    // Returns true if the currently selected subchannel's address is still present in the new address list.
+    fn sticky(&self, new_addresses: &[Address]) -> bool {
+        self.selected
+            .as_ref()
+            .map(|sc| new_addresses.contains(&sc.address()))
+            .unwrap_or(false)
+    }
+
+    // Cancels the connection pacing timer if it is active.
+    fn abort_pacing_timer(&mut self) {
```
arjan-bal: nit: The word "pacing" is not mentioned in the gRFC, so we could rename this to `abort_happy_eyeballs_timer` instead.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    /// managed by this policy experiences a connectivity state transition.
+    ///
+    /// # Parameters
+    /// * `subchannel`: The specific backend connection instance (`Arc<dyn Subchannel>`) that triggered the event.
+    ///   It identifies *which* transport lane is reporting telemetry.
+    /// * `state`: The new connectivity status snapshot (`SubchannelState`) being reported.
+    ///   It details *what* happened (e.g., transitioned to `READY`, `IDLE`, or encountered a `TransientFailure`).
+    /// * `channel_controller`: The internal control plane interface used to update the channel's RPC picker
+    ///   or signal the Name Resolver to fetch new addresses.
+    ///
```
arjan-bal: We don't need to have rustdocs for trait methods in the `impl` blocks since they are redundant.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    /// Returns true if the given subchannel matches the currently selected active subchannel.
+    fn subchannel_is_selected(&self, subchannel: &Arc<dyn Subchannel>) -> bool {
+        self.selected
+            .as_ref()
+            .map_or(false, |sel| sel.address() == subchannel.address())
+    }
+
+    /// Returns true if the subchannel's address is present in the most recently received address list.
+    // This compares against the current list of subchannels the LB is attempting to connect to. To
+    // see if the LB already connected to the channel, use 'subchannel_is_selected'.
```
arjan-bal: style: Please wrap comments at 80 columns for readability.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    fn subchannel_is_current(&self, subchannel: &Arc<dyn Subchannel>) -> bool {
```
arjan-bal, line 366: nit+optional: This method could be inlined since there's currently only one caller. While I recall Uncle Bob's *Clean Code* recommending helper methods to keep the level of abstraction consistent, personally, I feel that's overkill for a method this short.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        // Update the cache for all updates.
+        self.subchannel_states
+            .insert(subchannel.address(), state.clone());
+
+        // If the subchannel being updated is the selected one, it affects handling.
+        let is_selected = self
+            .selected
+            .as_ref()
+            .map_or(false, |s| s.address() == subchannel.address());
```
arjan-bal: Locally, rust_analyzer is showing me a warning about [unnecessary map_or](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_map_or) here.  We can instead use `is_some_and` here:
```rust
        let is_selected = self
            .selected
            .as_ref()
            .is_some_and(|s| s.address() == subchannel.address());
```

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+        &mut self,
+        subchannel: Arc<dyn Subchannel>,
+        state: &SubchannelState,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        // Failover triggers only if the failure comes from the subchannel currently at the frontier.
+        if let Some(attempting) = self.subchannels.get(self.frontier_index) {
+            if attempting.address() == subchannel.address()
+                && state.connectivity_state == ConnectivityState::TransientFailure
+            {
```
arjan-bal: Locally rust_analyzer is warning me that the nested `if` can be merged as follows:
```rust
        if let Some(attempting) = self.subchannels.get(self.frontier_index)
            && attempting.address() == subchannel.address()
            && state.connectivity_state == ConnectivityState::TransientFailure
```

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+                    // Pass exhausted: enter policy-level TRANSIENT_FAILURE and switch to steady state.
+                    let error = state
+                        .last_connection_error
+                        .clone()
+                        .expect("gRPC Contract Violation: last_connection_error must be populated when in TransientFailure");
+
+                    // Cancel the pacing timer since this connection pass is over.
+                    self.abort_pacing_timer();
+
+                    self.last_connection_error = Some(error.clone());
```
arjan-bal: This assignment seems redundant since the `error` was assigned from `state.last_connection_error` a couple of lines above.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+
+                    self.steady_state = Some(SteadyState::new(self.subchannels.len()));
+
+                    // Trigger connection attempts on any subchannels that transitioned to IDLE
+                    // during the first pass, ensuring they don't get stuck.
+                    for sc in &self.subchannels {
+                        let is_idle = self
+                            .subchannel_states
+                            .get(&sc.address())
+                            .map_or(false, |s| s.connectivity_state == ConnectivityState::Idle);
```
arjan-bal: Same here, we can replace `map_or` with `is_ok_and`.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    /// Assumes the subchannel is in a non-READY state.
+    /// If the failure is from the subchannel at the frontier, advances the frontier and triggers a connection on the next subchannel.
+    fn update_first_pass(
+        &mut self,
+        subchannel: Arc<dyn Subchannel>,
+        state: &SubchannelState,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        // Failover triggers only if the failure comes from the subchannel currently at the frontier.
+        if let Some(attempting) = self.subchannels.get(self.frontier_index) {
```
arjan-bal: nit: Can we use early returns to reduce indentation in this method?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+            let addr = sc.address();
+            let state = self
+                .subchannel_states
+                .get(&addr)
+                .map(|s| s.connectivity_state)
+                .expect("Expected non-None subchannel state");
+
+            match state {
+                // Push the frontier if sc is in TransientFailure, otherwise return the sc.
+                ConnectivityState::TransientFailure => self.frontier_index += 1,
```
arjan-bal, line 345: I believe we need to update `self.last_resolver_error` here. Otherwise, if all the addresses are already in TF, the first pass may end without `self.last_resolver_error` being set.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+}
+
+#[derive(Debug)]
+struct IdlePicker {
+    work_scheduler: Arc<dyn WorkScheduler>,
+}
+
+impl Picker for IdlePicker {
+    fn pick(&self, _: &RequestHeaders) -> PickResult {
+        self.work_scheduler.schedule_work();
```
arjan-bal: In Java and Go, we use an atomic to make the `IdlePicker` call schedule work at most once. This avoids redundant work and ensure calling the `Idle` picker repeatedly doesn't reset the first pass. We should do the same here as well.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,114 +38,1381 @@ use crate::client::load_balancing::LbPolicy;
...
+    /// Triggers a connection on the subchannel, and starts the 250ms timer.
+    /// If no connection succeeds before the timer expires, the frontier will advance to
+    /// the next subchannel.
+    fn trigger_subchannel_connection(
+        &mut self,
+        sc: Arc<dyn Subchannel>,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        let sc_clone = sc.clone();
+        self.connectivity_state = ConnectivityState::Connecting;
```
arjan-bal: We should not set the channel state to `CONNECTING` unconditionally. [A62](https://github.com/grpc/proposal/blob/master/A62-pick-first.md#sticky-transient-failure) describes the "sticky-TF" behavior of `pick_first`. In short, if the subchannel is in TF, it must remain in TF until a subchannel reports `READY` to avoid repeated `CONNECTING->TF->CONNECTING` transitions. 


Based on the Go implementation, `pick_first` should transition the channel to CONNECTING in the following cases:
1. When the LB policy is initially built.
2. The previous resolver address list was empty and the new address list is non-empty.
3. The channel was ready, and a resolver update removed all the ready addresses.
4. When exiting the `IDLE` state. `pick_first` should NOT wait for a subchannel to transition to CONNECTING to avoid [this bug](https://github.com/grpc/grpc-go/pull/8615). 

When a subchannel transitions to CONNECTING, I believe `pick_first` would already be in CONNECTING due to one of the above conditions, and `pick_first` can simply ignore the subchannel transition.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: In `resolver_update` and `compile_address` this error is automatically propogated up.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -22,8 +22,14 @@
...
 
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::fmt::Debug;
 use std::sync::Arc;
-use std::time::Duration;
+
+use crate::metadata::MetadataMap;
+use rand::seq::SliceRandom;
+use std::sync::atomic::{AtomicBool, Ordering};
```
arjan-bal: Import grouping seems off here. Each import should be in a new line. This is how I've configured rust analyzer on neovim, you may need to add similar config for your editor:
```
settings = {
        ["rust-analyzer"] = {
            diagnostics = {
                styleLints = {
                    enable = true,
                },
            },
            cargo = {
                allFeatures = true,
            },
            check = {
                command = "clippy",
            },
            imports = {
                granularity = {
                    group = "item",
                },
            },
            rustfmt = {
                extraArgs = {
                    "+nightly",
                    "--config",
                    "unstable_features=true,group_imports=StdExternalCrate,imports_granularity=Item",
                },
            },
        },
    }
```

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+    connectivity_state: ConnectivityState,
+
+    // Subchannel information.
+    subchannels: Vec<Arc<dyn Subchannel>>,
+    subchannel_states: HashMap<Address, SubchannelState>, // Cached states for all subchannels by address.
+    selected: Option<Arc<dyn Subchannel>>,
+    frontier_index: usize,
+
+    // Detailed error tracking.
+    last_resolver_error: Option<String>,
```
arjan-bal: This field seems unused now. It can be removed.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+            new_states.insert(addr, state);
+        }
+
+        self.subchannels = new_subchannels;
+        self.subchannel_states = new_states;
+        ready_subchannel
+    }
+
+    /// Call when the selected subchannel is dropped or loses connection.
+    // This causes the LB to go IDLE.
```
arjan-bal: > when the selected subchannel is dropped

When does this situation happen? I'm asking since there's only one caller for this method which seems to be for the second case, i.e. the subchannel disconnecting.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+                break;
+            }
+        }
+
+        // If we have no addresses, clear subchannels and set TRANSIENT_FAILURE.
+        if interleaved.is_empty() {
+            self.subchannels.clear();
+            self.selected = None;
+            self.set_transient_failure(channel_controller, Some("empty address list".to_string()))?;
+        }
```
arjan-bal: The fact that `compile_addresses` has a side-effect of also setting the channel state is a little unexpected. I think we should have `compile_addresses` be infallible and move the empty endpoint check to the `resolver_update` method.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+
+    /// Starts a connection pass through the address list.
+    // This clears the selected subchannel.
+    fn start_connection_pass(&mut self, channel_controller: &mut dyn ChannelController) {
+        self.selected = None;
+
+        // If there is a viable subchannel at the frontier, connect to it and
+        // update picker to CONNECTING.
+        if let Some(sc) = self.advance_frontier(true) {
+            let sc = sc.clone(); // Avoid borrow issues.
```
arjan-bal: nit: All the callers of `advance_frontier` need to clone the returned `Arc`. I think it's more ergonomic to have `advance_frontier` return an owned value directly. Since these clones are not on the RPC path (e.g. in the picker.Pick() method), the performance impact would be negligible.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+
+    /// Triggers a connection on the subchannel, and starts the 250ms timer.
+    /// If no connection succeeds before the timer expires, the frontier will
+    /// advance to the next subchannel.
+    fn trigger_subchannel_connection(
+        &mut self,
+        sc: Arc<dyn Subchannel>,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        let sc_clone = sc.clone();
```
arjan-bal: This clone seems necessary and can be removed.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        // If there is a viable subchannel at the frontier, connect to it and
+        // update picker to CONNECTING.
+        if let Some(sc) = self.advance_frontier(true) {
+            let sc = sc.clone(); // Avoid borrow issues.
+            self.trigger_subchannel_connection(sc, channel_controller);
+
+            channel_controller.update_picker(LbState {
+                connectivity_state: ConnectivityState::Connecting,
+                picker: Arc::new(QueuingPicker {}),
```
arjan-bal: This can lead to sending redundant queuing pickers when the channel was already in `CONNECTING`. Java and Go avoid this by sending picker updates through a method that handles de-duplication. You can add a TODO to address this in a follow-up PR.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+    fn start_connection_pass(&mut self, channel_controller: &mut dyn ChannelController) {
+        self.selected = None;
+
+        // If there is a viable subchannel at the frontier, connect to it and
+        // update picker to CONNECTING.
+        if let Some(sc) = self.advance_frontier(true) {
+            let sc = sc.clone(); // Avoid borrow issues.
+            self.trigger_subchannel_connection(sc, channel_controller);
+
+            channel_controller.update_picker(LbState {
```
arjan-bal: Sending a `CONNECTING` update here could cause a channel currently in `TF` to transition to `CONNECTING` upon receiving a resolver update. This would result in the channel flapping between `TF` and `CONNECTING`, which the sticky-TF behavior was introduced to prevent. Since this is not a correctness concern, we can add a `TODO` and address the sticky-TF changes in a follow-up PR.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+                connectivity_state: ConnectivityState::Connecting,
+                picker: Arc::new(QueuingPicker {}),
+            });
+        } else {
+            // Otherwise all addresses are in transient failure: update picker.
+            _ = self.set_transient_failure(channel_controller, None);
```
arjan-bal, line 268: From an API design perspective, calling `set_transient_failure` without an error seems wrong. I think a more explicit design would be to have a helper method that gets the last connectivity error and passes it to `set_transient_failure`, which should accept a non-optional error.

Since this is not a correctness issue, you can add a TODO and address it in a follow-up PR.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+    /// Triggers a connection on the subchannel, and starts the 250ms timer.
+    /// If no connection succeeds before the timer expires, the frontier will
+    /// advance to the next subchannel.
+    fn trigger_subchannel_connection(
+        &mut self,
+        sc: Arc<dyn Subchannel>,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        let sc_clone = sc.clone();
+        self.connectivity_state = ConnectivityState::Connecting;
```
arjan-bal: Here, the `connectivity_state` field is updated without sending a picker update. IIUC, `connectivity_state` should track the state of the channel. Updating `connectivity_state` without simultaneously sending a picker update can cause the two to diverge, potentially introducing bugs if future transitions depend on `connectivity_state`. I think we should introduce a helper method that updates `connectivity_state` and sends a picker update together. I have another comment about de-duplicating picker updates, which can be addressed using this helper.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +541,973 @@ impl LbPolicy for PickFirstPolicy {
...
+        }
+    }
+
+    fn work(&mut self, channel_controller: &mut dyn ChannelController) {
+        if self.connectivity_state == ConnectivityState::Idle {
+            self.exit_idle(channel_controller);
+        } else if self.connectivity_state == ConnectivityState::Connecting {
+            // Check if timer expired
+            if self.timer_expired.load(Ordering::SeqCst) {
+                self.timer_expired.store(false, Ordering::SeqCst); // Reset
```
arjan-bal: Resetting `timer_expired` only when the `connectivity_state` is `Connecting` may be problematic with the sticky TF behavior, where the channel remains in TF even when a new first-pass is started. 

I believe the `timer_expired` field is only meaningful when a timer is running. Therefore, we could tie the timer task handle and the `timer_expired` boolean together into an optional field. When the timer is started, the entire optional field can be set to a fresh value. This ensures the boolean is reset every time a timer is run and the LB policy can't read the boolean when no timer is running. We could have Drop on this new struct to abort the task, making aborting the task setting the field to None.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+
+        // Check if First Pass is fully exhausted (frontier exhausted AND zero connecting).
+        if self.frontier_index >= self.subchannels.len() {
+            let any_connecting = self.subchannels.iter().any(|sc| {
+                self.subchannel_states
+                    .get(&sc.address())
+                    .is_some_and(|s| s.connectivity_state == ConnectivityState::Connecting)
+            });
+
+            if !any_connecting && self.steady_state.is_none() {
```
arjan-bal: The check `self.steady_state.is_none()` seems redundant as the `update_first_pass` must not have been called if `steady_state.is_some()`. If this is the case, we should avoid the check.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -32,80 +38,501 @@ use crate::client::load_balancing::LbPolicy;
...
+    }
+
+    // Returns true if the currently selected subchannel's address is still
+    // present in the new address list.
+    fn sticky(&self, new_addresses: &[Address]) -> bool {
+        self.selected
+            .as_ref()
+            .map(|sc| new_addresses.contains(&sc.address()))
+            .unwrap_or(false)
+    }
```
arjan-bal: This method seems to be unused now. It can be removed.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +541,973 @@ impl LbPolicy for PickFirstPolicy {
...
+                    self.update_first_pass(subchannel, state, channel_controller);
+                }
+            }
+        }
+    }
+
+    fn work(&mut self, channel_controller: &mut dyn ChannelController) {
+        if self.connectivity_state == ConnectivityState::Idle {
+            self.exit_idle(channel_controller);
+        } else if self.connectivity_state == ConnectivityState::Connecting {
```
arjan-bal: What is the motivation behind this check? Based on how the guarded timer is handled, it feels like a better check would be `self.steady_state.is_none()`. On second thought, if `steady_state.is_some()`, the timer shouldn't be running anyway, so we could directly check if the timer handle is present and has timer has expired.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Fixed formatting

grpc/src/client/load_balancing/pick_first.rs
dfawley: Removed

grpc/src/client/load_balancing/pick_first.rs
dfawley: Updated the comment.  @nathanielford let us know if maybe a call site was missing or the comment was just stale.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
dfawley: Left a TODO. I think we are supposed to leave TF, though, if we get a resolver update.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Fixed per below.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Agreed; added todo.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Removed the check.  Added a comment.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
dfawley: Fixed; added helper

grpc/src/client/load_balancing/pick_first.rs
dfawley: Agreed. Changed

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
dfawley: We do check if the timer expired. I removed the connecting check.

Also I added a TODO in the policy struct to consider a "mode" enum instead of just `Option<SteadyState>`

grpc/src/client/load_balancing/pick_first.rs
dfawley: Implemented.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Part of my modification of this was actually to have a connection mode enum, so I'm glad we're thinking along the same lines! Will do a followup with that.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: It was just stale - my last update was addressing the majority of the previous comments, but I wanted to checkpoint it before working on the enum, since that was likely to be invasive. I missed this before I pushed the commit.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Sounds good. For now I just want to something that has good enough behavior in the main use cases, even if it's suboptimal.  We can save bigger cleanups for another PR.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -191,16 +192,34 @@ impl PickFirstPolicy {
...
     }
 
-    /// Call when the selected subchannel is dropped or loses connection.
+    /// Call when the selected subchannel loses connection.
     // This causes the LB to go IDLE.
     fn subchannel_drop(&mut self, channel_controller: &mut dyn ChannelController) {
         self.selected = None;
-        self.connectivity_state = ConnectivityState::Idle;
+        self.update_picker(
+            ConnectivityState::Connecting,
```
arjan-bal: The state should be updated to `Idle` instead of `Connecting` here.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+                },
+                None,
+                controller.as_mut(),
+            )
+            .unwrap();
+
+        // Expect NewSubchannel/addr, Connect, UpdatePicker(Connecting).
+        for _ in 0..(addrs_len + 2) {
+            rx.recv().unwrap();
+        }
```
arjan-bal: nit: Can we assert the type of events also, i.e. `addrs_len` subchannel creations, one subchannel connection and one picker update?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+        // Should create new subchannel for addr3 (was not in previous list).
+        match rx.recv().unwrap() {
+            TestEvent::NewSubchannel(sc) => assert_eq!(sc.address().address.to_string(), "addr3"),
+            other => panic!("unexpected event {:?}", other),
+        }
+
+        // Should NOT have any more events (no Connect, no UpdatePicker),
+        // because it stuck to the original selected subchannel.
+        std::thread::sleep(Duration::from_millis(50));
```
arjan-bal: We should use `tokio::time::timeout` or `tokio::time::sleep` to avoid blocking the entire thread. Same comment for other tests.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+                Some(&config),
+                controller.as_mut(),
+            )
+            .unwrap();
+
+        let resulting_addrs: Vec<String> = policy
+            .subchannels
+            .iter()
+            .map(|sc| sc.address().address.to_string())
+            .collect();
```
arjan-bal: Instead of checking the unexported `subchannels` field, could we check the order of the subchannel connections instead? This would make the tests much more robust against internal refactoring. Same comment for other test references to the `subchannels` field.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+    }
+
+    // If the resolver update contains no addresses, the LB should clear
+    // subchannels, update to TransientFailure, and request re-resolution.
+    #[tokio::test]
+    async fn test_pick_first_empty_update_clears_state() {
+        let (rx, mut policy, mut controller) =
+            simulate_successful_connection(vec!["addr1", "addr2"], None);
+        assert_eq!(policy.subchannels.len(), 1);
+        assert!(policy.selected.is_some());
```
arjan-bal: Instead of depending on unexported fields, could we replace this with a check for a ready picker so we only test through the public API?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+                    found = Some(event);
+                    break;
+                }
+                Err(std::sync::mpsc::TryRecvError::Empty) => {
+                    // Yield to runtime to allow timer task to run.
+                    tokio::time::sleep(std::time::Duration::from_millis(10)).await;
+                }
+                Err(e) => panic!("error recv: {:?}", e),
+            }
+        }
```
arjan-bal: Instead of looping, we can use a `tokio::time::timout` of 10s to wrap the `rx.recv`. We should define 10s as a global const named `DEFAULT_TEST_DURATION` and re-use it for other tests.  For events not expected to happen, we should define a constant `DEFAULT_TEST_SHORT_DURATION` and set it to 10 millis.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+    // If all addresses fail during a connection pass, the LB should enter
+    // steady state and monitor for backoff expiry to retry connections.
+    #[tokio::test]
+    async fn test_pick_first_steady_state_retries() {
+        let (rx, mut policy, mut controller) = simulate_failed_connection(vec!["addr1"], None);
+        let sc1 = policy.subchannels[0].clone();
+
+        // Expect UpdatePicker(TransientFailure) and RequestResolution.
+        rx.recv().unwrap();
+        rx.recv().unwrap();
```
arjan-bal: nit: Can we assert the event type here? Same for other places the return type of `rx.recv()` is being ignored.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+
+        // Should automatically call connect() again.
+        match rx.recv().unwrap() {
+            TestEvent::Connect(addr) => assert_eq!(addr.address.to_string(), "addr1"),
+            other => panic!("unexpected event {:?}", other),
+        }
+    }
+
+    // If the LB is in steady state, and a new address becomes ready, it should
+    // switch to it immediately. If the current active address goes idle, it
```
arjan-bal, line 1264: Even though the comment mentions a subchannel becoming ready, the test doesn't seem to have that. Can we add that case in the implementation?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+
+        // This should be a different subchannel due to different attributes.
+        // Therefore, expect a new TestEvent::NewSubchannel event to be emitted.
+        let mut found_new_subchannel = false;
+        while let Ok(event) = rx.try_recv() {
+            if let TestEvent::NewSubchannel(_) = event {
+                found_new_subchannel = true;
+                break;
+            }
+        }
```
arjan-bal, line 1438: We should use a `tokio::time::timeout` instead of a busy loop here.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -114,32 +534,1016 @@ impl LbPolicy for PickFirstPolicy {
...
+        // Check that re-resolution was requested.
+        match rx.recv().unwrap() {
+            TestEvent::RequestResolution => {}
+            other => panic!("unexpected event {:?}", other),
+        }
+    }
+
+    // If the timer expires during a connection pass, the LB should advance to
+    // the next subchannel and trigger a connection attempt.
+    #[tokio::test(start_paused = true)]
```
arjan-bal: Is there a reason to use `start_paused` for this test specifically?

arjan-bal (commented):
Non-test changes look good. Left some minor comments on the test code.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: We should have test for the channel entering Idle  when a subchannel disconnects and exiting Idle state when an RPC is made.

dfawley: Oh thanks for catching that. Fixed & added the missing test.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done. Also added a timeout.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Good catch.  Tokio only runs tests with a single thread, so there's no way for the code under test to fail if the thread is blocked.  (https://docs.rs/tokio/latest/tokio/attr.test.html#current-thread-runtime)

I've replaced everything to use helpers to pull from the channel.

This might be less perfect for test cases that _know_ a call into the policy should trigger an event inline, but I think it's OK. Let me know what you think.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done.  But note that we can't use recv, as that will block the thread.  PTAL.

There's an unfortunate difference between go and rust, which is that we don't get the concept of a context with a deadline in it from the start.  So if we use the 10s timeout for every operation, it doesn't apply to the whole test. We could follow this convention, but I don't know if it's worth it.  For now I'm just making every thing that could timeout use 10s. I don't think we'll have an issue where things will run really slowly and take many tens of seconds and then timeout, so I think this is fine.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done!

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
dfawley: I'm not sure why it would have been like that. Maybe something about the sleep stuff below?  I think this is fixed now.

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done.

grpc/src/client/load_balancing/pick_first.rs
dfawley: I believe this is correct. It drains the event channel of all events.

arjan-bal: Yes, it's correct. I misunderstood the loop.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: We can revert the Cargo.toml changes now.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: > This might be less perfect for test cases that know a call into the policy should trigger an event inline.

Looks good to me. I didn't notice that the mpsc channel was not an async channel, so I didn't consider the result of calling recv.

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: > But note that we can't use recv, as that will block the thread. PTAL.

I didn't notice that the mpsc channel was not an async channel.  I suspect this is because the `TestChannelController` is using a `std::sync::mpsc` instead of tokio mpsc channel, making the receiver blocking. In the DNS resolver tests, I've used an unbounded tokio mspc channel for which sends are sync, but receives can be awaited:
https://github.com/hyperium/tonic/blob/ae8ab11a13006b551b8ec88383417cecc05d6331/grpc/src/client/name_resolution/test_utils.rs#L62-L69

Maybe a similar pattern can avoid the need for polling in the pickfirst tests? This can be attempted in a later PR.

arjan-bal (approved):
LGTM

grpc/src/client/load_balancing/pick_first.rs
dfawley: I _believe_ the test event channel should just be synchronous since all accesses are done synchronously (from non-async lb policy methods).  But the work scheduler should probably be an async channel.  We can do that in a follow-up?

grpc/src/client/load_balancing/pick_first.rs
dfawley: Done

grpc/src/client/load_balancing/pick_first.rs
arjan-bal: Yes, sounds good!

arjan-bal (approved):
LGTM

PR #2583, opened 2026-04-07
tonic-xds: Add xds retry support 

gu0keno0:
original issue: https://github.com/hyperium/tonic/issues/2444

## Motivation

As title, add xds retry support to tonix-xds, the implementation is based on https://github.com/grpc/proposal/blob/master/A44-xds-retry.md

## Solution
 Make the request cloneable and reuse tower::retry layer

### DIFF
--- a/tonic-xds/src/client/retry.rs
+++ b/tonic-xds/src/client/retry.rs
@@ -7,13 +7,13 @@ use std::task::{Context, Poll};
 use std::time::Duration;
 
 use arc_swap::ArcSwap;
-use backoff::backoff::Backoff;
 use backoff::ExponentialBackoffBuilder;
+use backoff::backoff::Backoff;
 use http::{Request, Response};
 use http_body::Body;
 use shared_http_body::{SharedBody, SharedBodyExt};
-use tower::retry::Retry;
 use tower::retry::Policy;
+use tower::retry::Retry;
 use tower::{Layer, Service};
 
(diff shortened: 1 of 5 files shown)


tonic-xds/src/client/retry.rs
```diff
@@ -0,0 +1,685 @@
...
+pub(crate) fn is_retryable_grpc_status_code(
+    code: tonic::Code,
+    retryable_codes: &[tonic::Code],
+) -> bool {
+    retryable_codes.contains(&code)
+}
+
+/// Check if a request should be retried, either because of a retryable connection error
+/// or because the gRPC response status code is in the retryable set.
+/// TODO: gRPC retriability is based on gRPC status code by default, in practice this may
```
LucioFranco, line 50: This is safe for now since this is contained within the xds protocol so its a close world of what is idempotent or not?

LucioFranco (approved):

tonic-xds/src/client/retry.rs
gu0keno0: It's safe as long as user return unavailable only for idempotent methods.

I did a small poc offline, with something like (tested on Java, Go, Cpp):
```
handle_rpc() {
  side_effect()
  abort(...)
}
```

In this case, client will return unavailable when server panics, even though side effect is already done. Therefore I was thinking, could we build a better way for clients to get clear signal about retriability, and added the TODO there.

YutaoMa (approved):

PR #2584, opened 2026-04-07
feat(tonic-xds): grpc channel builder

YutaoMa:
## Motivation

Ref: #2444 

Last PR to complete A27 gRFC support for basic xDS service discovery: wire all xDS components into `XdsChannelBuilder`.

## Solution

Key design decisions looking for in-depth reviews:

1. **One xDS client per channel**: each channel now creates its own `XdsClient` with its own ADS connection. Other gRPC implementation like `grpc-go` uses a singleton shared across channels. The plan is to switch to, or at least provide an option for, the same singleton pattern, but in a later PR.
2. **Default retry policy**: currently the channel builder does not expose API to configure the retry policy and always use the default values. Didn't include it in this PR because it's not part of the xDS spec, but worth discussing how will we plan to support it in future PRs.

## Testing

Aside from UTs, added example runnables that serve as a local integration test: 

```bash
% ./tonic-xds/examples/run_xds_example.sh

[xds] xDS server listening on 0.0.0.0:18000
[xds]   listener: my-service
[xds]   cluster:  my-cluster
[xds]   endpoints: 127.0.0.1:50051
[greeter] Greeter server listening on 0.0.0.0:50051
[client] Building xDS channel for target: xds:///my-service
[client] Channel built. Sending requests (Ctrl-C to stop)...
[client]
[xds]   -> envoy.config.listener.v3.Listener: 1 resource(s)
[xds]   -> envoy.config.cluster.v3.Cluster: 1 resource(s)
[xds]   -> envoy.config.endpoint.v3.ClusterLoadAssignment: 1 resource(s)
[greeter] Received request: name=request-1
[client] [1] Response: Hello request-1 from 0.0.0.0:50051
[greeter] Received request: name=request-2
[client] [2] Response: Hello request-2 from 0.0.0.0:50051
[greeter] Received request: name=request-3
[client] [3] Response: Hello request-3 from 0.0.0.0:50051
```

### DIFF
--- a/tonic-xds/examples/channel.rs
+++ b/tonic-xds/examples/channel.rs
@@ -4,28 +4,33 @@
 //! The channel discovers endpoints via the xDS management server and
 //! load-balances across them.
 //!
-//! # Prerequisites
+//! # Quick start
 //!
-//! 1. Start one or more greeter backends:
-//!    ```sh
-//!    PORT=50051 cargo run -p tonic-xds --example greeter_server
-//!    ```
+//! Run all three examples (greeter backend, xDS server, this client) together:
 //!
-//! 2. Start an xDS control plane (e.g., go-control-plane) configured to
-//!    return LDS/RDS/CDS/EDS pointing at the greeter backends.
+//! ```sh
+//! ./tonic-xds/examples/run_xds_example.sh
+//! ```
+//!
+//! # Running individually
+//!
+//! ```sh
+//! # Terminal 1: greeter backend
+//! PORT=50051 cargo run -p tonic-xds --example greeter_server --features testutil
+//!
+//! # Terminal 2: xDS control plane
+//! cargo run -p tonic-xds --example xds_server
+//!
+//! # Terminal 3: xDS client
+//! GRPC_XDS_BOOTSTRAP_CONFIG='{"xds_servers":[{"server_uri":"http://localhost:18000"}],"node":{"id":"test"}}' \
+//!     cargo run -p tonic-xds --example channel --features testutil
+//! ```
 //!
 //! # Configuration
 //!
 //! - `GRPC_XDS_BOOTSTRAP` — path to a bootstrap JSON file, **or**
 //! - `GRPC_XDS_BOOTSTRAP_CONFIG` — inline bootstrap JSON
 //! - `XDS_TARGET` — xDS target URI (default: `xds:///my-service`)
-//!
-//! # Usage
-//!
-//! ```sh
-//! GRPC_XDS_BOOTSTRAP_CONFIG='{"xds_servers":[{"server_uri":"localhost:18000"}],"node":{"id":"test"}}' \
-//!     cargo run -p tonic-xds --example channel
-//! ```
 
 use tonic_xds::testutil::proto::helloworld::{HelloRequest, greeter_client::GreeterClient};
 use tonic_xds::{XdsChannelBuilder, XdsChannelConfig, XdsUri};
@@ -44,8 +49,6 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("Channel built. Sending requests (Ctrl-C to stop)...\n");
 
     for i in 1.. {
-        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
-
         let request = HelloRequest {
             name: format!("request-{i}"),
         };
@@ -58,6 +61,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 eprintln!("[{i}] Error: {status}");
             }
         }
+
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
     }
 
     Ok(())
--- a/tonic-xds/examples/channel.rs
+++ b/tonic-xds/examples/channel.rs
@@ -1,0 +1,64 @@
+//! Example: send gRPC requests through an xDS-aware channel.
+//!
+//! Builds an xDS channel, then sends HelloRequest RPCs through it in a loop.
+//! The channel discovers endpoints via the xDS management server and
+//! load-balances across them.
+//!
+//! # Prerequisites
+//!
+//! 1. Start one or more greeter backends:
+//!    ```sh
+//!    PORT=50051 cargo run -p tonic-xds --example greeter_server
+//!    ```
+//!
+//! 2. Start an xDS control plane (e.g., go-control-plane) configured to
+//!    return LDS/RDS/CDS/EDS pointing at the greeter backends.
+//!
+//! # Configuration
+//!
+//! - `GRPC_XDS_BOOTSTRAP` — path to a bootstrap JSON file, **or**
+//! - `GRPC_XDS_BOOTSTRAP_CONFIG` — inline bootstrap JSON
+//! - `XDS_TARGET` — xDS target URI (default: `xds:///my-service`)
+//!
+//! # Usage
+//!
+//! ```sh
+//! GRPC_XDS_BOOTSTRAP_CONFIG='{"xds_servers":[{"server_uri":"localhost:18000"}],"node":{"id":"test"}}' \
+//!     cargo run -p tonic-xds --example channel
+//! ```
+
+use tonic_xds::testutil::proto::helloworld::{HelloRequest, greeter_client::GreeterClient};
+use tonic_xds::{XdsChannelBuilder, XdsChannelConfig, XdsUri};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let target_str = std::env::var("XDS_TARGET").unwrap_or_else(|_| "xds:///my-service".into());
+    let target = XdsUri::parse(&target_str)?;
+
+    println!("Building xDS channel for target: {target_str}");
+
+    let channel = XdsChannelBuilder::new(XdsChannelConfig::new(target)).build_grpc_channel()?;
+
+    let mut client = GreeterClient::new(channel);
+
+    println!("Channel built. Sending requests (Ctrl-C to stop)...\n");
+
+    for i in 1.. {
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
+
+        let request = HelloRequest {
+            name: format!("request-{i}"),
+        };
+
+        match client.say_hello(request).await {
+            Ok(response) => {
+                println!("[{i}] Response: {}", response.into_inner().message);
+            }
+            Err(status) => {
+                eprintln!("[{i}] Error: {status}");
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/tonic-xds/examples/xds_server.rs
+++ b/tonic-xds/examples/xds_server.rs
@@ -1,0 +1,278 @@
+//! Example: minimal xDS control plane for testing.
+//!
+//! Serves a static set of LDS/CDS/EDS resources via the ADS
+//! (Aggregated Discovery Service) SotW protocol. Useful for testing
+//! the `channel` example without an external control plane.
+//!
+//! # Quick start
+//!
+//! ```sh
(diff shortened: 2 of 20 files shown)


tonic-xds/examples/channel.rs
```diff
@@ -0,0 +1,64 @@
...
+//!
+//! Builds an xDS channel, then sends HelloRequest RPCs through it in a loop.
+//! The channel discovers endpoints via the xDS management server and
+//! load-balances across them.
+//!
+//! # Prerequisites
+//!
+//! 1. Start one or more greeter backends:
+//!    ```sh
+//!    PORT=50051 cargo run -p tonic-xds --example greeter_server
```
gu0keno0: provide a shell script maybe

tonic-xds/examples/channel.rs
```diff
@@ -0,0 +1,64 @@
...
+//!    ```
+//!
+//! 2. Start an xDS control plane (e.g., go-control-plane) configured to
```
gu0keno0: probably add another script to build go-control-plane in a temp folder, then start it with pre-defined resources.

tonic-xds/examples/channel.rs
```diff
@@ -0,0 +1,64 @@
...
+    println!("Building xDS channel for target: {target_str}");
+
+    let channel = XdsChannelBuilder::new(XdsChannelConfig::new(target)).build_grpc_channel()?;
+
+    let mut client = GreeterClient::new(channel);
+
+    println!("Channel built. Sending requests (Ctrl-C to stop)...\n");
+
+    for i in 1.. {
+        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
```
gu0keno0: no need for this?

tonic-xds/src/lib.rs
```diff
@@ -1,53 +1,115 @@
...
 //!
-//! // Use with your generated gRPC client
-//! // let client = MyServiceClient::new(xds_channel);
-//! // client.my_rpc_method(...).await;
+//! ### Using inline JSON
+//!
+//! ```rust,no_run
+//! use tonic_xds::{BootstrapConfig, XdsChannelBuilder, XdsChannelConfig, XdsUri};
+//!
+//! let bootstrap = BootstrapConfig::from_json(r#"{
```
gu0keno0, line 71: I feel that if someone decides to initialize xds in code, then native rust structs can be directly used and it's more straightforward, wdyt?

tonic-xds/src/lib.rs
```diff
@@ -1,53 +1,115 @@
...
-//! This crate provides an xDS-enabled [`tonic::client::GrpcService`] implementation ([`XdsChannelGrpc`])
-//! that automatically discovers, routes and load-balances across endpoints using the xDS protocol.
-//! The implementation will align with the
-//! [gRPC xDS features](https://github.com/grpc/grpc/blob/master/doc/grpc_xds_features.md).
+//! This crate provides [`XdsChannelGrpc`], a [`tonic::client::GrpcService`]
+//! that connects to an xDS management server (via ADS) and automatically
+//! discovers, routes, and load-balances requests across endpoints. The
+//! implementation follows the [gRPC xDS features] specification.
 //!
-//! In addition to gRPC, this crate also provides a generic [`tower::Service`] implementation ([`XdsChannel`])
```
gu0keno0, line 11: IMO this is an important feature, although less developed. We use it internally as well.

tonic-xds/src/lib.rs
```diff
@@ -1,53 +1,115 @@
...
+//!
+//! [gRFC A27]: https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md
+//!
+//! ## Bootstrap configuration
+//!
+//! The bootstrap can be supplied in three ways (in order of precedence):
+//!
+//! | Method | How |
+//! |--------|-----|
+//! | Programmatic | [`BootstrapConfig::from_json`] then [`XdsChannelConfig::with_bootstrap`] |
```
gu0keno0, line 33: Similar comment as below, if user decides to do this by code, then structured data is always better than unstructured json.

tonic-xds/src/xds/bootstrap.rs
```diff
@@ -13,19 +12,39 @@ const ENV_BOOTSTRAP_FILE: &str = "GRPC_XDS_BOOTSTRAP";
...
+/// ```
+///
+/// [gRFC A27]: https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md
 #[derive(Debug, Clone, Deserialize)]
 #[non_exhaustive]
-pub(crate) struct BootstrapConfig {
+pub struct BootstrapConfig {
     /// xDS management servers to connect to.
-    pub xds_servers: Vec<XdsServerConfig>,
+    pub(crate) xds_servers: Vec<XdsServerConfig>,
```
gu0keno0, line 43: Do we have accessors for these? They might be useful in unit / integ tests.

tonic-xds/src/xds/cluster_discovery.rs
```diff
@@ -0,0 +1,60 @@
...
+
+use tonic::transport::{Channel, Endpoint};
+
+use crate::client::endpoint::{EndpointAddress, EndpointChannel};
+use crate::client::lb::{BoxDiscover, ClusterDiscovery};
+use crate::xds::cache::XdsCache;
+use crate::xds::endpoint_manager::EndpointManager;
+
+/// Shared connector function that creates endpoint services from addresses.
+pub(crate) type EndpointConnector =
```
gu0keno0, line 19: Nit: use this type in endpointmanager? Or, shall we make this a factory trait? I'm considering doing it for new LB as well. We need such a factory to deal with reconnects. So we have at least 3 places where this will be used.

tonic-xds/src/client/channel.rs
```diff
@@ -1,168 +1,227 @@
...
+    pub fn with_bootstrap(mut self, bootstrap: BootstrapConfig) -> Self {
+        self.bootstrap = Some(bootstrap);
         self
     }
+
+    /// Eagerly loads bootstrap configuration from environment variables.
+    ///
+    /// This is optional — [`XdsChannelBuilder::build_grpc_channel`] falls back
+    /// to env vars automatically if no bootstrap is set. Use this method when
+    /// you want to surface bootstrap errors at config time rather than build time.
```
gu0keno0, line 54: Nit: strictly speaking env variable is a runtime stuff, because they are only read when binary is running.

tonic-xds/src/client/channel.rs
```diff
@@ -1,168 +1,227 @@
...
     #[must_use]
-    pub fn with_config(config: XdsChannelConfig) -> Self {
+    pub fn new(config: XdsChannelConfig) -> Self {
         Self {
             config: Arc::new(config),
         }
     }
 
-    /// Builds an `XdsChannel`, which takes generic request, endpoint, and service types and can be
-    /// used for generic HTTP services.
```
gu0keno0, line 147: This API will be needed in the future for HTTP use cases.

tonic-xds/src/client/channel.rs
```diff
@@ -429,4 +502,169 @@ mod tests {
...
+            EndpointConnector, XdsClusterDiscovery, default_endpoint_connector,
+        };
+        use crate::xds::routing::XdsRouter;
+
+        let router: Arc<dyn Router> = Arc::new(XdsRouter::new(&cache));
+        let connector: EndpointConnector = Arc::new(default_endpoint_connector);
+        let discovery: Arc<dyn ClusterDiscovery<EndpointAddress, EndpointChannel<Channel>>> =
+            Arc::new(XdsClusterDiscovery::new(cache, connector));
+
+        tokio::task::yield_now().await;
```
gu0keno0: what is this for?

gu0keno0 (approved):
LGTM overall

tonic-xds/examples/channel.rs
YutaoMa: `go-control-plane` required some go code changes that's hard to include here. I'm switching to a Rust example xDS server approach in the new commit.

tonic-xds/examples/channel.rs
YutaoMa: Related to the next comment.

tonic-xds/examples/channel.rs
YutaoMa: It's for spreading out the requests so the logs don't just spam the stdout.

It's before each request instead of after because of a separate issue: the current channel design doesn't block RPC until the config is valid. This is a miss when wiring the routing layer in, which I'm addressing in the next commit. It should have the same behavior as other grpc implementations which blocks RPC until a valid channel can be constructed.

tonic-xds/src/lib.rs
YutaoMa: Also related to other Bootstrap config comments below: valid suggestion, we already have a `new` placeholder consturctor in code, hidden behind `pub(crate)`. I'm hesitant to expose that API right now because with so many knobs existing in xDS config, we'll need more design discussion on how to best expose the API (while maintaining a good backward compatibility, likely with some sort of builder pattern). I'm noting this as a TODO for the future PRs.

tonic-xds/src/lib.rs
YutaoMa: HTTP, or more generic channels, are still in scope, just that we'll add them in future releases.

tonic-xds/src/lib.rs
YutaoMa: see above comment.

tonic-xds/src/xds/bootstrap.rs
YutaoMa: related to the BootstrapConfig API discussion. I'll add them along with the change to add public programmatic constructor.

tonic-xds/src/xds/cluster_discovery.rs
YutaoMa: Good suggestion. Will extract this to a trait in the TLS support PR as we'll be dealing with more transport options then.

tonic-xds/src/client/channel.rs
YutaoMa: The config time and build time here refers to the BootstrapConfig construction time and XdsChannel construction time, I'll update the wording to avoid the confusion with code build time.

tonic-xds/src/client/channel.rs
YutaoMa: see comment above, will add HTTP API back eventually.

tonic-xds/src/client/channel.rs
YutaoMa: see the channel blocking comment, adding proper channel construction blocking to avoid needing this kind of flaky awaits.

tonic-xds/examples/xds_server.rs
```diff
@@ -0,0 +1,278 @@
...
+const TYPE_CLUSTER: &str = "type.googleapis.com/envoy.config.cluster.v3.Cluster";
+const TYPE_ENDPOINTS: &str = "type.googleapis.com/envoy.config.endpoint.v3.ClusterLoadAssignment";
+const TYPE_HCM: &str = "type.googleapis.com/envoy.extensions.filters.network.http_connection_manager.v3.HttpConnectionManager";
+
+/// Static xDS resource snapshot, keyed by type URL.
+struct Snapshot {
+    resources: HashMap<String, Vec<Any>>,
+}
+
+impl Snapshot {
```
gu0keno0, line 71: What about loading these from yaml files? It should be more readable than constructing them directly in code.

gu0keno0 (approved):

tonic-xds/examples/xds_server.rs
YutaoMa: The `envoy-types` crate we are using doesn't support json/yaml parsing to proto types. I'll refactor this to config file once we switch to in-crate binding generation where we'll be able to add pbjson support.

PR #2587, opened 2026-04-11
tonic-xds: Implement subchannel state machine

gu0keno0:
Implement the type-statemachine for load balancer subchannels. This is preparing for implementing the actual load balancer.

Ref: https://github.com/hyperium/tonic/issues/2444

## Motivation

A real world xDS load-balancer needs to deal with outlier detection and reconnects. Therefore, subchannels managed by load balancer needs to implement a state machine. This PR implements it via type-states.

In the next PR, the load balancer tower layer will drive the state transitions in its poll_ready() implementation. This way, we get full control and visibility in to the managed channels. Therefore, we opens up the flexibility of implementing different LB strategies (e.g. sticky, round-robin, etc), compared to the hardcoded P2C stategy offered by tower::balance that we are currently using.

## Solution
Define new type for each state, state transition is implemented by consuming the type for the source state. ConnectingChannel and EjectedChannel needs to be implemented as Streams so that load balancer can use StreamMap to poll them to advance their state transitions.

### DIFF
--- a/tonic-xds/src/client/loadbalance/channel_state.rs
+++ b/tonic-xds/src/client/loadbalance/channel_state.rs
@@ -4,16 +4,17 @@
 //! This prevents using a channel in an invalid state at compile time.
 //!
 //! ```text
-//!                +---reconnect---+
-//!                |               |
-//!                v               |
-//! Idle --> Connecting --> Ready --+--eject--> Ejected
-//!                ^                               |
-//!                +----------reconnect------------+
+//!                +-----------+
+//!                |           |
+//!                v           |
+//! Idle --> Connecting --> Ready <--+--> Ejected
+//!                ^                       |
+//!                |                       |
+//!                +-----------------------+
 //! ```
 //!
-//! State changes are all one-shot: [`ConnectingChannel`] and [`EjectedChannel`] are
-//! [`Future`]s, not streams. The caller (typically a pool) uses [`KeyedFutures`] to
+//! State changes are all one-shot. [`ConnectingChannel`] and [`EjectedChannel`] are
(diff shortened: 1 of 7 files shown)


gu0keno0:
WIP, will replace stream with futures, because streammap's poll is O(N).

tonic-xds/src/client/loadbalance/channel_state.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl<S, Req> Service<Req> for ReadyChannel<S>
+where
+    S: Service<Req> + Clone + Send + 'static,
+    S::Future: Send + 'static,
+    S::Error: Into<BoxError>,
+    Req: Send + 'static,
+{
+    type Response = S::Response;
+    type Error = BoxError;
```
ankurmittal: why not S::Future and S::Error?

gu0keno0: It's because ReadyChannel.inner is LbChannel, and LbChannel wraps S::future so that in-flight guard can be used.

ankurmittal (commented):
rest LGTM

tonic-xds/src/client/loadbalance/channel_state.rs
ankurmittal: still S::future and S::Error should work right? let inner make decision about its type?

gu0keno0: Originally ReadyChannel wraps a fixed type of LbChannel, which returns BoxFuture and BoxError.

Now I've changed it and let ReadyChannel just wraps generic S, so it can now use S::Future.

ankurmittal (approved):

PR #2589, opened 2026-04-13
feat(grpc): Add name resolver for the unix scheme

arjan-bal:
This PR adds a name resolver for parsing targets with the `unix` and `unix-abstract` schemes. These formats are specified in the [gRPC naming documentation](https://github.com/grpc/grpc/blob/2ce8152a535f251ed105c957472a1fbded8f22c3/doc/naming.md?plain=1#L37-L50).

Support for connecting to unix sockets will be added in follow-up PRs.

### DIFF
--- a/grpc/src/client/name_resolution/dns/test.rs
+++ b/grpc/src/client/name_resolution/dns/test.rs
@@ -27,16 +27,12 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
 
-use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::mpsc::{self};
 use url::Host;
 
-use crate::client::name_resolution::ChannelController;
 use crate::client::name_resolution::Resolver;
 use crate::client::name_resolution::ResolverOptions;
-use crate::client::name_resolution::ResolverUpdate;
 use crate::client::name_resolution::Target;
-use crate::client::name_resolution::WorkScheduler;
 use crate::client::name_resolution::backoff::BackoffConfig;
 use crate::client::name_resolution::backoff::DEFAULT_EXPONENTIAL_CONFIG;
 use crate::client::name_resolution::dns::DnsOptions;
@@ -48,7 +44,8 @@ use crate::client::name_resolution::dns::get_resolving_timeout;
 use crate::client::name_resolution::dns::parse_endpoint_and_authority;
 use crate::client::name_resolution::dns::reg;
 use crate::client::name_resolution::global_registry;
-use crate::client::service_config::ServiceConfig;
+use crate::client::name_resolution::test_utils::TestChannelController;
+use crate::client::name_resolution::test_utils::TestWorkScheduler;
 use crate::rt::BoxFuture;
 use crate::rt::GrpcRuntime;
 use crate::rt::TcpOptions;
@@ -173,42 +170,13 @@ pub(crate) fn target_parsing() {
     }
 }
 
-struct FakeWorkScheduler {
-    work_tx: UnboundedSender<()>,
-}
-
-impl WorkScheduler for FakeWorkScheduler {
-    fn schedule_work(&self) {
-        self.work_tx.send(()).unwrap();
-    }
-}
-
-struct FakeChannelController {
-    update_result: Result<(), String>,
-    update_tx: UnboundedSender<ResolverUpdate>,
-}
-
-impl ChannelController for FakeChannelController {
-    fn update(&mut self, update: ResolverUpdate) -> Result<(), String> {
-        println!("Received resolver update: {:?}", &update);
-        self.update_tx.send(update).unwrap();
-        self.update_result.clone()
-    }
-
-    fn parse_service_config(&self, _: &str) -> Result<ServiceConfig, String> {
-        Err("Unimplemented".to_string())
-    }
-}
-
 #[tokio::test]
 pub(crate) async fn dns_basic() {
     reg();
     let builder = global_registry().get("dns").unwrap();
     let target = &"dns:///localhost:1234".parse().unwrap();
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let opts = ResolverOptions {
         authority: "ignored".to_string(),
         runtime: rt::default_runtime(),
@@ -219,10 +187,7 @@ pub(crate) async fn dns_basic() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
     // A successful endpoint update should be received.
     let update = update_rx.recv().await.unwrap();
@@ -235,9 +200,7 @@ pub(crate) async fn invalid_target() {
     let builder = global_registry().get("dns").unwrap();
     let target = &"dns:///:1234".parse().unwrap();
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let opts = ResolverOptions {
         authority: "ignored".to_string(),
         runtime: rt::default_runtime(),
@@ -248,10 +211,7 @@ pub(crate) async fn invalid_target() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
     // An error endpoint update should be received.
     let update = update_rx.recv().await.unwrap();
@@ -327,9 +287,7 @@ pub(crate) async fn dns_lookup_error() {
     let builder = global_registry().get("dns").unwrap();
     let target = &"dns:///grpc.io:1234".parse().unwrap();
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let runtime = FakeRuntime {
         inner: TokioRuntime::default(),
         dns: FakeDns {
@@ -347,10 +305,7 @@ pub(crate) async fn dns_lookup_error() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
     // An error endpoint update should be received.
     let update = update_rx.recv().await.unwrap();
@@ -360,9 +315,7 @@ pub(crate) async fn dns_lookup_error() {
 #[tokio::test]
 pub(crate) async fn dns_lookup_timeout() {
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let runtime = FakeRuntime {
         inner: TokioRuntime::default(),
         dns: FakeDns {
@@ -388,10 +341,7 @@ pub(crate) async fn dns_lookup_timeout() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
 
     // An error endpoint update should be received.
@@ -402,9 +352,7 @@ pub(crate) async fn dns_lookup_timeout() {
 #[tokio::test]
 pub(crate) async fn rate_limit() {
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let opts = ResolverOptions {
         authority: "ignored".to_string(),
         runtime: rt::default_runtime(),
@@ -426,10 +374,7 @@ pub(crate) async fn rate_limit() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
     // A successful endpoint update should be received.
     let update = update_rx.recv().await.unwrap();
@@ -452,9 +397,7 @@ pub(crate) async fn rate_limit() {
 #[tokio::test]
 pub(crate) async fn re_resolution_after_success() {
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let opts = ResolverOptions {
         authority: "ignored".to_string(),
         runtime: rt::default_runtime(),
@@ -476,10 +419,7 @@ pub(crate) async fn re_resolution_after_success() {
     // Wait for schedule work to be called.
     work_rx.recv().await.unwrap();
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Ok(()),
-    };
+    let mut channel_controller = TestChannelController::new(update_tx);
     resolver.work(&mut channel_controller);
     // A successful endpoint update should be received.
     let update = update_rx.recv().await.unwrap();
@@ -496,9 +436,7 @@ pub(crate) async fn re_resolution_after_success() {
 #[tokio::test]
 pub(crate) async fn backoff_on_error() {
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
     let opts = ResolverOptions {
         authority: "ignored".to_string(),
         runtime: rt::default_runtime(),
@@ -525,10 +463,8 @@ pub(crate) async fn backoff_on_error() {
     let mut resolver = DnsResolver::new(dns_client, opts, dns_opts);
 
     let (update_tx, mut update_rx) = mpsc::unbounded_channel();
-    let mut channel_controller = FakeChannelController {
-        update_tx,
-        update_result: Err("test_error".to_string()),
-    };
+    let mut channel_controller =
+        TestChannelController::new_with_result(update_tx, Err("test_error".to_string()));
 
     // As the channel returned an error to the resolver, the resolver will
     // backoff and re-attempt resolution.
--- a/grpc/src/client/name_resolution/dns/test.rs
+++ b/grpc/src/client/name_resolution/dns/test.rs
@@ -24,7 +24,6 @@
 
 use std::future::Future;
 use std::pin::Pin;
-use std::sync::Arc;
 use std::time::Duration;
 
 use tokio::sync::mpsc::{self};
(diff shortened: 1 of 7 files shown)


grpc/src/client/name_resolution/mod.rs
```diff
@@ -346,6 +351,16 @@ impl Resolver for NopResolver {
     }
 }
 
+fn nop_resolver_for_err(err: String, options: ResolverOptions) -> Box<dyn Resolver> {
```
dfawley: Can this be `NopResolver::new_with_err(err: String, work_scheduler: WorkScheduler)` or `StaticResolver::new(update: ResolverUpdate, work_scheduler: WorkScheduler)`?

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+use crate::client::name_resolution::ResolverUpdate;
+use crate::client::name_resolution::Target;
+use crate::client::name_resolution::UNIX_NETWORK_TYPE;
+use crate::client::name_resolution::global_registry;
+use crate::client::name_resolution::nop_resolver_for_err;
+
+#[derive(Debug, Copy, Clone)]
+enum UnixScheme {
+    Standard,
+    #[cfg(target_os = "linux")]
```
dfawley: Any harm in leaving this in all builds?

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+    Abstract,
+}
+
+impl UnixScheme {
+    const fn as_str(&self) -> &'static str {
+        match self {
+            UnixScheme::Standard => "unix",
+            #[cfg(target_os = "linux")]
```
dfawley: Same?

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+///   included it in `abstract_path`.
+/// - Note: Abstract sockets are a Linux-specific kernel feature.
+fn parse_endpoint_and_authority(target: &Target, scheme: UnixScheme) -> Result<Address, String> {
+    let host_port = target.authority_host_port();
+    if !host_port.is_empty() {
+        return Err(format!("invalid (non-empty) authority: {host_port}"));
+    }
+    let addr_string = match scheme {
+        UnixScheme::Standard => target.path().to_owned(),
+        #[cfg(target_os = "linux")]
```
dfawley: And this as well

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+/// - In the first form (`unix:path`), the path may be relative or absolute.
+/// - In the second form (`unix:///absolute_path`), the path must be absolute.
+///   The last of the three slashes is treated as the root of the filesystem
+///   path (e.g., `/absolute_path`).
+///
+/// ### `unix-abstract` (Abstract Namespace)
+/// Valid format: `unix-abstract:abstract_path`
+/// - `abstract_path` indicates a socket name in the abstract namespace.
+/// - The name has no connection with filesystem pathnames and bypasses standard
+///   filesystem permissions; any process or user may access the socket.
```
dfawley: I don't think I realized this.  So it means local creds should consider a unix-abstract connection as insecure?

dfawley (changes requested):

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+use crate::client::name_resolution::ResolverBuilder;
+use crate::client::name_resolution::ResolverOptions;
+use crate::client::name_resolution::ResolverUpdate;
+use crate::client::name_resolution::Target;
+use crate::client::name_resolution::UNIX_NETWORK_TYPE;
+use crate::client::name_resolution::global_registry;
+use crate::client::name_resolution::nop_resolver_for_err;
+
+#[derive(Debug, Copy, Clone)]
+enum UnixScheme {
```
sauravzg: Assuming we plan on continuing with os specific builds (like abstract sockets for linux only), we might want to consider use `target_family` to make this unix specific build only?

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+    Standard,
+    #[cfg(target_os = "linux")]
+    Abstract,
```
sauravzg: What are the advantages of having a single resolver handling unix  and unix-abstract vs having two different name resolvers?

So, if we can move this into two different resolvers, we can make the entire resolver modules os conditional instead of sprinkling a bunch of cfg blocks throughout the entire code.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Unix sockets are also supported on Windows 10+, though abstract Unix sockets are Linux-specific. I’ve added a guard around the abstract socket code, as these sockets are also protected by an OS-specific guard in Tokio: https://github.com/tokio-rs/tokio/blob/1b11840f539ced93855df362eb51149ba08903a0/tokio/src/net/unix/stream.rs#L76-L78

I’m not sure what the failure message would be if we pass an abstract socket path (beginning with `\0`) to Tokio on a non-Linux platform, so I’ll test this out.

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,259 @@
...
+            Ok(())
+        }
+
+        fn parse_service_config(&self, _: &str) -> Result<ServiceConfig, String> {
+            Err("Unimplemented".to_string())
+        }
+    }
+
+    #[tokio::test]
+    async fn test_unix_resolver() {
```
sauravzg: Are there any style guides around table driven testing around rust?

While go style guide recommends table driven tests: go/go-style/decisions#table-driven-tests 

It's usually frowned upon in other languages because it introduces control structures : http://go/unit-testing-practices?polyglot=cpp#logic , go/parameterized-testing?polyglot=python#loops .
With the recommendation being to use parameterized testing frameworks, which in rust is going to be something like rstest I assume.

grpc/src/client/name_resolution/unix.rs
sauravzg: https://github.com/tokio-rs/mio/issues/1609 - I don't think tokio supports unix sockets on windows. The code doesn't have any `std::os::windows` . 

So, maybe we stay safe and not support it either for now? by conditionally flagging it for unix family.

sauravzg (changes requested):

grpc/src/client/name_resolution/mod.rs
arjan-bal: Yes, that looks better. Changed.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Started a thread on the cross-language maintainers chat group. The security team is looking into this.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Based on Saurav's feedback, I've removed the enum and added two independent modules with resolver builders for the two schemes. Each module has its own platform guard (unix / linux).

grpc/src/client/name_resolution/unix.rs
arjan-bal: Based on Saurav's feedback, I've removed the enum and added two independent modules with resolver builders for the two schemes. Each module has its own platform guard (unix / linux).

grpc/src/client/name_resolution/unix.rs
arjan-bal: Based on Saurav's feedback, I've removed the enum and added two independent modules with resolver builders for the two schemes. Each module has its own platform guard (unix / linux).

grpc/src/client/name_resolution/unix.rs
arjan-bal: You're right, Tonic supports unix sockets only on Unix platforms. I've added two independent modules with resolver builders for the two schemes. Each module has its own platform guard (unix / linux).

grpc/src/client/name_resolution/unix.rs
arjan-bal: The single resolver was intended for better code-reuse. It was based on Go's implementation, but the guard blocks make the code worse in Rust.  Now I've removed the enum and added two independent modules with resolver builders for the two schemes. Each module has its own platform guard (unix / linux).

grpc/src/client/name_resolution/unix.rs
arjan-bal: > it's usually frowned upon in other languages because it introduces control structures

I don't believe table-driven tests inherently result in control structures. In fact, we should aim to avoid control structures even when writing table-driven tests. 

> With the recommendation being to use parameterized testing frameworks

It seems to me that parameterized testing frameworks are simply tools for writing table-driven tests with less boilerplate. For instance, the [Go tab in the parameterized testing doc](http://go/parameterized-testing?polyglot=go#setting-up-a-parameterized-test) highlights simple table-driven tests as the standard.

Coming from a Go background, I don't feel the need for a dedicated testing framework yet, as there isn't much boilerplate code to manage. Furthermore, `rstest` doesn't seem to be widely adopted; my concern is that if we adopt the framework and the project becomes unmaintained, migrating away could involve significant work.

---

Is the concern that the `want_success` field introduces a branch in the validation logic? If so, I can split the success and failure cases to avoid this branching entirely.

grpc/src/client/name_resolution/dns/test.rs
```diff
@@ -173,42 +170,13 @@ pub(crate) fn target_parsing() {
...
 #[tokio::test]
 pub(crate) async fn dns_basic() {
     reg();
     let builder = global_registry().get("dns").unwrap();
     let target = &"dns:///localhost:1234".parse().unwrap();
     let (work_tx, mut work_rx) = mpsc::unbounded_channel();
-    let work_scheduler = Arc::new(FakeWorkScheduler {
-        work_tx: work_tx.clone(),
-    });
+    let work_scheduler = Arc::new(TestWorkScheduler::new(work_tx.clone()));
```
dfawley: Why is this being cloned? I don't think it's being re-used.

(This is related to my next comment...)

grpc/src/client/name_resolution/test_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+use crate::client::service_config::ServiceConfig;
+
+/// A work scheduler for testing.
+pub struct TestWorkScheduler {
+    pub work_tx: UnboundedSender<()>,
+}
+
+impl TestWorkScheduler {
+    /// Creates a new `TestWorkScheduler`.
+    pub fn new(work_tx: UnboundedSender<()>) -> Self {
```
dfawley: Can we make this return the `UnboundedReceiver<()>` and own the channel construction instead?  That makes the call sites simpler.  It doesn't work if something existing needs to use the channel to send other things, too, but I'm not sure that should be needed, or is happening either.

grpc/src/client/name_resolution/test_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+
+/// A channel controller for testing.
+pub struct TestChannelController {
+    pub update_result: Result<(), String>,
+    pub update_tx: UnboundedSender<ResolverUpdate>,
+}
+
+impl TestChannelController {
+    /// Creates a new `TestChannelController` that returns `Ok(())` on update.
+    pub fn new(update_tx: UnboundedSender<ResolverUpdate>) -> Self {
```
dfawley: Similar idea.

grpc/src/client/name_resolution/test_utils.rs
```diff
@@ -0,0 +1,87 @@
...
+        Self {
+            update_result: Ok(()),
+            update_tx,
+        }
+    }
+
+    /// Creates a new `TestChannelController` with a specified update result.
+    pub fn new_with_result(
```
dfawley: Do we want to make this return an `UnboundedSender<Result<(), String>>` instead, and allow the test the ability to change the response for each call to `update`?

dfawley (commented):
LGTM, just some small stuff about the testing helpers.

grpc/src/client/name_resolution/unix.rs
sauravzg: >  I don't believe table-driven tests inherently result in control structures. In fact, we should aim to avoid control structures even when writing table-driven tests.
- We have a for loop and an if in the test.

go/parameterized-testing?polyglot=go#loops is actually noted as an exception because that's the popular model in go that's not used in any other language. Most languages actually call it out as an anti pattern.
So, I am not sure if we should use go'd model of parameterised tests in rust.


So here's my issue with table driven tests and why there's a need for test frameworks. 
- Table driven tests hide correct error messages. Testcases are simply an entry in the loop, with no formal  integration with the test framework. When a testcase fails, I see "test case x fails" or I am at the mercy of test developer, but when integrated with a parameterized test framework, such a test is integrated into the framework with a well defined name. go/unit-test-practices#naming . Ex: c++ param tests are essentially a macro that repeat the test body n times and require the author to provide a test name, so each test case is now a test and not an entry in a loop.
- Table driven tests don't guarantee state reset and are prone to bugs since this responsibility moves to the developer. Same, c++ param tests convert test cases into test , which ensures the guarantees of an empty slate and appropriate cleanup of fixture.
- Table driven tests tend to be misused when n equivalent tests can be used to provide smaller, simpler and more readable small tests. Tests should read like prose. If most of the logic spend on writing the driver the tests become less readable. Although this is a subjective take, unlike the other two.


My recommendation
- Prefer writing n small tests with each given their unique name. It's okay if we end up repeating some code across tests which table driven tests save. The benefit of a well named small test that's easy to fix in isolation is very underappreciated. go/tott/643
- If we decide that we need parameterized test(which I don't think we need in the current case), consult a rust style guide expert within(preferred) or outside google to see their opinion on how to go about it. Defaulting to go patterns for this may not be the correct choice, since it's the exception and not the norm. 



While my opinions around this are strong, we should still be okay to submit this and I'll LGTM.

sauravzg (commented):
LGTM. But please consult a readability expert before deciding to follow go's test patterns or create a todo to revisit it in future.

sauravzg (approved):

grpc/src/client/name_resolution/test_utils.rs
arjan-bal: Yes, that simplifies things. I've also changed the constructor to return an `Arc<dyn WorkScheduler>` since it was being type erased by every caller. I had to change the name from `new` since clippy warned that `new` functions are expected to return `Self`.

grpc/src/client/name_resolution/test_utils.rs
arjan-bal: Updated.

grpc/src/client/name_resolution/test_utils.rs
arjan-bal: There is a test that updates the result midway. Previously, this worked because the `update_result` field was `pub`. I've removed the second constructor in favor of a setter for `update_result`, which tests can now use to update the returned result at any time.

grpc/src/client/name_resolution/dns/test.rs
arjan-bal: The clone is unnecessary, I don't recall why it's there. It's removed after the refactoring the constructor.

grpc/src/client/name_resolution/unix.rs
arjan-bal: While Problems 2 and 3 can be mitigated by enforcing best practices during code review, Problem 1 is an inherent limitation of the test framework that cannot be bypassed without macros. Unlike Go's `t.Run()` method, `cargo test` does not support dynamic subtest reporting within a single function. Instead, we must generate individual test functions for every test case at compile time.

---

I think this test is a perfect candidate for table driven tests because all the test cases share the same code and only differ in the input and output values.

---

While looking into Google's recommended test frameworks, I found the Rust port of Google Test (`googletest-rust`). Although it does not natively support parameterized tests, the documentation recommends integrating it with `rstest`: https://github.com/google/googletest-rust?tab=readme-ov-file#interoperability. The repository includes integration tests to guarantee compatibility, and I found multiple examples in google3 using `#[gtest]` and `#[rstest]` together, which validates this.

I've added `rstest` as a dev dependency and updated the test. I also split the success and failure test cases to remove conditional logic from the validation steps.

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,158 @@
...
+    use crate::client::name_resolution::test_utils::TestChannelController;
+    use crate::client::name_resolution::test_utils::TestWorkScheduler;
+    use crate::rt;
+
+    #[rstest]
+    #[case::relative_path("unix:path/to/socket", "path/to/socket")]
+    #[case::absolute_path("unix:/absolute/path", "/absolute/path")]
+    #[case::absolute_path_with_slashes("unix:///absolute/path", "/absolute/path")]
+    #[tokio::test]
+    async fn unix_resolver_success(#[case] input: &str, #[case] want_addr: &str) {
```
dfawley: What's the advantage of this over:

```rust
async fn run_success(input: &str, want_addr: &str) { ... }

#[tokio::test]
async success_relative_path() {
   run_success("unix:path/to/socket", "path/to/socket").await;
}

...
```

(And you could _very_ easily use a tiny macro that creates the 1-line test function if you didn't want to waste 4 lines per test case.)

I can think of one disadvantage of using rstest here without even looking at the crate, which is that `grep -r <test_name>` won't work anymore.

Generally I dislike frameworks, as it ultimately requires everyone to become an expert in some new DSL, and it makes debugging errors harder as things tend to get obfuscated by the framework.

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,158 @@
...
+        let endpoints = update.endpoints.expect("Should have succeeded");
+        assert_eq!(endpoints.len(), 1);
+
+        let addr = &endpoints[0].addresses[0];
+        assert_eq!(addr.network_type, UNIX_NETWORK_TYPE);
+        assert_eq!(&*addr.address, want_addr);
+    }
+
+    #[tokio::test]
+    async fn unix_resolver_failure() {
```
dfawley: Nit: rename to `unix_resolver_error_with_authority` or something like that.

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,158 @@
...
+        let (mut channel_controller, mut update_rx) = TestChannelController::new_pair();
+        resolver.work(&mut channel_controller);
+
+        let update = update_rx.recv().await.unwrap();
+        let endpoints = update.endpoints.expect("Should have succeeded");
+        assert_eq!(endpoints.len(), 1);
+
+        let addr = &endpoints[0].addresses[0];
+        assert_eq!(addr.network_type, UNIX_NETWORK_TYPE);
+        assert_eq!(&*addr.address, want_addr);
```
dfawley: Endpoints has Eq so can't we do simply:
```rust
let want_endpoint = Endpoint::new(UNIX_NETWORK_TYPE, want_addr);
assert_eq!(update.endpoints, Ok(vec![want_endpoint]), "did not receive expected endpoint");
```

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,158 @@
...
+
+        let target: Target = input.parse().expect("Failed to parse target");
+        let (work_scheduler, mut work_rx) = TestWorkScheduler::new_pair();
+        let opts = ResolverOptions {
+            authority: "ignored".to_string(),
+            runtime: rt::default_runtime(),
+            work_scheduler: work_scheduler.clone(),
+        };
+
+        let builder = global_registry().get("unix").expect("scheme not found");
```
dfawley, line 107: Nit: we should be consistent about our capitalization of `expect` strings (`"Failed"` is capitalized above).

grpc/src/client/name_resolution/unix.rs
```diff
@@ -0,0 +1,158 @@
...
+        let mut resolver = builder.build(&target, opts);
+
+        // Wait for work to be scheduled.
+        work_rx.recv().await.unwrap();
+
+        let (mut channel_controller, mut update_rx) = TestChannelController::new_pair();
+        resolver.work(&mut channel_controller);
+
+        let update = update_rx.recv().await.unwrap();
+        let endpoints = update.endpoints.expect("Should have succeeded");
```
dfawley: Is `expect("should have succeeded")` any more useful than `unwrap()`?

Let's make sure we're following:

https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

> We recommend that expect messages are used to describe the reason you expect the Result should be Ok.

https://doc.rust-lang.org/std/option/enum.Option.html#recommended-message-style

> We recommend that expect messages are used to describe the reason you expect the Option should be Some.

And if the string adds no value then we should just use `unwrap()` instead.

grpc/src/client/name_resolution/unix.rs
dfawley: > Table driven tests don't guarantee state reset and are prone to bugs since this responsibility moves to the developer. Same, c++ param tests convert test cases into test , which ensures the guarantees of an empty slate and appropriate cleanup of fixture.

Conversely, restarting every test with an empty slate can prevent discovering real bugs where state unintentionally leaked between iterations unexpectedly.  (I don't think this argument is an obvious win for either side is my point.)

I think the biggest argument against table-driven tests in Rust is that the error messages are kind of awful unless you are _very_ meticulous about making sure you never use `unwrap` and that every `expect`/`assert` explains what case you were running when the error happened. It's probably OK for testing _pure functions_ that require no setup/teardown (or none that could error?), but otherwise it does seem like we should avoid them.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Fixed throughout.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Removed the `expect` here and from other places where the message was redundant.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Yes, changed. I didn't think of this. We don't have any constructors for `Endpoint` presently, so I'm creating a struct and using `Default::default()..`.

grpc/src/client/name_resolution/unix.rs
arjan-bal: Renamed.

grpc/src/client/name_resolution/unix.rs
arjan-bal: I can't think of any disadvantages to this pattern. I hadn't considered making the entire test body a single helper function. My main concern with helpers was that splitting the test into multiple smaller helpers for reuse would hurt readability.

I've removed `rstest` dependency and updated the code to follow this approach.

FYI @sauravzg

sauravzg: Yep. This is something that I am alligned with i.e. prefer n tests over parameterized ones. Abstract within tests as necessary.

grpc/src/client/name_resolution/unix.rs
```diff
@@ -152,7 +167,7 @@ mod tests {
         resolver.work(&mut channel_controller);
 
         let update = update_rx.recv().await.unwrap();
-        let err = update.endpoints.expect_err("Should have failed");
+        let err = update.endpoints.err().unwrap();
```
dfawley: Can be `unwrap_err()` instead.

dfawley (approved):

grpc/src/client/name_resolution/unix.rs
arjan-bal: Changed.

PR #2590, opened 2026-04-14
Modify the ServerMessage abstractions and Generic Streams API

sauravzg:
- Add a new type `SendMessage`. Mostly the same as client `RecvStream` without the `StreamClosed` because that isn't needed.
- Change the API for RecvStream `next` to return Option<Result<(),()>> .
   - `Some(OK())` - Successful next
   - `Some(Err())` - End of stream with error
   - `None` - Eos 
   - Subsequent calls after end stream result in error.
- Change the API for SendStream to return `Result<(),()>` 
- All CallOptions to the `Handle` and friends.

### DIFF
--- a/grpc/src/core/mod.rs
+++ b/grpc/src/core/mod.rs
@@ -129,9 +129,15 @@ pub enum ResponseStreamItem<M> {
 /// void as the received message is passed in via the `next` method.
 pub type ClientResponseStreamItem = ResponseStreamItem<()>;
 
-/// The server's view of a ResponseStream in a SendStream: the message type is
-/// part of the payload provided to the `send` method.
-pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
+/// The server's view of a ResponseStream in a SendStream, using references to avoid allocations.
+pub enum ServerResponseStreamItem<'a> {
+    /// Indicates the headers for the stream.
+    Headers(ResponseHeaders),
+    /// Indicates a message on the stream.
+    Message(&'a dyn SendMessage),
+    /// Indicates trailers were received on the stream and includes the trailers.
+    Trailers(Trailers),
+}
 
 /// Contains all information transmitted in the response headers of an RPC.
 #[derive(Debug, Clone, Default)]
--- a/grpc/src/inmemory/mod.rs
+++ b/grpc/src/inmemory/mod.rs
@@ -61,8 +61,8 @@ use crate::core::ClientResponseStreamItem;
 use crate::core::RecvMessage;
 use crate::core::RequestHeaders;
 use crate::core::ResponseHeaders;
-use crate::core::ResponseStreamItem;
 use crate::core::SendMessage;
+use crate::core::ServerResponseStreamItem;
 use crate::core::Trailers;
 use crate::credentials::SecurityLevel;
 use crate::credentials::client::ClientConnectionSecurityContext;
@@ -195,17 +195,16 @@ pub struct InMemoryServerSendStream {
 impl ServerSendStream for InMemoryServerSendStream {
     async fn send<'a>(
         &mut self,
-        item: crate::core::ServerResponseStreamItem<'a>,
+        item: ServerResponseStreamItem<'a>,
         _options: ServerSendOptions,
     ) -> Result<(), ()> {
         let inmemory_item = match item {
-            ResponseStreamItem::Headers(h) => InMemoryResponseStreamItem::Headers(h),
-            ResponseStreamItem::Message(m) => {
+            ServerResponseStreamItem::Headers(h) => InMemoryResponseStreamItem::Headers(h),
+            ServerResponseStreamItem::Message(m) => {
                 let buf = m.encode().map_err(|_| ())?;
                 InMemoryResponseStreamItem::Message(buf)
             }
-            ResponseStreamItem::Trailers(t) => InMemoryResponseStreamItem::Trailers(t),
-            ResponseStreamItem::StreamClosed => InMemoryResponseStreamItem::StreamClosed,
+            ServerResponseStreamItem::Trailers(t) => InMemoryResponseStreamItem::Trailers(t),
         };
 
         self.tx.send(inmemory_item).map_err(|_| ())
@@ -217,12 +216,16 @@ pub struct InMemoryServerRecvStream {
 }
 
 impl ServerRecvStream for InMemoryServerRecvStream {
-    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Result<(), ()> {
+    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Option<Result<(), ()>> {
         match self.rx.recv().await {
             Some(InMemoryRequestStreamItem::Message(mut buf)) => {
-                msg.decode(&mut buf).map_err(|_| ())
+                if msg.decode(&mut buf).is_err() {
+                    return Some(Err(()));
+                }
+                Some(Ok(()))
             }
-            _ => Err(()),
+            Some(InMemoryRequestStreamItem::StreamClosed) => None,
+            None => None,
         }
     }
 }
--- a/grpc/src/server/mod.rs
+++ b/grpc/src/server/mod.rs
@@ -23,9 +23,9 @@
  */
 
 use std::sync::Arc;
-
 use tonic::async_trait;
 
+use crate::client::CallOptions;
 use crate::core::RecvMessage;
 use crate::core::RequestHeaders;
 use crate::core::ServerResponseStreamItem;
@@ -62,11 +62,12 @@ impl Server {
     pub async fn serve(&self, l: &impl Listener) {
         while let Some(call) = l.accept().await {
             let mut send: Box<dyn DynSendStream> = Box::new(call.send);
-            let recv: Box<dyn DynRecvStream> = Box::new(call.recv);
+            let recv = BoxedRecvStream(Box::new(call.recv));
+            let options = CallOptions::default();
             self.handler
                 .as_ref()
                 .unwrap()
-                .dyn_handle(call.headers, &mut *send, BoxedRecvStream(recv))
+                .dyn_handle(call.headers, options, &mut *send, recv)
                 .await;
         }
     }
@@ -88,6 +89,7 @@ pub trait Handle: Send + Sync {
     async fn handle(
         &self,
         headers: RequestHeaders,
+        options: CallOptions,
         tx: &mut impl SendStream,
         rx: impl RecvStream + 'static,
     );
@@ -98,21 +100,22 @@ trait DynHandle: Send + Sync {
     async fn dyn_handle(
         &self,
         headers: RequestHeaders,
+        options: CallOptions,
         tx: &mut dyn DynSendStream,
         rx: BoxedRecvStream,
     );
 }
+
 #[async_trait]
 impl<T: Handle> DynHandle for T {
     async fn dyn_handle(
         &self,
         headers: RequestHeaders,
+        options: CallOptions,
         mut tx: &mut dyn DynSendStream,
         rx: BoxedRecvStream,
     ) {
-        // Wrap `rx` in BoxedRecvStream here
-        self.handle(headers, &mut tx, BoxedRecvStream(Box::new(rx)))
-            .await
+        self.handle(headers, options, &mut tx, rx).await
     }
 }
 
@@ -131,7 +134,7 @@ struct BoxedRecvStream(Box<dyn DynRecvStream + 'static>);
 
 // Implement RecvStream for the wrapper instead of the Box directly
 impl RecvStream for BoxedRecvStream {
-    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Result<(), ()> {
+    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Option<Result<(), ()>> {
         self.0.dyn_next(msg).await
     }
 }
@@ -141,7 +144,12 @@ impl RecvStream for BoxedRecvStream {
 /// order in which they must be sent.
 #[trait_variant::make(Send)]
 pub trait SendStream {
-    /// Sends the next item on the stream.
+    /// Sends the next item on the stream. Returns `Ok(())` on success, or
+    /// `Err(())` on failure. `Err(())` is a terminal state.
+    /// Sending is also considered complete after successfully sending
+    /// `ServerResponseStreamItem::Trailers`.
+    /// Calling this method after an error or after sending trailers should be
+    /// avoided and is unspecified.
     ///
     /// # Cancel safety
     ///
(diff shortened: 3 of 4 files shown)


grpc/src/core/mod.rs
dfawley, line 116: The only reason this is generic is because the client & server were sharing it.

If they aren't sharing it anymore it should be removed.  The `Message` variant should have no value with it.

Also just rename this to `ClientResponseStreamItem` and delete the alias below, and same for `Server...`.

grpc/src/core/mod.rs
```diff
@@ -129,9 +129,18 @@ pub enum ResponseStreamItem<M> {
 /// void as the received message is passed in via the `next` method.
 pub type ClientResponseStreamItem = ResponseStreamItem<()>;
 
-/// The server's view of a ResponseStream in a SendStream: the message type is
-/// part of the payload provided to the `send` method.
-pub type ServerResponseStreamItem<'a> = ResponseStreamItem<&'a dyn SendMessage>;
+/// The server's view of a ResponseStream in a SendStream, generic over the message type.
+pub enum ServerResponseItem<M> {
```
dfawley: Similarly this should probably be generic on `'a`, and `Message(&'a dyn SendMessage)`, instead of `M`, to limit how it's used.

grpc/src/inmemory/mod.rs
```diff
@@ -195,20 +197,23 @@ pub struct InMemoryServerSendStream {
 impl ServerSendStream for InMemoryServerSendStream {
     async fn send<'a>(
         &mut self,
-        item: crate::core::ServerResponseStreamItem<'a>,
+        item: ServerResponseStreamItem<'a>,
         _options: ServerSendOptions,
-    ) -> Result<(), ()> {
+    ) -> Result<(), Status> {
```
dfawley: I don't particularly like the use of `Status` for any non-RPC-result errors.  They have too much potential to be misunderstood.

If it's actually important to explain why the error occurred to the application directly (why? what will it do differently?), as opposed to via metrics, then should we define a special error type instead?

grpc/src/inmemory/mod.rs
```diff
@@ -217,12 +222,15 @@ pub struct InMemoryServerRecvStream {
 }
 
 impl ServerRecvStream for InMemoryServerRecvStream {
-    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Result<(), ()> {
+    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Result<Option<()>, Status> {
```
dfawley: As above.  Also `Result<Option<()>, ..>`  seems confusing?

grpc/src/inmemory/mod.rs
sauravzg: So, there needs to be a signal for the user to be able to understand an error happened and stop.
Think of a use case where we need to return 1 million rows for a database query response, and the first row itself fails, since the user doesn't receive a signal to it, they'll continue try to send a million more rows.
A status or an error provides a signal to them.

grpc/src/inmemory/mod.rs
sauravzg: Agreed, I couldn't think of a better option(pun not intended).
So, assumptions
-  We need to signal error
-  We need to signal EOS
-  We need to signal success

So, we need a tristate enum. The client handles it in exactly the same way, it returns an enum with various states. 

So, I can either go 
- enum : Downside being that `?` would not work and status will need to be handled explicitly.
- Result<bool, Status> - Folks can ignore bool
- Resut<Option<()>,Status> - Forces people to handle all three states explicitly. So idiot proof.

LMK what you'd like.

grpc/src/inmemory/mod.rs
dfawley: What we had before accomplished that use case already.  `Err(())` meant "you should stop sending now; the stream ended".

If the application needs more information that just "operation didn't work; stop sending more", then we can talk about why it needs that, and how to communicate what it needs.

grpc/src/inmemory/mod.rs
dfawley: So the thing it wasn't differentiating before was "client closed its send stream (normal)" vs "stream died (error)", right?

I wonder if this is important to distinguish.  Because in a "stream died" scenario, the handler's future is also getting dropped by gRPC ASAP.  If this is some other task handling the incoming messages, does it care whether it was case 1 or 2?

If we do need to distinguish, I don't think I like the `Ok()` part of the `Result` here being a boolean, since I think `Ok()` should be reserved for "got a message and wrote it to `msg`" so that you can loop over it cleanly.  The "client closed" situation is more like an error in the context of the receive stream, even if it's not an error in the context of the entire RPC.  I see this as similar to [`mpsc::Receiver::recv()`](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv).

I think any of these could work (if it is even needed to change):

* `Result<(), RecvStreamError>` where `RecvStreamError` has `EOS`/`Error` variants.
* `Option<Result<(), ()>>` -- `None` = `EOS`, `Some(Ok(()))` = got a message, and `Some(Err(()))` = error.
* tri-state enum - `Ok`, `EOS`, `Error` - none of which need a payload.

sauravzg: And what about Result<Option<()>, ()>  ? This is virtually identical to Option<Result<(),()>> , but it prioritizes "safety" check before "presence" check.  OK(Som()) = got message , OK(None) = error , ERR() = error.

The advantage being you get to be able to use `?`.

I think being able to distinguish between error and successful termination might be relevant for the applications to take different decisions based on things, like updating different application metrics.

dfawley: I believe in the context of a stream of values, which this is, `Option<Result>` is preferred.  Here's an example:

https://doc.rust-lang.org/std/io/struct.Lines.html#method.next

This method returns `None` on EOF, or `Some(Err(Error))` on an unexpected error.  That basically makes it match `Iiterator::next` with `Item=Result<>`.  That makes the operation logically: "Is there more information to process?" followed by "Was that a value or an error?".

OTOH for single value lookups, `Result<Option>` seems to be preferred. So logically: "Did the operation error?" followed by "Was the value I requested found?".

`Stream` is just like `Iterator`, btw -- so if we wanted to adapt this to a `Stream`, it would want to be implemented as `Stream::Item=Result<(), ()>`, and that's easier to do if this API is `Option<Result>`.

dfawley: > I wonder if this is important to distinguish. Because in a "stream died" scenario, the handler's future is also getting dropped by gRPC ASAP. If this is some other task handling the incoming messages, does it care whether it was case 1 or 2?

I feel we should discuss on this point more, first.  Is there a need to distinguish between "client stopped sending" and "stream died" when we have "your handler's future was dropped" to indicate the latter already.  I assumed the latter when I wrote this, but if you have counter examples that show where this is insufficient, I'm totally fine with changing this.

grpc/src/core/mod.rs
sauravzg: I've removed the server side generic. I'll tacket the realiasing of `ClientResponseStreamItem` in a differnt PR to keep this PR focused and isolated. Changing this  broke some tests that dependended on `M()`.

grpc/src/core/mod.rs
sauravzg: Done.

grpc/src/inmemory/mod.rs
sauravzg: I could swear I replied yesterday about it, but maybe I didn't press "comment" :(. 

Anyways, while there may not be a value in understanding the distinction between "client stopped sending" and "http stream died", there is value in distinguishing "client stopped sending" vs "error in grpc processing or interceptors".

For http stream died, the task running the handler task gets cancelled and that's fine. But , we could have sources of error like "serialization failed", "interceptor returned an error". These errors will be in the recv stream layer, and since we are not using push streams(or anything that ties production and consumption together) they won't autotrigger short circuit and stream cancellation. 

This implies grpc will need a layer that's able to detect abnormal stream failures on grpc side and then take action upon it to close the http stream.

grpc/src/inmemory/mod.rs
sauravzg: This should now be fixed. (unless I forgot uploading)

grpc/src/inmemory/mod.rs
sauravzg: I believe on the client side it currently happens 
- sendstream: explicit end of stream signal in the "sendoptions"
- recvstream: Trailers + StreamClosed combination .

dfawley: > But , we could have sources of error like "serialization failed", "interceptor returned an error". These errors will be in the recv stream layer, and since we are not using push streams(or anything that ties production and consumption together) they won't autotrigger short circuit and stream cancellation.

I don't know about this.  I don't think we want to rely on the application to forward errors from receive to send in order for us to have the proper stream behavior.  When there is an error in the stream, whether it happens on the wire or via our local processing, that the whole stream should be cancelled.  That means that if an interceptor detects this situation (e.g. it's deserializing a request and gets an error), the interceptor should drop its delegate's future, push an error to the SendStream that describes the failure, and return.

---

> Is there a need to distinguish between "client stopped sending" and "stream died" when we have "your handler's future was dropped" to indicate the latter already.

We discussed this a bunch in the office yesterday, and I think I'm convinced that it is important.  Mainly because the `RecvStream` can be passed to another task, meaning it could easily outlive the main handler's task, and it would not be trivial for it to discover whether the main handler task's future had been dropped already.  More importantly, we would also need to ensure the timing happens from our end properly, which seems hard, and even harder to expect interceptors to follow.

(Eric had another idea which is that we could simply hang this function in the event of a stream error.  I suspect this is a bad idea -- it means the application _must_ manually abort the task doing the receives to avoid leaking the task -- but it at least prevents any ordering issues.)

sauravzg: > When there is an error in the stream, whether it happens on the wire or via our local processing, that the whole stream should be cancelled. That means that if an interceptor detects this situation (e.g. it's deserializing a request and gets an error), the interceptor should drop its delegate's future, push an error to the SendStream that describes the failure, and return.


Assuming this is what the interceptor does, the interceptor doesn't ever know when the error happens it's only the recvstream wrapper that does.
```
async fn intercept(recvstream, sendstream, next){
   return next.call(wrap_recv(recvstream), wrap_send(sendstream), next);
}
```

So, the let's say the error happens in `wrap_recv`'s implementation of `next`  and it needs to send then send the error to sendstream would require all recvstream interceptors to hold a reference to sendstream which is a reference and no recvstream would ever be static and become painful to deal with. We could use channel as workarounds  maybe to avoid the reference shenanigans , but channels aren't zero cost. Again, a single channel might not be an issue but requiring all interceptors to have to end up doing it isn't reasonable.

I think we are much better  in a situation where we allow interceptors to propagate error and then the final layer that converts generic API to proto API is responsible for getting the error status from the interceptors via the recvstream objects.  So, the error from interceptors bubbles up and grpc is able to appropriately terminate the stream  by cancelling the handler future and shortcircuiting using something like a `race` on either recvstream error or the handler success. 

Something like this is able to intercept the errors and finally break the chain
```
async fn proto_handler(recvstream, sendstream){
    (tx , rx) = oneshot()
    concrete_stream = wrap_recvstream(recvstream,rx);
    race( handler(concrete_stream, concretesend_stream) , rx);
}
```
This roughly ensures the application handler is always dropped on stream failure.

sauravzg: > We discussed this a bunch in the office yesterday, and I think I'm convinced that it is important.

So I assume we are okay with Option<Result<(),()> ?

dfawley: I think something like you wrote in `proto_handler` should be pretty reasonable to expect of any interceptor that wants to be able to create a stream failure.  There is one sort-of-questionable part of it in terms of: is it OK for it to write a Trailers with an error in it if the delegate already wrote another Trailers frame?  If so, then it's trivial.  If not, then it needs to wrap the stream it gives to the delegate so it can know whether the delegate wrote Trailers.

> So I assume we are okay with Option<Result<(),()> ?

I _think_ this is my current preference since it lines up with Iterator/Stream best.

sauravzg: > I think this is my current preference since it lines up with Iterator/Stream best.


So, this should wrap up questions around the PR I assume.

We can take the other conversation offline, but right now I am probably missing the benefit of having each interceptor being forced to follow a pattern give us, if we get similar results be letting the interceptor do its own thing and propagate errors, while the final `proto_handler` takes care of state management complexity.

Regarding trailers: I think the interface currently says that writing trailers twice is an error and the first one wins. So, yes we may need some wrapping here as well, which is probably again why the make `proto_handler` do all complex stuff sounds fine to me.  Since, the application handler cannot deal with trailers at all, we don't deal with the problem of two trailer writes, the `proto_handler` is the sole writer and the interceptors delegates are suposed to intercept and modify.

grpc/src/server/mod.rs
```diff
@@ -141,7 +144,9 @@ impl RecvStream for BoxedRecvStream {
 /// order in which they must be sent.
 #[trait_variant::make(Send)]
 pub trait SendStream {
-    /// Sends the next item on the stream.
+    /// Sends the next item on the stream. Returns `Ok(())` on success, or `Err(())` on failure. `Err(())` is a terminal state.
+    /// Sending is also considered complete after successfully sending `ServerResponseStreamItem::Trailers`.
+    /// Calling this method after an error or after sending trailers should be avoided and is unspecified.
```
dfawley, line 149: Please apply 80-column line wrapping on columns throughout.

grpc/src/server/mod.rs
```diff
@@ -209,31 +214,31 @@ pub struct SendOptions {
 /// Represents the receiving side of a server stream.
 #[trait_variant::make(Send)]
 pub trait RecvStream {
-    /// Returns the next message on the stream.  If an error is returned, the
-    /// stream ended or the client closed the send side of the request stream.
+    /// Returns the next message on the stream. Returns `Some(Ok(()))` on success, `None` on normal stream end, or `Some(Err(()))`. Both `None` and `Some(Err(()))` are terminal states.
```
dfawley, line 217: Nit:

"....or `Some(Err(()))` if the stream encountered an error before the client's final request message"

(or something like that.

dfawley (approved):

grpc/src/inmemory/mod.rs
dfawley: Yeah I think we're good now - I just have a couple small nits.

> We can take the other conversation offline, but right now I am probably missing the benefit of having each interceptor being forced to follow a pattern give us

If nothing else, not returning a `Status` here (which I'm pretty vehement about not doing) means we don't even know what kind of error to propagate back to the send side of the stream.  That makes it a non-starter in the first place.

> Regarding trailers: I think the interface currently says that writing trailers twice is an error and the first one wins. So, yes we may need some wrapping here as well, which is probably again why the make proto_handler do all complex stuff sounds fine to me.

It's our API, though, so we can just say "the first one wins" which is exactly what we would expect, and make it so that anything sent after Trailers is silently ignored.  That would simplify the structure of interceptors that need to do this kind of thing, which would be a good reason to change it.

When gRPC receives the trailers from the application, it's going to write them ~immediately (unless the stream is blocked on sending data), and anything sent later would be too late to do anything about besides panic, anyway, so ignoring seems fine.

sauravzg: I'll take this conversation offline and proceed to submit this PR.

PR #2592, opened 2026-04-14
transport: add max_frame_size to client Endpoint

timlee0119:
## Motivation

tonic already exposes `Server::max_frame_size()` to let servers configure this parameter, but the
client-side `Endpoint` struct has no equivalent API. The underlying hyper HTTP/2 `Builder` already
supports `.max_frame_size()`, so this change simply wires it through from the client-side endpoint
configuration.

I've seen production speedup in our gRPC service by increasing this client-side `max_frame_size` to allow the server to send a large response (multi-MB) without excessive framing overheads caused by the default 16 KiB setting.

## Solution

Add a new `max_frame_size` field to the `Endpoint` struct and expose it via a public builder method,
mirroring the server-side pattern:

- Add `max_frame_size: Option<u32>` field to `Endpoint`
- Add public `pub fn max_frame_size(self, frame_size: impl Into<Option<u32>>) -> Self` builder method
- Wire the field through to hyper's `Builder::max_frame_size()` in `connection.rs`
- Include integration test verifying the feature works end-to-end

Refs: #264

### DIFF
--- a/tests/integration_tests/tests/max_frame_size.rs
+++ b/tests/integration_tests/tests/max_frame_size.rs
@@ -1,0 +1,52 @@
+use tokio::net::TcpListener;
+use tokio::sync::oneshot;
+
+use integration_tests::pb::{test_client::TestClient, test_server, Input, Output};
+use tonic::transport::{server::TcpIncoming, Channel, Server};
+use tonic::{Request, Response, Status};
+
+struct Svc;
+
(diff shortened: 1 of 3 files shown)


timlee0119:
@dfawley gentle ping on this tiny PR. Thanks!

LucioFranco (approved):

PR #2593, opened 2026-04-15
feat(tonic-xds): gRFC A29 TLS foundation - cert provider & control plane TLS

YutaoMa:
## Motivation

Ref: #2444

PR 1 of 3 to implement [gRFC A29 xDS TLS security](https://github.com/grpc/proposal/blob/master/A29-xds-tls-security.md) support in `tonic-xds`.

## Solution

1. Add `CertificateProvider` trait and outline `file_watcher` plugin with its config shape. Background file watch refresher is not implemented yet.
2. Feature flags gating setup in `xds-client` and `tonic-xds`. Allow optional use of one of the crypto backends from `tonic`.
3. Extend `TonicTransportBuilder` with `with_tls_config()` to allow subsequent TLS changes.
4. Parse bootstrap `certificate_providers` and enable TLS in xDS connection to xDS management server conditionally. Currently because the file watcher is not fully implemented, the setup only enables TLS with system root CA only, no custom certs yet.

## Next Steps

Two more PRs are in plan to complete A29:

1. PR 2: data-plane endpoint connection TLS support, by supporting CDS `UpstreamTlsContext` and SAN matching.
2. PR 3: Background cert refresh to complete `file_watcher`.

### DIFF
--- a/tonic-xds/src/xds/cert_provider/file_watcher.rs
+++ b/tonic-xds/src/xds/cert_provider/file_watcher.rs
@@ -1,0 +1,383 @@
+//! `file_watcher` certificate provider plugin.
+//!
+//! Reads PEM-encoded certificates and keys from local files. This is the
+//! only built-in certificate provider plugin per gRFC A29.
+//!
+//! # Bootstrap configuration
+//!
+//! ```json
+//! {
+//!   "plugin_name": "file_watcher",
+//!   "config": {
+//!     "certificate_file": "/path/to/cert.pem",
+//!     "private_key_file": "/path/to/key.pem",
+//!     "ca_certificate_file": "/path/to/ca.pem",
+//!     "refresh_interval": "60s"
+//!   }
+//! }
+//! ```
+
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+use std::time::Duration;
+
+use arc_swap::ArcSwap;
+use serde::Deserialize;
+
+use super::{CertProviderError, CertificateData, CertificateProvider};
+
+/// Plugin name used in the bootstrap `certificate_providers` JSON.
+pub(crate) const PLUGIN_NAME: &str = "file_watcher";
+
+/// Configuration for the `file_watcher` certificate provider.
+#[derive(Debug, Clone, Default, Deserialize)]
+pub(crate) struct FileWatcherConfig {
+    /// Path to PEM X.509 identity certificate or certificate chain.
+    #[serde(default)]
+    pub certificate_file: Option<PathBuf>,
+    /// Path to PEM PKCS private key.
+    #[serde(default)]
+    pub private_key_file: Option<PathBuf>,
+    /// Path to PEM X.509 CA trust bundle (root certificates).
+    #[serde(default)]
+    pub ca_certificate_file: Option<PathBuf>,
+    /// How often to re-read the files. Default: 600s.
+    /// Parsed from protobuf JSON duration format (e.g., `"60s"`, `"0.5s"`).
+    #[serde(default, deserialize_with = "deserialize_proto_duration")]
+    pub refresh_interval: Option<Duration>,
+}
+
+/// Deserialize a protobuf JSON duration string (e.g., `"60s"`, `"0.5s"`) into a `Duration`.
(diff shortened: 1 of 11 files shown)


tonic-xds/src/xds/cert_provider/file_watcher.rs
```diff
@@ -0,0 +1,383 @@
...
+
+/// Deserialize a protobuf JSON duration string (e.g., `"60s"`, `"0.5s"`) into a `Duration`.
+fn deserialize_proto_duration<'de, D>(deserializer: D) -> Result<Option<Duration>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let Some(s) = Option::<String>::deserialize(deserializer)? else {
+        return Ok(None);
+    };
+    let num = s.strip_suffix('s').ok_or_else(|| {
```
gu0keno0, line 58: A29 specifies google.protobuf.Duration, which should support more time formats

gu0keno0: NVM, it is indeed the case

tonic-xds/src/xds/cert_provider/file_watcher.rs
```diff
@@ -0,0 +1,383 @@
...
+        .ca_certificate_file
+        .as_deref()
+        .map(read_file)
+        .transpose()?;
+
+    let identity_cert_chain = config
+        .certificate_file
+        .as_deref()
+        .map(read_file)
+        .transpose()?;
```
gu0keno0, line 127: PTAL https://github.com/grpc/proposal/blob/master/A65-xds-mtls-creds-in-bootstrap.md#proposal
TLDR is that these values can be optional depending on 1. whether mTLS is requires (versus TLS), 2. whether we want to default to system ca root certs

gu0keno0: It should be fine to require them for now as MVP

tonic-xds/src/xds/cert_provider/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use serde::Deserialize;
+
+use crate::xds::bootstrap::CertProviderPluginConfig;
+
+/// Certificate material returned by a [`CertificateProvider`] plugin.
+#[derive(Debug, Clone)]
+pub(crate) struct CertificateData {
```
gu0keno0, line 24: This probably should be a enum depending on if MTLS is being configured.

tonic-xds/src/xds/cert_provider/file_watcher.rs
```diff
@@ -0,0 +1,383 @@
...
+        let data = provider.fetch().unwrap();
+
+        assert_eq!(data.root_certs.as_deref(), Some(b"ca-pem".as_slice()));
+        assert_eq!(
+            data.identity_cert_chain.as_deref(),
+            Some(b"cert-pem".as_slice())
+        );
+        assert_eq!(data.identity_key.as_deref(), Some(b"key-pem".as_slice()));
+    }
+
```
gu0keno0, line 234: Once we implement a65, we should have more test cases to verify the combinations of ca, cert and key.

tonic-xds/src/xds/cert_provider/file_watcher.rs
YutaoMa: Updated to use an enum design to enforce the following requirements from A29 + A65:
1. cert and key are present or missing at the same time
2. at least one of ca cert and identity cert needs to be present

tonic-xds/src/xds/cert_provider/mod.rs
YutaoMa: see above comment, introduced enum as suggested.

tonic-xds/src/xds/cert_provider/file_watcher.rs
YutaoMa: makes sense, also added more test along with the new enum approach. And given now the enum enforces invariants at the type level, the only validation boundary left is the JSON config -> cert data layer which I added some test for too.

gu0keno0 (approved):

PR #2594, opened 2026-04-16
feat(grpc): Unix transport

arjan-bal:
This PR introduces the following changes:
* Runtime Trait: Adds a new method to enable connecting to Unix sockets, along with the corresponding `tokio` runtime implementation.
* Tonic Transport: Introduces support for connecting to Unix sockets.
* Bugfix (Tonic Channel Creation): Updates the Tonic transport to send the authority rather than the address when creating a channel. This prevents errors since Unix addresses are not always valid authorities. Note that the Unix resolver overrides the authority to `localhost`.
* Add support for unix addresses in LocalCredentials.

### DIFF
--- a/grpc/src/client/name_resolution/dns/test.rs
+++ b/grpc/src/client/name_resolution/dns/test.rs
@@ -51,6 +51,7 @@ use crate::client::name_resolution::global_registry;
 use crate::client::service_config::ServiceConfig;
 use crate::rt::BoxFuture;
 use crate::rt::GrpcRuntime;
+use crate::rt::Runtime;
 use crate::rt::TcpOptions;
 use crate::rt::tokio::TokioRuntime;
 use crate::rt::{self};
@@ -288,7 +289,7 @@ struct FakeRuntime {
     dns: FakeDns,
 }
 
-impl rt::Runtime for FakeRuntime {
+impl Runtime for FakeRuntime {
     fn spawn(
         &self,
         task: Pin<Box<dyn Future<Output = ()> + Send + 'static>>,
--- a/grpc/src/client/name_resolution/dns/test.rs
+++ b/grpc/src/client/name_resolution/dns/test.rs
@@ -24,19 +24,13 @@
 
 use std::future::Future;
 use std::pin::Pin;
-use std::sync::Arc;
 use std::time::Duration;
 
-use tokio::sync::mpsc::UnboundedSender;
-use tokio::sync::mpsc::{self};
 use url::Host;
 
-use crate::client::name_resolution::ChannelController;
 use crate::client::name_resolution::Resolver;
 use crate::client::name_resolution::ResolverOptions;
-use crate::client::name_resolution::ResolverUpdate;
 use crate::client::name_resolution::Target;
-use crate::client::name_resolution::WorkScheduler;
 use crate::client::name_resolution::backoff::BackoffConfig;
 use crate::client::name_resolution::backoff::DEFAULT_EXPONENTIAL_CONFIG;
 use crate::client::name_resolution::dns::DnsOptions;
@@ -48,13 +42,12 @@ use crate::client::name_resolution::dns::get_resolving_timeout;
 use crate::client::name_resolution::dns::parse_endpoint_and_authority;
 use crate::client::name_resolution::dns::reg;
 use crate::client::name_resolution::global_registry;
-use crate::client::service_config::ServiceConfig;
-use crate::rt::BoxFuture;
+use crate::client::name_resolution::test_utils::TestChannelController;
+use crate::client::name_resolution::test_utils::TestWorkScheduler;
+use crate::rt;
 use crate::rt::GrpcRuntime;
 use crate::rt::Runtime;
-use crate::rt::TcpOptions;
 use crate::rt::tokio::TokioRuntime;
-use crate::rt::{self};
 
 const DEFAULT_TEST_SHORT_TIMEOUT: Duration = Duration::from_millis(10);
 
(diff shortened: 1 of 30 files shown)


grpc/src/credentials/mod.rs
```diff
@@ -153,6 +153,18 @@ pub(crate) mod common {
         pub fn port(&self) -> Option<u16> {
             self.port
         }
+
+        pub fn host_port_string(&self) -> String {
```
sauravzg, line 157: unit test?

grpc/src/credentials/local.rs
```diff
@@ -57,10 +58,15 @@ pub struct LocalChannelCredentials {
...
 
 impl LocalChannelCredentials {
-    /// Creates a new instance of `InsecureChannelCredentials`.
+    /// Creates a new instance of `LocalChannelCredentials`.
     pub fn new() -> Self {
         Self { _private: () }
     }
+
+    /// Creates a new ref-counted instance of `LocalChannelCredentials`.
+    pub fn new_arc() -> Arc<Self> {
```
sauravzg, line 67: Why do we need this function? How is it different from our user calling `Arc::new(LocalChannelCredentials::new())` themselves?

grpc/src/rt/tokio/mod.rs
```diff
@@ -126,6 +129,33 @@ impl Runtime for TokioRuntime {
             Ok(stream)
         })
     }
+
+    #[cfg(unix)]
+    fn unix_stream(
```
sauravzg, line 134: Can we unit test this ?  No is a very valid answer , given that its tested indirectly via tonic tests.

But maybe, we should add a small unit test that  creates a unix stream and verify if arbitrary data is being sent correctly.

grpc/src/client/transport/tonic/test.rs
```diff
@@ -265,6 +261,134 @@ async fn grpc_invoke_tonic_unary() {
...
+        let target = format!("unix-abstract:{}", abstract_path);
+
+        run_unix_test(&PathBuf::from(bind_path), &target).await;
+    }
+
+    // Calculates the relative path from a `base` directory to a `target` path.
+    ///
+    /// Both paths should be absolute. This operation is infallible on Unix
+    /// systems due to the presence of a single root directory.
+    fn get_relative_path(target: &Path, base: &Path) -> Result<PathBuf, String> {
```
sauravzg, line 349: How's it different from https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix ?

sauravzg (approved):
lgtm. Please address the comments.

grpc/src/client/transport/tonic/test.rs
arjan-bal: `strip_prefix` only handles cases where `base` is a prefix of `target`. The temp directories are created in `/tmp` while the current directory is usually not in `/tmp`. The relative path takes the form `../.../../tmp/random/grpc-test-abstract-socket-relative.sock`. This function handles reaching a common ancestor and then going down until the target file.

grpc/src/credentials/local.rs
arjan-bal: It’s the same; it just means callers don’t have to call `Arc::new()`. Since there’s no configuration to pass and the channel accepts an `Arc<ChannelCredentials>`, all callers were ending up calling `Arc::new(LocalChannelCredentials::new())` anyway.

grpc/src/rt/tokio/mod.rs
arjan-bal: The tonic transport tests provide coverage for this code path. Given that the implementation primarily interacts with tonic with minimal custom logic, I didn't include additional dedicated tests.

grpc/src/credentials/mod.rs
arjan-bal: The logic for creating the call credentials authority was very similar, except it omitted port 443. To avoid duplication, I refactored it to use authority.host_port_string() for non-443 ports. I've also added tests here.

grpc/src/rt/tokio/mod.rs
sauravzg: Fair enough.

grpc/src/client/transport/tonic/test.rs
sauravzg: Got it. Pretty annoying, but I got no better ideas here.

grpc/src/client/name_resolution/dns/test.rs
```diff
@@ -46,6 +46,7 @@ use crate::client::name_resolution::test_utils::TestChannelController;
 use crate::client::name_resolution::test_utils::TestWorkScheduler;
 use crate::rt;
```
dfawley, line 47: Delete this `use`?  Or are there other things like what you converted?  If so, change them, too?

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -332,39 +354,47 @@ impl Transport for TransportBuilder {
...
+        };
+        let transport = if let Some(deadline) = opts.connect_deadline {
             let timeout = deadline.saturating_duration_since(Instant::now());
             tokio::select! {
-            _ = runtime.sleep(timeout) => {
-                return Err("timed out waiting for TCP stream to connect".to_string())
-            }
-            tcp_stream = tcp_stream_fut => { tcp_stream? }
+                _ = runtime.sleep(timeout) => {
+                    return Err("timed out waiting for TCP stream to connect".to_string());
```
dfawley, line 377: I guess it's not necessarily a TCP stream anymore, right?  Probably we wouldn't expect a unix socket to timeout, but I'm not actually sure - maybe it's possible.

grpc/src/client/transport/tonic/test.rs
```diff
@@ -265,6 +261,134 @@ async fn grpc_invoke_tonic_unary() {
...
+    #[tokio::test]
+    async fn unix_abstract_socket() {
+        let abstract_path = format!("grpc-test-abstract-socket-{}", rand::random::<u64>());
+        let bind_path = format!("\0{}", abstract_path);
+        let target = format!("unix-abstract:{}", abstract_path);
+
+        run_unix_test(&PathBuf::from(bind_path), &target).await;
+    }
+
+    // Calculates the relative path from a `base` directory to a `target` path.
```
dfawley, line 345:
```suggestion
    /// Calculates the relative path from a `base` directory to a `target` path.
```

grpc/src/credentials/local.rs
```diff
@@ -89,7 +95,9 @@ fn security_level_for_endpoint(
     {
         return Ok(SecurityLevel::NoSecurity);
     }
-    // TODO: Add support for unix sockets.
+    if network_type == UNIX_NETWORK_TYPE {
```
dfawley, line 98: Did we get an answer on whether we need to consider abstract sockets as `NoSecurity`?

If not I think it would be best to err on that side for now until we find out otherwise.

grpc/src/client/name_resolution/dns/test.rs
arjan-bal: I updated several more occurrences. The remaining instances use the `rt` crate name because the `dns` crate contains symbols with the exact same name.

grpc/src/client/transport/tonic/mod.rs
arjan-bal: Changed the message to say "transport stream" instead of "TCP stream".

grpc/src/credentials/local.rs
arjan-bal: I've set the security level for abstract sockets to `NoSecurity` and added a TODO to increase it in the future. Gregory mentioned that while abstract sockets can be authorized using peer information via `SO_PEERCRED`, gRPC doesn't utilize this information across all languages yet. I've asked him to confirm if `NoSecurity` is the right approach for Rust.

arjan-bal: Gregory confirmed that Rust should use `NoSecurity` for abstract sockets.

grpc/src/client/name_resolution/dns/test.rs
dfawley: BTW I'm fine with keeping `rt::____` wherever you think it helps disambiguate things.  There definitely seems to be an art to these things.

dfawley (approved):
Sorry I somehow lost track of this one. LGTM, thanks!

PR #2595, opened 2026-04-17
feat(grpc): Add server-side stream validators

sauravzg:
Introduces `ServerSendStreamValidator` and `ServerRecvStreamValidator` to enforce strict gRPC semantics on server streams. These are intended to wrap raw transport streams before passing them to application-level handlers, ensuring that user-provided service logic cannot violate protocol rules.

- `ServerSendStreamValidator` ensures proper response sequencing (Headers -> Messages -> Trailers) and prevents invalid state transitions, while fully supporting trailers-only responses.
- `ServerRecvStreamValidator` safely manages terminal states, ensuring that any subsequent polls after stream completion consistently return an error to prevent undefined behavior.
- Adds comprehensive unit tests to verify state machine correctness and protocol compliance.

### DIFF
--- a/grpc/examples/inmemory.rs
+++ b/grpc/examples/inmemory.rs
@@ -41,6 +41,7 @@ use grpc::credentials::InsecureChannelCredentials;
 use grpc::inmemory;
 use grpc::server;
 use grpc::server::Handle;
+use grpc::core::Trailers;
 
 struct Handler {
     id: String,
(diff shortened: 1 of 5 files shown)


dfawley:
As discussed offline, we likely should move `Trailers` to be a return value from the handler instead of being an item on the stream, which would impact this PR (likely making it simpler since there's less ordering to enforce).

sauravzg:
Back from OOO. ETA for addressing comment: 1d , since I need to catch up to a few things first.

sauravzg:
I'll close this PR and create a new one, since it's painful to do stacked PR workflow from fork.

PR #2596, opened 2026-04-19
tonic-reflection: fix panic when client drops connection early

bmwill:
## Motivation

This patch fixes a panic due to an `expect` call when sending on a channel. It is possible for the client to close half of the bidi streaming request early leading to the response stream being closed and dropped before the request stream is terminated.

## Solution

This patch addresses this by not panicking if the recieving side of the channel has been close and instead gracefully ends the task.


seanmonstar:
Should it log the error somehow? If not that's fine, I'm less familiar with the workings of the reflection crate.

seanmonstar (approved):

PR #2601, opened 2026-04-23
feat(grpc): Remove trailers from `SendStream` and make Handle return …

sauravzg:
- Server sendstream doesn't write trailers anymore
- Handle returns Trailer
- Compatibility
   - current server's `Call` had to be updated to provide a method to write trailers
   - in memory transport had to get a few major changes including how client receives the data, requiring oneshots to handle receiving trailer from in memory server.

### DIFF
--- a/grpc/src/core/mod.rs
+++ b/grpc/src/core/mod.rs
@@ -101,8 +101,7 @@ impl dyn RecvMessage + '_ {
     }
 }
 
-/// ResponseStreamItem represents an item in a response stream (either server
-/// sending or client receiving).
+/// ResponseStreamItem represents an item in a response stream from the client's view.
 ///
 /// A response stream must always contain items exactly as follows:
 ///
@@ -135,8 +134,6 @@ pub enum ServerResponseStreamItem<'a> {
     Headers(ResponseHeaders),
     /// Indicates a message on the stream.
     Message(&'a dyn SendMessage),
-    /// Indicates trailers were received on the stream and includes the trailers.
-    Trailers(Trailers),
 }
 
 /// Contains all information transmitted in the response headers of an RPC.
--- a/grpc/src/inmemory/mod.rs
+++ b/grpc/src/inmemory/mod.rs
@@ -35,6 +35,8 @@ use tokio::sync::Notify;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 
+use crate::Status;
+use crate::StatusCode;
 use crate::attributes::Attributes;
 use crate::client::CallOptions;
 use crate::client::DynRecvStream as ClientDynRecvStream;
@@ -85,6 +87,7 @@ struct InMemoryServerCall {
     headers: RequestHeaders,
     req_rx: mpsc::UnboundedReceiver<InMemoryRequestStreamItem>,
     resp_tx: mpsc::UnboundedSender<InMemoryResponseStreamItem>,
+    trailer_tx: oneshot::Sender<Trailers>,
 }
 
 enum InMemoryRequestStreamItem {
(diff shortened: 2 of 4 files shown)


grpc/src/server/mod.rs
```diff
@@ -146,10 +152,9 @@ impl RecvStream for BoxedRecvStream {
 pub trait SendStream {
     /// Sends the next item on the stream. Returns `Ok(())` on success, or
     /// `Err(())` on failure. `Err(())` is a terminal state.
-    /// Sending is also considered complete after successfully sending
-    /// `ServerResponseStreamItem::Trailers`.
-    /// Calling this method after an error or after sending trailers should be
-    /// avoided and is unspecified.
+    /// Sending is considered complete when the `handle` method returns
+    /// `Trailers`.
```
dfawley: I wouldn't mention stuff about `handle` in here.  The `SendStream` works until it's dropped, really.

One thing we should probably mention instead is that if `SendOptions.final_msg` is set, then no subsequent calls should be made.

grpc/src/server/mod.rs
```diff
@@ -38,6 +40,7 @@ pub struct Call<SS, RS> {
     pub headers: RequestHeaders,
     pub send: SS,
     pub recv: RS,
+    pub trailers_tx: oneshot::Sender<Trailers>,
```
dfawley, line 43: I'm not really sure how I feel about this `Listener` API now, because of this.  I wonder if there's a nicer design here.  But, we can debate about it later.

grpc/src/inmemory/mod.rs
```diff
@@ -309,8 +315,14 @@ impl ClientRecvStream for InMemoryClientRecvStream {
                 msg.decode(&mut buf).unwrap();
                 ClientResponseStreamItem::Message(())
             }
-            Some(InMemoryResponseStreamItem::Trailers(t)) => ClientResponseStreamItem::Trailers(t),
-            _ => ClientResponseStreamItem::StreamClosed,
+            _ => {
+                if let Some(trailer_rx) = self.trailer_rx.take() {
+                    if let Ok(trailers) = trailer_rx.await {
+                        return ClientResponseStreamItem::Trailers(trailers);
+                    }
```
dfawley: Should this just `.expect` instead or something?  It's a bug if no trailers were sent on the channel and the stream is done, isn't it?

grpc/src/inmemory/mod.rs
```diff
@@ -95,7 +96,6 @@ enum InMemoryRequestStreamItem {
 enum InMemoryResponseStreamItem {
     Headers(ResponseHeaders),
     Message(Box<dyn Buf + Send + Sync>),
-    Trailers(Trailers),
     StreamClosed,
```
dfawley: Does this one not also go away?

dfawley (changes requested):

grpc/src/server/mod.rs
sauravzg: I'll close it off here. I'll think about it when we get to it later.

grpc/src/server/mod.rs
sauravzg: Fixed.

grpc/src/inmemory/mod.rs
sauravzg: I chose to return trailer with error, along the same lines as what client side recvstream validators do on incorrect state transitions.

panic fro expect seemed too extreme. LMK if this is fine.

grpc/src/core/mod.rs
```diff
@@ -101,8 +101,7 @@ impl dyn RecvMessage + '_ {
     }
 }
 
-/// ResponseStreamItem represents an item in a response stream (either server
-/// sending or client receiving).
+/// ResponseStreamItem represents an item in a response stream from the client's view.
```
dfawley, line 104: We still need to fix this type, too.  It should be renamed to `Client...`, generic removed, and delete the alias.

Next PR?

dfawley (approved):

grpc/src/core/mod.rs
sauravzg: Yes. I tried changing this, but it broke many other files  and I wanted to keep this focussed change focused and small, so will send a separate PR for this.

PR #2602, opened 2026-04-23
feat(grpc): add server-side stream validators

sauravzg:
Introduces `ServerSendStreamValidator` and `ServerRecvStreamValidator` to enforce strict gRPC semantics on server streams. These are intended to wrap raw transport streams before passing them to application-level handlers, ensuring that user-provided service logic cannot violate protocol rules.

- `ServerSendStreamValidator` ensures proper response sequencing (Headers -> Messages -> Trailers) and prevents invalid state transitions, while fully supporting trailers-only responses.
- `ServerRecvStreamValidator` safely manages terminal states, ensuring that any subsequent polls after stream completion consistently return an error to prevent undefined behavior.
- Adds comprehensive unit tests to verify state machine correctness and protocol compliance.

This is recreation of https://github.com/hyperium/tonic/pull/2595 to allow using a stacked PR workflow

### DIFF
--- a/grpc/src/server/handler_validation.rs
+++ b/grpc/src/server/handler_validation.rs
@@ -1,0 +1,598 @@
+/*
+ *
+ * Copyright 2026 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use crate::client::CallOptions;
+use crate::core::{RecvMessage, RequestHeaders, ServerResponseStreamItem, Trailers};
+use crate::server::interceptor::Intercept;
+use crate::server::{Handle, RecvStream, SendOptions, SendStream};
+use crate::{StatusCodeError, StatusError};
+use tokio::sync::mpsc::channel;
+
+struct ServerSendStreamValidator<S> {
+    inner: S,
+    state: SendStreamState,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SendStreamState {
+    Init,
+    HeadersSent,
+    MessagesSent,
+    Done,
+}
+
+impl<S: SendStream> ServerSendStreamValidator<S> {
+    fn new(inner: S) -> Self {
+        Self {
+            inner,
+            state: SendStreamState::Init,
+        }
+    }
+}
+
+impl<S: SendStream> SendStream for ServerSendStreamValidator<S> {
+    async fn send<'a>(
+        &mut self,
+        item: ServerResponseStreamItem<'a>,
+        options: SendOptions,
+    ) -> Result<(), ()> {
+        if self.state == SendStreamState::Done {
+            // Protocol error: Attempted to send an item on a completed or failed stream.
+            return Err(());
+        }
+
+        let next_state = match &item {
+            ServerResponseStreamItem::Headers(_) => match self.state {
+                SendStreamState::Init => SendStreamState::HeadersSent,
+                _ => {
+                    // Protocol error: Received multiple headers frames.
+                    self.state = SendStreamState::Done;
+                    return Err(());
+                }
+            },
+            ServerResponseStreamItem::Message(_) => match self.state {
+                SendStreamState::HeadersSent | SendStreamState::MessagesSent => {
+                    SendStreamState::MessagesSent
+                }
+                _ => {
+                    // Protocol error: Attempted to send a message before headers.
+                    self.state = SendStreamState::Done;
+                    return Err(());
+                }
+            },
+        };
+
+        let res = self.inner.send(item, options).await;
+        match res {
+            Ok(()) => self.state = next_state,
+            Err(_) => {
+                self.state = SendStreamState::Done;
+            }
+        }
+        res
+    }
+}
+
+struct ServerRecvStreamValidator<R> {
+    inner: R,
+    done: bool,
+}
+
+impl<R: RecvStream> ServerRecvStreamValidator<R> {
+    fn new(inner: R) -> Self {
+        Self { inner, done: false }
+    }
+}
+
+impl<R: RecvStream> RecvStream for ServerRecvStreamValidator<R> {
+    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Option<Result<(), ()>> {
+        if self.done {
+            // Protocol error: Attempted to receive a message after reaching a terminal state (EOF or error).
+            return Some(Err(()));
+        }
+
+        let res = self.inner.next(msg).await;
+        match res {
+            Some(Ok(())) => Some(Ok(())),
+            None => {
+                self.done = true;
+                None
+            }
+            Some(Err(())) => {
+                self.done = true;
+                Some(Err(()))
+            }
+        }
+    }
+}
+
+struct ChannelAwareSendStreamValidator<S> {
+    inner: ServerSendStreamValidator<S>,
+    error_tx: tokio::sync::mpsc::Sender<()>,
+}
+
+impl<S: SendStream> ChannelAwareSendStreamValidator<S> {
+    fn new(inner: S, error_tx: tokio::sync::mpsc::Sender<()>) -> Self {
+        Self {
+            inner: ServerSendStreamValidator::new(inner),
+            error_tx,
+        }
+    }
+
+    fn report_error(&self) {
+        let _ = self.error_tx.try_send(());
+    }
+}
+
+impl<S: SendStream> SendStream for ChannelAwareSendStreamValidator<S> {
+    async fn send<'a>(
+        &mut self,
+        item: ServerResponseStreamItem<'a>,
+        options: SendOptions,
+    ) -> Result<(), ()> {
+        let res = self.inner.send(item, options).await;
+        if res.is_err() {
+            self.report_error();
+        }
+        res
+    }
+}
+
+struct ChannelAwareRecvStreamValidator<R> {
+    inner: ServerRecvStreamValidator<R>,
+    error_tx: tokio::sync::mpsc::Sender<()>,
+}
+
+impl<R: RecvStream> ChannelAwareRecvStreamValidator<R> {
+    fn new(inner: R, error_tx: tokio::sync::mpsc::Sender<()>) -> Self {
+        Self {
+            inner: ServerRecvStreamValidator::new(inner),
+            error_tx,
+        }
+    }
+
+    fn report_error(&self) {
+        let _ = self.error_tx.try_send(());
+    }
+}
+
+impl<R: RecvStream> RecvStream for ChannelAwareRecvStreamValidator<R> {
+    async fn next(&mut self, msg: &mut dyn RecvMessage) -> Option<Result<(), ()>> {
+        let res = self.inner.next(msg).await;
+        if let Some(Err(())) = res {
+            self.report_error();
+        }
+        res
+    }
+}
+
+pub struct StreamValidationInterceptor;
+
+impl Intercept for StreamValidationInterceptor {
+    async fn intercept(
+        &self,
+        headers: RequestHeaders,
+        options: CallOptions,
+        tx: &mut impl SendStream,
+        rx: impl RecvStream + 'static,
+        next: &impl Handle,
+    ) -> Trailers {
+        let (error_tx, mut error_rx) = channel::<()>(1);
+        let mut wrapped_tx = ChannelAwareSendStreamValidator::new(tx, error_tx.clone());
+        let wrapped_rx = ChannelAwareRecvStreamValidator::new(rx, error_tx);
+
+        tokio::select! {
+            res = next.handle(headers, options, &mut wrapped_tx, wrapped_rx) => {
+                if error_rx.try_recv().is_ok() {
+                    Trailers::new(Err(StatusError::new(StatusCodeError::Internal, "Stream validation error")))
+                } else {
+                    res
+                }
+            }
+            _ = error_rx.recv() => {
+                Trailers::new(Err(StatusError::new(StatusCodeError::Internal, "Stream validation error")))
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::StatusCodeError;
+    use crate::client::CallOptions;
+    use crate::core::{
+        RecvMessage, RequestHeaders, ResponseHeaders, SendMessage, ServerResponseStreamItem,
+        Trailers,
+    };
+    use crate::server::SendOptions;
+    use crate::server::interceptor::HandleExt;
+    use bytes::{Buf, Bytes};
+
+    impl SendMessage for () {
+        fn encode(&self) -> Result<Box<dyn Buf + Send + Sync>, String> {
+            Ok(Box::new(Bytes::new()))
+        }
+    }
+
+    struct NopRecvMessage;
+    impl RecvMessage for NopRecvMessage {
+        fn decode(&mut self, _data: &mut dyn Buf) -> Result<(), String> {
+            Ok(())
+        }
+    }
+
+    struct MockSendStream;
+    impl SendStream for MockSendStream {
+        async fn send<'a>(
+            &mut self,
+            _item: ServerResponseStreamItem<'a>,
+            _options: SendOptions,
+        ) -> Result<(), ()> {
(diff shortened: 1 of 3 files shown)


grpc/src/server/stream_util.rs
```diff
@@ -0,0 +1,375 @@
...
+        if self.state == SendStreamState::Done {
+            return Err(());
+        }
+
+        let next_state = match &item {
+            ServerResponseStreamItem::Headers(_) => match self.state {
+                SendStreamState::Init => SendStreamState::HeadersSent,
+                _ => {
+                    self.state = SendStreamState::Done;
+                    return Err(());
```
dfawley: Please add comments throughout explaining the scenario we encountered that is an error.  Here it would be something like `// Received multiple headers frames`.

dfawley (commented):
I'm trying to figure out how exactly this is going to be used.

Most likely it will be a server-side interceptor?  In which case, when an error occurs, I would think we'd want to terminate the RPC with a status that indicates the cause of the problem.

grpc/src/server/stream_util.rs
sauravzg: Fixed.

sauravzg:
> I'm trying to figure out how exactly this is going to be used.
> 
> Most likely it will be a server-side interceptor? In which case, when an error occurs, I would think we'd want to terminate the RPC with a status that indicates the cause of the problem.

Yes on the server side interceptor.

RecvStreamVal.. is somewhat straight forward. This'll be used closer to the protobuf API to ensure that we transition our generic stream request receiver to the protobuf API stream object.

SendStreamVal.. is closer to the transport to validate the outgoing data after all other interceptors.

I somewhat agree with the need for status. We need to update the `SendStream` API on the server side to return status. This wasn't a problem previously because `Trailers` were a part of the `SendStream` and would be used to send trailers and terminate the API. 

Now that we are returning trailers instead of sending them, we need more details from SendStream to propagate to trailers . Does it sound okay to add `Status` to the `SendStream` API itself, because we initially had some reservations on it , but I guess the situation has changed.

dfawley:
> I somewhat agree with the need for status. We need to update the SendStream API on the server side to return status. This wasn't a problem previously because Trailers were a part of the SendStream and would be used to send trailers and terminate the API.

I was expecting we'd just need an interceptor that can terminate the RPC underneath the handler when either the SendStream or RecvStream has any problem.  I.e. we can't do stream validation via something that simply wraps each stream independently.

sauravzg:
> > I somewhat agree with the need for status. We need to update the SendStream API on the server side to return status. This wasn't a problem previously because Trailers were a part of the SendStream and would be used to send trailers and terminate the API.
> 
> I was expecting we'd just need an interceptor that can terminate the RPC underneath the handler when either the SendStream or RecvStream has any problem. I.e. we can't do stream validation via something that simply wraps each stream independently.

I mean we are definitely gonna need an interceptor for this. But I was planning to separate it into two responsibilities , one for lifecycle for stream and another for lifecycle of handler.

Something along the lines of `RecvStreamValidator` and moving the logic of stream termination closer to the protobuf API. 
and then
`SendStreamValidator` and moving the logic of handling errors from the `SendStream` closer to the transport to handle error from interceptors.

But we might need some `status` on the SendStream now, because if sendstream terminates, we'll need to return Trailers. I could either terminate  the stream with some arbitrary `Internal` stream broken error, or we could try and "safely" propagate the error to trailers?

dfawley:
> But I was planning to separate it into two responsibilities , one for lifecycle for stream and another for lifecycle of handler.

Sure but they will need to be linked, so it would make more sense to me to make them all together.

> Something along the lines of RecvStreamValidator and moving the logic of stream termination closer to the protobuf API.
> and then SendStreamValidator and moving the logic of handling errors from the SendStream closer to the transport to handle error from interceptors.

So you're thinking two different interceptors might be needed?

> or we could try and "safely" propagate the error to trailers?

Yes we should make sure the RPC error includes an explanation of the protocol violation.  I expect the interceptor would basically select on the handler future and a receive from a oneshot channel that the validator writes to, and then terminates with whichever status comes first, dropping the other future.

sauravzg:
So, we discussed offline. It'd make sense for both stream and handler validation to be a part of the same PR.

I initially thought we'd need different handlers, but based on my progress so far, I believe a single one suffices.

> Yes we should make sure the RPC error includes an explanation of the protocol violation. I expect the interceptor would basically select on the handler future and a receive from a oneshot channel that the validator writes to, and then terminates with whichever status comes first, dropping the other future.

Since , we don't expose error status from sendstream or recvstream right now, we cannot propagate anything beyond an arbitrary "internal error: stream broken" . Would you want to tweak our stream APIs to be able to propagate errors Err(Status) or Err(String) instead of Err()

sauravzg:
The validation implementation has been updated to be interceptor based and the stream validation utilities are now private.

There's still a question about if we should introduce "status" or "error messages" into the sendstream and recvstream to be propagated to trailer or will an arbitrary "internal error" suffice. 

Also, this now is stacked on top of https://github.com/hyperium/tonic/pull/2634

grpc/src/server/handler_validation.rs
```diff
@@ -0,0 +1,603 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SendStreamState {
+    Init,
+    HeadersSent,
+    MessagesSent,
+    Done,
+}
+
+impl<S: SendStream> ServerSendStreamValidator<S> {
+    pub fn new(inner: S) -> Self {
```
dfawley: I think this needs to not be public since the type isn't public?  (And same for the other types below)

grpc/src/server/handler_validation.rs
```diff
@@ -0,0 +1,603 @@
...
+}
+
+struct ServerRecvStreamValidator<R> {
+    inner: R,
+    state: RecvStreamState,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum RecvStreamState {
+    Init,
```
dfawley: Nit: `Init` isn't quite right here, because you stay in init even after getting some messages.  Maybe `Active`?  Or you could replace this enum with a bool, maybe.

grpc/src/server/handler_validation.rs
```diff
@@ -0,0 +1,603 @@
...
+                self.state = RecvStreamState::Done;
+                Some(Err(()))
+            }
+        }
+    }
+}
+
+struct ChannelAwareSendStreamValidator<S> {
+    inner: ServerSendStreamValidator<S>,
+    error_tx: tokio::sync::mpsc::Sender<()>,
```
dfawley, line 132: Would an `Option<Oneshot>` be preferable here?  It might make the access simpler, but I'm not sure.

grpc/src/server/handler_validation.rs
```diff
@@ -0,0 +1,603 @@
...
+    ) -> Result<(), ()> {
+        let res = self.inner.send(item, options).await;
+        if res.is_err() {
+            self.report_error();
+        }
+        res
+    }
+}
+
+struct ChannelAwareRecvStreamValidator<R> {
```
dfawley, line 162: Maybe we don't need this? It seems OK if the application keeps receiving after getting an error.  We want our wrapper to avoid propagating that read down further, since it would be illegal-per-the-API to do so, but I don't think that needs to error the whole call?

grpc/src/server/handler_validation.rs
```diff
@@ -0,0 +1,603 @@
...
+        next: &impl Handle,
+    ) -> Trailers {
+        let (error_tx, mut error_rx) = channel::<()>(1);
+        let mut wrapped_tx = ChannelAwareSendStreamValidator::new(tx, error_tx.clone());
+        let wrapped_rx = ChannelAwareRecvStreamValidator::new(rx, error_tx);
+
+        tokio::select! {
+            res = next.handle(headers, options, &mut wrapped_tx, wrapped_rx) => {
+                if error_rx.try_recv().is_ok() {
+                    Trailers::new(Err(StatusError::new(StatusCodeError::Internal, "Stream validation error")))
```
dfawley, line 208: On the one hand, it's unfortunate that we aren't being any more specific than this.  "Server attempted to send multiple headers / messages before headers / etc"

On the other hand, maybe it's for the best, since the client maybe shouldn't get details about server bugs?

sauravzg: With the current API that's likely the best we can do. The issue may exaggerate when dealing with interceptors because there would be no way to propagate error from an interceptor's stream  which may be a painful debugging experience. 

But maybe the right answer is that server interceptor implementations can log errors when as needed, and we treat all of it as internal error with a generic error message to the client of some form.

grpc/src/server/handler_validation.rs
sauravzg: Done.

grpc/src/server/handler_validation.rs
sauravzg: Done.

grpc/src/server/handler_validation.rs
sauravzg: So, this is mpsc because the sender is shared between send and recv streams. So, this is subject to the conversation below, if we can eliminate Chann...Recv...Valida...

grpc/src/server/handler_validation.rs
sauravzg: Can you help me understand the reason behind the asymmetry in how we treat recvstream errors vs sendstream errors? 

Probably worth noting that this could happen on interceptor side , in which case maybe we want to send out an error.

dfawley: On the sending side, if the server application attempts to violate the gRPC protocol by sending the wrong data on the stream (e.g. transmit headers after messages), the stream is corrupted and we can't reasonably proceed.  We should end the stream with an error.

On the receiving side, if the server application requests data after we told it there was no more data....well that's wrong, but it doesn't actually cause any problems.  The overall call can still function just fine.  We just keep telling it "no, you have no more data" over and over if they ask.

So I believe this situation is fine.  But if you have reasons why it's more dangerous than that then I'm happy to discuss further.

sauravzg: What about other types of error, beyond polling the empty stream again? What if an interceptor sends an error, which doesn't necessarily mean end of stream? I assume we want to terminate the call on this one with some form of internal error instead of letting the application continue. 

This'd imply we don't need the validator, but would still the channel error signal which is able to distinguish between repolling and an interceptor error?

dfawley: Hmm, I think it's not clear to me what the role of this interceptor is.  I was assuming it was validating the application's use of the stream.  But maybe it's intended for something else? On the client side, it's mainly there to assert the generated code's assumptions in case an interceptor below it did something wrong.

sauravzg: So here are what I think rough requirements for grpc handler are:

1. Is the incoming stream valid grpc request? If no, short circuit stream and return error . The source of errors for incoming stream errors could be 
   - polling after completion
   - bad stream data(let's say corrupted message)
   - application interceptor's recvstream returning error (due to arbitrary business logic dependent reasons)
2. Is the outgoing stream valid grpc response? Similarly, sources of error
     - application interceptor's sendstream errors(arbitrary business logic in intercetor)
     - bad data from application application(incorrect grpc order)
     - anything down stream maybe?
3. Any type of error that breaks either side of the stream terminates the handler and sends a RST_STREAM 


1 and 2 should be the same as on the client side (an interceptor could do something wrong on the server side), 3 is the new requirement which needs a channel to listen on either error and then terminate the handler somehow with an error, which can then be used to cancel the call on transport.

dfawley: Specifically, though, where are you putting this interceptor?

On the client, it's employed by the generated code to ensure the stream in from the channel follows its expectations.  The generated code ensures that protocol isn't violated in the outbound direction.

dfawley: And to be clear, I'm not saying this needs to be the analog of that for the server.  I just want to know where you're seeing this going.

Maybe this is going at the perimeter of the server library to make sure the application is behaving correctly instead?  But depending on where it goes and what its job is, the code would look very different.

sauravzg: I am planning for it to site closer to the transport rather than codegen i.e. be the first interceptor.

I'd like the  ServerValidationInterceptor to be able to monitor each and every error , be it from codegen or from application interceptors.

This also allows it to handle all potential error cases and immediately terminate the server handler on any error and cancel the rpc. 

If I were to put it closer to the codegen, I couldn't guarantee the cancellation on error, since another interceptor on could receive the validation error and choose to ignore it and do some shenanigans that we wouldn't want.

dfawley: OK, that's what I was thinking this was at first.  In that case, I also think what I was saying before is reasonable: this is validating the application, not the gRPC library, so we don't need to deal with errors that occur inside the library, like "data is corrupt".  If the data is corrupt, then the channel should see the error from the (de)serializing of the messages and propagate errors & drop the handler.

So I think this only needs to validate the send side of the stream.  If the recv side is used incorrectly by the application (read after error), then it doesn't cause any problems, per my above argument.

And:

> application interceptor's recvstream returning error (due to arbitrary business logic dependent reasons)

If this is sitting at the edge of the server, _before_ the application's interceptors/handler, then we won't see their recvstreams at all.  We can't alter or respond to their behavior.  If those send arbitrary recv errors to the application, we'll never know about them.  We only can observe the channel's recvstream from here.


If we _also_ need something like the client side interceptor, which the generated code uses to enforce the semantics it assumes about the incoming data (e.g. that a unary request has exactly one message), then it would be separate from this.  This is validation logic that we must have, too.  Whether it uses an interceptor for that, though, is another decision.

sauravzg: So, this is probably where the recvstream /sendstream implementation assymmetry kicks in.

IIUC recvstreams are usually supposed to be like. This is obviously to the discretion of the recvstream implementation , but doing stuff to x before calling inner.recv may not make a lot of sense.
```
recv(x){
inner.recv(x)?;
x.modify();
}
```
while sendstreams are usually opposite . The same applies i.e. applications can modify post inner call but likely won't do anything in particular. 

```
send(x){
x.modify();
inner.send(x);
}
```

So, even if the validation sits closer to the transport even on recv side, it actually will run at the end of all interceptors and can catch those.

dfawley: The application's interceptors will wrap this interceptor's send & recv streams before calling the final handler.

That means we can't observe any errors they introduce between them and the application; we can only see their usage of this interceptor's streams.  I.e. if they return an error from their `recv` back to the application after calling our `recv`, we have no visibility into that.

Aside: you're right those should be the most common patterns, but I could see this kind of thing happening, too:

```rust
fn recv(..) -> .. {
  let my_message = wrap(x);
  inner.recv(&my_message)
}
```

What situation are you still worried about?

sauravzg: Yes. We have zero visibility to what they do beyond streams. So, I want to understand that's the expected behavior when an interceptor does

```
fn recv(x) {
  if num_messages> 5:
     return Err()
  inner.recv(x);
}
```

Are we supposed to treat it like a regular recvstream failure and propagate the error, or are we supposed to close the handler with internal stream error of some sort? 

I am leaning towards the latter.

dfawley: > So, I want to understand that's the expected behavior when an interceptor does

If this interceptor is _inside_ our channel at the boundary, we don't ever see that error anyway.  So we don't need to think about it for this.

If their interceptor is doing that, we'd normally expect that interceptor to also terminate the handler future ASAP and return an error from the `intercept` method.

The handler itself should also see that error and know the RPC was cancelled or errored, and return from their handler ASAP, but they can't return a very useful status based on this.

sauravzg: Okay, you are right. Roughly putting this closer to the transport and expecting to prevent interceptor doing bad things from there is essentially lost cause. 

So, I've actually been thinking of the behavior that we want for the last couple of days for corner cases.

1. Should recvs and send side interceptors receive some guarantees on the state machine from the transport or  the application? Likely no, since  this seems like a lost cause since a buggy interceptor can create out of order messages for downstream interceptors? 
2. Should the transport be protected from bad sendstream from interceptors? Likely yes, because while hyper provides some safety, it doesn't guarantee the full grpc spec safety. 
3. Should the application handler be protected from bad recvstream from interceptors? Likely yes, this'll probably be somewhere in codegen?
4. What's the behavior we want on some interceptor's individual `recvstream` message failure(think throttling interceptor mid streaming request?) to the application, should this end up cancelling the  stream or should the interceptor be responsible for handling that and grpc just propagates the error to the application handler? leaning towards, making interceptor's problem. 
5. What's the behavior we want on some interceptor's individual `sendstream` message failure(think throttling and not a protocol ordering violation ), does this drop handler and cancel stream?


Here's what I am currently leaning towards
- Some validation on the codegen side for recvstream. Mostly focusing on ensuring terminal states are terminal. That ensures that the API doesn't allow.
- Some validation on the codegen side for sendstream. This will be focussed on error handling from interceptors, all errors are fallible and terminate the handler with reset stream.
- Some validation closer to the transport side for sendstream (roughly this PR). This ensure that we send valid grpc to transport.

Does that sound good ?

dfawley: > * Some validation on the codegen side for recvstream. Mostly focusing on ensuring terminal states are terminal. That ensures that the API doesn't allow.

On the server the recvstream is simple.  There are two aspects of it:

1. No more messages received after end.  I actually don't care too much about this.  The application/etc should not be reading anymore once they get `None`.  If they do and there is somehow more data there, then "oh well"?  I don't think we should worry about this at all.

2. Unary request vs. streaming request.  This needs to be enforced.  The generated code should be validating this -- a unary request handler (which includes server-streaming RPCs) should read from the stream twice and confirm the second read returns `None`.  If multiple messages are received, that's an error - the handler should never be called and it should immediately return a cardinality violation error (UNIMPLEMENTED per https://github.com/grpc/grpc/blob/master/doc/statuscodes.md) that the client application will see.  This behavior probably doesn't need to be in an interceptor, but if it works out more nicely that way then I'm fine with it.

On the sendstream, it's a little more complex, since it _must_ begin with headers, then only messages (any number):

> * Some validation on the codegen side for sendstream. This will be focussed on error handling from interceptors, all errors are fallible and terminate the handler with reset stream.

The codegen doesn't need to care about any of this, since it never gives the stream out directly -- it only calls the proto-based handler which is type-constrained to send the right number of messages and no headers/trailers.  This ensures response stream cardinality without needing a validator.

> Some validation closer to the transport side for sendstream (roughly this PR). This ensure that we send valid grpc to transport.

The channel needs to enforce the application (interceptors+handler) is following the gRPC protocol.  So the channel needs what you have here for the send side.  If the application violates the protocol, this interceptor needs to terminate the RPC with INTERNAL.  (So: what you said SGTM.)

PR #2604, opened 2026-04-24
feat(tonic-xds): add OutlierDetectionConfig types (gRFC A50)

LYZJU2019:
## Summary

First of a 4-PR series implementing [gRFC A50: xDS Outlier Detection](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) in `tonic-xds`. This PR is **type definitions only — no behavior change**.

Defines three validated config types that the algorithm will consume:

- `OutlierDetectionConfig` — sweep interval, base/max ejection time, max ejection percent, plus the two optional sub-configs.
- `SuccessRateConfig` — `stdev_factor`, `enforcement_percentage`, `minimum_hosts`, `request_volume`.
- `FailurePercentageConfig` — `threshold`, `enforcement_percentage`, `minimum_hosts`, `request_volume`.
- `OutlierDetectionConfig::is_enabled()` — true iff at least one ejection algorithm is configured.

Sub-configs are `Option<_>` so the absence of either algorithm is part of the type, not a sentinel value.

## Why split this out

The full A50 implementation pulls in (a) proto parsing, (b) the statistical algorithms + sweep engine, (c) per-RPC outcome interception, and (d) wiring into the load-balancing pipeline. Reviewing them in one PR is hard. This series:

1. **This PR** — config types only.
2. Outlier-detection algorithm + sweep engine (pure, unit-tested in isolation).
3. Per-endpoint RPC outcome interception (a `Service` wrapper).
4. Proto parsing into `ClusterResource`, wire-up in `XdsClusterDiscovery` (preserving connections on ejection), end-to-end tests, and the `lib.rs` feature-table update.

## Test plan

- `cargo test -p tonic-xds --lib outlier` — 3 tests for `is_enabled()`.
- `cargo fmt -p tonic-xds --check` clean.
- `cargo clippy -p tonic-xds --lib --all-features -- -D warnings` clean.

### DIFF
--- a/tonic-xds/src/xds/resource/outlier_detection.rs
+++ b/tonic-xds/src/xds/resource/outlier_detection.rs
@@ -1,0 +1,114 @@
+//! Outlier-detection configuration types (gRFC A50).
+//!
+//! These are the validated config inputs consumed by the outlier-detection
+//! algorithm. Parsing them from `envoy.config.cluster.v3.OutlierDetection`
+//! and exposing them on `ClusterResource` lands in a follow-up PR alongside
+//! the wiring into the load-balancing pipeline.
+//!
+//! [gRFC A50]: https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md
+
+use std::time::Duration;
+
+/// Validated A50 outlier-detection configuration for a cluster.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct OutlierDetectionConfig {
+    /// How often the ejection sweep runs.
+    pub interval: Duration,
+    /// Base duration for a single ejection; actual ejection time is
+    /// `base_ejection_time * multiplier`, capped by `max_ejection_time`.
+    pub base_ejection_time: Duration,
+    /// Upper bound on `base_ejection_time * multiplier`. The spec guarantees
+    /// this is at least `base_ejection_time`.
+    pub max_ejection_time: Duration,
+    /// Maximum percentage of endpoints that may be ejected at any time (0-100).
+    pub max_ejection_percent: u32,
+    /// Success-rate ejection parameters. `None` if the algorithm is disabled.
+    pub success_rate: Option<SuccessRateConfig>,
+    /// Failure-percentage ejection parameters. `None` if the algorithm is
+    /// disabled.
+    pub failure_percentage: Option<FailurePercentageConfig>,
+}
+
+/// Success-rate ejection parameters (gRFC A50).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SuccessRateConfig {
+    /// Ejection threshold factor, scaled by 1000 (so `1900` means `1.9`).
+    /// An endpoint is a candidate for ejection when its success rate falls
+    /// below `mean - stdev * (stdev_factor / 1000.0)`.
+    pub stdev_factor: u32,
+    /// Probability (0-100) that a candidate is actually ejected.
+    pub enforcement_percentage: u32,
+    /// Minimum number of candidate endpoints required to run the algorithm.
+    pub minimum_hosts: u32,
+    /// Minimum number of requests an endpoint must have seen in the last
+    /// interval to be considered a candidate.
+    pub request_volume: u32,
+}
+
+/// Failure-percentage ejection parameters (gRFC A50).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct FailurePercentageConfig {
+    /// Failure rate (0-100) at or above which an endpoint is a candidate
+    /// for ejection.
+    pub threshold: u32,
+    /// Probability (0-100) that a candidate is actually ejected.
+    pub enforcement_percentage: u32,
+    /// Minimum number of candidate endpoints required to run the algorithm.
+    pub minimum_hosts: u32,
+    /// Minimum number of requests an endpoint must have seen in the last
+    /// interval to be considered a candidate.
+    pub request_volume: u32,
+}
+
+impl OutlierDetectionConfig {
+    /// True when at least one ejection algorithm is enabled and the detector
+    /// should do work. If false, the cluster can skip instantiating detection.
+    pub(crate) fn is_enabled(&self) -> bool {
+        self.success_rate.is_some() || self.failure_percentage.is_some()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn base_config() -> OutlierDetectionConfig {
+        OutlierDetectionConfig {
+            interval: Duration::from_secs(10),
+            base_ejection_time: Duration::from_secs(30),
+            max_ejection_time: Duration::from_secs(300),
+            max_ejection_percent: 10,
+            success_rate: None,
+            failure_percentage: None,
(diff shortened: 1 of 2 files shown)


tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+/// Failure-percentage ejection parameters (gRFC A50).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct FailurePercentageConfig {
+    /// Failure rate (0-100) at or above which an endpoint is a candidate
+    /// for ejection.
+    pub threshold: u32,
+    /// Probability (0-100) that a candidate is actually ejected.
+    pub enforcement_percentage: u32,
```
ankurmittal: maybe also add validation logic? also u16 should suffice here? as max is 100? same with other config values

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,114 @@
...
+/// Success-rate ejection parameters (gRFC A50).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SuccessRateConfig {
+    /// Ejection threshold factor, scaled by 1000 (so `1900` means `1.9`).
+    /// An endpoint is a candidate for ejection when its success rate falls
+    /// below `mean - stdev * (stdev_factor / 1000.0)`.
+    pub stdev_factor: u32,
+    /// Probability (0-100) that a candidate is actually ejected.
+    pub enforcement_percentage: u32,
+    /// Minimum number of candidate endpoints required to run the algorithm.
```
ankurmittal, line 73: what about enforcing_success_rate as per the spec?

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// this is at least `base_ejection_time`.
+    pub max_ejection_time: Duration,
+    /// Maximum percentage of endpoints that may be ejected at any time (0-100).
+    pub max_ejection_percent: u32,
+    /// Success-rate ejection parameters. `None` if the algorithm is disabled.
+    pub success_rate: Option<SuccessRateConfig>,
+    /// Failure-percentage ejection parameters. `None` if the algorithm is
+    /// disabled.
+    pub failure_percentage: Option<FailurePercentageConfig>,
+}
```
ankurmittal, line 62: no child_policy ?

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,114 @@
+//! Outlier-detection configuration types (gRFC A50).
+//!
+//! These are the validated config inputs consumed by the outlier-detection
+//! algorithm. Parsing them from `envoy.config.cluster.v3.OutlierDetection`
+//! and exposing them on `ClusterResource` lands in a follow-up PR alongside
```
YutaoMa: For "progress tracking" comments like these better put them in TODO comments so they can be cleaned up later, or just leave them as PR comments instead.

tonic-xds/src/xds/resource/outlier_detection.rs
YutaoMa: ^, better enforce invariants through type system for all `Resouce` types since we have the validation and NACK workflow

tonic-xds/src/xds/resource/outlier_detection.rs
LYZJU2019: Discussed offline with @ankurmittal. For now we only support P2C balancer so omit `child_policy` in this PR. We can add this field later if needed.

tonic-xds/src/xds/resource/outlier_detection.rs
LYZJU2019: Done. Added a `Percentage(u8)` newtype with a fallible `new(value: u32) -> Result<Self, PercentageError>` constructor (returns the proto wire type `u32` so the parser doesn't have to cast at every site, then narrows to `u8` once validated). The four percentage fields now hold `Percentage` instead of `u32`: `max_ejection_percent`, `enforcing_success_rate`, `threshold`, `enforcing_failure_percentage`. Once constructed, the algorithm never has to re-validate.

tonic-xds/src/xds/resource/outlier_detection.rs
ankurmittal: we will still need child policy to wrap the balancer so that we are not picking failed endpoint again and again, but can be done in future PR

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,159 @@
+//! Validated configuration types for [gRFC A50] outlier detection.
+//!
+//! [`OutlierDetectionConfig`] is the input to the outlier-detection
+//! algorithm. The two sub-configs gate which ejection algorithms run.
+//!
+//! Note: A50 specifies outlier detection as a load-balancing policy
+//! wrapping a `child_policy`. `tonic-xds` currently runs P2C as its only
+//! load balancer and integrates outlier detection as a filter on the
+//! `Discover` stream feeding it, so there is no `child_policy` field
```
gu0keno0, line 9: This is not entirely accurate. Tonic-xds outlier detection will likely be implemented as a stream / mpsc channel that can be polled by LoadBalancer layer in https://github.com/hyperium/tonic/pull/2607 . The implementation will need to leverage the EjectedChannel type added in https://github.com/hyperium/tonic/pull/2587

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+/// Success-rate ejection parameters (gRFC A50).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SuccessRateConfig {
+    /// Ejection threshold factor, scaled by 1000 (so `1900` means `1.9`).
+    /// An endpoint is a candidate for ejection when its success rate falls
+    /// below `mean - stdev * (stdev_factor / 1000.0)`.
+    pub stdev_factor: u32,
+    /// Probability that a candidate is actually ejected.
```
gu0keno0, line 71: I think this is a threshold for ejecting a host if RPC success rate drops below it?

gu0keno0 (approved):

gu0keno0 (commented):
We should also support consecutive 5xx as well. It's not in the original grpc spec but is useful.

Grpc status are not http error codes, however, envoy xds has error matcher config https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/upstreams/http/v3/http_protocol_options.proto#extensions-upstreams-http-v3-httpprotocoloptions-outlierdetection to allow us to specify how to map headers to 5xx.

I think you can proceed with basic A50, and then add support for consecutive 5xx.

tonic-xds/src/xds/resource/outlier_detection.rs
LYZJU2019: Thanks for this comment @gu0keno0! Will update the module description in subsequent PRs when #2607 is merged.

PR #2607, opened 2026-04-28
[tonic-xds] Add the loadbalancer tower service

gu0keno0:
Add load balancer tower service that's extensive to support outlier detection and customized channel pickers.

Issue: https://github.com/hyperium/tonic/issues/2444

## Motivation
tonic-xds currently uses tower::balance, which is hard-coded to using P2C LB algorithm. For xDS, we need more flexibility and need full control over the underlying channels so that we can implement customized LB, outlier detection, etc.

## Solution
Leverage the channel states (Idle, Connecting and Ready for now), build a new LoadBalance tower service whose poll_ready() handles channel state transitions and call() handles load balance picking algorithm.

Also, added a simple ChannelPicker trait to abstract channel picking algorithm. In the future, customized LB can implement this trait and tonic-xds will provide a way to register customized LB and make it available for LoadBalance layer.

### DIFF
--- a/tonic-xds/src/client/loadbalance/pickers/mod.rs
+++ b/tonic-xds/src/client/loadbalance/pickers/mod.rs
@@ -10,5 +10,5 @@ use crate::client::endpoint::EndpointAddress;
 /// The picker only needs to observe `S`'s load — it doesn't depend on any
 /// specific channel state type.
 pub(crate) trait ChannelPicker<S, Req> {
-    fn pick(&self, req: &Req, ready: &IndexMap<EndpointAddress, S>) -> Option<usize>;
+    fn pick<'a>(&self, req: &Req, ready: &'a IndexMap<EndpointAddress, S>) -> Option<&'a S>;
 }
--- a/tonic-xds/src/client/loadbalance/pickers/mod.rs
+++ b/tonic-xds/src/client/loadbalance/pickers/mod.rs
@@ -10,9 +10,5 @@ use crate::client::endpoint::EndpointAddress;
 /// The picker only needs to observe `S`'s load — it doesn't depend on any
 /// specific channel state type.
 pub(crate) trait ChannelPicker<S, Req> {
-    fn pick(
-        &self,
-        req: &Req,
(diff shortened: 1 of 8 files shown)


tonic-xds/src/client/loadbalance/pickers/mod.rs
```diff
@@ -0,0 +1,14 @@
...
+
+use crate::client::endpoint::EndpointAddress;
+
+/// Trait for selecting a channel to handle a request.
+///
+/// Generic over `S` (the channel type in the ready set) and `Req` (the request).
+/// The picker only needs to observe `S`'s load — it doesn't depend on any
+/// specific channel state type.
+pub(crate) trait ChannelPicker<S, Req> {
+    fn pick(&self, req: &Req, ready: &IndexMap<EndpointAddress, S>) -> Option<usize>;
```
YutaoMa: If this returns a lifetime-scoped borrow to a map entry, then it simplifies the usage by removing the need for "picker returned invalid index" error branch

gu0keno0: I thought about this earlier, the constraint is that S needs to be mutable because later call() needs to forward the call to S.call(), and S.call() takes mut S. So ... seems to me that the best trade-off between complexity and safety for now is to return index. Open to suggestions.

YutaoMa: but the current `loadbalancer.rs` `call()` design clones the `svc` before dispatching the inner call, so I don't think borrowing up to the clone point will be an issue?

gu0keno0: Makes sense, I was too attached to the svc in the map, will look into it and change.

LYZJU2019 (approved):
Approve from my side

tonic-xds/src/client/loadbalance/pickers/mod.rs
gu0keno0: Done.

YutaoMa (approved):

PR #2610, opened 2026-04-28
feat(grpc): Google call credentials

arjan-bal:
This PR introduces a `CallCredentials` implementation that injects authentication tokens based on GCP Application Default Credentials (ADC), leveraging the official [`google-cloud-auth`](https://github.com/googleapis/google-cloud-rust/tree/main/src/auth) crate.

To avoid pulling in Google-specific dependencies by default, the implementation is housed in a new `grpc-google` crate. Additionally, to facilitate unit testing, the following types from the `grpc` crate are now exported:
* The `attributes` module.
* Constructors for the static data structs `CallDetails` and `ClientConnectionSecurityInfo`.

The code was manually tested by running [this example](https://github.com/arjan-bal/tonic/pull/11/changes/599f5ba4bd55620a8bfc6597c9f8179f94756568), which will be included in a follow-up PR.


grpc-google/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+use tonic::metadata::AsciiMetadataValue;
+use tonic::metadata::MetadataMap;
+
+const DEFAULT_CLOUD_PLATFORM_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
+
+/// An abstraction for fetching authentication tokens.
+#[trait_variant::make(Send)]
+pub trait TokenProvider: Sync + Debug + 'static {
+    /// Returns an authentication token.
+    async fn get_token(&self) -> Result<String, String>;
```
sauravzg, line 50: nit: Are we sure about `String` error type? It loses some context error context.

grpc-google/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// specified scopes.
+    ///
+    /// [ADC-link]: https://cloud.google.com/docs/authentication/application-default-credentials
+    pub fn new_application_default_with_scope<I, S>(scopes: I) -> Result<Self, String>
+    where
+        I: IntoIterator<Item = S>,
+        S: Into<String>,
+    {
+        let credentials = google_cloud_auth::credentials::Builder::default()
+            .with_scopes(scopes)
```
sauravzg, line 80: For my education: Is scopes a mandatory argument and the default scope documented somewhere? 
What happens if this isn't specified?

sauravzg (approved):

grpc-google/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+*/
+
+//! GCP Credentials implementation for gRPC.
```
dfawley: Please spell out what GCP stands for and ideally link to a site with more info.

grpc-google/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+//!
+//! This crate provides a way to create gRPC channel credentials that
+//! automatically fetch and attach Google Cloud Platform (GCP) authentication
```
dfawley: Oh, just move the explanation from here then?

grpc-google/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "grpc-google"
+version = "0.9.0-alpha.1"
```
dfawley: `0.0.0-alpha`?

grpc-google/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+            .get_token()
+            .await
+            .map_err(|e| Status::new(StatusCode::Unavailable, e))?;
+        let mut value: AsciiMetadataValue = format!("Bearer {}", token).parse().map_err(|e| {
+            Status::new(
+                StatusCode::Internal,
+                format!("invalid values in authorization header value: {}", e),
+            )
+        })?;
+        value.set_sensitive(true);
```
dfawley, line 114: Outside the scope of this PR: Is this not the default? I suspect we would want it to be.

dfawley (approved):
Looks good except for the nit about the comments.

grpc-google/src/lib.rs
arjan-bal: Spelled out the full term here and used the abbreviation in the next paragraph.

grpc-google/src/lib.rs
arjan-bal: Done.

grpc-google/src/lib.rs
arjan-bal: Generally, we only need specific error types if callers must handle errors differently—a requirement we don't currently have. That said, returning the underlying library's errors directly makes them part of our public API. We could introduce a custom error enum with a single variant for now to protect our API boundary and preserve the ability to add more types later. Do you think we should go this route?

grpc-google/src/lib.rs
arjan-bal: [Scopes](https://developers.google.com/identity/protocols/oauth2/scopes) restrict which APIs can be called using the generated token. I've set the default Cloud Platform scope based on [Go's implementation](https://github.com/grpc/grpc-go/blob/850cc543aa3c5b86b82697b91f3ec5a7210829a0/credentials/google/google.go#L124-L126).

If scopes are not provided, the [IAM API](https://docs.cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken) rejects the request. However, when scopes are omitted on the metadata server, it seems to default to [using all scopes](https://github.com/googleapis/google-cloud-rust/blob/8efbc5527361243af5d3ccf5e01e82a68f104682/src/auth/src/credentials/mds.rs#L208-L216) configured for that service account.

grpc-google/src/lib.rs
arjan-bal: [http::HeaderValue](https://docs.rs/http/latest/src/http/header/value.rs.html#212-225) defaults to `false`, and Tonic doesn't appear to have any logic to override this to `true`. The gRPC metadata map also defaults to `false`. Should I change the behaviour of the gRPC map?

grpc-google/src/lib.rs
sauravzg: Let's keep it as string for now. I was looking for something more generic like `absl::Status` . Might be worth revisiting when the library is ready.

grpc-google/src/lib.rs
sauravzg: I tend to have a personal preference towards providing the same behavior as the underlying library to avoid any surprises for the end user.

Unless of course the practice of "default Cloud Platform" scope is something that's a part of grpc's practices across majority of languages, or if there's a valid historical reason behind the use of this default.

arjan-bal: Go and [C++](https://github.com/grpc/grpc/blob/61b93b03f64d2e29d2ca1504201fcffb8e201246/src/core/credentials/call/external/external_account_credentials.cc#L594-L603) use the GCP scope by default, whereas Java [expects the application](https://github.com/grpc/grpc-java/blob/01a43ec40c621435940a2512d97353591ccadc31/examples/example-gauth/src/main/java/io/grpc/examples/googleAuth/GoogleAuthClient.java#L121) to explicitly create the GCP credentials object with the required scopes. Since most users rely on the default GCP scope, a second constructor is available for those who need to specify custom scopes.

sauravzg: lgtm

grpc-google/Cargo.toml
arjan-bal: Changed. I was matching the version with the `grpc` crate earlier.

grpc-google/src/lib.rs
arjan-bal: Discussed offline. We want to make values sensitive by default to avoid accidental logging of credentials. I'll raise a follow-up PR for this.

PR #2613, opened 2026-04-28
grpc: add route_guide example and make minor tweaks to the generated code APIs

dfawley:
Also add a temporary `client_only` flag to our generated code to prevent the output of the tonic-protobuf server-side code.

### DIFF
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -97,8 +97,8 @@ fn codegen_grpc_routeguide(root_dir: &Path) {
     let path = std::env::var("PATH").unwrap_or_default();
     unsafe {
         std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
-        // OUT_DIR must be set in CodeGen::new, so we set it correctly here
-        // instead of using `.output_dir()` on CodeGen.
+        // TODO: Once the codegen is updated to allow the `OUT_DIR` env var to
+        // be unset, change this to use `.output_dir` instead.
         std::env::set_var("OUT_DIR", root_dir.join("src/grpc-routeguide/generated"));
     }
 
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -81,7 +81,35 @@ fn main() {
         true,
         true,
     );
+
+    // grpc-routeguide example
+    codegen_grpc_routeguide(
+        &PathBuf::from(std::env!("CARGO_MANIFEST_DIR"))
+            .parent()
+            .unwrap()
+            .join("examples"),
+    );
+
     println!("Codgen completed: {}ms", start.elapsed().as_millis());
+}
+
+fn codegen_grpc_routeguide(root_dir: &Path) {
+    let path = std::env::var("PATH").unwrap_or_default();
+    unsafe {
+        std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
+        // OUT_DIR must be set in CodeGen::new, so we set it correctly here
+        // instead of using `.output_dir()` on CodeGen.
+        std::env::set_var("OUT_DIR", root_dir.join("src/grpc-routeguide/generated"));
+    }
+
+    std::env::set_current_dir(root_dir).unwrap();
+
+    grpc_protobuf_build::CodeGen::new()
+        .include("proto/routeguide")
+        .inputs(["route_guide.proto"])
+        .client_only()
+        .compile()
+        .unwrap();
 }
 
 fn codegen(
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -1,6 +1,8 @@
 use std::{env, path::PathBuf};
 
 fn main() {
+    build_grpc();
+
     tonic_prost_build::configure()
         .compile_protos(&["proto/routeguide/route_guide.proto"], &["proto"])
         .unwrap();
@@ -64,4 +66,25 @@ fn build_json_codec_service() {
         .build();
 
     tonic_prost_build::manual::Builder::new().compile(&[greeter_service]);
+}
+
+fn build_grpc() {
+    let proto = "proto/routeguide/route_guide.proto";
+
+    eprintln!("{}", grpc_protobuf_build::protoc());
+    let path = std::env::var("PATH").unwrap_or_default();
+    unsafe {
+        std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
+    }
+
+    grpc_protobuf_build::CodeGen::new()
+        .include("proto/routeguide")
+        .inputs(["route_guide.proto"])
+        .output_dir("src/grpc-routeguide/generated")
+        .client_only()
+        .compile()
+        .unwrap();
+
+    // prevent needing to rebuild if files (or deps) haven't changed
+    println!("cargo:rerun-if-changed={proto}");
 }
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -1,8 +1,6 @@
 use std::{env, path::PathBuf};
 
 fn main() {
-    build_grpc();
-
     tonic_prost_build::configure()
         .compile_protos(&["proto/routeguide/route_guide.proto"], &["proto"])
         .unwrap();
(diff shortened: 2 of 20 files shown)


examples/build.rs
```diff
@@ -65,3 +67,24 @@ fn build_json_codec_service() {
...
+fn build_grpc() {
+    let proto = "proto/routeguide/route_guide.proto";
+
+    eprintln!("{}", grpc_protobuf_build::protoc());
+    let path = std::env::var("PATH").unwrap_or_default();
+    unsafe {
+        std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
+    }
+
+    grpc_protobuf_build::CodeGen::new()
```
arjan-bal: While writing the GCP call creds example, I was concerned that adding a gRPC build step to the existing examples would mean users must have `protoc-gen-rust-grpc` in their `PATH`, even if they are just trying to run standard `tonic` examples. I wanted to avoid this poor user experience, so I tried to work around it by using a feature flag in the build file (see the reverted change here: https://github.com/hyperium/tonic/pull/2610/changes/6f588add514baa557bc3fc0c90723244a951d777).

However, I ran into a CI failure where `udeps` flagged `grpc_protobuf_build` as an unused dependency, likely because of the feature flag. I haven't spent time resolving that failure yet, and I'm also unsure if using feature flags in build files is an encouraged practice here. 

Maybe we should just have a separate directory specifically for gRPC examples?

examples/build.rs
```diff
@@ -65,3 +67,24 @@ fn build_json_codec_service() {
...
+        .include("proto/routeguide")
+        .inputs(["route_guide.proto"])
+        .output_dir("src/grpc-routeguide/generated")
+        .client_only()
+        .compile()
+        .unwrap();
+
+    // prevent needing to rebuild if files (or deps) haven't changed
+    println!("cargo:rerun-if-changed={proto}");
```
arjan-bal: Both the message codegen and the service codegen emit these instructions:
https://github.com/hyperium/tonic/blob/368361b54c1f09bf652d87b11833860b016be872/grpc-protobuf-build/src/lib.rs#L234-L243

So this may be redundant.

examples/src/grpc-routeguide/generated/route_guide_grpc.pb.rs
arjan-bal, line 1: For tonic examples, the generated code is not checked in. Do we want to check-in generated code for grpc examples? If yes, tonic uses [this binary](https://github.com/hyperium/tonic/blob/master/codegen/src/main.rs) to re-generate the code. This binary also runs as part of CI to detect when the checked in code needs to be updated.

examples/build.rs
dfawley: Hmm, but why is that predicated on `!generate_message_code` (which we are leaving set here).

examples/build.rs
dfawley: Yeah I was thinking about this, too.

For the grpc demos I think this dependency is OK since users will realistically need to set it up eventually anyway.  But I agree it shouldn't apply to the tonic demos.  I would hope we could make the dependency depend on the feature flag, but IDK if that's possible.  I'll try that out later.

examples/build.rs
arjan-bal: This part was copied from [protobuf codegen](https://github.com/protocolbuffers/protobuf/blob/86f66fed851fce03f4fb7276ecaff75686a811cb/rust/release_crates/protobuf_codegen/src/lib.rs#L173-L180). I put it behind the `!self.generate_message_code` check to avoid printing duplicate lines that the protobuf crate was also printing.

examples/src/grpc-routeguide/generated/route_guide_grpc.pb.rs
LucioFranco: I don't think it matters much anyways just creates some less noise which may make it easier for users to navigate. That said, people do tend to get tripped up when they can't see the code gen. But even looking at the codegen its not super readable anyways.

dfawley: For this example anyway, I like the idea of checking in the code and using a separate binary to regenerate.  Users then don't need protoc or to compile the grpc-rust plugin (in C++), or worry about having stuff in their path.

I'm not sure how we expect people to be using this most commonly in OSS, but I _think_ that would be a more common workflow than generating the code via `build.rs`.  I'd expect most library/binary owners would rather not have their users need those dependencies to compile their project.

examples/build.rs
LucioFranco: can just add the ignore for that udeps thing since its just in the examples

examples/src/grpc-routeguide/generated/route_guide_grpc.pb.rs
LucioFranco: yeah checking in the generated code seems to be able to accelerate users here which I like as well

dfawley:
I've moved the code generation logic to the `codegen` binary as discussed (and updated CI to reuse the cache for the plugin, and fixed udeps).  PTAL.  A very minor nit about this approach is that the build-related `rerun-if` lines will be printed to stdout when running the `codegen` binary, because protobuf's rust codegen and our grpc-rust codegen don't have the magic that tonic has to avoid it:

https://github.com/hyperium/tonic/blob/b77a50691d291b837bbc67dcf73b4e6c09d62f0d/tonic-prost-build/src/lib.rs#L83

But I think this is totally fine for now, it just means you get a couple harmless lines of stdout.  I did play around with the other option of running `protoc` with the right flags, but ultimately decided against it once I got to the point where it was working but not formatting, since the code formatting happens outside of the protoc plugins (because those are written in C++, not Rust, so they can't).

LucioFranco (approved):

examples/src/grpc-routeguide/generated/crate_mapping.txt
arjan-bal, line 1: We don't need to check this in. It’s only required for code generation and isn't used at runtime.

codegen/src/main.rs
```diff
@@ -81,9 +81,37 @@ fn main() {
...
+fn codegen_grpc_routeguide(root_dir: &Path) {
+    let path = std::env::var("PATH").unwrap_or_default();
+    unsafe {
+        std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
+        // OUT_DIR must be set in CodeGen::new, so we set it correctly here
+        // instead of using `.output_dir()` on CodeGen.
+        std::env::set_var("OUT_DIR", root_dir.join("src/grpc-routeguide/generated"));
+    }
+
+    std::env::set_current_dir(root_dir).unwrap();
```
arjan-bal, line 105: I’m not following the issue here. Is this an area where the `grpc_protobuf_build` crate could be improved? Could you clarify the specific failure mode when using `output_dir()`?

dfawley: There's an `unwrap()` on reading the `OUT_DIR` env var. That isn't set when running this binary.

https://github.com/hyperium/tonic/blob/a92deef9d59b62b428d66520c5fb32b5f0244799/grpc-protobuf-build/src/lib.rs#L132

examples/src/grpc-routeguide/generated/crate_mapping.txt
dfawley: If I don't check it in, then I assume I need to specifically delete it in the codegen binary so it doesn't appear as a change?

codegen/src/main.rs
arjan-bal: `tonic-prost-build` delays the unwrap call until the compile call: https://github.com/hyperium/tonic/blob/a92deef9d59b62b428d66520c5fb32b5f0244799/tonic-prost-build/src/lib.rs#L749-L753

We could follow the same pattern in `grpc_protobuf_build` too. We can leave a TODO and handle this in a follow-up.

arjan-bal (approved):
LGTM

examples/src/grpc-routeguide/generated/crate_mapping.txt
arjan-bal: I see, we can leave it checked-in then.

codegen/src/main.rs
dfawley: Done

PR #2614, opened 2026-04-29
feat(build): Preserve error from ready() as Status source

jrose-signal:
Client APIs await `ready()` on their inner Service before sending anything, but currently any errors from `ready()` are flattened to a message on an `Unknown` Status. Changing to `Status::from_error` would be a more drastic change and lose the custom message, but we can at least save the original error as an `Error::source`.

This comes up when e.g. a connection is no longer accepting new requests; getting a plain `Unknown` back for that is rather inscrutable, and doubly so when not logging the message out of an abundance of caution around user data.

Still to do: add a test, because I wasn't sure where it should live. But I'm happy to write it! I'm also not sure if this qualifies as a semantically-breaking change that can't go into v0.14.x.

### DIFF
--- a/grpc/src/generated/grpc_examples_echo.rs
+++ b/grpc/src/generated/grpc_examples_echo.rs
@@ -101,9 +101,12 @@ pub mod echo_client {
                 .ready()
                 .await
                 .map_err(|e| {
-                    tonic::Status::unknown(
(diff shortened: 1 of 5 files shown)


jrose-signal:
Hm, the Clippy failure is my fault (and easy enough to fix), but the Windows one certainly isn't.

PR #2615, opened 2026-04-29
refactor(grpc): eliminate ResponseStreamItem in favor of ClientResponseStreamItem

sauravzg:
The generic `ResponseStreamItem<M>` was only used on the client side with `M = ()`. This change eliminates the generic type and redefines `ClientResponseStreamItem` as a non-generic enum to simplify the API.

As part of this change, the unit type `()` has been removed from the `Message` variant, simplifying `ClientResponseStreamItem::Message(())` to `ClientResponseStreamItem::Message`.

Fixed all occurrences and test breakages across the workspace:

- `grpc/src/core/mod.rs`
- `grpc/src/client/stream_util.rs`
- `grpc/src/inmemory/mod.rs`
- `grpc/src/client/transport/tonic/mod.rs`
- `grpc/src/client/transport/tonic/test.rs`
- `grpc-protobuf/src/client/mod.rs`
- `grpc/examples/inmemory.rs`

### DIFF
--- a/grpc-protobuf/src/client/mod.rs
+++ b/grpc-protobuf/src/client/mod.rs
@@ -139,7 +139,7 @@ where
         // 2. If we receive Trailers, we will only ever receive StreamClosed.
         match i {
             ClientResponseStreamItem::Headers(_) => unreachable!(),
-            ClientResponseStreamItem::Message(_) => Ok(()),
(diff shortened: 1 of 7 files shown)


dfawley (approved):
Thanks!

PR #2616, opened 2026-04-29
feat(tonic-xds): gRFC A29 cert validation building blocks

YutaoMa:
## Motivation

Ref: #2444

Continues the gRFC A29 (xDS-Based TLS Security) work in `tonic-xds`. The cert provider foundation merged in #2593 gives us a pluggable source of trust roots and identity. The next step toward end-to-end mTLS in the data plane is to parse the cluster's TLS config from xDS, validate server certs against the configured trust roots, and apply SAN matching on top of WebPKI chain validation.

This PR adds the building blocks for that validation. Integration with the
connector factory is deferred to a follow-up PR — see below.

## Solution

- **StringMatcher extraction** (`xds/resource/string_matcher.rs`): pulled out of `routing.rs` so it can be reused. The SAN matcher uses it for the string-comparison primitives.
- **UpstreamTlsContext parsing** (`xds/resource/security.rs`): parses `envoy.extensions.transport_sockets.tls.v3.UpstreamTlsContext` into a typed config.
- **SAN matcher** (`xds/resource/san_matcher.rs`): RFC 6125 wildcard DNS matching, RFC 5952 IP canonicalization, plus URI, EMAIL, and OTHER_NAME paths matching using the above extracted `StringMatcher`.
- **A29 `ServerCertVerifier`** (`xds/cert_provider/verifier.rs`): wraps `WebPkiServerVerifier` and applies SAN matching after standard chain validation. Sources trust roots from the existing `CertProviderRegistry`.

All new types are tested in isolation against synthetic certs.

### Why integration is deferred

The tonic-side hook for installing a custom rustls `ServerCertVerifier` on a `Channel` is still in flight as #2612. Without that API, the verifier can't be wired into the per-cluster connector from a downstream crate.


tonic-xds/src/xds/resource/security.rs
```diff
@@ -0,0 +1,492 @@
...
+            ctx.tls_certificate_certificate_provider_instance.is_some(),
+            "CommonTlsContext.tls_certificate_certificate_provider_instance",
+        )?;
+    }
+    Ok(())
+}
+
+/// Reject `CertificateValidationContext` fields whose silent absence would
+/// weaken peer verification relative to what the config requested.
+fn reject_unsupported_validation_fields(
```
LYZJU2019, line 198: Add `reject(ctx.custom_validator_config.is_some(), "custom_validator_config")` since it's also a NACK required field?

tonic-xds/src/xds/cert_provider/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+        .collect())
+}
+
+fn convert_general_name(gn: &GeneralName<'_>) -> Option<SanEntry> {
+    match gn {
+        GeneralName::DNSName(s) => Some(SanEntry::Dns(s.to_string())),
+        GeneralName::URI(s) => Some(SanEntry::Uri(s.to_string())),
+        GeneralName::RFC822Name(s) => Some(SanEntry::Email(s.to_string())),
+        GeneralName::IPAddress(bytes) => parse_ip_san(bytes).map(SanEntry::IpAddress),
+        GeneralName::OtherName(oid, value) => Some(SanEntry::OtherName {
```
LYZJU2019: should be something like:

```
                                                                                                                                               
  GeneralName::OtherName(oid, raw) => {
      // raw = `[0] EXPLICIT ANY` DER. Pull out the inner Any, then take its
      // content bytes — typically a UTF8String / IA5String / PrintableString.                                                                 
      let (_, outer) = der_parser::ber::parse_ber(raw).ok()?;                                                                                  
      let inner = outer.as_slice().ok()?;          // contents of [0] EXPLICIT                                                                 
      let (_, any) = der_parser::ber::parse_ber(inner).ok()?;                                                                                  
      let value = any.as_slice().ok()?.to_vec();   // inner string bytes                                                                       
      Some(SanEntry::OtherName { oid: oid.to_id_string(), value })                                                                             
  }
```

tonic-xds/src/xds/resource/security.rs
```diff
@@ -0,0 +1,492 @@
...
+    )?;
+    reject(
+        !ctx.verify_certificate_hash.is_empty(),
+        "verify_certificate_hash",
+    )?;
+    reject(
+        ctx.require_signed_certificate_timestamp.is_some(),
+        "require_signed_certificate_timestamp",
+    )?;
+    reject(ctx.crl.is_some(), "crl")?;
```
ankurmittal, line 213: do we need to reject custom_validator_config ?

tonic-xds/src/xds/cert_provider/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+impl ServerCertVerifier for XdsServerCertVerifier {
+    fn verify_server_cert(
+        &self,
+        end_entity: &CertificateDer<'_>,
+        intermediates: &[CertificateDer<'_>],
+        server_name: &ServerName<'_>,
+        ocsp_response: &[u8],
+        now: UnixTime,
+    ) -> Result<ServerCertVerified, RustlsError> {
+        self.inner.verify_server_cert(
```
ankurmittal: will this support SPIFFE-only cert ?

tonic-xds/src/xds/resource/security.rs
```diff
@@ -0,0 +1,492 @@
...
+        .map(|m| StringMatcher::from_proto(m).map(SanMatcher::Dns))
+        .collect()
+}
+
+/// Reject `CommonTlsContext` fields we don't support to avoid silently
+/// degrading security relative to what the config requested.
+fn reject_unsupported_common_fields(ctx: &CommonTlsContext) -> xds_client::Result<()> {
+    reject(ctx.tls_params.is_some(), "CommonTlsContext.tls_params")?;
+    reject(
+        !ctx.tls_certificates.is_empty(),
```
ankurmittal: spec says


> If the [tls_certificate_provider_instance](https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L247) field is unset but either of the [tls_certificates](https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L226) or [tls_certificate_sds_secret_configs](https://github.com/envoyproxy/envoy/blob/b29d6543e7568a8a3e772c7909a1daa182acc670/api/envoy/extensions/transport_sockets/tls/v3/tls.proto#L239) fields are set, gRPC will NACK the CDS update.


do we follow it here? I mean do we check if `tls_certificate_provider_instance` is unset?

tonic-xds/src/xds/resource/san_matcher.rs
```diff
@@ -0,0 +1,332 @@
...
+    };
+    let Some(label) = stripped.and_then(|s| s.strip_suffix('.')) else {
+        return false;
+    };
+    !label.is_empty() && !label.contains('.')
+}
+
+fn strip_suffix_ignore_ascii_case<'a>(s: &'a str, suffix: &str) -> Option<&'a str> {
+    let split = s.len().checked_sub(suffix.len())?;
+    let (head, tail) = s.split_at(split);
```
ankurmittal: Panics if mid is not on a UTF-8 code point boundary, or if it is past the end of the last code point of the string slice. as value comes from cert, can this be a attack vector for causing panic?

tonic-xds/src/xds/resource/string_matcher.rs
```diff
@@ -0,0 +1,207 @@
...
+            Self::Exact { value, ignore_case } => {
+                if *ignore_case {
+                    v.eq_ignore_ascii_case(value)
+                } else {
+                    v == value
+                }
+            }
+            Self::Prefix { value, ignore_case } => {
+                if *ignore_case {
+                    v.to_ascii_lowercase()
```
ankurmittal: to_ascii_lowercase will cause allocations, maybe try slice and .eq_ignore_ascii_case with len checks first

tonic-xds/src/xds/cert_provider/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+        )?;
+
+        // A29 SAN matching uses "any" semantics: at least one matcher must match
+        // some SAN entry. Empty matcher list means CA-trust-only authorization.
+        if !self.san_matchers.is_empty() {
+            let sans = extract_sans(end_entity)
+                .map_err(|e| RustlsError::General(format!("failed to extract SANs: {e}")))?;
+            if !self.san_matchers.iter().any(|m| m.matches_any(&sans)) {
+                return Err(RustlsError::General(
+                    "no SAN matcher matched the presented certificate".into(),
```
ankurmittal: is there no Error for this? do we need custom strings?

tonic-xds/src/xds/cert_provider/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+    match gn {
+        GeneralName::DNSName(s) => Some(SanEntry::Dns(s.to_string())),
+        GeneralName::URI(s) => Some(SanEntry::Uri(s.to_string())),
+        GeneralName::RFC822Name(s) => Some(SanEntry::Email(s.to_string())),
+        GeneralName::IPAddress(bytes) => parse_ip_san(bytes).map(SanEntry::IpAddress),
+        GeneralName::OtherName(oid, value) => Some(SanEntry::OtherName {
+            oid: oid.to_id_string(),
+            value: value.to_vec(),
+        }),
+        _ => None,
```
gu0keno0, line 163: Should we pass it along to SanEntry as a "Unspecified" variant? I'm not familiar with the other variants, yet I feel maybe there is a chance that the custom verifier may need to validate against them?

tonic-xds/src/xds/cert_provider/verifier.rs
```diff
@@ -0,0 +1,355 @@
...
+    }
+}
+
+fn parse_ip_san(bytes: &[u8]) -> Option<IpAddr> {
+    match bytes.len() {
+        4 => <[u8; 4]>::try_from(bytes).ok().map(IpAddr::from),
+        16 => <[u8; 16]>::try_from(bytes).ok().map(IpAddr::from),
+        _ => None,
+    }
+}
```
gu0keno0, line 177: According to https://github.com/rusticata/x509-parser/blob/b7dcc9397b596cf9fa3df65115c3f405f1748b2a/src/extensions/mod.rs#L824 , ip addr can also be 8 bytes or 32 bytes because of network masks.

gu0keno0 (commented):
Looks good overall, put a few comments for details.

tonic-xds/src/xds/resource/security.rs
YutaoMa: good catch, added `custom_validator_config` case.

tonic-xds/src/xds/resource/security.rs
YutaoMa: good catch, added `custom_validator_config` case.

tonic-xds/src/xds/resource/string_matcher.rs
YutaoMa: good point, created starts_with/ends_with/contains_ignore_ascii_case functions for this

tonic-xds/src/xds/resource/security.rs
YutaoMa: 👍 , moved the check to be guarded by `tls_certificate_provider_instance` check first.

tonic-xds/src/xds/resource/san_matcher.rs
YutaoMa: valid point, switched to `split_at_checked` and propagate the `None`, safe to do so since the code point error imply no match.

tonic-xds/src/xds/cert_provider/verifier.rs
YutaoMa: switched to use `CertificateError:ApplicationVerificationFailure` which now makes rustls send `AccessDenied (49)` on the wire properly.

tonic-xds/src/xds/cert_provider/verifier.rs
YutaoMa: yeah if we want to properly parse `OtherName` in SAN this will be needed. To avoid this complexity, since A29 only defined support for the above 4 SAN types (DNS, URI, Email, IP), I've removed all logic related to `OtherName` and rejects it at config time instead.

tonic-xds/src/xds/cert_provider/verifier.rs
YutaoMa: Related to the above comment, no SAN matching behavior of other variants is defined in A29 and other gRPC impls (grpc-go, grpc-java) or skip those types. So I'll keep the `None` case here and also dropping the `OtherName` variant as well. Updated comments to reflect this.

tonic-xds/src/xds/cert_provider/verifier.rs
YutaoMa: according to RFC 5280 4.2.1.6, IP SAN is 4/16 bytes. The 8/32 bytes is for CA nameConstraints which happens to be represented with the same extensions type here in `x509-parser`. I've updated the comments here to reflect this.

tonic-xds/src/xds/cert_provider/verifier.rs
YutaoMa: Thanks for pointing this out, actually the hostname-checking nature of the `WebPkiServerVerifier` would cause SPIFFE cert to be rejected always. I refactored the verifier to only use `rustls`'s chain verification impl and now all SAN checking is handled by our xDS matcher logic, allowing URI-only SPIFFE certs use case.

gu0keno0 (approved):

gu0keno0 (approved):

PR #2617, opened 2026-04-29
feat(tonic-xds): node metadata & AddOrigin layer

YutaoMa:
## Motivation

tonic-xds doesn't currently interoperate with Istio's istiod. Two specific gaps surfaced during local Istio testing, and both affect any other real xDS control plane with the same shape:

1. **Node metadata not sent.** istiod gates its proxyless-gRPC-style config behind `node.metadata.GENERATOR = "grpc"` (gRFC A27). Without it, istiod serves sidecar-style config that doesn't fit.
2. **Outgoing requests carry no `:authority`.** The routing layer falls through to wildcard matches, so any control plane returning non-wildcard virtual hosts (Istio's `*.svc.cluster.local`, FQDN-based domains in general) doesn't route correctly. tonic's per-endpoint internal `AddOrigin` runs after LB selection (keyed by endpoint IP:port, not service authority), so it can't substitute at the routing layer.

## Solution

- **Node.metadata field & Bootstrap propagation** (`tonic-xds/src/xds/bootstrap.rs`) —
  `NodeConfig` parses `node.metadata` from the bootstrap JSON;
  `From<NodeConfig> for Node` carries it through.
- **`AddOriginLayer`** (`tonic-xds/src/client/add_origin.rs`) — a tower
  service that rewrites each request's scheme + authority before
  forwarding. Wired as the outermost layer in the xDS service stack, with
  authority taken from the xDS target. Falls back to a no-op (with
  `tracing::warn`) if the target isn't a valid authority. (`grpc-go` does the equivalent in `clientconn.go::initAuthority`. )


tonic-xds/src/client/add_origin.rs
```diff
@@ -0,0 +1,166 @@
+//! Sets the URI scheme and authority on outgoing requests so the routing
+//! layer can match against virtual hosts.
+//!
+//! When a tonic-generated client calls into the xDS channel, the outgoing
+//! request URI is typically just the gRPC method path with no `:authority`.
+//! The routing layer downstream matches `:authority` against the virtual
```
gu0keno0: This might have an effect on TLS? Could you verify if this causes TLS verification failure?

I guess maybe it's fine for now because as we discussed in another PR for A29, Tonic does not use :authority as SNI to validate against server's certs?

tonic-xds/src/client/add_origin.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /// Build an `AddOriginLayer` for the given authority string, defaulting
+    /// to the HTTP scheme. Returns `None` (with a `tracing::warn`) if the
+    /// string is not a valid URI authority — outgoing requests will then
+    /// carry no `:authority` and non-wildcard virtual-host matching will
+    /// fail at the routing layer.
+    pub(crate) fn for_authority(authority: &str) -> Option<Self> {
+        match Authority::try_from(authority) {
+            Ok(parsed) => Some(Self::new(Scheme::HTTP, parsed)),
```
gu0keno0: Does Scheme affect routing works?

Asking because the code cannot decide whether it is HTTP or HTTPS by just the xDS target, it can only be decided when cluster is resolved. So physically Scheme cannot be determined here.

gu0keno0 (commented):
a few questions, generally looks good to me

tonic-xds/src/client/add_origin.rs
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    fn call(&mut self, req: Request<B>) -> Self::Future {
+        // Mirrors tonic's per-endpoint AddOrigin layer
+        // (channel/service/add_origin.rs). The Scheme/Authority clones are
+        // refcount bumps on internal Bytes — no string copies.
+        let (mut head, body) = req.into_parts();
+        head.uri = {
+            let mut parts: Parts = head.uri.into();
+            parts.scheme = Some(self.scheme.clone());
```
gu0keno0: should we just use head.uri.scheme?

tonic-xds/src/client/add_origin.rs
YutaoMa: xDS gRFCs don't use scheme for routing.

The HTTP value here is mainly a placeholder needed because `Uri::from_parts` fails when `authority` is set but `scheme` is empty.

I can switch to use request's `head.uri.scheme`, but technically becuase all requests in tonic grpc channel have empty `scheme`, we'll need to have a fallback of adding `HTTP` and for all of the gprc requests this will be the actual code path it executes.

tonic-xds/src/client/add_origin.rs
YutaoMa: see above comment, we should be good to use head.uri.scheme but the actual code path will require a default value nonetheless.

tonic-xds/src/client/add_origin.rs
YutaoMa: Update: after further analysis the problem, I identified the better solution is to not use this `:authority` pseudo header on request and instead just wires `authority` as a field on LB layer to be accessed for routing. See the latest commit. This should address both concern on whether we'll accidentally break TLS and the mandatory scheme placeholder.

tonic-xds/src/client/add_origin.rs
YutaoMa: Update: per update in above comment, switched to instead plumbing the authority through a dedicated field on routing layer to avoid messing with the TLS stuff

AvrahamO:
I tested this PR on our istio production environment, and it throws the following error:
```Bootstrap(InvalidJson(Error("invalid type: map, expected a string", line: 18, column: 21)))```

This is because `.metadata` contains several different type.
Here is a sample metadata object:
```
{
  "xds_servers": [
    {
      "server_uri": "unix:///etc/istio/proxy/XDS",
      "channel_creds": [
        {
          "type": "insecure"
        }
      ],
      "server_features": [
        "xds_v3"
      ]
    }
  ],
  "node": {
    "id": "test",
    "metadata": {
      "ANNOTATIONS": {
        "inject.istio.io/templates": "grpc-agent",
        "istio.io/rev": "default",
      },
      "CLUSTER_ID": "Kubernetes",
      "ENVOY_PROMETHEUS_PORT": 15090,
      "ENVOY_SKIP_DEPRECATED_LOGS": "true",
      "ENVOY_STATUS_PORT": 15021,
      "LABELS": {
        "app": "my-app",
        "service.istio.io/canonical-name": "my-app",
        "service.istio.io/canonical-revision": "latest",
      },
      "PILOT_SAN": [
        "istiod.istio-system.svc"
      ]
    }
  }
}
```

**Comment by AvrahamO** [comment]
We also need to support UDS in `server_uri`, as this is what istio grpc-agent puts in bootstrap file (see above example)
It's just a one-line change in [xds-client/src/transport/tonic.rs](https://github.com/hyperium/tonic/blob/master/xds-client/src/transport/tonic.rs#L177)
```
-        let endpoint = Channel::from_shared(server.uri().to_string())
+        let endpoint = Endpoint::from_shared(server.uri().to_string())
```

Thank you

**Comment by YutaoMa** [comment]
@AvrahamO thanks for validating this with a production istio environment, I only have access to a test one so this really helps a lot. I've validated your comments and suggestions and implemented the change in the latest commit. Feel free to test against it again.

**Comment by AvrahamO** [comment]
@YutaoMa thank you for the quick fix. I tested again on production and verified it works as expected.

File: tonic-xds/src/client/route.rs [hunk 4]
```diff
@@ -95,6 +101,73 @@ impl<S> Layer<S> for XdsRoutingLayer {
...
+        assert_eq!(
+            captured.lock().unwrap().as_deref(),
+            Some("greeter.svc:50051"),
+        );
+
+        // Case 2: request with a different authority on the URI — the layer
+        // must still use its own configured authority.
+        *captured.lock().unwrap() = None;
+        let req = Request::builder()
+            .uri("http://other.example:443/pkg.Greeter/SayHello")
```
gu0keno0, line 164: In principle this cannot happen when using our client?

tonic-xds/src/xds/bootstrap.rs
```diff
@@ -123,12 +123,32 @@ pub(crate) struct NodeConfig {
...
+}
+
+/// Convert a `serde_json::Value` to the codec-agnostic [`MetadataValue`].
+fn json_to_metadata(value: serde_json::Value) -> MetadataValue {
+    match value {
+        serde_json::Value::Null => MetadataValue::Null,
+        serde_json::Value::Bool(b) => MetadataValue::Bool(b),
+        // `as_f64` is documented to return `Some` for any `Number` parsed
+        // from valid JSON; `unwrap_or_default` is a defensive fallback.
+        serde_json::Value::Number(n) => MetadataValue::Number(n.as_f64().unwrap_or_default()),
```
gu0keno0: Since this is for bootstrap, shall we just be strict and return error when conversion fails?

gu0keno0 (approved):

tonic-xds/src/client/route.rs
YutaoMa: right, when using xDS channel user don't get to modify the `request::Uri`.

tonic-xds/src/xds/bootstrap.rs
YutaoMa: makes sense, I've refactored this to emit `BoostrapError` instead.

PR #2618, opened 2026-04-30
fix(status): prevent panic on invalid base64 in grpc-status-details-bin

sauravzg:
Instead of panicking with `.expect()` when base64 decoding of the `grpc-status-details-bin` header fails, fall back to setting the status code to `Unknown` and include the decoding error in the message. This is consistent with how Tonic handles failures to decode the `grpc-message` header.

Added a unit test `details_invalid_base64` to verify this behavior.

** I am not entirely sure if this panic was an intentional choide or this change counts as a breaking change, but I'd assume that not panicking would be considered good behavior**

### DIFF
--- a/tonic/src/status.rs
+++ b/tonic/src/status.rs
@@ -473,12 +473,11 @@ impl Status {
             None => Ok(String::new()),
         };
 
-        let details = match header_map.get(Self::GRPC_STATUS_DETAILS) {
+        let error_details = match header_map.get(Self::GRPC_STATUS_DETAILS) {
             Some(header) => crate::util::base64::STANDARD
                 .decode(header.as_bytes())
-                .expect("Invalid status header, expected base64 encoded value")
-                .into(),
-            None => Bytes::new(),
+                .map(Bytes::from),
+            None => Ok(Bytes::new()),
         };
 
         let other_headers = {
@@ -489,12 +488,17 @@ impl Status {
             header_map
         };
 
-        let (code, message) = match error_message {
-            Ok(message) => (code, message),
-            Err(e) => {
+        let (code, message, details) = match (error_message, error_details) {
+            (Ok(msg), Ok(det)) => (code, msg, det),
+            (Err(e), _) => {
                 let error_message = format!("Error deserializing status message header: {e}");
                 warn!(error_message);
-                (Code::Unknown, error_message)
+                (Code::Unknown, error_message, Bytes::new())
+            }
+            (Ok(_msg), Err(e)) => {
+                let error_message = format!("Error deserializing status details header: {e}");
+                warn!(error_message);
+                (Code::Unknown, error_message, Bytes::new())
             }
         };
 
@@ -1076,6 +1080,26 @@ mod tests {
         let status = Status::from_header_map(&header_map).unwrap();
 
         assert_eq!(status.details(), DETAILS);
+    }
+
+    #[test]
+    fn details_invalid_base64() {
+        let mut header_map = HeaderMap::new();
+        header_map.insert(Status::GRPC_STATUS, HeaderValue::from_static("14")); // Unavailable
+        header_map.insert(
+            Status::GRPC_STATUS_DETAILS,
+            HeaderValue::from_static("invalid base64!"),
+        );
+
+        let status = Status::from_header_map(&header_map).unwrap();
+
+        assert_eq!(status.code(), Code::Unknown);
+        assert!(
+            status
+                .message()
+                .contains("Error deserializing status details header")
+        );
+        assert!(status.details().is_empty());
     }
 }
 


tonic/src/status.rs
```diff
@@ -489,12 +488,17 @@ impl Status {
...
                 let error_message = format!("Error deserializing status message header: {e}");
                 warn!(error_message);
-                (Code::Unknown, error_message)
+                (Code::Unknown, error_message, Bytes::new())
+            }
+            (Ok(_msg), Err(e)) => {
+                let error_message = format!("Error deserializing status details header: {e}");
+                warn!(error_message);
+                (Code::Unknown, error_message, Bytes::new())
             }
```
arjan-bal, line 502: In gRPC Go, if decoding binary metadata fails, `Code::Internal` is returned (see [here](https://github.com/grpc/grpc-go/blob/850cc543aa3c5b86b82697b91f3ec5a7210829a0/internal/transport/http2_client.go#L1522-L1527) and [here](https://github.com/grpc/grpc-go/blob/850cc543aa3c5b86b82697b91f3ec5a7210829a0/internal/transport/http2_client.go#L1582-L1586)). According to Gemini, Java and C++ also have the same behavior, though I haven't verified this. We should probably return the same status for consistency.

sauravzg: Fixed. It was set to "unknown" for consistency with tonic. Changed it to "internal" for consistency with other grpc impls. 
Haven't changed the "unkown" for tonic's existing behavior.

tonic/src/status.rs
```diff
@@ -473,12 +473,11 @@ impl Status {
             None => Ok(String::new()),
         };
 
-        let details = match header_map.get(Self::GRPC_STATUS_DETAILS) {
+        let details_result = match header_map.get(Self::GRPC_STATUS_DETAILS) {
```
dfawley: Note that `error_message` is also actually a `Result`.  Should we rename it to `message_result` for clarity and symmetry?

tonic/src/status.rs
```diff
@@ -489,12 +488,17 @@ impl Status {
             header_map
         };
 
-        let (code, message) = match error_message {
-            Ok(message) => (code, message),
-            Err(e) => {
+        let (code, message, details) = match (error_message, details_result) {
+            (Ok(msg), Ok(det)) => (code, msg, det),
+            (Err(e), _) => {
                 let error_message = format!("Error deserializing status message header: {e}");
```
dfawley, line 494: Note that Go doesn't consider it an error that changes the status code in this case.  I'm not sure about Java/C++/the spec.

tonic/src/status.rs
sauravzg: I've actually renamed `details_result` -> `error_details` , since adding type suffix for errors to a statically typed language might not be the best thing(at least the c++ style guide discourages it )

tonic/src/status.rs
sauravzg: For now, I've decided to keep things consistent with how tonic handles it after some discussion with arjan on it i.e. "unknown error" which is what tonic does for incorrect error message.

I'll loop lucio in to the review and get his opinion here, keeping the comment open.

LucioFranco: Ah okay I see I think its fine to keep the prev status code I don't think anyone is depending on this behavior and I think this is prob just an impl gap. So happy to have this align with what other grpc crates do.

sauravzg: So, it seems that this may not be consistent across languages.
- Go : https://github.com/grpc/grpc-go/blob/1c63fa5f5492fb6cb4cabf9847999ce469505f49/internal/status/status.go#L54-L58  - Seems to just ignore the errors here
- Java: https://github.com/grpc/grpc-java/blob/0248e6f75605ac5007fa71c2a62e4f325a61d32d/protobuf-lite/src/main/java/io/grpc/protobuf/lite/ProtoLiteUtils.java#L271-L278 ,  https://github.com/grpc/grpc-java/blob/0248e6f75605ac5007fa71c2a62e4f325a61d32d/protobuf/src/main/java/io/grpc/protobuf/StatusProto.java#L180 indicate that it just gets bubbled up to application as runtime exception.
- C++ : This was more difficult to find so I asked   gemini, which seems to think it ends up returning internal error.

Anyways, given the lack of consistency of error handling ,  personally lean towards keeping the same behavior as "grpc message" for tonic , but I don't mind changing it to something else. @dfawley  LMK what you'd like here.

dfawley: Sure, keeping the tonic behavior for now is fine, but we should figure out the "right" behavior, too.  I think C++ (or gemini's version of c++) is most likely what we want, but I'm not sure.

dfawley (approved):

tonic/src/status.rs
```diff
@@ -489,12 +488,17 @@ impl Status {
...
-            Err(e) => {
+        let (code, message, details) = match (error_message, error_details) {
+            (Ok(msg), Ok(det)) => (code, msg, det),
+            (Err(e), _) => {
                 let error_message = format!("Error deserializing status message header: {e}");
                 warn!(error_message);
-                (Code::Unknown, error_message)
+                (Code::Unknown, error_message, Bytes::new())
+            }
+            (Ok(_msg), Err(e)) => {
```
LucioFranco, line 498: do we want to do anything with the message instead of throwing it away?

sauravzg: So, that's probably a behavior we should define. Unfortunately, that's not consistently defined in grpc as one of the other comment discusses.

- This approach treats invalid status details as a header and returns a completely different error unrelated to the RPC error. (This might be similar to what gemini thinks c++ behavior is)
- The other approach would be to do something similar to go, where this is treated as a silent failure and we simply return original code and message.

I wouldn't want to change the error code but reuse the original error message.

tonic/src/status.rs
dfawley: Looking at the places in Go that you pointed to, I think we're talking about different things.

This error is about failing to percent-decode the `grpc-message` header, whereas you seem to be looking at the `grpc-status-details-bin` proto.

In Go we just do our best to percent-decode, ignoring any errors while emitting, e.g. `%yq` comes out verbatim if the header contains `%yq`, since it's a not valid percent-encoding.  Or `%%2F` would emit `%/`.    And we leave the error code the same.  It looks like tonic will convert the error into `Unknown` and modify the strong to include the prefix of percent-decoding failing.  The code in Go for this is here:

https://github.com/grpc/grpc-go/blob/bb023f846c01c094fede9bfe591ef5c5369e59ca/internal/transport/http2_client.go#L1515

https://github.com/grpc/grpc-go/blob/bb023f846c01c094fede9bfe591ef5c5369e59ca/internal/transport/http_util.go#L268-L299

tonic/src/status.rs
dfawley: Ultimately I don't think we want to be doing anything with grpc-status-details-bin at this low of a level.  That requires a protobuf dependency to parse it.  The generated code should deal with this trailer somehow instead.

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#appendix-a---grpc-for-protobuf

(This clause only requires the use of a `google.rpc.Status` proto message in this header when you are using protobuf with gRPC.  gRPC can transmit messages of any type.)

tonic/src/status.rs
sauravzg: > https://github.com/grpc/grpc-go/blob/bb023f846c01c094fede9bfe591ef5c5369e59ca/internal/transport/http2_client.go#L1515
> 
> https://github.com/grpc/grpc-go/blob/bb023f846c01c094fede9bfe591ef5c5369e59ca/internal/transport/http_util.go#L268-L299
> 
> 

The code in go that you linked to talks about `grpc-message` field. For this PR I am mostly concerned with "if we don't crash on invalid b64 for status details, what's the behavior that we'd like" . 

> It looks like tonic will convert the error into Unknown and modify the strong to include the prefix of percent-decoding failing. 

Doesn't tonic simply return the utf error from `decode` with a new message?

tonic/src/status.rs
sauravzg: I don't think there's a protobuf dep here in tonic, tonic simply returns the serialized bin.

dfawley: > I don't think there's a protobuf dep here in tonic, tonic simply returns the serialized bin.

OK that's good..  So this is just a base64 decoding error on the `grpc-status-details-bin` header.  I don't think we want to treat this any differently than decoding any other `-bin` header, ideally.  In Go at least, any -bin header decoding error leads to `Internal` + "malformed header <name>: <base64 decoding error info>":

https://github.com/grpc/grpc-go/blob/609310837bbc7fab1553fa53f2d1312bd7d85275/internal/transport/http2_client.go#L1522-L1527
and
https://github.com/grpc/grpc-go/blob/609310837bbc7fab1553fa53f2d1312bd7d85275/internal/transport/http2_client.go#L1582-L1585

I guess in Tonic the other decoding happens only when the application accesses it from the metadata?  But this one happens specially in the transport?  If that's true there's probably not much more we can do here.  I think the behavior for this header roughly matches Go's behavior (though we call it INTERNAL -- https://github.com/grpc/grpc/blob/master/doc/statuscodes.md doesn't specify -- "could not decompress" is INTERNAL, which seems closest?).  But I don't think this exact status code is too important for now considering it isn't well defined.  So I'd be fine with leaving it alone.

tonic/src/status.rs
dfawley: > The code in go that you linked to talks about grpc-message field. For this PR I am mostly concerned with "if we don't crash on invalid b64 for status details, what's the behavior that we'd like" .

I get that, but this comment is on the part of the code where Tonic is changing the status code when the `grpc-message` header fails to parse.  I believe that behavior is incorrect.  It should keep the status code the same instead of converting it to UNKNOWN.  I'll try to confirm with the other languages today.

PR #2619, opened 2026-04-30
feat(tonic-xds): implement gRFC A50 outlier detection (failure-percentage) + LoadBalancer integration

LYZJU2019:
## Summary

Implements [gRFC A50: xDS Outlier Detection](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) (failure-percentage algorithm) in `tonic-xds` and integrates it into `LoadBalancer`. Config types landed in #2604.

## Architecture

- **Data path** (`OutlierStatsRegistry::record_outcome`, called from `LoadBalancer::call`) — increments the picked channel's success/failure counter. Nothing else.
- **Sweep** (`OutlierStatsRegistry::run_housekeeping`, called by the housekeeping actor on each `config.interval` tick) — snapshots all channel counters, runs the failure-percentage algorithm against the snapshot population (applying `minimum_hosts`, `max_ejection_percent`, threshold, enforcement roll), dispatches eject addresses on an mpsc, then resets counters and decrements multipliers for non-ejected channels.
- **Load balancer** — drains the eject mpsc in `poll_ready`, ejects via `ReadyChannel::eject`, tracks the resulting `EjectedChannel` in `KeyedFutures<_, UnejectedChannel<_>>`. The picker only sees `ready`, so ejected channels are unpickable by construction.

Un-ejection is timer-driven per channel: each `EjectedChannel`'s `Sleep` fires at `min(base × multiplier, max(base, max_ejection_time))` and yields an `UnejectedChannel`; the LB routes the resolved channel back to `ready`.

## Constructor interface

`LoadBalancer::new` takes `Arc<ArcSwap<OutlierDetectionConfig>>`. `OutlierDetectionConfig::default()` is the disabled config (both algorithms `None`) — no actor spawned, `record_outcome` short-circuits at the per-state counter increment. The `ArcSwap` shape reserves the slot for the future xDS-driven config-update path.

## A50 compliance

- Algorithm runs at the interval sweep, not per RPC (§6).
- Failure-percentage uses strict `>` against the threshold.
- Multiplier decrements at the same transition that un-ejects (§6.b).
- `max_ejection_percent` floors at 1 for non-empty pools (spec: "will eject at least one address regardless of the value").
- Outlier state survives `Change::Insert` for an already-tracked address.

## Deferred

- Success-rate algorithm.
- Live config-update plumbing (`ArcSwap::store` is supported but the actor doesn't observe swaps yet).
- Wiring from `ClusterResource` into LB construction.

## Test plan

- [x] `cargo test -p tonic-xds --lib --all-features` — 324 lib tests pass
- [x] `cargo fmt -p tonic-xds`
- [x] `cargo clippy` clean on changed files


tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+/// state snapshots), so the consumer must process every event in order
+/// to stay in sync with the detector. We therefore can't drop or
+/// coalesce — but we don't want unbounded growth either if the consumer
+/// stalls. With sweep cadence on the order of seconds and at most
+/// `2 * num_endpoints` decisions per sweep, 256 buffers several sweeps'
+/// worth of decisions for clusters of typical size. When the buffer
+/// fills, `tx.send().await` parks the sweep task, which naturally
+/// throttles sweep cadence to whatever rate the consumer can drain —
+/// the right behavior, since computing more decisions than the consumer
+/// can apply just widens the desync.
```
YutaoMa: need to summarize these comments in a user/developer facing way. Don't leave LLM thinking context like "the right behavior" in here.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+}
+
+/// Probability source for `enforcing_*` rolls. Abstracted so tests can
+/// inject deterministic outcomes.
+pub(crate) trait Rng: Send + Sync + 'static {
+    /// Return a uniform random `u32` in `0..100`.
+    fn pct_roll(&self) -> u32;
+}
+
+/// Default RNG backed by `fastrand` (already a workspace dep).
```
YutaoMa: comment again, remove the `already a workspace dep` part

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+impl EndpointCounters {
+    pub(crate) fn record_success(&self) {
+        self.success.fetch_add(1, Ordering::Relaxed);
+    }
+
+    pub(crate) fn record_failure(&self) {
+        self.failure.fetch_add(1, Ordering::Relaxed);
+    }
+
+    /// Atomically read and zero both counters. Returns `(success, failure)`.
```
YutaoMa: these two swaps are not atomic in terms of the pairs' execution, so the snapshot comes with a possible gap of more failure comes in between line 76 and 77. depends on how strict the spec is regarding this calculation we may need to switch to a truly atomic calculation on both

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+
+        let threshold = u64::from(cfg.threshold.get());
+        for c in qualifying {
+            if *budget == 0 {
+                break;
+            }
+            // failure_pct = 100 * failure / total. A50 specifies a strict
+            // "greater than" comparison: an address sitting exactly at
+            // the threshold is not ejected.
+            let failure_pct = 100 * c.failure / c.total;
```
YutaoMa: need to handle divide by zero case to avoid panic

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+}
+
+/// Background task: runs `detector.run_sweep` on each interval tick and
+/// forwards each decision on the channel. The task ends (and `tx` is
+/// dropped, closing the receiver) when [`AbortOnDrop`] is dropped or
+/// when the receiver itself is dropped.
+///
+/// `tx.send().await` is fallible (returns `Err` if the receiver was
+/// dropped) and may park briefly when the channel is full — see
+/// [`DECISIONS_CHANNEL_CAPACITY`].
```
YutaoMa: the const is named `DEFAULT_DECISIONS_CHANNEL_CAPACITY`

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+            config,
+            state: Mutex::new(HashMap::new()),
+            rng,
+        })
+    }
+
+    /// Sort a decision list deterministically so equality checks can rely
+    /// on a canonical order without coupling to `HashMap` iteration order.
+    fn sort(mut ds: Vec<EjectionDecision>) -> Vec<EjectionDecision> {
+        ds.sort_by(|a, b| format!("{a:?}").cmp(&format!("{b:?}")));
```
YutaoMa: relying on `Debug` for sort behavior is not ideal, better to impl/derive `Ord/PartialOrd`

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+        let total_endpoints = state.len();
+        let max_ejections = (total_endpoints as u64
+            * u64::from(self.config.max_ejection_percent.get())
+            / 100) as usize;
+        let already_ejected = state.values().filter(|ep| ep.ejected_at.is_some()).count();
+        let mut budget = max_ejections.saturating_sub(already_ejected);
+
+        // Steps 3 & 4: run the algorithms on the snapshot. Hosts that are
+        // currently ejected naturally fail the `request_volume` gate
+        // because they receive no traffic in production, so iterating
```
YutaoMa: but there can be a case where hosts that got ejected still have in-flight traffic, so re-ejection happens on already ejected hosts, causing double count in budget decrement. Better to add some logic to remove this double counting code path for already ejected endpoint.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,862 @@
...
+    /// sweep task on the current Tokio runtime. The sweep runs every
+    /// `config.interval` until the returned [`AbortOnDrop`] is dropped.
+    pub(crate) fn spawn(
+        config: OutlierDetectionConfig,
+    ) -> (Arc<Self>, mpsc::Receiver<EjectionDecision>, AbortOnDrop) {
+        Self::spawn_with(config, OutlierDetectorOptions::default())
+    }
+
+    /// Variant of [`Self::spawn`] that accepts custom runtime options.
+    pub(crate) fn spawn_with(
```
gu0keno0: What about we just don't use these spawned loops at all, and just using RPCs as ticks for the outlier detection window? You can still use wallclock time, just that instead of ticking every second, tick it with every RPC call.

This works because tonic-xds only need to deal with all ejection signals in loadbalancer.poll_ready(), which already take mut self of loadbalancer, the decision of outlier detection will be global and without race conditions.

YutaoMa (approved):

ankurmittal (approved):

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,852 @@
...
+/// successes, low 32 bits: failures), so each record is one `fetch_add`
+/// and a snapshot is one `swap(0)`. Each counter is capped at
+/// `u32::MAX` per sweep interval; exceeding that carries into the
+/// other counter's bits but is unreachable for realistic workloads.
+#[derive(Debug, Default)]
+pub(crate) struct EndpointCounters {
+    /// High 32 bits: successes since last sweep.
+    /// Low 32 bits: failures since last sweep.
+    packed: AtomicU64,
+}
```
gu0keno0: \o/ Let's avoid such low-level dances and prioritize human readability. Packing them in the same atomic does not add much value because in reality, concurrent calls can succeed / fail before they bump up the packed counter. Forcing a snapshot view of the world by packing the counters together does not give us much.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,852 @@
...
+        (v >> 32, v & FAILURE_MASK)
+    }
+}
+
+/// A decision emitted by an [`OutlierDetector`] sweep.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) enum EjectionDecision {
+    /// Eject this endpoint from the load-balancing pool. The caller
+    /// should keep its underlying connection alive (A50 requires
+    /// preserving connections across ejection).
```
gu0keno0: We probably should extend A50 and allow the transition from ejection to connecting. Yet we can discuss this more offline in the future. For now it's fine.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,852 @@
...
+    Eject(EndpointAddress),
+    /// Restore a previously-ejected endpoint to the pool.
+    Uneject(EndpointAddress),
+}
+
+/// Probability source for `enforcing_*` rolls.
+pub(crate) trait Rng: Send + Sync + 'static {
```
gu0keno0: Do wo really need this abstraction? Can we just remove this trait?

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,852 @@
...
+}
+
+/// gRFC A50 outlier detector.
+///
+/// Held by `&mut`; the consumer drives sweeps by calling
+/// [`Self::maybe_run_sweep`] from its own event loop (typically the
+/// load balancer's `poll_ready`).
+pub(crate) struct OutlierDetector {
+    config: OutlierDetectionConfig,
+    state: HashMap<EndpointAddress, EndpointState>,
```
gu0keno0: How do you plan to integrate this with LoadBalancer?

After reading the Envoy doc https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/outlier.html , I think there are 3 types of outlier detections:

1. consecutive failures: these should be accumulated inline after ReadyChannels are called, and should be handled directly during poll_ready of LoadBalancer(), we can implement this in the future.

2. Succsss / Failure threshold -- these need to be re-calculated as call()s are performed on the channels. O(n) scans on atomic counters are the most straightforward ways. But we should avoid doing O(n) scans in poll_ready(), because it's the performance-sensitive path.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,852 @@
...
+            ejected_at: None,
+        }
+    }
+}
+
+/// gRFC A50 outlier detector.
+///
+/// Held by `&mut`; the consumer drives sweeps by calling
+/// [`Self::maybe_run_sweep`] from its own event loop (typically the
+/// load balancer's `poll_ready`).
```
gu0keno0: We should avoid doing O(N) scans in poll_ready(). We can do this in a spawned actor, it can share a DashMap of ReadyLbChannels with the LoadBalancer.

The outlier detection actor can notify LoadBalancer about ejections via a ReadyLbChannel.ejected future. The LoadBalancer can listen to these futures via a UnorderedFutures, so everything is O(1) in critical path.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,749 @@
...
+        // failure_pct = 100 * failure / total. A50 uses strict ">".
+        let failure_pct = 100 * f / total;
+        if failure_pct <= u64::from(fp.threshold.get()) {
+            return;
+        }
+        if !roll(fp.enforcing_failure_percentage.get()) {
+            return;
+        }
+
+        if state.try_eject(Instant::now()) {
```
mingley: This evaluates failure-percentage on every RPC completion, but A50 says the timer sweep swaps the counters and then runs failure-percentage over one interval's bucket. With interval=10s, an endpoint that receives 50 failures in the first 100ms can be ejected immediately instead of at the next sweep. The data path should only count outcomes; the interval actor should snapshot and run the ejection algorithm.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,749 @@
...
+            return;
+        };
+
+        let (s, f) = state.counters();
+        let total = s + f;
+        let request_volume = u64::from(fp.request_volume);
+
+        // Bump `qualifying_count` exactly once per channel per
+        // interval so the `minimum_hosts` gate is a single atomic load.
+        if total >= request_volume && state.mark_qualifying() {
```
mingley: This can mutate registry-level counters for a stale state after the endpoint has been removed. LoadBalancer::call keeps an Arc<OutlierChannelState>; if discovery removes the address before that RPC finishes, remove_channel() has already removed/decremented the map entry, but this late record_outcome can still mark qualifying, increment qualifying_count/ejected_count, and enqueue an eject for an address no longer in channels. Please verify the state is still the registered Arc for this address (or mark removed states inactive) before changing registry-level counters.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,749 @@
...
+        if state.is_ejected() {
+            return;
+        }
+        if total < request_volume {
+            return;
+        }
+        if self.qualifying_count.load(Ordering::Relaxed) < u64::from(fp.minimum_hosts) {
+            return;
+        }
+        if self.ejected_count.load(Ordering::Relaxed) >= self.max_ejections(&config) {
```
mingley: This cap check is racy across concurrent RPC completions. With max_ejections == 1, two endpoints can both read ejected_count == 0, both pass this check, and both successfully try_eject() before incrementing, leaving two ejected endpoints. The cap needs an atomic reservation (for example, CAS ejected_count from n to n+1 before try_eject, with rollback if the state was already ejected) or equivalent serialization.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,749 @@
...
+            }
+            if !state.is_ejected() {
+                state.decrement_multiplier();
+            }
+        }
+    }
+
+    /// Resolve `max_ejection_percent` against the current channel count.
+    fn max_ejections(&self, config: &OutlierDetectionConfig) -> u64 {
+        self.channels.len() as u64 * u64::from(config.max_ejection_percent.get()) / 100
```
mingley: A50's config text says max_ejection_percent still permits at least one address to be ejected regardless of the percentage. This floors to zero for common small clusters, e.g. the default 10% with 5 endpoints returns 0, so the cap check prevents every ejection. Return 1 when len > 0 (or otherwise apply the minimum-one rule) before enforcing the cap.

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Good catch. Changed to match grpc-go and envoy. [1c6d119](https://github.com/grpc/grpc-rust/pull/2619/commits/1c6d119e7de191edd72fb7c5098d9b4832b00305)

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: commit [1c6d119](https://github.com/grpc/grpc-rust/pull/2619/commits/1c6d119e7de191edd72fb7c5098d9b4832b00305) resolved this.

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: commit [1c6d119](https://github.com/grpc/grpc-rust/pull/2619/commits/1c6d119e7de191edd72fb7c5098d9b4832b00305) resolved this.

tonic-xds/src/xds/resource/outlier_detection.rs
```diff
@@ -2,27 +2,19 @@
...
 //! [`OutlierDetectionConfig`] is the input to the outlier-detection
 //! algorithm. The two sub-configs gate which ejection algorithms run.
-//!
-//! Note: A50 specifies outlier detection as a load-balancing policy
-//! wrapping a `child_policy`. `tonic-xds` currently runs P2C as its only
-//! load balancer and integrates outlier detection as a filter on the
-//! `Discover` stream feeding it, so there is no `child_policy` field
-//! here yet. It will be added when more balancers are supported.
+//! The `child_policy` field from A50 is not modeled — `tonic-xds`
+//! currently runs P2C as its only load balancer.
```
gu0keno0: This might change soon

tonic-xds/src/client/loadbalance/channel_state.rs
```diff
@@ -26,16 +26,176 @@
...
+    pub(crate) fn snapshot_and_reset(&self) -> (u64, u64) {
+        let s = self.success.swap(0, Ordering::Relaxed);
+        let f = self.failure.swap(0, Ordering::Relaxed);
+        (s, f)
+    }
+}
+
+/// Per-channel outlier-detection state, shared via `Arc` between the
+/// data path (per-RPC outcome recording + threshold-based ejection),
+/// the housekeeping actor, and the load balancer.
```
gu0keno0: LB does not need to know this directly?

tonic-xds/src/client/loadbalance/channel_state.rs
```diff
@@ -26,16 +26,176 @@
...
+        let nanos = now
+            .saturating_duration_since(self.epoch)
+            .as_nanos()
+            .min(u64::MAX as u128) as u64;
+        // 0 means "not ejected"; use 1 as a sentinel if the channel
+        // was created at exactly `now`.
+        let stamp = nanos.max(1);
+        if self
+            .ejected_at_nanos
+            .compare_exchange(0, stamp, Ordering::AcqRel, Ordering::Relaxed)
```
gu0keno0, line 156: Sweeper seems to be the only arbitrator for ejection / unejection, so is this CAS operation necessary?

tonic-xds/src/client/loadbalance/channel_state.rs
```diff
@@ -26,16 +26,176 @@
...
+            return false;
+        }
+        self.ejection_multiplier.fetch_add(1, Ordering::Relaxed);
+        true
+    }
+
+    /// Atomically clear the ejection. Returns `true` on the
+    /// ejected → not-ejected transition.
+    pub(crate) fn try_uneject(&self) -> bool {
+        self.ejected_at_nanos.swap(0, Ordering::AcqRel) != 0
```
gu0keno0, line 168: Same question here.

tonic-xds/src/client/loadbalance/channel_state.rs
```diff
@@ -47,9 +207,13 @@ pub(crate) struct EjectionConfig {
...
 pub(crate) enum UnejectedChannel<S> {
-    /// The channel is ready to serve again (ejection expired, no reconnect needed).
+    /// Cooldown elapsed; the original connection is reused with its
+    /// outlier state reattached.
     Ready(ReadyChannel<S>),
-    /// A fresh connection has been started.
+    /// A fresh connection has been started. The LB looks the
+    /// existing outlier state back up via the registry to call
+    /// `note_uneject` — `add_channel` is idempotent, so no Arc needs
+    /// to ride along on the variant.
```
gu0keno0: This comment is too much detailed about the internals.

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -1,28 +1,48 @@
...
-//! machine, and routes requests to ready endpoints via a [`ChannelPicker`].
+//! Receives endpoint updates via [`tower::discover::Discover`],
+//! manages the connection lifecycle via the channel state machine,
+//! and routes requests to ready endpoints via a [`ChannelPicker`].
+//!
+//! Outlier detection (gRFC A50) is integrated via [`OutlierDetector`],
+//! which is always present. The actor inside is conditionally spawned
+//! based on the [`OutlierDetectionConfig`] passed at construction —
+//! `OutlierDetectionConfig::default()` is the disabled config, in
+//! which case `record_outcome` short-circuits and no eject signals
```
gu0keno0: Maybe add a TODO about dynamic config via xDS, this needs to be supported in the next PR

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -1,28 +1,48 @@
...
+//! ever fire. Eject requests arrive on an mpsc channel from the data
```
gu0keno0: The issue is, how big is this MPSC's buffer?

How about just use a watch with a Hashmap/Hashset as its value? You can always just reconstruct the ejected/unejected hashmap in each swap run. This is memory allocation, but since it's not in data path, and the sweeper runs at low QPS, it should be fine.

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -105,22 +147,137 @@ where
...
+        loop {
+            let addr = match self.outlier.poll_eject_request(cx) {
+                Poll::Ready(Some(a)) => a,
+                _ => return,
+            };
+            let registry = self.outlier.registry().clone();
+            // Channel may have been removed by discovery in the
+            // meantime; if so, nothing to eject.
+            let Some(ch) = self.ready.swap_remove(&addr) else {
+                continue;
```
gu0keno0, line 242: add a debug message

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -105,22 +147,137 @@ where
...
+            match unejected {
+                UnejectedChannel::Ready(ready) => {
+                    registry.note_uneject(ready.outlier());
+                    tracing::debug!("outlier detection: uneject {addr}");
+                    self.ready.insert(addr, ready);
+                }
+                // `needs_reconnect = false` for A50; this arm is
+                // reserved for future policies.
+                UnejectedChannel::Connecting(connecting) => {
+                    let state = registry.add_channel(addr.clone());
```
gu0keno0, line 287: debug message

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Resolved.

tonic-xds/src/xds/resource/outlier_detection.rs
LYZJU2019: Agree. Add a note after this comment [(c468439)](https://github.com/grpc/grpc-rust/pull/2619/commits/c468439152692e756b12c9a97a7bfe0925deb9d8)

tonic-xds/src/client/loadbalance/channel_state.rs
LYZJU2019: Yes, updated in [324b374](https://github.com/grpc/grpc-rust/pull/2619/commits/324b3748301cd057be49671963c9edaf5aa2a3cd)

tonic-xds/src/client/loadbalance/channel_state.rs
LYZJU2019: Discussed offline and CAS operation is necessary to make sure there's only one changer at one time.

tonic-xds/src/client/loadbalance/channel_state.rs
LYZJU2019: Same reply as above.

tonic-xds/src/client/loadbalance/channel_state.rs
LYZJU2019: updated in [4f8469c](https://github.com/grpc/grpc-rust/pull/2619/commits/4f8469cca9d84eb80f94e6267aa114a16354985c)

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Added a TODO in [2022535](https://github.com/grpc/grpc-rust/pull/2619/commits/2022535e30f320b0208c5df39d8523777dbc2c55)

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Added in [5c93365](https://github.com/grpc/grpc-rust/pull/2619/commits/5c9336529033846baf50a729a8fd44330cdc5232)

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Added in [f83b9f6](https://github.com/grpc/grpc-rust/pull/2619/commits/f83b9f6ea62b090a14943e75c2a9bfb4540d7c72)

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -16,6 +16,11 @@
 //! resolved [`UnejectedChannel`] is routed back into `ready` or
 //! `connecting`.
 //!
+//! TODO: wire dynamic outlier-detection config updates from xDS
+//! (`ArcSwap::store` on the config + actor restart on `interval`
```
gu0keno0: Nit: maybe just let the actor pick up new configs.

tonic-xds/src/client/loadbalance/loadbalancer.rs
gu0keno0: @LYZJU2019 update comment?

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,850 @@
...
+
+    /// Number of registered channels.
+    pub(crate) fn len(&self) -> usize {
+        self.channels.len()
+    }
+
+    /// Per-RPC entry point. Records the outcome on the channel's
+    /// counter. Ejection decisions are deferred to the next sweep
+    /// (gRFC A50 §6) — see [`Self::run_housekeeping`].
+    pub(crate) fn record_outcome(&self, state: &OutlierChannelState, success: bool) {
```
gu0keno0: This should be a method of OutlierChannelState itself ?

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,850 @@
...
+            .channels
+            .iter()
+            .map(|e| {
+                let state = e.value().clone();
+                let (s, f) = state.counters();
+                (state, s, f)
+            })
+            .collect();
+
+        if let Some(fp) = config.failure_percentage.as_ref() {
```
gu0keno0, line 195: Seems like we only do failure percentage detection for now?

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,850 @@
...
+            let qualifying = snapshots
+                .iter()
+                .filter(|(_, s, f)| s + f >= request_volume)
+                .count() as u64;
+            if qualifying >= u64::from(fp.minimum_hosts) {
+                let max_ejections = self.max_ejections(&config);
+                let now = Instant::now();
+                let threshold = u64::from(fp.threshold.get());
+                let enforcing = fp.enforcing_failure_percentage.get();
+                for (state, s, f) in &snapshots {
```
gu0keno0, line 206: For each snapshot, probably add a debug print

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -0,0 +1,850 @@
...
+    /// 4. If the ejected-set version changed (sweep ejected at least
+    ///    one channel, or the LB unejected between ticks), rebuild
+    ///    the snapshot of ejected addresses and broadcast it on the
+    ///    `watch` channel. Quiet ticks skip the rebuild via an O(1)
+    ///    version compare.
+    ///
+    /// Un-ejection is *not* driven from here — each `EjectedChannel`
+    /// owns its own `Sleep` timer.
+    pub(crate) fn run_housekeeping(&self) {
+        let config = self.config.load();
```
gu0keno0, line 178: debug print the config

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Updated [fb753d1](https://github.com/grpc/grpc-rust/pull/2619/commits/fb753d120231819629e66046de50fa9a435d9042)

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Yes, this PR only support failure percentage algorithm

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Updated [f00849f](https://github.com/grpc/grpc-rust/pull/2619/commits/f00849f11e35266f3503d364649f7cd1e8e98144)

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Updated [498edd4](https://github.com/grpc/grpc-rust/pull/2619/commits/498edd490e5149ac8f64903f8c8a3f4aa44f06a1)

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: Updated [4db92be](https://github.com/grpc/grpc-rust/pull/2619/commits/4db92be25008d2d89218817487875796222223ef).

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -346,14 +348,13 @@ where
         // Cheap clones (all Arc-shared internals) so the async block
         // can take ownership without holding the picker borrow.
         let mut svc = picked.clone();
-        let outlier_state = picked.outlier().clone();
-        let registry = self.outlier.registry().clone();
+        let recorder = picked.clone();
```
gu0keno0: Why a second clone? (already cloned a few lines before)

gu0keno0 (approved):

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: updated [3435b4f](https://github.com/grpc/grpc-rust/pull/2619/commits/3435b4f6ca53fccbe374131f6d2083cd5f18ebea)

gu0keno0 (approved):

ejona86:
Hmm... Seems this might have added a flaky test. I got a test failure on https://github.com/grpc/grpc-rust/pull/2685 . Re-running passed.

```
    running 1 test
    test client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state ... FAILED

    failures:

    failures:
        client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 335 filtered out; finished in 0.00s

  stderr ───

    thread 'client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state' (7160) panicked at tonic-xds/src/client/loadbalance/loadbalancer.rs:1017:9:
    assertion `left == right` failed: counters must survive re-insert
      left: (0, 0)
     right: (3, 0)
```

gu0keno0:
> Hmm... Seems this might have added a flaky test. I got a test failure on #2685 . Re-running passed.
> 
> ```
>     running 1 test
>     test client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state ... FAILED
> 
>     failures:
> 
>     failures:
>         client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state
> 
>     test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 335 filtered out; finished in 0.00s
> 
>   stderr ───
> 
>     thread 'client::loadbalance::loadbalancer::tests::test_outlier_detection_reinsert_preserves_state' (7160) panicked at tonic-xds/src/client/loadbalance/loadbalancer.rs:1017:9:
>     assertion `left == right` failed: counters must survive re-insert
>       left: (0, 0)
>      right: (3, 0)
>     note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

I'll take a look!

ejona86:
I found a flake on master where the logs are still intact.
https://github.com/grpc/grpc-rust/actions/runs/26962845254/job/79557584053

(But since you've fixed the flake, this doesn't really matter now.)

gu0keno0:
> I found a flake on master where the logs are still intact. https://github.com/grpc/grpc-rust/actions/runs/26962845254/job/79557584053
> 
> (But since you've fixed the flake, this doesn't really matter now.)

Yeah, I'm looking at it now, and looks like there is another flaky test "test_web::grpc smoke_server_stream" https://github.com/grpc/grpc-rust/actions/runs/27624202442/job/81681297096 . Will keep monitoring.

ejona86:
Yeah, I noticed the `smoke_server_stream` test failure, but it seemed it might be an older flake with a low flake rate.

PR #2620, opened 2026-05-01
grpc: Update Status to be a Result<> and make StatusErr which holds non-OK codes only.

dfawley:
This PR is stacked on #2613.  All the unique changes for this PR are in the latest commit.

### DIFF
--- a/grpc/src/status.rs
+++ b/grpc/src/status.rs
@@ -26,11 +26,11 @@ mod server_status;
 mod status_code;
 
 pub use server_status::ServerStatusErr;
-pub use status_code::StatusCode;
+pub use status_code::StatusCodeError;
 
 /// Represents either a failing gRPC status or a successful result containing
 /// `T`.
-pub type StatusOr<T> = Result<T, StatusErr>;
+pub type StatusOr<T> = Result<T, StatusError>;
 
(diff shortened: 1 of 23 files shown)


grpc/src/status.rs
```diff
@@ -25,31 +25,39 @@
...
+pub type StatusOr<T> = Result<T, StatusErr>;
+
+/// The representation of a gRPC status.  OK statuses may not contain a status
+/// message, while error values may.
+pub type Status = StatusOr<()>;
+
 /// Represents a gRPC status.
 #[derive(Debug, Clone)]
-pub struct Status {
+pub struct StatusErr {
```
sauravzg: I am not sure if it's worth it but the current absl API has  the following public types
- StatusError
- StatusCodeError

Do we want to mimic the public API(or a subset of the exposed public API) for the currently expose types from absl status? This may limit the amount of breaking changes for us in the migration path, assuming that absl doesn't break.

sauravzg (approved):
Do we have any remaining instances of `Result<(),StatusErr>` that we want to cleanup  and covert to Status ? Or that's a task for another day?

grpc/src/status.rs
dfawley: Forgot to note here: this is done; I renamed these types.

PR #2629, opened 2026-05-05
refactor(grpc): Use gRPC metadata instead of tonic metadata

arjan-bal:
This PR migrates the gRPC crates to use the gRPC `MetadataMap` introduced in #2567, replacing the dependency on `tonic` types.

This PR also includes the following fixes:
* Explicit failure on invalid binary metadata: Converting `http::HeaderMap` to the gRPC `MetadataMap` now fails explicitly when encountering invalid base64-encoded binary metadata, rather than silently dropping the values. The `tonic` transport now returns an `INTERNAL` status in these cases to ensure consistency with other gRPC implementations.
* Stream cancellation: The HTTP/2 stream is now cancelled whenever message or header decoding fails, or when the application drops the `RecvStream`.

### DIFF
--- a/grpc/src/client/transport/tonic/mod.rs
+++ b/grpc/src/client/transport/tonic/mod.rs
@@ -213,17 +213,7 @@ fn trailers_from_tonic_status(
             status.message(),
         )),
     };
-
-    let trailers = match md.map(TryInto::try_into) {
-        Some(Err(e)) => Trailers::new(Err(StatusError::new(
-            StatusCodeError::Internal,
-            format!("failed to parse metadata: {e}"),
-        ))),
-        Some(Ok(metadata)) => Trailers::new(status_res).with_metadata(metadata),
-        None => Trailers::new(status_res),
-    };
-
-    ClientResponseStreamItem::Trailers(trailers)
+    trailers_from_status(status_res, md)
 }
 
 // Builds a trailers with a status
--- a/grpc/src/client/transport/tonic/mod.rs
+++ b/grpc/src/client/transport/tonic/mod.rs
@@ -56,7 +56,7 @@ use tonic::codec::Codec;
 use tonic::codec::Decoder;
 use tonic::codec::EncodeBuf;
 use tonic::codec::Encoder;
-use tonic::metadata::MetadataMap;
+use tonic::metadata::MetadataMap as TonicMeta;
 use tower::ServiceBuilder;
 use tower::buffer::Buffer;
 use tower::buffer::future::ResponseFuture as BufferResponseFuture;
@@ -154,7 +154,7 @@ impl Invoke for TonicTransport {
         let request_stream = ReceiverStream::new(req_rx);
         let mut request = TonicRequest::new(Box::pin(request_stream));
         let (method, metadata) = headers.into_parts();
-        *request.metadata_mut() = metadata;
+        *request.metadata_mut() = metadata.into();
 
         let Ok(path) = PathAndQuery::from_maybe_shared(method) else {
             return err_streams(StatusError::new(StatusCodeError::Internal, "invalid path"));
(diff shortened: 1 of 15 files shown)


grpc/src/client/transport/tonic/mod.rs
```diff
@@ -192,28 +192,38 @@ impl Invoke for TonicTransport {
...
+    let trailers = match md.map(TryInto::try_into) {
+        Some(Err(e)) => Trailers::new(Err(StatusError::new(
+            StatusCodeError::Internal,
+            format!("failed to parse metadata: {e}"),
+        ))),
+        Some(Ok(metadata)) => Trailers::new(status_res).with_metadata(metadata),
+        None => Trailers::new(status_res),
+    };
+
     ClientResponseStreamItem::Trailers(trailers)
```
dfawley: Can this just call `trailers_from_status` to do all of this part?

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -262,11 +272,33 @@ impl RecvStream for TonicRecvStream {
...
+                            // Start streaming and return the headers.
+                            self.state = StreamState::Streaming(stream);
+                            ClientResponseStreamItem::Headers(
+                                ResponseHeaders::new().with_metadata(md),
+                            )
+                        }
+                        // TODO: in this case, tonic believes the stream is
+                        // still running, but our parsing failed -- do we need
+                        // to terminate the request stream now even though the
+                        // Streaming is dropped?
```
dfawley: Let's answer this before merging this PR if we can.

grpc/src/client/transport/tonic/mod.rs
arjan-bal: Yes, that removed the redundant error handling. Updated.

grpc/src/client/transport/tonic/mod.rs
arjan-bal: I initially assumed that either Tonic, Hyper, or h2 would automatically close the underlying HTTP/2 stream when the client drops the inbound stream. However, this is not the case. 

I verified this by modifying the `route_chat` client in Tonic's `routeguide` example and running a Go gRPC server to log all HTTP/2 frames:

<details>
<summary> Modified route_chat client </summary>

```rust
async fn run_route_chat(client: &mut RouteGuideClient<Channel>) -> Result<(), Box<dyn Error>> {
    let start = time::Instant::now();

    let outbound = async_stream::stream! {
        let mut interval = time::interval(Duration::from_secs(1));

        loop {
            let time = interval.tick().await;
            let elapsed = time.duration_since(start);
            let note = RouteNote {
                location: Some(Point {
                    latitude: 409146138 + elapsed.as_secs() as i32,
                    longitude: -746188906,
                }),
                message: format!("at {elapsed:?}"),
            };
            println!("Sending message: {:?}", note.clone());

            yield note;
        }
    };

    let response = client.route_chat(Request::new(outbound)).await?;
    let mut inbound = response.into_inner();
    
    // Drop the inbound stream immediately
    drop(inbound);

    tokio::time::sleep(std::time::Duration::from_secs(60)).await;

    Ok(())
}
```
</details>

Despite dropping `inbound`, the outbound stream continued to be polled by Hyper/h2, and the Go server continued logging incoming `DATA` frames.

---

I updated the Tonic transport to explicitly close the outbound stream when it fails to decode response messages or headers on the inbound stream. With this change, the Go server receives a RST_STREAM frame with code `CANCELLED` once decoding fails on the client.

To implement this without introducing a mutex or spawning an additional background task, I used the [take_until](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take_until) stream combinator from the `futures` crate.

Update: Also ensured dropping `TonicRecvStream` results in stream cancellation.

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -151,7 +154,15 @@ impl Invoke for TonicTransport {
...
-        let request_stream = ReceiverStream::new(req_rx);
+        let stop_notify = Arc::new(Notify::new());
+
+        // Tonic runs the outbound request stream in a background task. It will
+        // NOT automatically stop sending if the inbound response stream is
+        // dropped. We use `take_until` with this Notify to explicitly force the
+        // stream to yield `None`, which tells Tonic to cancel the stream.
+        let stop_notify_clone = stop_notify.clone();
+        let request_stream = ReceiverStream::new(req_rx)
+            .take_until(Box::pin(async move { stop_notify_clone.notified().await }));
```
dfawley: Would this work?

```rust
.take_until(Box::pin(stop_notify_clone.notified()));
```

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -341,11 +347,20 @@ impl RecvStream for TonicRecvStream {
     }
 }
 
+impl Drop for TonicRecvStream {
+    fn drop(&mut self) {
+        if let Some(notify) = self.stop_notify.take() {
```
dfawley: I believe this `take` is unnecessary, but it's probably not significant so feel free to not change it.

dfawley (approved):

grpc/src/client/transport/tonic/mod.rs
arjan-bal: Because `notified()` takes `&self`, the `async` block was originally needed to take ownership of the `Notify` instance and ensure it lived long enough. Tokio's `notified_owned()` method handles this by taking ownership directly, which cleanly removes the need for the `async` wrapper. I've updated the code to use it, and since `Box::pin` was also unnecessary, I've removed that as well.

grpc/src/client/transport/tonic/mod.rs
arjan-bal: Yes, removed the take.

PR #2631, opened 2026-05-06
name_resolution: include structural attributes in Address PartialEq

nathanielford:
This PR is for the grpc-Rust official library, and not for tonic.

## Motivation

In the process of writing the Pick First LB policy, I realized that there were edge cases characterized by the following:
* The name resolver provides a new list of addresses
* Some of those addresses match to the previous list of addresses based on the hashed fields, even though they should not because they have different attribute fields, causing a false positive match.

The reason for this is that the addresses cannot be included in the hashing, because they are type-erased. I think all load balancers are likely to run into this issue. 

## Solution

Updating PartialEq in order to validate that the address attributes are equal in addition to the network type and address field. This means new Addresses with different attributes should be hashed into the same bucket but that looking an address with attribute set A will not fetch an address with attribute set B. 

There may be a better way to handle this, or a reason it hasn't been done already - I'd be grateful for any feedback that anyone can provide on this front.

### DIFF
--- a/grpc/src/client/name_resolution/mod.rs
+++ b/grpc/src/client/name_resolution/mod.rs
@@ -323,7 +323,6 @@ impl PartialEq for Address {
     }
 }
 
-
 impl Hash for Address {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.network_type.hash(state);
--- a/grpc/src/client/name_resolution/mod.rs
+++ b/grpc/src/client/name_resolution/mod.rs
@@ -298,7 +298,7 @@ impl Hash for Endpoint {
 
 /// An Address is an identifier that indicates how to connect to a server.
 #[non_exhaustive]
-#[derive(Debug, Clone, Default, Ord, PartialOrd)]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Ord, PartialOrd)]
 pub(crate) struct Address {
     /// The network type is used to identify what kind of transport to create
     /// when connecting to this address.  Typically TCP_IP_ADDRESS_TYPE.
@@ -313,15 +313,6 @@ pub(crate) struct Address {
     pub attributes: Attributes,
 }
 
-impl Eq for Address {}
-
-impl PartialEq for Address {
-    fn eq(&self, other: &Self) -> bool {
-        self.network_type == other.network_type
-            && self.address == other.address
-            && self.attributes == other.attributes
-    }
-}
 
 impl Hash for Address {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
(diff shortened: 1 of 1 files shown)


dfawley (commented):
I don't recall if there is a reason it was like this before, but this SGTM.

Can we not `derive(PartialEq,Eq)`?  (If not why not?)

nathanielford:
No reason not to use derive! I just didn't think of it right off since I was in the weeds. Have fixed that. Also included a perhaps not needed unit test in order to at least capture the thinking. If we feel it's not worth the bytes, though, I can pull it.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -452,4 +443,57 @@ mod test {
             assert_eq!(&target.to_string(), tc.want_str);
         }
     }
+
+    // This test ensures that the Address struct correctly maintains its asymmetric PartialEq and Hash contracts.
+    // Specifically, two addresses with the same physical coordinates but different metadata attributes must
+    // hash to the same HashMap bucket (intentional collision) but fail strict equality, forcing collection layers 
+    // (like load balancers and connection pools) to safely treat them as distinct endpoints without corrupting the map.
```
dfawley: Nit: please wrap comments at 80 columns for readability.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -452,4 +443,57 @@ mod test {
...
+    #[test]
+    fn test_address_hashmap_asymmetric_collision() {
+        use std::collections::HashMap;
+        use std::hash::{Hash, Hasher};
+        use std::collections::hash_map::DefaultHasher;
+        use crate::client::name_resolution::Address;
+        use crate::attributes::Attributes;
+        use crate::byte_str::ByteStr;
```
dfawley: Move to module-level?

grpc/src/client/name_resolution/mod.rs
```diff
@@ -452,4 +443,57 @@ mod test {
...
+        // Address B: Identical text/type coordinates, but WITH metadata attributes
+        let attrs = Attributes::new().add("metadata_payload".to_string());
+        let addr_b = Address {
+            network_type: "tcp",
+            address: ByteStr::from(addr_base.to_string()),
+            attributes: attrs,
+        };
+
+        // Invariant Verification: Stricter equality must say they differ
+        assert_ne!(addr_a, addr_b, "Address equality must be distinct when attributes mutate!");
```
dfawley: I don't think an equality test is useful since we're deriving it.

However, we are implementing hash manually, so some tests there are probably warranted.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -452,4 +443,57 @@ mod test {
...
+        
+        // Invariant Verification: Hashing must ignore attributes (intentional collision)
```
dfawley: Is this AI?

grpc/src/client/name_resolution/mod.rs
nathanielford: Done!

grpc/src/client/name_resolution/mod.rs
nathanielford: Done!

grpc/src/client/name_resolution/mod.rs
nathanielford: I removed that equality check. I added a few more to 'check' the manual hash, but it all seems a little trite, tbh. I don't want to belabor the testing of this - the main reason to have it is to prevent a change that breaks this assumption from happening silently. So let me know if this is too much, or fine. I'd like to get this in so I can settle the LB pick first PR, which needs this for it's e2e stuff.

PR #2639, opened 2026-05-12
grpc: improve crate documentation and hide non-preview symbols

dfawley:
cc @emil10001

A deploy preview of the documentation in this PR is currently available at https://dfawley.github.io/doc-preview/grpc


grpc/src/credentials/call.rs
```diff
@@ -68,6 +73,8 @@ pub struct ClientConnectionSecurityInfo {
 }
 
 impl ClientConnectionSecurityInfo {
+    /// Constructs a new instance containing the provided security info.
+    // TODO: change to builder to allow it to be stable?
```
arjan-bal: Does this mean the builder will be fallible if new required fields are unset? If not, optional fields can be added using methods that accept the new option, consume `self`, and return `self`.

grpc/src/credentials/mod.rs
```diff
@@ -34,8 +47,10 @@ pub(crate) mod server;
 use std::sync::Arc;
 
 pub use client::CompositeChannelCredentials;
-pub use insecure::InsecureChannelCredentials;
-pub use insecure::InsecureServerCredentials;
+#[allow(unused_imports)]
+pub(crate) use insecure::InsecureChannelCredentials;
+#[allow(unused_imports)]
+pub(crate) use insecure::InsecureServerCredentials;
```
arjan-bal: Should we simply remove them?

grpc/src/client/interceptor.rs
```diff
@@ -49,15 +52,12 @@ pub trait Intercept<I>: Sync {
...
+/// A single-use form of [`Intercept`] that consumes `self`.
 #[trait_variant::make(Send)]
 pub trait InterceptOnce<I>: Sync {
     type SendStream: SendStream + 'static;
     type RecvStream: RecvStream + 'static;
 
-    /// Intercepts the start of a call.  Implementations should generally use
-    /// next to create and start a call whose streams are optionally wrapped
-    /// before being returned.
     async fn intercept_once(
```
arjan-bal, line 68: Was the rustdoc removed intentionally?

grpc/src/lib.rs
```diff
@@ -23,26 +23,45 @@
...
+//!
+//! The only currently-supported feature flags are the defaults.
+//!
+//! # Modules
+//!
+//! * [`client`] - Creating and working with gRPC client-side channels
+//! * [`credentials`] - Securing connections and providing access tokens
+//! * [`metadata`] - Data sent with all RPCs typically used by interceptors
+//! * [`core`] - Common types shared between clients and servers
+//! * [`attributes`] - Generic key/value storage used by gRPC plugins.
```
arjan-bal: Since this isn't a complete sentence, we should omit the period at the end.

grpc/src/client/mod.rs
```diff
@@ -22,20 +22,37 @@
...
+use crate::core::Trailers;
 
 pub mod channel;
 pub mod interceptor;
 pub mod metadata_utils;
-pub mod service_config;
+pub(crate) mod service_config;
 pub mod stream_util;
 
 pub use channel::Channel;
```
arjan-bal, line 58: Since `mod channel` and `channel::Channel` are already `pub`, this re-export seems redundant. Users can access the same type via both `client::Channel` and `client::channel::Channel`. Should we hide one of these paths to simplify the public API?

grpc/src/credentials/mod.rs
```diff
@@ -34,8 +47,10 @@ pub(crate) mod server;
 use std::sync::Arc;
 
 pub use client::CompositeChannelCredentials;
```
arjan-bal, line 48: I think we should add a short doc comment for the constructor of `CompositeChannelCredentials`. We should also add a basic usage example in the rustdoc of `CompositeChannelCredentials`.

grpc/src/credentials/mod.rs
```diff
@@ -34,8 +47,10 @@ pub(crate) mod server;
...
-pub use insecure::InsecureChannelCredentials;
-pub use insecure::InsecureServerCredentials;
+#[allow(unused_imports)]
+pub(crate) use insecure::InsecureChannelCredentials;
+#[allow(unused_imports)]
+pub(crate) use insecure::InsecureServerCredentials;
 pub use local::LocalChannelCredentials;
 pub use local::LocalServerCredentials;
```
arjan-bal, line 50: We need to add a rustodoc for `LocalServerCredentials::new`.

grpc/src/credentials/rustls/mod.rs
arjan-bal, line 1: The aliases `IdentityList`, `StaticIdentityProvider`, `StaticRootCertificatesProvider` are missing rustdocs.

arjan-bal (commented):
`RUSTDOCFLAGS="-W missing_docs" cargo doc --no-deps` lists the public symbols without a doc. We may want to add docs to some of them.

grpc/src/credentials/mod.rs
dfawley: I didn't really understand this at the time, but I think I figured it out.

The insecure credentials are used in tests only, currently.

I'll change those tests to use local credentials instead and just remove the insecure credentials for now.  I don't think we want to provide them at all.

grpc/src/client/interceptor.rs
dfawley: Well, it was, because I wanted to just reference `Intercept` instead.  But I'll copy/paste it back over since it's probably worse with nothing.  Same for the invoke variants.

grpc/src/client/mod.rs
dfawley: My thinking was `grpc::client::channel::Channel` was too redundant, especially since clients are also channels in grpc.  For now I'll un-pub that module and bring it into `channel`.

grpc/src/credentials/rustls/mod.rs
dfawley: Done

grpc/src/credentials/call.rs
dfawley: I believe I was thinking "attributes is optional, which means Arjan expects all fields to be passed here" or something.  So maybe we just remove "attributes" from here?  I updated the TODO to suggest that instead.  Thoughts?

I would have (and did?) `pub(crate)`'d this for the preview but it's needed by the google credentials crate.

grpc/src/credentials/mod.rs
dfawley: Done for the first.  But for the second, we don't currently have an implementation of call credentials that I can use from an example.  We should add JWT token support and then do it.

grpc/src/credentials/mod.rs
dfawley: Done

grpc/src/lib.rs
dfawley: Done

dfawley (commented):
> RUSTDOCFLAGS="-W missing_docs" cargo doc --no-deps lists the public symbols without a doc. We may want to add docs to some of them.

As part of this, I somehow noticed that we have `SendStream::send` but `RecvStream::next`.  So I renamed `next` to `recv` to match.  That's a pretty big change, but it's just text, so I think we're OK.  I'll have to take another pass through all the documentation to make sure everything else is updated, too.

emil10001:
I had gone through the test site yesterday afternoon and verified that all the links work and skimmed the documentation. Overall looks good to me.

grpc/src/credentials/call.rs
arjan-bal: The new message is clearer. I considered making `attributes` optional but decided against it because they must be propagated from the output of the channel credentials to the parameters of the call credentials. Keeping them required ensures the compiler will catch it if the attributes are not passed for some reason.

We can address the TODO after discussing this in a later PR.

grpc/src/credentials/client.rs
```diff
@@ -153,11 +152,9 @@ pub struct CompositeChannelCredentials<T> {
 }
 
 impl<T: ChannelCredentials> CompositeChannelCredentials<T> {
+    /// Constructs a new instance that combines `channel_creds` and `call_creds`
+    /// so that both can be provided to a [`Channel`](crate::client::Channel).
     pub fn new(channel_creds: T, call_creds: Arc<dyn CallCredentials>) -> Result<Self, String> {
```
arjan-bal: This constructor no longer seems to be fallible. Should we remove the `Result` from the return type?

arjan-bal (approved):
LGTM!

grpc/src/credentials/client.rs
dfawley: Done

PR #2642, opened 2026-05-13
Extending Interop tests

znaczko:
Motivation
The current interop client and server test harnesses in Tonic lacked support for several core command-line input parameters defined in the [official gRPC interoperability test specification.](https://github.com/grpc/grpc/blob/master/doc/interop-test-descriptions.md)

Solution
This PR implements all remaining interop test specifications, fully handles network configuration inputs, and enables dynamic ASCII/Binary metadata injection.


interop/src/bin/client.rs
```diff
@@ -62,53 +103,102 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let test_cases = matches.test_case;
 
+    let additional_metadata = if let Some(ref am) = matches.additional_metadata {
```
mswierq, line 111: I think Opts should keep `Option<MetadataMap>`. This parsing code could be implemented as a FromStr trait for MetadataMapWrapper,  where MetadataMapWrapper wraps MetadataMap in order to satisfy the orphan rule.

interop/src/bin/client.rs
```diff
@@ -16,11 +16,28 @@ use tonic::transport::Certificate;
 use tonic::transport::ClientTlsConfig;
 use tonic::transport::Endpoint;
 
+#[allow(dead_code)]
```
mswierq, line 19: maybe add a comment which members and why are currently unused or apply `#[allow(dead_code)]` only to the members that are unused and make a general comment why they are currently unused

interop/src/client_prost.rs
```diff
@@ -384,6 +388,312 @@ impl InteropTest for TestClient {
...
+        assertions.push(test_assert!(
+            "Call must be cancelled",
+            match &result {
+                Err(e) => e.is_cancelled(),
+                _ => false,
+            },
+            format!("result={:?}", result)
+        ));
+
+        drop(tx);
```
mswierq, line 655: I do not think that we need an explicit drop here, `tx` is going out of the scope here and should be dropped automatically. **Unless**  `rx` is still awaited somehow and the default order of dropping local variables would hang this function. If so, please add a comment to explain it.

interop/src/client_prost.rs
```diff
@@ -384,6 +388,312 @@ impl InteropTest for TestClient {
...
+                tokio::time::timeout(std::time::Duration::from_millis(50), stream.next()).await;
+
+            assertions.push(test_assert!(
+                "Stream must time out (DEADLINE_EXCEEDED)",
+                stream_result.is_err(),
+                format!("stream_result={:?}", stream_result)
+            ));
+        }
+
+        drop(tx);
```
mswierq, line 704: Please look at the comment [here](https://github.com/hyperium/tonic/pull/2642/changes#r3247149141), it seems to be similar situation

mswierq (commented):
I have assumed that some of the newly added `Opts` members are yet unused, since this PR does not provide the interop test cases that cover scenarios that would use those. The main goal here is to provide support for handling all the interop args to be in parity with grpc implementation and the missing test cases will be provided later on?

dfawley (commented):
@znaczko, could you please remove the executables from this PR?  I can't see the actual diffs in the github UI, and I suspect that's the reason.  If the Go binaries need to be updated, then let's do that as a separate PR, or later in this PR.

(Also I'm not convinced they needed to be committed in the first place vs. building and caching them in CI.  These binaries take up significant space compared to source files.  As best I can tell, this directory consumes 80% of the tonic git repo's size.)

PR #2643, opened 2026-05-14
grpc: fix duplicated "to" in graceful_switch.rs doc-comment

oab24413gmai:
One-line typo fix in `grpc/src/client/load_balancing/graceful_switch.rs`: doc-comment "/// to to active and tear down the previously active policy." → "/// to active and tear down the previously active policy.". No code/behavior change.


dfawley:
Thank you for the PR.  We're in the middle of making some changes to the repo here, so I would like to wait until those resolve before we move forward with this.  (Probably within the next week.)

dfawley:
Hi @oab24413gmai, sorry for the extra work for a trivial change, but would you mind accepting the CLA so that we can merge this PR?

dfawley:
Actually, #2656 has the same fix and the CLAs are good there, so I merged it and this is no longer needed.

PR #2648, opened 2026-05-19
feat(codec): add compression API, standard algorithms, and global reg…

sauravzg:
…istry

This change introduces the foundational compression abstractions and implementations for the grpc server component.

- **Compression API (`grpc/src/codec/compression.rs`)**: Defines the `Compressor` and `Decompressor` traits for handling the compression and decompression of gRPC payloads via byte buffers.
- **Simplified Design Philosophy**: The API is intentionally kept simple. It deliberately avoids handling concerns like limiting input and output buffer sizes, instead delegating these safety boundaries to the higher-level gRPC business logic.
- **Performance Improvements**: By relying on direct buffer manipulation, this implementation avoids the hidden memory copying overhead present in the `tonic` crate (which utilizes 8KB intermediate buffers for I/O).
- **Standard Algorithms**: Implements `gzip`, `deflate`, and `zstd` compression logic in their respective modules. These are conditionally compiled based on feature flags.
- **Global Registry (`registry.rs`)**: Introduces a thread-safe `GlobalCompressionRegistry` (backed by `RwLock` and `Arc`) to manage registered compressors and decompressors. It handles safe concurrent access and automatically updates the broadcasted `grpc-accept-encoding` headers.
- **Module Integration**: Exposes the new `codec` and `compression` modules to the library's root structure.

### DIFF
--- a/grpc/src/codec/compression.rs
+++ b/grpc/src/codec/compression.rs
@@ -1,0 +1,65 @@
+/*
+ *
+ * Copyright 2026 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use bytes::Buf;
+use bytes::BufMut;
+
+#[cfg(feature = "gzip")]
+mod gzip;
+
+pub(crate) mod registry;
+
+// TODO: Evaluate defining custom buffer types instead of exposing `bytes::Buf` and `bytes::BufMut`
+// directly before exposing these traits publicly in stable API.
+/// A trait for compressing outgoing gRPC payloads.
+pub trait Compressor: Send + Sync + 'static {
+    /// The canonical gRPC content coding name (e.g., "gzip").
+    fn name(&self) -> &str;
+
+    /// Compress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if compression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn compress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String>;
+}
+
+// TODO: Evaluate defining custom buffer types instead of exposing `bytes::Buf` and `bytes::BufMut`
+// directly before exposing these traits publicly in stable API.
+/// A trait for decompressing incoming gRPC payloads.
+pub trait Decompressor: Send + Sync + 'static {
+    /// The canonical gRPC content coding name (e.g., "gzip").
+    fn name(&self) -> &str;
+
+    /// Decompress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if decompression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn decompress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String>;
+}
--- a/grpc/src/codec/compression/gzip.rs
+++ b/grpc/src/codec/compression/gzip.rs
@@ -1,0 +1,414 @@
+/*
+ *
+ * Copyright 2026 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use bytes::Buf;
+use bytes::BufMut;
+use flate2::Compress;
+use flate2::Compression as FlateCompression;
+use flate2::Decompress;
+use flate2::FlushCompress;
+use flate2::FlushDecompress;
+use flate2::Status;
+
+use crate::codec::compression::Compressor;
+use crate::codec::compression::Decompressor;
+
+/// The base-2 logarithm of the gzip sliding window size. 15 is the maximum
+/// (and standard) value; `flate2` adds the gzip framing on top of it.
+const GZIP_WINDOW_BITS: u8 = 15;
+
+/// A gzip compression implementation.
+#[derive(Debug, Clone, Copy)]
+pub struct Gzip {
+    level: FlateCompression,
+}
+
+impl Gzip {
+    /// Creates a new gzip compression implementation.
+    pub fn new() -> Self {
+        Self {
+            level: FlateCompression::new(6),
+        }
+    }
+}
+
+impl Default for Gzip {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Compressor for Gzip {
+    fn name(&self) -> &str {
+        "gzip"
+    }
+
+    /// Compresses `source` into `destination`, writing the gzip output directly
+    /// into the destination's spare capacity (no intermediate buffer).
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if compression fails or if `destination` is a
+    /// fixed-capacity sink too small to hold the output. On error, the contents
+    /// of `destination` are unspecified and should be discarded by the caller.
+    fn compress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String> {
+        let mut compressor = Compress::new_gzip(self.level, GZIP_WINDOW_BITS);
+        loop {
+            let input = source.chunk();
+            // Once every source byte lives in the current chunk, ask the codec
+            // to emit the final gzip trailer via `Finish`.
+            let flush = if source.remaining() == input.len() {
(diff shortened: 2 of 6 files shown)


dfawley:
> Standard Algorithms: Implements gzip, deflate, and zstd compression logic in their respective modules. These are conditionally compiled based on feature flags.

gRPC-Go only implements gzip.  Do any other languages offer these other ones out of the box?

dfawley (commented):
Please make sure all files include the copyright header.

grpc/src/codec/compression/gzip.rs
```diff
@@ -0,0 +1,96 @@
...
+        let data = Bytes::from_static(b"hello world");
+        let mut compressed = Vec::new();
+        compressor
+            .compress(&mut data.clone(), &mut compressed)
+            .unwrap();
+        let mut decompressed = Vec::new();
+        compressor
+            .decompress(&mut compressed.as_slice(), &mut decompressed)
+            .unwrap();
+        assert_eq!(data, decompressed.as_slice());
```
dfawley, line 214: Let's also assert that `compressed != data` above, maybe?

grpc/src/codec/compression/registry.rs
```diff
@@ -0,0 +1,338 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::sync::LazyLock;
+use std::sync::RwLock;
+
+use crate::codec::compression::Compressor;
+use crate::codec::compression::Decompressor;
+
+/// A read-only interface to query supported compression algorithms.
+pub trait CompressionResolver: Send + Sync + 'static {
```
dfawley: What problem is solved by putting traits around the registry?

grpc/src/lib.rs
```diff
@@ -33,6 +33,7 @@
 
 pub mod attributes;
 pub mod client;
+pub mod codec;
```
dfawley: Since the only thing in `codec` is `pub(crate)`, there's no need to make this bring it in as `pub`.  Let's keep it limited to crate for now.

grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    /// Compress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if compression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn compress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut)
+    -> Result<(), io::Error>;
```
dfawley: `String` might be fine as the error type?  gRPC isn't going to do anything of interest with the error -- just put it in the message portion of a `Status`.

grpc/src/codec/compression/gzip.rs
sauravzg: Done.

grpc/src/codec/compression.rs
sauravzg: Done.

grpc/src/lib.rs
sauravzg: I was hoping it didn't matter since everything inside codec is pub(crate) right now.  Done.

grpc/src/codec/compression/registry.rs
sauravzg: The traits serve two main purposes:

Testing: Global singletons are painful to test with Rust's parallel test runner. An interface allows for easy mocks or fakes in unit tests. 

Interface Segregation/data hiding: Interceptors only need 'get' APIs, so I needed a trait for read-only things. I could've probably achieved this via making the write methods `&mut self` , but it seems like for global registries interior mutability was more common (tracing lib). 

Although, I must admit it's mostly the former, global singletons are painful during testing.

dfawley: But this is just a glorified map -- surely we don't need a mock implementation of the map, do we?  Whether we have a global instance of it or one that gets passed around is independent from whether we have a trait for it.  The trait here feels like overkill.

grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn name(&self) -> &'static str;
+
+    /// Compress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if compression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn compress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String>;
```
dfawley, line 47: `SendMessage` has the return type of `Result<Box<dyn Buf + Send + Sync>, String>`.  I believe these two traits should match (one way or the other).  Is there any advantage to having gRPC allocate the destination `BufMut` as opposed to letting the codec produce the `Buf`?  The latter seems more flexible than the former depending on the implementation of the codec itself -- it seems like if the compression library's API isn't as we expect, then you could end up having to do a byte-wise copy to get it into the `BufMut` we allocate (`put<T:Buf>` does a copy).

I believe if we return a `Buf` that we could have gzip allocate a `BytesMut` and use that with the gzip library (it impls `BufMut`) and return it (it also impls `Buf`).

grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn name(&self) -> &'static str;
+
+    /// Decompress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if decompression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn decompress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String>;
```
dfawley, line 64: (The same decision would apply here I believe.)

grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use bytes::Buf;
+use bytes::BufMut;
+
+#[cfg(feature = "deflate")]
+mod deflate;
+#[cfg(feature = "gzip")]
+mod gzip;
+#[cfg(feature = "zstd")]
+mod zstd;
```
dfawley: It seems grpc-C++ supports deflate?  But I don't think anything supports zstd:

https://github.com/grpc/grpc/blob/dc5bf65c0eae98dd47cf0ba4a8f2468b7708958b/include/grpc/impl/compression_types.h#L63

It might even be best to remove deflate for now unless we have an interop test to confirm they work together.

grpc/src/codec/compression/registry.rs
sauravzg: It being a glorified global map is implementation detail. 
Using global instance might be painful in unit tests, because rust runs them parallely.

The trait helps abstract the implementation detail, while allowing polymorphism. Injecting a map might leak too much.

grpc/src/codec/compression.rs
sauravzg: The current `sendMessage` API forces an allocation per message which may not be optimal. I believe we may want to benchmark before taking a call.

So, having non allocating APIs can help amortize allocation costs. We allocate once per rpc and then keep chopping of slices for each message as needed.

grpc/src/codec/compression.rs
sauravzg: I don't mind. But given that tonic supports it, it might harm our future migration story.

dfawley: I think interop testing should be a prerequisite for things that go on the wire.  We have some "compression" tests in the interop framework, but they only do gzip, currently.

grpc/src/codec/compression/registry.rs
dfawley: > It being a glorified global map is implementation detail.

Not really.  The API is a map-like API.  Global vs. not has nothing to do with whether we use a concrete type or a trait to implement it.  Adding a trait adds complexity for no benefit if we don't intend for a very different implementation of it (which I can't see us needing).

grpc/src/codec/compression.rs
dfawley: > The current sendMessage API forces an allocation per message which may not be optimal. I believe we may want to benchmark before taking a call.

How are we avoiding an allocation otherwise?  gRPC would have to allocate something to pass into the `encode` method, wouldn't it?  We could probably devise some kind of scheme to reuse them, but that would probably be more trouble than it's worth.  I'm more worried about `BufMut::put()` performing a full copy than that one extra allocation (that we probably have in our library anyway).

I think we should keep these consistent.  If we want to do benchmarking and it shows better performance with `BufMut` then we can change all 3 of them together.

sauravzg: For unary, likely doesn't matter.

For a streaming RPC, we allocate a large-ish block. We treat the large-ish block as a queue ,  make the message write to this pre-allocated buffer, cut from beginning and expand from end as needed. This is a common patter that h2 and tonic use. 

This reduces the number of allocations from O(messages) to O(max(rpcs, stream size)). 

The compression API doesn't use `BufMut::put()` , so that's a non issue here.

grpc/src/codec/compression.rs
sauravzg: Fixed. Only gzip remains.

grpc/src/codec/compression/registry.rs
sauravzg: Removed the trait. 
Replaced with a `builder` pattern for mutation and the built instance to be immutable.

The global immutable instance will default to identity+gzip.
The idea would be that provide the ability to inject and propagate immutable registries into the relevant locations, as applicable.

PTAL

grpc/src/codec/compression/gzip.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+}
+
+impl Decompressor for Gzip {
+    fn name(&self) -> &str {
+        "gzip"
+    }
+
+    fn decompress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String> {
+        let mut decoder = GzDecoder::new(destination.writer());
```
dfawley: The gzip writer implementation used here will first allocate a buffer ([ref](https://docs.rs/flate2/latest/src/flate2/zio.rs.html#169)) to hold bits of the decompressed data, then call `destination.writer().write` with the contents, which will do a copy.  The `flate2` crate instead offers a `Compress` (and also a `Decompress`) type that I believe we'll want to use to avoid that extra copy by writing directly into a destination buffer:

https://docs.rs/flate2/latest/flate2/struct.Compress.html#method.new_gzip

The crate recommends using the reader/writer types because it assumes you are doing I/O on the other side, but we are going memory->memory and don't want to pay the penalty of copying into the `BufMut` from a side buffer.

Realistically it's probably not a big cost next to the gzip operation itself, but I think it's still worth doing.

grpc/src/codec/compression.rs
dfawley: It still seems `BufMut` is not what we want since it doesn't let us avoid a copy into its `Bytes` buffers.  Eric & I were playing around with this some and it seems like we could have our own trait instead.  Maybe something like this?

```rust
fn compress/decompress(
    &self,
    input: &mut dyn Buf,
    output: &mut dyn WriterAndAllocator) -> Result<(), String>;

pub trait WriterAndAllocator {
    fn write_u8(&mut self, data: u8) -> Result<(), String>;
    fn write_u8_slice(&mut self, data: &[u8]) -> Result<(), String>;
    fn write_bytes(&mut self, data: Bytes) -> Result<(), String>;

    fn allocate_bytes(&mut self, size: u32) -> Result<data, String>; // best effort to use this for data blocks
}
```

grpc/src/codec/compression/gzip.rs
sauravzg: Fixed. The implementation now uses Decompress/Compress structs with unsafe usage of `BufMut` and it should probably now not have unnecessary copies or buffering(subject to deflates gzip underlying implementation).

grpc/src/codec/compression.rs
sauravzg: Can you TAL at the current implementation ?

It should probably be sufficient for compression, since I'd assume most of the compression libraries (at least the ones present in the PR previously) , tend to be fine with writing to pre-allocated buffers. It uses, unsafe items from BufMut to directly write to uninitialized slice. I don't think we have additional buffering or copies right now.

There were some discussions about limiting per segment length yesterday, but that should be something that we can implement behind BufMut using vector of BytesMut.

`SendMessage` is a separate conversation, but I feel that we should be able to use `BufMut` + inherited `put(Bytes)` backed by `VecDeque<BytesMut> + metadata` to be able to handle it.

grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,61 @@
...
+    fn name(&self) -> &str;
+
+    /// Compress data from `source` into `destination`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an `io::Error` if compression fails. Implementations should gracefully
+    /// handle constrained `destination` buffers by returning an error rather than panicking
+    /// (e.g., by verifying `destination.remaining_mut()` is sufficient before writing).
+    fn compress(&self, source: &mut dyn Buf, destination: &mut dyn BufMut) -> Result<(), String>;
```
dfawley, line 47: Let's leave a TODO behind to potentially define our own types instead of `Buf` and `BufMut` before exposing these publicly.  (We should also add a similar one to `SendMessage` and `RecvMessage`, too, but you don't have to do it here if you don't want to.)

dfawley (approved):

grpc/src/codec/compression.rs
sauravzg: Done.

grpc/src/codec/compression.rs
sauravzg: Discussed offline. Keepig BufMut with a todo on re-evaluating it later.

grpc/src/codec/compression.rs
sauravzg: Discussed offline. Keepig BufMut with a todo on re-evaluating it later.

PR #2651, opened 2026-05-21
feat: Implement Least Request Load Balancing Policy (gRFC A48)

emil10001:
Implements the "all weights equal" Least Request Load Balancing policy in gRPC-Rust, in compliance with gRFC A48. The Least Request policy improves tail latencies in heterogeneous environments by tracking active request counts per endpoint and directing new requests to the backend with the lowest load.

Detailed Changes:

1. Core Load Balancing Policy (`least_request.rs`):
   - Defined `LeastRequestLoadBalancingConfig` to parse and validate the `choiceCount` config parameter (default = 2, clamped from 2 to 10).
   - Implemented `LeastRequestBuilder` registering policy name `least_request_experimental`.
   - Implemented `LeastRequestPolicy` managing endpoint-level connections via `ChildManager` children delegating to `pick_first`.
   - Maintained a persistent mapping of weak subchannel references to active request counters (`subchannel_counters`) so that outstanding request metrics survive picker updates and name re-resolutions.
   - Implemented `LeastRequestPicker` utilizing a random sampling selection algorithm over `choice_count` subchannels.

2. Active Request Cancellation Safety:
   - Identified and resolved a request counter leak bug where async task cancellations during `dyn_invoke.await` dropped the `Pick` closure without calling it.
   - Implemented a custom, defusable `ActiveRequestGuard` using an `AtomicBool` inside `LeastRequestPicker::pick`. The guard guarantees that the active request count is decremented upon drop if the picker's `on_complete` callback is never invoked.

3. Channel & Service Config Integration:
   - Registered the builder with the global LB registry in `Channel::new` inside `channel.rs`.
   - Added `CallbackRecvStream` wrapping the stream in the channel's `Invoke` implementation to trigger `on_complete` callbacks when client streams are completed or dropped.
   - Added `LeastRequest` variant to `LbPolicyType` enum in `service_config.rs`.
   - Mapped `LbPolicyType::LeastRequest` configuration inside `ResolverChannelController::update` in `channel.rs`.

4. Test Additions & Verification:
   - Added comprehensive unit tests in `least_request.rs` covering configuration parsing/clamping/validation, least request selection, tie-breaking, fewer subchannels than choice count, and cancellation drop-guard safety.
   - Modified the `InMemoryResolver` in `inmemory/mod.rs` to dynamically set the `LeastRequest` load-balancing policy based on target URI path prefixes.
   - Wrote a robust E2E integration test `test_in_memory_least_request_load_balancing` in `inmemory/mod.rs` verifying dynamic load balancing across multiple in-memory backends concurrently.


grpc/src/client/service_config.rs
```diff
@@ -35,4 +35,5 @@ pub enum LbPolicyType {
     #[default]
     PickFirst,
     RoundRobin,
+    LeastRequest,
```
dfawley, line 39: This is a pre-defined list of LB policies limited to just PF and RR.  But I think we can do this for now if we put a TODO to remove.  cc @nathanielford for when he implements support for the newer field that lets you specify any policy.

grpc/src/inmemory/mod.rs
```diff
@@ -392,9 +392,20 @@ pub struct InMemoryResolverBuilder {}
 impl ResolverBuilder for InMemoryResolverBuilder {
     fn build(&self, target: &Target, options: ResolverOptions) -> Box<dyn Resolver> {
         let path = target.path().strip_prefix('/').unwrap_or(target.path());
-        let ids: Vec<String> = path.split(',').map(|s| s.to_string()).collect();
+        let (lb_policy, rest) = if let Some(stripped) = path.strip_prefix("leastrequest/") {
```
dfawley, line 396: The LB policy _should_ get set by specifying a default service config, not through the in memory resolver.  We don't have that yet, though, so this is fine for now since this module is unexported and unused except in tests.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    fn name(&self) -> &'static str {
+        POLICY_NAME
+    }
+
+    fn parse_config(
+        &self,
+        config: &ParsedJsonLbConfig,
+    ) -> Result<Option<<Self::LbPolicy as LbPolicy>::LbConfig>, String> {
```
dfawley: I _believe_ you can just name the type here if you want to avoid all the colons and "as"es..  `Result<LeastRequestLoadBalancingConfig, String>`?

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+        let parsed: LeastRequestLoadBalancingConfig = config
+            .convert_to()
+            .map_err(|e| format!("failed to parse least_request config: {e}"))?;
+
+        if parsed.choice_count < 2 {
+            return Err("choice_count must be at least 2".to_string());
+        }
+
+        let choice_count = parsed.choice_count.min(10);
+        Ok(Some(LeastRequestLoadBalancingConfig { choice_count }))
```
dfawley: Optional: A style choice.  This could also be `parsed.choice_count = ...min(10)` and then `return Ok(Some(parsed))`.  (`parsed` would need to be declared as `mut`)

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct LeastRequestPolicy {
+    child_manager: ChildManager<Endpoint>,
+    pick_first_builder: Arc<DynLbPolicyBuilder>,
+    choice_count: u32,
```
dfawley: Consider storing an `Option<Config>` here instead of the flat config fields.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+
+    fn resolver_update(
+        &mut self,
+        update: ResolverUpdate,
+        config: Option<&Self::LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), String> {
+        if let Some(cfg) = config {
+            self.choice_count = cfg.choice_count;
+        }
```
dfawley: Every update should include the lb config or the policy should error.  If the channel/resolver isn't doing this correctly, then I think we need changes there.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+
+        if update.endpoints.is_err() {
+            return self.handle_resolver_error(update, channel_controller);
+        }
+
+        // Shard the update by endpoint.
+        let updates = update.endpoints.as_ref().unwrap().iter().map(|e| {
+            let update = ResolverUpdate {
+                attributes: crate::attributes::Attributes::default(),
```
dfawley: Make this a `use` instead please.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+        if !self.child_manager.child_updated() {
+            return;
+        }
+        let aggregate_state = self.child_manager.aggregate_states();
+
+        if aggregate_state == ConnectivityState::Ready {
+            let mut ready_subchannels = Vec::new();
+            for child in self.child_manager.children() {
+                if child.state.connectivity_state == ConnectivityState::Ready {
+                    if let PickResult::Pick(pick) = child
```
dfawley: Per clippy, these can be collapsed

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+            counter: counter.clone(),
+            active: active.clone(),
+        };
+
+        let counter_clone = counter.clone();
+        let on_complete = Box::new(move || {
+            if active.swap(false, Ordering::Relaxed) {
+                counter_clone.fetch_sub(1, Ordering::Relaxed);
+            }
+            let _ = &guard;
```
dfawley: Shouldn't we just move the guard into this callback and then when it gets dropped, the decrement happens automatically via its drop implementation?  You shouldn't need this `active` bool at all; drops happen synchronously.

`on_complete` doesn't appear to be an `FnOnce`, but it should be.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+        });
+
+        PickResult::Pick(crate::client::load_balancing::Pick {
+            subchannel: selected.subchannel.clone(),
+            metadata: crate::metadata::MetadataMap::new(),
```
dfawley: `use` please.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+        let aggregate_state = self.child_manager.aggregate_states();
+
+        if aggregate_state == ConnectivityState::Ready {
+            let mut ready_subchannels = Vec::new();
+            for child in self.child_manager.children() {
+                if child.state.connectivity_state == ConnectivityState::Ready {
+                    if let PickResult::Pick(pick) = child
+                        .state
+                        .picker
+                        .pick(&crate::core::RequestHeaders::default())
```
dfawley: We should not be making picks for fake RPCs.  If we can't map via the child itself, then we'll need another way to retrieve its active subchannel.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+                            subchannel: pick.subchannel.clone(),
+                            active_requests: counter,
+                        });
+                    }
+                }
+            }
+
+            // Clean up stale counters
+            self.subchannel_counters
+                .retain(|weak, _| weak.upgrade().is_some());
```
dfawley: I think this would be better if it reused the known-active subchannels that we got above instead of keeping all the ones that happen to still upgrade.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    // Sends an aggregate picker based on states of children.
+    fn update_picker(&mut self, channel_controller: &mut dyn ChannelController) {
+        if !self.child_manager.child_updated() {
+            return;
+        }
+        let aggregate_state = self.child_manager.aggregate_states();
+
+        if aggregate_state == ConnectivityState::Ready {
```
dfawley: Optional: maybe `let picker = if ...` and send the update commonly between the if/else.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+        } else {
+            // Forward the child picker for non-ready aggregate state
+            let picker = self
+                .child_manager
+                .children()
+                .find(|cs| cs.state.connectivity_state == aggregate_state)
+                .map(|cs| cs.state.picker.clone())
+                .unwrap_or_else(|| {
+                    Arc::new(crate::client::load_balancing::QueuingPicker) as Arc<dyn Picker>
+                });
```
dfawley: What picker is coming out of this?  Is it just the first child whose state is the same as the aggregate state?  We should round-robin over all of those children that match instead.  You should ideally share the existing `RoundRobin` picker for this.

dfawley (commented):
Thanks for the PR!  I haven't looked at the tests yet, but here's a first round of review comments.

grpc/src/client/service_config.rs
emil10001: I think I'd prefer to drop it now. It can be added back if it makes sense down the road.

grpc/src/inmemory/mod.rs
emil10001: Leaving this as a TODO comment.

grpc/src/client/load_balancing/least_request.rs
emil10001: It would not build with `Result<LeastRequestLoadBalancingConfig, String>`, but does build with `Result<Option<LeastRequestLoadBalancingConfig>, String>`.

grpc/src/client/load_balancing/least_request.rs
emil10001: I changed this a bit, ended up with:

```
        parsed.choice_count = parsed.choice_count.min(10);
        Ok(Some(parsed))
```

It would not build with `...min(10)`.

grpc/src/client/load_balancing/least_request.rs
emil10001: I was able to swap `config` with `choice_count`.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,647 @@
...
+            picker: Arc::new(FailingPicker { error }),
+        });
+        channel_controller.request_resolution();
+    }
+
+    // Sends an aggregate picker based on states of children.
+    fn update_picker(&mut self, channel_controller: &mut dyn ChannelController) {
+        if !self.child_manager.child_updated() {
+            return;
+        }
```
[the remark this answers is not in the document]
emil10001: fixed

grpc/src/client/load_balancing/least_request.rs
emil10001: I think this has been addressed.

grpc/src/client/load_balancing/least_request.rs
emil10001: I think I've addressed this.

grpc/src/client/service_config.rs
emil10001: Huh, it won't build without this being there. I've put the TODO in.

grpc/src/client/load_balancing/least_request.rs
emil10001: The logic around this was changed, updated behavior looks reasonable to me based on other feedback.

grpc/src/client/load_balancing/least_request.rs
emil10001: This has been updated to handle an empty state, a case where there's only 1, and a case where we use the existing `RoundRobin` picker. It could possibly be simplified further to drop the case where there's only 1.

emil10001:
Ok, I believe all of the comments have been addressed, and there aren't any more merge conflicts. However, testing is currently failing for some apparently unrelated reason, it's complaining about TLS, which I don't have in my PR. There also appear to be some CI failures around unresolved dependencies which are also unrelated to my PR.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -65,10 +65,17 @@ pub(crate) struct Child<T> {
...
     policy: Box<DynLbPolicy>,
     work_scheduler: Arc<ChildWorkScheduler>,
 }
 
+impl<T> Child<T> {
+    pub fn subchannels(&self) -> impl Iterator<Item = Arc<dyn Subchannel>> + '_ {
+        self.subchannels.iter().filter_map(|weak| weak.upgrade())
+    }
+}
+
```
dfawley, line 78: Sorry, I should have caught this sooner.

Please see the updates to the least-request gRFC in A61:

https://github.com/grpc/proposal/blob/master/A61-IPv4-IPv6-dualstack-backends.md#wrr-in-javago (and the subsequent section about least-request specifically)

We should not be looking at the subchannels of the children at all.  We should store the request counts for the endpoint instead.

grpc/src/client/load_balancing/least_request.rs
```diff
@@ -0,0 +1,1195 @@
...
+
+        let state = CompletionState {
+            counter: selected.active_requests.clone(),
+            active: AtomicBool::new(true),
+        };
+
+        let on_complete = Box::new(move || {
+            if state.active.swap(false, Ordering::Relaxed) {
+                state.counter.fetch_sub(1, Ordering::Relaxed);
+            }
```
dfawley, line 405: This is redundant logic.  The drop for `state` already does this.  So you can just do either:


```suggestion
            let _ = state;
```

Which captures `state` in the closure and its `drop` will be run automatically when the closure finishes.

Or:


```suggestion
            drop(state);
```

PR #2654, opened 2026-05-23
grpc: crate maintenance

dfawley:
1. Rework the codegen crate:

- `protoc-gen-rust-grpc` is now a real Rust crate that is able to build `protoc` (for now) and our plugin binary.
- `grpc-protobuf-build` uses `protoc-gen-rust-grpc` to produce the binaries, instead of requiring the user to put them in their path.
- Reworked the caching of the binaries a little.  I thought we'd be able to use the standard Rust caching, but it didn't work for various reasons.  So maybe this isn't much better than what was there before.  LMK.
- We now run `cargo build` for `interop` inside the CI's environment instead of in the `run` environment which uses MSYS and for some reason doesn't like the way we're doing the caching.  (Maybe this is a good enough reason to go back to how the cache was working before?)

2. Add a `grpc/README.md`, and add more info to `Cargo.toml`s, remove `publish=false` to allow them to be published, etc.


protoc-gen-rust-grpc/build.rs
```diff
@@ -0,0 +1,63 @@
...
+            fs::copy(&src_protoc, out_bin_dir.join(protoc_filename)).unwrap();
+            fs::copy(&src_plugin, out_bin_dir.join(plugin_filename)).unwrap();
+            return;
+        }
+    }
+
+    cmake::build("src/cpp_source");
+
+    // If a prebuilt binary directory is specified but was empty, save the newly compiled
+    // binaries there so they can be cached and reused.
```
arjan-bal: The [cargo docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) discourage writing artifacts outside of `OUT_DIR`. I suspect this may also lead to race conditions if there are multiple versions of the crate being built concurrently.

Is this being done only to speed up CI? I feel it's better to manually build the binaries using CMake once in CI instead of having a flag in the production build script to support this. Please let me know if you think otherwise.

Also, why does the regular Rust build cache not work in CI? Is it something that can be addressed by spending more time on the configuration, or is it a limitation in `Swatinem/rust-cache@v2`?

codegen/src/main.rs
```diff
@@ -94,11 +94,7 @@ fn main() {
 }
 
 fn codegen_grpc_routeguide(root_dir: &Path) {
-    let path = std::env::var("PATH").unwrap_or_default();
     unsafe {
-        std::env::set_var("PATH", format!("{}:{}", path, grpc_protobuf_build::bin()));
-        // TODO: Once the codegen is updated to allow the `OUT_DIR` env var to
-        // be unset, change this to use `.output_dir` instead.
```
arjan-bal, line 101: Since the OUT_DIR override is still present, we should leave a comment explaining why it's required.

grpc-protobuf-build/Cargo.toml
```diff
@@ -1,16 +1,22 @@
...
+keywords = ["rpc", "grpc", "async", "protobuf", "proto"]
+categories = ["web-programming", "network-programming", "asynchronous"]
 rust-version = { workspace = true }
+edition = "2024"
 
 [dependencies]
 prettyplease = "0.2.35"
 protobuf-codegen = { version = "4.34.0-release" }
 syn = "2.0.104"
+protoc-gen-rust-grpc = { version = "0.9.0-alpha.1", path = "../protoc-gen-rust-grpc" }
```
arjan-bal: Should the version of the `protoc-gen-rust-grpc` dependency also be `0.9.0-alpha.2`?

grpc-protobuf/Cargo.toml
```diff
@@ -17,7 +22,7 @@ allowed_external_types = [
 
 [dependencies]
 bytes = "1.11.1"
-grpc = { path = "../grpc" }
+grpc = { version = "0.9.0-alpha.1", path = "../grpc" }
```
arjan-bal: Should the version of the `grpc` dependency be `0.9.0-alpha.2`? Or is it intentionally using the second latest version?

interop/Cargo.toml
```diff
@@ -39,3 +39,4 @@ rustls = { version = "0.23", default-features = false, features = ["ring"] }
 [build-dependencies]
 tonic-prost-build = {path = "../tonic-prost-build"}
 grpc-protobuf-build = {path = "../grpc-protobuf-build"}
+protoc-gen-rust-grpc = {path = "../protoc-gen-rust-grpc"}
```
arjan-bal: nit: No newline at end of file.

protoc-gen-rust-grpc/src/lib.rs
```diff
@@ -0,0 +1,52 @@
+/*
+ *
+ * Copyright 2025 gRPC authors.
```
arjan-bal: nit: The copyright year needs to be 2026.

protoc-gen-rust-grpc/README.md
```diff
@@ -1,13 +1,29 @@
...
+A protoc plugin that generates Rust gRPC service code for use with the [`grpc`
+crate](https://crates.io/crates/grpc).
 
-## Build
+> NOTE: This version is a preview and not recommended for any production
+> use.  All APIs are unstable.  Proceed at your own risk.
+
+Note: as part of compiling, the source files for `protoc` are downloaded from
+the [protobuf github repository](https://github.com/protocolbuffers/protobuf).
+The archive's checksum is verified before compiling.
```
arjan-bal, line 14: There is a fallback which skips hash verification for unknown versions, should we remove that to avoid supply chain attacks?

https://github.com/grpc/grpc-rust/blob/af473358574197f4a70e804080e052a1210e3c0c/protoc-gen-rust-grpc/cmake/FetchProtobuf.cmake#L23-L26

.github/workflows/CI.yml
```diff
@@ -401,46 +188,30 @@ jobs:
...
-        fi
-
-        # Set the protoc include path only if it exists
-        if [ -d "${PROTOC_DIR}/include" ]; then
-          echo "PROTOC_INCLUDE=${PROTOC_DIR}/include" >> $GITHUB_ENV
-        fi
+        path: protoc-cache
+        key: protoc-binaries-${{ runner.os }}-${{ hashFiles('protoc-gen-rust-grpc/src/cpp_source/**') }}
     - uses: Swatinem/rust-cache@v2
+    - run: cargo build -p interop
```
arjan-bal, line 265: Is this done to avoid building the binaries twice, once for TLS and once for non-TLS tests? Maybe we should add this to the comment in test.sh. It took me a couple of minutes to understand the use reason for caching the artifacts in CI.

protoc-gen-rust-grpc/build.rs
dfawley: There were many issues with the `Swatinem/rust-cache@v2` crate.  First, I had to give it a shared cache location, and I had to set `cache-workspace-crates`.  I ran into more problems, and ultimately I couldn't get it working, and also the cache files were very large.  I think maybe we should revert back to the manual caching that it was doing before, and just change the way the compilation worked to use cargo instead of cmake from the commandline?

.github/workflows/CI.yml
dfawley: See PR description:

> We now run cargo build for interop inside the CI's environment instead of in the run environment which uses MSYS and for some reason doesn't like the way we're doing the caching. (Maybe this is a good enough reason to go back to how the cache was working before?)

I can try to expand on this in the comment in test.sh if we don't decide to change things further.

protoc-gen-rust-grpc/build.rs
dfawley: So it's easy enough to build this crate, then find the binary locations from the OUT_DIR (there's functions in `lib.rs` to tell us that).  So we can find them and copy them out.  But I think we have to be able to read them from the cache location and copy them back from there as part of `build.rs` -- if we're doing it outside of the build process, I don't know how we can determine where we should put them, and compiling the crate to call the functions it has would require building them.  So I guess we can't completely go back to the way it was before.  We _could_ move the copying-out into it, though, though I suspect it's fine.

Also maybe we should publish our binary and download it / protoc instead of building by default?

grpc-protobuf-build/Cargo.toml
dfawley: Fixed

grpc-protobuf/Cargo.toml
dfawley: Fixed.  We'll have to remember to keep these all in sync.

protoc-gen-rust-grpc/README.md
dfawley: Fixed

protoc-gen-rust-grpc/src/lib.rs
dfawley: Fixed

codegen/src/main.rs
dfawley: Fixed

dfawley:
PR is updated.  Here is the current state:

1. protoc-gen-rust-grpc is a rust crate that always does cmake for protoc and our plugin except when running on docs.rs or in CS (via a different env var)

2. grpc-protobuf-build now has three ways of working:

- it can be configured in the `CodeGen` itself where the binaries are, and it will use those first, then.....
- if a "build-plugin" feature flag is set, it will dep on the `protoc-gen-rust-grpc` crate and use the binaries it makes
- if a custom env var is set (`GRPC_RUST_PROTOC_DIR`), it will look there for both binaries.
- lastly, it will scan your path for both binaries

3. The examples are updated to use `grpc-protobuf-build` now when running a grpc example.  `build.rs` uses it instead of needing to externally run `protoc`.  It doesn't use the `build-plugin` flag which means the user needs protoc and our plugin in their path (or in `GRPC_RUST_PROTOC_DIR`; or they can edit cargo.toml to make it add the build-plugin feature back in).

4. I added a workflow that detects a release of the plugin, which builds all 10 of the targets that protoc is built for, and attaches those binaries to the release.  Those don't compile protoc, just our plugin.  And I added support for a `--version` flag that will print the release version, which it either reads from Cargo.toml when building from the rust crate, or from the tag name when building for a release.

LMK what you think.

grpc-protobuf-build/src/lib.rs
```diff
@@ -214,10 +222,83 @@ impl CodeGen {
...
+        if let Some(paths) = std::env::var_os("PATH") {
+            for path in std::env::split_paths(&paths) {
+                let candidate = path.join(binary_name);
+                if candidate.is_file() {
+                    return Some(candidate);
+                }
+            }
+        }
+        None
+    }
```
arjan-bal: This may fail when the fix exists but is not executable or when running on windows which has a PATHEXT environment variable and may check the cwd. Gemini mentioned that the which crate is the idiomatic way of finding binary paths. We could use it to simplify our code.

Instead of checking, we could also call the `protoc` or `protoc-gen-rust-grpc` binaries with the `--version` flag to check if the binaries are available.

grpc-protobuf-build/src/lib.rs
```diff
@@ -214,10 +222,83 @@ impl CodeGen {
...
+            return Ok((protoc.clone(), plugin.clone()));
+        }
+
+        // 2. Compiled via protoc-gen-rust-grpc (build-plugin feature)
+        #[cfg(feature = "build-plugin")]
+        {
+            let compiled_protoc = PathBuf::from(protoc_gen_rust_grpc::protoc());
+            let compiled_plugin = PathBuf::from(protoc_gen_rust_grpc::protoc_gen_rust_grpc());
+            if compiled_protoc.exists() && compiled_plugin.exists() {
+                // The files may not exist if a build setting instructed
```
arjan-bal: Is this sentence incomplete?

protoc-gen-rust-grpc/README.md
```diff
@@ -1,13 +1,29 @@
...
+
+## Using from Rust
+
+A `build.rs` script will ensure the `protoc` and `protoc-gen-rust-grpc` binaries
+are compiled, and the functions `protoc` and `protoc_gen_rust_grpc` can be used
+to find their locations.
+
+## Building binaries manually
```
arjan-bal, line 36: Just a thought that came up while reviewing the code: We could create a Rust binary crate that statically links to the C++ code (compiled via the cmake crate). The Rust binary would act as a thin wrapper, using FFI to forward command-line arguments to the C++ logic. This way, users can seamlessly build and install the protoc plugin using cargo install rather than invoking cmake manually.

.github/workflows/CI.yml
```diff
@@ -29,126 +31,33 @@ jobs:
...
-              else
-                cp "$protoc_path" "${TARGET_PATH}/protoc"
-                chmod +x "${TARGET_PATH}/protoc"
-                echo "Copied to: ${TARGET_PATH}/protoc"
-              fi
-              PROTOC_FOUND=true
-              break
+          mkdir -p protoc-cache
+          # Find where cmake built the binaries.
+          find target/ -name "protoc" -o -name "protoc.exe" -o -name "protoc-gen-rust-grpc" -o -name "protoc-gen-rust-grpc.exe" | while read -r file; do
```
arjan-bal: Since we're building `protoc` from source, we shouldn't need `taiki-e/install-action@protoc`, right?

examples/Cargo.toml
```diff
@@ -323,3 +323,4 @@ grpc = { path = "../grpc", optional = true }
 
 [build-dependencies]
 tonic-prost-build = { path = "../tonic-prost-build" }
+grpc-protobuf-build = { path = "../grpc-protobuf-build", default-features = false }
```
arjan-bal, line 326: Since the feature flags for the gRPC examples are enabled by default and the `grpc-protobuf-build` feature is disabled, running the Tonic examples now requires `protoc` and `protoc-gen-rust-grpc` to be installed and in the user's PATH. For instance, to run the Tonic `routeguide` server, I had to execute:

```bash
cargo run --bin routeguide-server --no-default-features --features=routeguide
```

Is this UX acceptable for Tonic users? If we want to improve this, we have a couple of options:

1. Use `required-features: Disable the gRPC example features by default and add `required-features` to each gRPC example binary. This ensures Cargo prints a clear error message instructing users to enable the exact features required.
2. Use a separate crate: Move the gRPC examples into a dedicated crate that has the necessary gRPC features enabled by default.

protoc-gen-rust-grpc/build.rs
arjan-bal: > Also maybe we should publish our binary and download it / protoc instead of building by default?

In the present state, only the first build takes a while to compile protobuf from source, subsequent builds are using cargo's cache. Maybe the first build times are not a huge problem? If it is, we could add such an option in the future.

arjan-bal (commented):
Mostly looks good, some minor comments.

protoc-gen-rust-grpc/build.rs
dfawley: It's not the time alone; it's also the C++/cmake dependencies.  But regarding the time: because Rust puts the target directory local -- unlike Go -- if you're using grpc in 5 different projects (or if you use worktrees like I do, and don't manually configure things like I do so your target directory is shared), then you will have to rebuild it in every directory.

grpc-protobuf-build/src/lib.rs
dfawley: Oops I forgot to finish my thought.  Finished.

examples/Cargo.toml
dfawley: Thanks for pointing this out.  I'd actually like even the grpc examples to be runnable without having proto in your path.  So what I've done is added an env var (`GRPC_RUST_REGENERATE_PROTO`) to build.rs to cause it to regenerate the protobuf output.  And if it's not set, then nothing is regenerated and nothing fails if the binaries aren't there.

grpc-protobuf-build/src/lib.rs
dfawley: Changed to use which and to execute the commands.

protoc-gen-rust-grpc/README.md
dfawley: Oh I like that idea!  Except that we need two binaries for now.  I talked to Em and he's open to the idea of publishing something similar for `protoc` itself.

.github/workflows/CI.yml
dfawley: Definitely not while building the plugin.  I removed it from there.  But the other ones might, because AIUI tonic requires protoc in your path.  But maybe our CI doesn't except for the `codegen` job?  I'll play around with this some more.

dfawley (commented):
Thanks for the review!

protoc-gen-rust-grpc/README.md
dfawley: (Sorry, to be clear; he was open to publishing something that can build protoc so we don't have to have something that builds both protoc and our plugin.  I haven't talked to him about the FFI idea yet.)

.github/workflows/CI.yml
dfawley: Aha!  It's needed for the other jobs because the install rule includes standard proto include files -- not just the `protoc` binary.

examples/build.rs
```diff
@@ -40,6 +40,30 @@ fn main() {
...
+            .unwrap();
+
+        grpc_protobuf_build::CodeGen::new()
+            .output_dir(manifest_dir.join("src/grpc-routeguide/generated"))
+            .input("route_guide.proto")
+            .include(manifest_dir.join("proto/routeguide"))
+            .client_only()
+            .compile()
+            .unwrap();
+    }
```
LucioFranco, line 67: I think we should get some of these env var settings in a readme for the examples folder?

grpc-protobuf-build/src/lib.rs
```diff
@@ -214,10 +244,78 @@ impl CodeGen {
...
+        #[cfg(feature = "build-plugin")]
+        {
+            let compiled_protoc = PathBuf::from(protoc_gen_rust_grpc::protoc());
+            let compiled_plugin = PathBuf::from(protoc_gen_rust_grpc::protoc_gen_rust_grpc());
+            if compiled_protoc.exists() && compiled_plugin.exists() {
+                // The files may not exist if a build setting instructed protoc-gen-rust-grpc to
+                // skip the C++ build (DOCS_RS / our CI setting).
+                return Ok((compiled_protoc, compiled_plugin));
+            }
+        }
```
LucioFranco, line 270: we need to add a `NO_VENDOR` style env var (prob something like `GRPC_RUST_NO_BUILD_PLUGIN`) that even if `build-plugin` is enabled you can still disable it since feature flags are additive across the whole workspace if another crate enables it it may get enabled in your code. Can see the no vendor version here for openssl

grpc-protobuf/src/lib.rs
```diff
@@ -22,6 +22,27 @@
  *
  */
 
+//! Protobuf integration for the [`grpc` crate](grpc).
```
LucioFranco:
```suggestion
//! Protobuf integration for the [`grpc`](grpc) crate.
```

interop/test.sh
```diff
@@ -38,8 +38,11 @@ esac
 
 ARG="${1:-""}"
 
-
-(cd interop && cargo build --bins)
+# Skip building in github actions as the workflow already builds it (to
```
LucioFranco: is this still needed? says to make sure cache works?

LucioFranco (changes requested):
Overall LGTM have a few changes I think we should get in first

examples/build.rs
dfawley: The `CARGO_FEATURE` env vars are cargo magic.  I added something for the `GRPC_RUST_REGENERATE_PROTO`.

grpc-protobuf-build/src/lib.rs
dfawley: There's already `GRPC_RUST_SKIP_CPP_BUILD` in `protoc-gen-rust-grpc` -- should we just give that a better name and document it?

Otherwise, because the cmake happens in build.rs, there is nothing we can do in _this crate_ to stop it from happening.

grpc-protobuf/src/lib.rs
dfawley: Done

LucioFranco (approved):

protoc-gen-rust-grpc/README.md
arjan-bal: I mentioned the FFI since each binary crate should have a `main.rs` file. In this case, it will simply call the main function defined in C++ and pass the command line arguements.

PR #2655, opened 2026-05-25
fix(codec): respect server's enabled encodings when selecting response compression

kv-cache:
## Motivation

`CompressionEncoding::from_accept_encoding_header` selects a response compression encoding based purely on the client's `grpc-accept-encoding` header and the *compile-time* feature set, not the *runtime* set of encodings the server enabled via `Server::send_compressed(...)`.

When the server has only zstd enabled but the client advertises `gzip,zstd,identity`, tonic compresses the response with gzip, which is an encoding the server never opted into and would not have advertised back.

The sibling `from_encoding_header` (used for request decompression) already has the correct `if enabled_encodings.is_enabled(...)` guards per match arm. This PR brings the accept-side in line.

**Observed in production** with tonic 0.14.6: server configured `send_compressed(Zstd)` only, ~38% of server CPU in `miniz_oxide::deflate`. Worked around downstream with a tower layer that rewrites the incoming `grpc-accept-encoding` header to `"zstd,identity"` to force the buggy selection onto the correct path. After the fix, that work-around will no longer be needed.

The bug affects all four RPC kinds — `from_accept_encoding_header` is called from `unary` / `server_streaming` / `client_streaming` / `streaming` in `tonic/src/server/grpc.rs`.

## Solution

Add a runtime `is_enabled` guard to each match arm of `from_accept_encoding_header` so the function returns the first encoding in the intersection of (server-enabled, client-accepted), preserving the client's preference order. Returns `None` if the intersection is empty so the server falls back to identity-encoded responses.

Adds three unit tests in `mod tests` of `tonic/src/codec/compression.rs`:

| Test | Server enabled | Client `grpc-accept-encoding` | Expected |
|---|---|---|---|
| `from_accept_encoding_header_respects_server_enabled_encodings` | `Zstd` | `gzip,zstd,identity` | `Some(Zstd)` |
| `from_accept_encoding_header_returns_none_when_no_overlap` | `Zstd` | `gzip,identity` | `None` |
| `from_accept_encoding_header_uses_client_preference_order` | `Zstd`, `Gzip` | `gzip,zstd,identity` | `Some(Gzip)` |

All three unit tests fail on master without the code fix; all pass with it. Existing `tonic` unit tests (79) and `compression` integration tests (63) continue to pass.

## Backport note

This fix applies cleanly. Flagging that the `v0.14.x` branch head appears to be behind the `tonic-v0.14.6` tag (latest commit there is `chore: prepare v0.14.5 release`), so I targeted master like v0.14.6 did. Happy to open a parallel PR against `v0.14.x` if maintainers want a backport.

### DIFF
--- a/tonic/CHANGELOG.md
+++ b/tonic/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- *(codec)* respect server's enabled encodings when selecting response compression, fixing a case where a server configured with `send_compressed(Zstd)` would still gzip responses when the client listed `gzip` before `zstd` in `grpc-accept-encoding`
+
 ## [0.14.6](https://github.com/hyperium/tonic/compare/tonic-v0.14.5...tonic-v0.14.6) - 2026-05-06
 
 ### Added
--- a/tonic/src/codec/compression.rs
+++ b/tonic/src/codec/compression.rs
@@ -118,11 +118,17 @@ impl CompressionEncoding {
 
         split_by_comma(header_value_str).find_map(|value| match value {
             #[cfg(feature = "gzip")]
-            "gzip" => Some(CompressionEncoding::Gzip),
+            "gzip" if enabled_encodings.is_enabled(CompressionEncoding::Gzip) => {
+                Some(CompressionEncoding::Gzip)
+            }
             #[cfg(feature = "deflate")]
-            "deflate" => Some(CompressionEncoding::Deflate),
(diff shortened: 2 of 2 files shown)


kv-cache:
@arjan-bal @dfawley can one of you take a look pls? 🙏

dfawley (approved):
@LucioFranco PTAL but this seems reasonable to me.  One thing to watch out for is that it _is_ a behavior change, and if `accept-encoding` wasn't setting gzip for some reason (or any other supported compressor), then responses that used to be compressed (and apparently have been working?) will no longer be.  This is the right behavior, but it's something worth calling out in release notes.

LucioFranco (approved):

PR #2662, opened 2026-05-28
protoc-gen-rust-grpc: fix macos universal build

dfawley:
The cmake flags were not properly quoted, so when the mac universal build used a semicolon in them, it was treated as the end of the command by bash.  This fixes that by quoting the flags.

### DIFF
--- a/.github/workflows/release-protoc-gen-rust-grpc.yml
+++ b/.github/workflows/release-protoc-gen-rust-grpc.yml
@@ -103,7 +103,7 @@ jobs:
             -DBUILD_PROTOC=OFF \
             -DBUILD_PLUGIN=ON \
             -DRUST_GRPC_VERSION="$VERSION_FROM_TAG" \
-            ${{ matrix.cmake_flags }}
+            "${{ matrix.cmake_flags }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/release-protoc-gen-rust-grpc.yml
```diff
@@ -103,7 +103,7 @@ jobs:
             -DBUILD_PROTOC=OFF \
             -DBUILD_PLUGIN=ON \
             -DRUST_GRPC_VERSION="$VERSION_FROM_TAG" \
-            ${{ matrix.cmake_flags }}
+            "${{ matrix.cmake_flags }}"
```
ejona86, line 106: Wouldn't this make `-DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=s390x -DCMAKE_C_COMPILER=s390x-linux-gnu-gcc -DCMAKE_CXX_COMPILER=s390x-linux-gnu-g++` be a _single_ argument. That is not going to do what you want. I don't know if Github allows us to use an array here.

PR #2663, opened 2026-05-28
examples: Generate code to its own dedicated dir

ejona86:
This allows us to delete the directory before re-generating code to avoid stale files from prior builds. This model works well if checking in the code or outputing to OUT_DIR; you'd only change between grpc::include_generated_proto! and grpc::include_proto!.

@dfawley, note that I excluded from this PR the change to input()/include() and such. This PR produces identical codegen output, just at the different location; the other changes change the generated code source.


examples/build.rs
```diff
@@ -47,18 +47,23 @@ fn main() {
     let grpc_routeguide = env::var_os("CARGO_FEATURE_GRPC_ROUTEGUIDE").is_some();
 
     if (grpc_helloworld || grpc_routeguide) && env::var_os("GRPC_RUST_REGENERATE_PROTO").is_some() {
-        let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+        let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+
+        let generated_dir = manifest_dir.join("generated");
+        if generated_dir.exists() {
+            std::fs::remove_dir_all(&generated_dir).unwrap();
```
dfawley: Nit: this should `expect()` instead with a message about how it failed trying to clear out your generated directory, so go clean it up by hand (or whatever we want the user to do).  Note the recommended style for `Result::expect()` messages: https://doc.rust-lang.org/std/result/enum.Result.html#recommended-message-style

dfawley (approved):

PR #2665, opened 2026-05-29
protoc-gen-rust-grpc: Fix well-known protos for prost

ejona86:
This makes the well-known protos available to the protoc built by protoc-gen-rust-grpc. It needed to "install" the protos, but also run the "installed" protoc executable so that it could find the protos in the include/ directory next to its own bin/. Note that the well-known protos don't yet work with grpc-protobuf-build because of missing crate mapping.

With that in place, it can be used in the examples.


examples/README.md
```diff
@@ -37,7 +37,8 @@ downloaded from [our releases].
 Alternatively, you can also set the `grpc-protobuf-build/build-plugin` feature
 flag when building via `cargo` using the `--features` flag.  In this case you
 will need a compatible C++ compiler and CMake as described in the
-`protoc-gen-rust-grpc` crate.
+`protoc-gen-rust-grpc` crate. When using this approach, you can also set the
+`protoc-gen-rust-grpc` feature to use the same protoc for the prost examples.
```
dfawley: As discussed offline, I think we should move this new bit up to the previous section since it's about how we can avoid the protoc install for tonic builds.

dfawley (approved):

examples/README.md
ejona86: Reworked.

dfawley (approved):

PR #2668, opened 2026-06-02
CI: make protoc-gen-rust-grpc build workflow always trigger

dfawley:
With this change, the workflow is always triggered on PRs and releases.  If it's running for a PR and the plugin's source files (and this yml file) are not changed, it ~immediately passes.  This allows us to make the plugin build step a required checker without imposing the cost of building it on everyone.


arjan-bal (approved):
LGTM

PR #2669, opened 2026-06-02
feat(xds-client): gRFC A78 XdsClient metrics + extension point

YutaoMa:
## Motivation

Ref: #2444 

Improve the production readiness of `tonic-xds` library by supporting metrics.

## Solution

This PR builds the foundation for the following metrics-related gRFCs support in `tonic-xds`:

1. [A78: XdsClient metrics](https://github.com/grpc/proposal/blob/master/A78-grpc-metrics-wrr-pf-xds.md)
2. [A79: framework-agnostic non-per-call metrics abstraction](https://github.com/grpc/proposal/blob/master/A79-non-per-call-metrics-architecture.md)

It adds the following:

1. `MetricsRecorder` trait, allowing external users to build any telemetry framework integration they need.
2.  5 A78 XdsClient metrics definition: `grpc.xds_client.{connected, server_failure, resource_updates_valid, resource_updates_invalid, resources}`

A follow-up PR to add a bundled OpenTelemetry implementation is planned.

### DIFF
--- a/xds-client/src/client/worker.rs
+++ b/xds-client/src/client/worker.rs
@@ -44,75 +44,89 @@ impl ClientAttrs {
         ]
     }
 
-    fn type_attrs(&self, type_url: &str) -> [KeyValue; 3] {
+    fn type_attrs(&self, type_url: &Arc<str>) -> [KeyValue; 3] {
         [
             KeyValue::str(metrics::attrs::GRPC_TARGET, Arc::clone(&self.target)),
             KeyValue::str(metrics::attrs::GRPC_XDS_SERVER, Arc::clone(&self.server)),
-            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, type_url.to_string()),
+            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, Arc::clone(type_url)),
         ]
     }
 
-    fn cache_state_attrs(&self, type_url: &str, cache_state: &'static str) -> [KeyValue; 4] {
+    fn cache_state_attrs(&self, type_url: &Arc<str>, cache_state: &'static str) -> [KeyValue; 4] {
         [
             KeyValue::str(metrics::attrs::GRPC_TARGET, Arc::clone(&self.target)),
             KeyValue::str(metrics::attrs::GRPC_XDS_SERVER, Arc::clone(&self.server)),
-            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, type_url.to_string()),
+            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, Arc::clone(type_url)),
             KeyValue::str(metrics::attrs::GRPC_XDS_CACHE_STATE, cache_state),
         ]
     }
 }
 
-/// A78 metric emission verbs on the worker's recorder.
-///
-/// Inherent methods on `dyn MetricsRecorder` (legal because the trait is
-/// defined in this crate). `pub(crate)` keeps them invisible to downstream
-/// consumers — the methods exist only for use by this crate's worker.
-///
-/// Call sites look like `self.recorder.record_X(&self.attrs, ...)`. The
-/// receiver and `attrs` borrow are split-borrow-friendly with `&mut self`
-/// borrows on other worker fields (e.g. `self.type_states`), so the mutation
-/// handlers don't need to be restructured to defer emission.
-impl dyn MetricsRecorder {
+/// Worker-side wrapper around an optional [`MetricsRecorder`] backend.
+pub(crate) struct RecorderHandle {
+    recorder: Option<Arc<dyn MetricsRecorder>>,
+    attrs: ClientAttrs,
+}
+
+impl RecorderHandle {
+    pub(crate) fn new(recorder: Option<Arc<dyn MetricsRecorder>>, target: Arc<str>) -> Self {
+        Self {
+            recorder,
+            attrs: ClientAttrs {
+                target,
+                server: Arc::from(""),
+            },
+        }
+    }
+
+    /// Update the `grpc.xds.server` attribute for subsequent emissions.
+    pub(crate) fn set_server(&mut self, server: Arc<str>) {
+        self.attrs.server = server;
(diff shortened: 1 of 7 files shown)


xds-client/src/client/worker.rs
```diff
@@ -20,10 +21,134 @@ use crate::client::watch::{ProcessingDone, ResourceEvent};
...
+        self.add_counter_u64(
+            &metrics::instruments::XDS_CLIENT_SERVER_FAILURE,
+            1,
+            &attrs.connection_attrs(),
+        );
+    }
+
+    /// `grpc.xds_client.resource_updates_valid` + `_invalid`, with aggregated
+    /// counts from a single response.
+    fn record_resource_updates(
```
madhurishgupta: We shouldn't call `record_*` functions when MetricsRecorder is NoOpRecorder

xds-client/src/client/worker.rs
```diff
@@ -20,10 +21,134 @@ use crate::client::watch::{ProcessingDone, ResourceEvent};
...
+            KeyValue::str(metrics::attrs::GRPC_XDS_SERVER, Arc::clone(&self.server)),
+            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, type_url.to_string()),
+        ]
+    }
+
+    fn cache_state_attrs(&self, type_url: &str, cache_state: &'static str) -> [KeyValue; 4] {
+        [
+            KeyValue::str(metrics::attrs::GRPC_TARGET, Arc::clone(&self.target)),
+            KeyValue::str(metrics::attrs::GRPC_XDS_SERVER, Arc::clone(&self.server)),
+            KeyValue::str(metrics::attrs::GRPC_XDS_RESOURCE_TYPE, type_url.to_string()),
```
madhurishgupta: Heap allocation of string (&str --> String) will happen on every record call which is in the hot path. Should we follow the same pattern as `ClientAttrs`

xds-client/src/client/worker.rs
```diff
@@ -20,10 +21,134 @@ use crate::client::watch::{ProcessingDone, ResourceEvent};
...
+            KeyValue::str(metrics::attrs::GRPC_XDS_CACHE_STATE, cache_state),
```
madhurishgupta, line 72: In A78 proposal, the labels for `grpc.xds_client.resources` are grpc.target, grpc.xds.authority, grpc.xds.cache_state and grpc.xds.resource_type but in current impl, `grpc.xds.server` is being emitted instead of `grpc.xds.authority`.

xds-client/src/client/worker.rs
madhurishgupta, line 624: We should record server failure here as well?
`self.recorder.record_server_failure(&self.attrs);`

madhurishgupta, line 639: We should record server failure here as well?
`self.recorder.record_server_failure(&self.attrs);`

madhurishgupta, line 894: We should do -1 emission when a resource leaves the cache by removal

xds-client/src/metrics.rs
```diff
@@ -0,0 +1,301 @@
...
+
+impl MetricsRecorder for NoOpRecorder {
+    fn add_counter_u64(&self, _: &'static Instrument, _: u64, _: &[KeyValue]) {}
+    fn add_up_down_counter_i64(&self, _: &'static Instrument, _: i64, _: &[KeyValue]) {}
+    fn record_histogram_f64(&self, _: &'static Instrument, _: f64, _: &[KeyValue]) {}
+    fn record_gauge_i64(&self, _: &'static Instrument, _: i64, _: &[KeyValue]) {}
+}
+
+/// Instrument descriptors for the gRFC A78 XdsClient metrics emitted by this crate.
+pub mod instruments {
```
madhurishgupta, line 232: The code below this can go into a different file as this is not related to recording metrics

xds-client/src/client/worker.rs
YutaoMa: Good point, I switched to an `Option<Recorder>` design and dropped `NoOpRecorder` impl.

xds-client/src/client/worker.rs
YutaoMa: Updated to `Arc<str>` clones for type url.

xds-client/src/client/worker.rs
YutaoMa: Good catch, switched to `grpc.xds.authority`. Federation is currently not supported so used the same placeholder value as `grpc-go`.

xds-client/src/metrics.rs
YutaoMa: The instruments/attrs modules are the A78 schema layer and the rest of the file is the recording framework, which is a fair distinction. For now there's only one consumer (the worker) and one schema (A78) so I'd prefer to defer the split until either (1) another component starts emitting metrics, or (2) the file grows enough that the framework gets harder to find

xds-client/src/client/worker.rs
```diff
@@ -405,10 +599,12 @@ where
                 Some(s) => s,
                 None => return, // No servers configured
             };
+            self.recorder.set_server(Arc::from(server.uri()));
 
             let transport = match self.transport_builder.build(server).await {
                 Ok(t) => t,
                 Err(_) => {
+                    self.recorder.record_server_failure();
```
ankurmittal: are we counting this multiple times due to retry? I mean if there is a unhealthy connection and we do multiple retires, will it be multi counted?

ankurmittal:
where is NoOpRecorder as mentioned in PR desc?

YutaoMa:
@ankurmittal , ^ updated the PR description. Have switched to `Option<Recorder>` instead of a `NoOpRecorder` impl for efficiency.

madhurishgupta (approved):
🚢

xds-client/src/client/worker.rs
YutaoMa: Aligned with A78 spec, each healthy->unhealthy transition counts as a failure so retry should be de-duped. Also need to exclude stream closure after a response is received. Updated code to reflect both behavior in the new commit.

ankurmittal (approved):

PR #2670, opened 2026-06-02
grpc: add payload to load_balancing's work/schedule_work

dfawley:
This allows for providing data directly to the LB policy that scheduled the work item -- note the pattern in child_manager for wrapping the payload in order to route it.  This will be leveraged for future things like subchannel health checking results.

API-wise, this is a two-line change:
- `WorkScheduler::schedule_work` adds a data payload field
- `LbPolicy::work` adds a matching field that is populated from the above

Aside from basic plumbing, some other changes are needed:
- We no longer squash duplicate calls in the channel
- The `ChildManager` policy leverages the payload to encode which child the payload should be sent to.  This is done with an anonymous handle instead of the current index of the child in the internal tracking vector, which makes it directly resilient to changes that could happen at any time.
- The `Lazy` policy's idle picker needs an atomic bool to squash the work calls produced by its picks to avoid flooding it if many RPCs happen at once.

cc @YutaoMa @gu0keno0 as FYI - this slightly changes the APIs we talked about yesterday.

### DIFF
--- a/grpc/src/client/load_balancing/child_manager.rs
+++ b/grpc/src/client/load_balancing/child_manager.rs
@@ -26,13 +26,11 @@
 //! purposes of forwarding channel updates.
 
 use std::collections::HashMap;
-use std::collections::HashSet;
 use std::error::Error;
 use std::fmt::Debug;
 use std::hash::Hash;
 use std::mem;
 use std::sync::Arc;
-use std::sync::Mutex;
 
 use crate::client::ConnectivityState;
 use crate::client::load_balancing::ChannelController;
@@ -43,6 +41,7 @@ use crate::client::load_balancing::LbPolicyOptions;
 use crate::client::load_balancing::LbState;
 use crate::client::load_balancing::Subchannel;
 use crate::client::load_balancing::SubchannelState;
+use crate::client::load_balancing::WorkData;
 use crate::client::load_balancing::WorkScheduler;
 use crate::client::load_balancing::subchannel::WeakSubchannel;
 use crate::client::name_resolution::Address;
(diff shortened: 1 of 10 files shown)


dfawley:
Also, @YutaoMa @gu0keno0: We are considering removing the `subchannel_update` function entirely and use this mechanism for delivering the state updates, too.  That's still up in the air, but could potentially impact future changes you're planning to make.  I'll let you know if that idea actually works out and if we eventually decide to make that change, _probably_ not for 3-4 weeks unless I'm extremely productive this week.

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -390,14 +382,19 @@ where
...
-        let child_idxes = mem::take(&mut *self.pending_work.lock().unwrap());
-        for child_idx in child_idxes {
-            let mut channel_controller = WrappedController::new(channel_controller);
-            self.children[child_idx]
-                .policy
-                .work(&mut channel_controller);
-            self.resolve_child_controller(channel_controller, child_idx);
+    pub fn work(&mut self, data: Option<WorkData>, channel_controller: &mut dyn ChannelController) {
+        if let Some(data) = data
+            && let Ok(child_work_item) = data.downcast::<ChildWorkItem>()
```
ejona86: Do we want to panic! or debug_assert! for failed downcast/None, to detect bugs?

If we do that here, and assert None in pick_first, then it seems we'd have pretty good coverage to detect bugs.

ejona86 (approved):

grpc/src/client/load_balancing/child_manager.rs
dfawley: Added debug asserts here and pick first.  Also added Debug so the printing of the data can work in case the downcast fails.  (And tons of rust boilerplate is needed for that.)

ejona86 (approved):

PR #2671, opened 2026-06-02
CI: fix plugin build to report status always

dfawley:
It seems I messed up in #2668.  It still skipped the build step.  This should actually work, and be even better.  Please see the description of the modified workflow behavior in the yml file.


arjan-bal (approved):
LGTM

PR #2673, opened 2026-06-04
feat(tonic-xds): implement gRFC A50 outlier detection success-rate algorithm

LYZJU2019:
## Motivation

#2619 landed the failure-percentage algorithm and the shared sweep plumbing. This PR fills in the second [gRFC A50](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md) ejection algorithm: **success-rate** (cross-host mean/stdev).

## Solution

In the housekeeping sweep, the success-rate algorithm now runs before the failure-percentage algorithm (A50 lists them in that order, and Envoy's implementation follows the same order; the two algorithms are otherwise independent and gated separately).

## Tests

Seven new unit tests covering:

- `success_rate_ejects_outlier_below_threshold` — 4× 100% + 1× 0%, mean=80, stdev=40, threshold@1900 = 4 → outlier ejected.
- `success_rate_uniform_population_does_not_eject` — stdev=0 → threshold=mean, nothing strictly below.
- `success_rate_minimum_hosts_gates_ejection` — qualifying<minimum_hosts → algorithm skipped.
- `success_rate_request_volume_filters_low_traffic` — low-volume outlier excluded from population *and* candidates.
- `success_rate_enforcement_zero_never_ejects` — `enforcing_success_rate = 0` short-circuits the roll.
- `success_rate_stdev_factor_zero_ejects_below_mean` — factor=0 collapses threshold to the mean.
- `success_rate_max_ejection_percent_caps_concurrent_ejections` — cap (with floor) holds.
- `success_rate_and_failure_percentage_compose` — both algorithms configured: success-rate ejects, failure-percentage skips already-ejected host (no double-count).

All 33 tests in `client::loadbalance::outlier_detection` pass, including the 26 from #2619.


tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -192,6 +196,51 @@ impl OutlierStatsRegistry {
...
+                    if rate >= threshold {
+                        continue;
+                    }
+                    if !roll(enforcing) {
+                        continue;
+                    }
+                    if state.try_eject(now) {
+                        self.ejected_count.fetch_add(1, Ordering::Relaxed);
+                        self.ejected_set_version.fetch_add(1, Ordering::Relaxed);
+                    }
```
madhurishgupta: This code is duplicated both for success_rate and failure_percentage logic except one line. Can you pleas create a new function for the common logic so that no drift is introduced in future between the two impls

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -192,6 +196,51 @@ impl OutlierStatsRegistry {
...
+                let enforcing = sr.enforcing_success_rate.get();
+                for (state, s, f) in &snapshots {
+                    let total = s + f;
+                    if total < request_volume || state.is_ejected() {
+                        continue;
+                    }
+                    if self.ejected_count.load(Ordering::Relaxed) >= max_ejections {
+                        break;
+                    }
+                    let rate = 100.0 * (*s as f64) / (total as f64);
```
madhurishgupta: If total and request_volume are 0, then rate will (100*0)/0 which is NaN. Can this might lead to ejection of all the hosts

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -621,6 +685,161 @@ mod tests {
...
+    }
+
+    /// max_ejection_percent applies before per-host eligibility, so
+    /// even when every host is below threshold the cap holds.
+    #[test]
+    fn success_rate_max_ejection_percent_caps_concurrent_ejections() {
+        let mut config = sr_config(1900, 10, 3);
+        config.max_ejection_percent = pct(20);
+        let registry = make_registry_only(config);
+        // 4 hosts at 100%, 1 at 0%. The outlier is the only candidate
```
madhurishgupta: There is only 1 bad host and max cap is also 1. So this test will never breach the cap. Maybe keep number of bad hosts > 1.

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 9438358a

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 9438358a

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 9438358a

madhurishgupta (approved):
🚢

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -282,6 +303,46 @@ impl OutlierStatsRegistry {
...
+        is_outlier: impl Fn(u64, u64, u64) -> bool,
+    ) {
+        let now = Instant::now();
+        for (state, s, f) in snapshots {
+            let (s, f) = (*s, *f);
+            let total = s + f;
+            if total == 0 || total < request_volume || state.is_ejected() {
+                continue;
+            }
+            if self.ejected_count.load(Ordering::Relaxed) >= max_ejections {
```
ankurmittal: This is the comparison, against a cap computed in `max_ejections`:

```rust
let cap = len * u64::from(config.max_ejection_percent.get()) / 100;
if len > 0 { cap.max(1) } else { 0 }
```

A50 does it differently:

> If the percentage of ejected addresses is greater than or equal to
> `max_ejection_percent`, stop.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -646,6 +722,163 @@ mod tests {
...
+            drive(&s, 80, 20);
+            all.push(s);
+        }
+        registry.run_housekeeping();
+        for s in &all {
+            assert!(!s.is_ejected());
+        }
+    }
+
+    /// minimum_hosts gate: only 2 hosts meet request_volume but
```
ankurmittal: `success_rate_minimum_hosts_gates_ejection` sets `minimum_hosts = 5` but only
registers 2 hosts. Both `2 >= 5` and `2 > 5` are false, so you could flip that
comparison and the test would still pass.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -282,6 +303,46 @@ impl OutlierStatsRegistry {
...
+            if self.ejected_count.load(Ordering::Relaxed) >= max_ejections {
+                break;
+            }
+            if !is_outlier(s, f, total) {
+                continue;
+            }
+            if !roll(enforcing) {
+                continue;
+            }
+            if state.try_eject(now) {
```
ankurmittal: If a host is removed while the sweep is running, we can still eject it, and
nothing ever un-ejects it.

1. The sweep snapshots host A. Not ejected yet.
2. An EDS update removes A. It wasn't ejected, so nothing decrements.
3. The sweep reaches it anyway, ejects it, and bumps `ejected_count`.

The load balancer drives un-ejection, and it no longer knows about A. The
broadcast rebuilds from `channels`, where A is gone. So the count stays up for
good.

tonic-xds/src/client/loadbalance/outlier_detection.rs
ankurmittal: We read the counters at the start of the sweep and reset them at the end. Any
request that finishes in between is dropped

A50 resets the counters first and runs the algorithms afterwards, so there's
nothing to lose. The window is small next to a 10s interval, but the
success-rate pass added here makes it longer.

LYZJU2019: addressed in f0115848

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 26923d90

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 3c6272f8

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in c0fac859

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -184,13 +181,20 @@ impl OutlierStatsRegistry {
...
+    ///    `request_volume`), gated by `minimum_hosts`; eject any host
+    ///    whose success rate is below `mean - stdev * stdev_factor /
+    ///    1000`, subject to `max_ejection_percent` and the enforcement
+    ///    roll.
+    /// 3. Run the failure-percentage algorithm against the same
+    ///    snapshot: apply `minimum_hosts` to the qualifying population,
+    ///    then `max_ejection_percent`, then per-channel threshold and
+    ///    the enforcement roll. Hosts already ejected by step 2 are
+    ///    skipped, and the `max_ejection_percent` cap accounts for them.
+    /// 4. Reset counters and decrement multipliers for non-ejected
```
ankurmittal: Counters aren't reset here anymore.

tonic-xds/src/client/loadbalance/outlier_detection.rs
```diff
@@ -37,15 +37,12 @@
...
-//!   counters, and decrements multipliers for non-ejected channels.
-//!   When the ejected-set membership changes, broadcasts a fresh
-//!   snapshot on the `watch` channel; quiet ticks skip the broadcast
-//!   via an O(1) version compare.
-//!
-//! Only the failure-percentage algorithm is implemented; success-rate
-//! (cross-endpoint mean/stdev) is left to a follow-up.
+//!   `config.interval` tick, runs the success-rate and
+//!   failure-percentage algorithms over a snapshot of counters, ejects
+//!   qualifying channels, resets counters, and decrements multipliers
```
ankurmittal: is this stale?

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 2ddbeb91

tonic-xds/src/client/loadbalance/outlier_detection.rs
LYZJU2019: addressed in 2ddbeb91

YutaoMa (approved):

YutaoMa:
Note for later: currently the error categorization code in https://github.com/grpc/grpc-rust/blob/2ddbeb9104a0445d99412a7a0a20acb4ac78922f/tonic-xds/src/client/loadbalance/loadbalancer.rs#L395-L396 only denotes transport level errors, thus failed rpc requests with grpc error status won't be caught. This is a higher up level issue so approving the success rate algorithm change as is, but we need to address this gap before wiring in OD in the LB stack.

LYZJU2019:
> Note for later: currently the error categorization code in
> 
> https://github.com/grpc/grpc-rust/blob/2ddbeb9104a0445d99412a7a0a20acb4ac78922f/tonic-xds/src/client/loadbalance/loadbalancer.rs#L395-L396
> 
> only denotes transport level errors, thus failed rpc requests with grpc error status won't be caught. This is a higher up level issue so approving the success rate algorithm change as is, but we need to address this gap before wiring in OD in the LB stack.

Good point. OD needs to be transport agnostic as well. See https://github.com/grpc/grpc-rust/pull/2849

PR #2675, opened 2026-06-05
grpc/client: Implement Channel builder

nathanielford:
Introduces a type-builder Channel builder for creating new Channels, using an idiomatic pattern for adding in configuration options.

## Motivation

In order to flexibly and ergonomically provide for channel configuration on construction, the type builder is introduced. This allows users to construct channels in a manner similar to:

```
let channel = Channel::builder("https://api.darkcave.com:443")
    .credential(my_creds)
    .runtime(my_runtime)
    .override_authority("arrow.wumpus.api")
    // ... any other options.
    .build();  // No error is returned here
``` 

## Solution
The solution uses a 'type builder' so that users can build up their channel configuration over various calls, and the compiler can catch type issues. This also allows us to encapsulate various ergonomics (such as default selection of runtime). This PR implements the critical values at this time and defers additional optional values not being used until a later point. It also leaves work around the internals of the `PersistentChannel`/`ActiveChannel` for a future PR. 

Because the public API is being altered, this requires a version bump.


grpc/src/client/channel.rs
```diff
@@ -225,54 +138,157 @@ impl Invoke for Channel {
...
-// some configurable timeout elapses without any any RPC activity.
-struct PersistentChannel {
-    target: Target,
-    resolver_builder: Arc<dyn ResolverBuilder>,
-    options: ChannelOptions,
-    active_channel: Mutex<Option<Arc<ActiveChannel>>>,
-    runtime: GrpcRuntime,
-    security_opts: SecurityOpts,
-    authority: String,
+pub struct MissingOpt;
```
arjan-bal: Since users shouldn't need to instantiate this struct, we should add zero-sized private field to prevent this.
```rust
pub struct MissingOpt(())
```

grpc/src/client/channel.rs
```diff
@@ -225,54 +138,157 @@ impl Invoke for Channel {
...
+pub struct PresentOpt<T>(pub T);
```
arjan-bal: Does the inner T need to be `pub` here? Since the struct is only instantiated by the gRPC crate, I think `pub struct PresentOpt<T>(T)` should work.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -183,6 +183,9 @@ pub(crate) struct ResolverOptions {
     /// This value is either the result of the `default_authority` method of
     /// this `ResolverBuilder`, or another string if the channel was explicitly
     /// configured to override the default.
+    // TODO(nford): This field is currently unused. When/if it is needed ensure the type is appropriate.
+    // - Channels reference SecurityOpts, which hold the value in an Authority object.
+    // - All other references currently use "ignored" as the authority.
```
arjan-bal: Across languages, there are a few ways to set the channel authority. In order of precedence (highest to lowest), they are:
1. User-specified authority in the channel options.
2. The authority from the name resolver builder.
3. The endpoint from a channel target formatted as `scheme://[authority]/endpoint`.

I believe our existing implementation already follows this behavior. I'd suggest we keep the authority handling logic as-is in this PR. Let me know if you have questions about the authority overriding.

arjan-bal:
As discussed offline, I've raised https://github.com/grpc/grpc-rust/pull/2694 with only the changes to the trait bounds for the credentials. I'll wait for Doug to review it.

arjan-bal:
Update: https://github.com/grpc/grpc-rust/pull/2703 has been merged to make the `ChannelCredentials` trait object safe. You should be able to use `Arc<dyn ChannelCredentials>` everywhere in builder.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -211,6 +211,11 @@ pub(crate) struct ResolverOptions {
     /// This value is either the result of the `default_authority` method of
     /// this `ResolverBuilder`, or another string if the channel was explicitly
     /// configured to override the default.
+    // TODO(nathanielford): This field is currently unused. Review type if/when
+    //     it is used.
+    // - Channels reference SecurityOpts, which hold the value in an Authority
+    //   object.
+    // - All other references currently use "ignored" as the authority.
     pub authority: String,
```
arjan-bal: This field is similar to the one defined in [Go's resolver.go](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fgrpc%2Fgrpc-go%2Fblob%2F1a80fca960d39ae4d7d6f2d9323ca2d243fd44bb%2Fresolver%2Fresolver.go%23L176-L178) and is intended for use by [the xDS resolver](https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fgrpc%2Fgrpc-go%2Fblame%2F1a80fca960d39ae4d7d6f2d9323ca2d243fd44bb%2Finternal%2Fxds%2Fresolver%2Fxds_resolver.go%23L172). Given this clear path, a TODO is likely unnecessary.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+
+type PresentCredentials = PresentOpt<Arc<dyn ChannelCredentials>>;
+type PresentRuntime = PresentOpt<GrpcRuntime>;
+
+pub trait IntoCredentialConfig {
```
arjan-bal: We should not need the `IntoCredentialConfig` trait anymore, we can directly use `Arc<dyn ChannelCredentials>` everywhere we need to accept the credentials.

grpc/src/client/channel.rs
```diff
@@ -616,3 +637,177 @@ fn parse_authority(host_and_port: &str) -> Authority {
...
+    #[cfg(unix)]
+    name_resolution::unix::reg();
+    #[cfg(target_os = "linux")]
+    name_resolution::unix_abstract::reg();
+    #[cfg(feature = "_runtime-tokio")]
+    tonic_transport::reg();
+}
+
+#[cfg(test)]
+mod tests {
```
arjan-bal: These tests were moved in master, I don't think they should be added back.

arjan-bal (commented):
I have left some new comments, and some of the previous feedback still needs to be addressed. Please take a look.

grpc/src/client/channel.rs
nathanielford: To be clear, this exists so that passing both arced and non-arced creds are possibe:
```
let my_creds: ChannelCredentials = ...
let my_other_creds: ChannelCredentials = ...
let my_arced_creds: Arc<dyn ChannelCredentials> = Arc::new(my_other_creds)

let my_channel = Channel::builder("https://api.darkcave.com:443")
    .credential(my_creds)
    .build()
let my_other_channel = Channel::builder("https://api.darkcave.com:443")
    .credential(my_arced_creds)
    .build()
```

In both cases they are, within the builder, coerced into a CredentialConfig, which is held for the lifetime of the builder but is not a thing the user needs even know about.

grpc/src/client/channel.rs
nathanielford: Done!

grpc/src/client/channel.rs
nathanielford: Done! I think this was held over from a previous attempt to get this working. But it compiles and tests fine now so we should be gtg.

grpc/src/client/channel.rs
nathanielford: Merge miss! Thanks!

grpc/src/client/name_resolution/mod.rs
nathanielford: I'll defer to you and remove this, but I think given that we are in a pre-scrub state our lack of signposts around what has been rigorously thought through and what was put in place for the prototype is leading to some confusion that then takes time to untangle.

grpc/src/client/channel.rs
arjan-bal: Introducing this trait adds an unnecessary layer of indirection for users trying to identify the concrete types their structs must implement. I recommend avoiding this approach.

If we want to improve ergonomics later, we can have the credential constructors return Arc-wrapped credentials directly, or add a builder method that accepts unwrapped credentials. However, it feels a bit too early to introduce these abstractions.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+    runtime: R, // Can be defaulted w/Tokio runtime feature.
+
+    // Optional values.
+    // TODO(nathanielford) In follow-up implement the following optional values:
+    // - default_service_config
+    // - http_proxy_cfg
+    // - disable_health_checks
+    // - idle_timeout
+    // - enable_channelz
+    // - keepalive_cfg
```
arjan-bal: Since many of these features are not yet implemented, adding a TODO to configure them feels premature. I recommend omitting the comment, as it lacks a tracking issue and is likely to become stale.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+    channel_authority: Option<String>, // TODO(nathanielford) Revist if this is subsumed by the SecurityOpts authority.
+}
+
+/// Impl for adding the (required) credentials to the builder.
+// This is provided as a separate builder function to allow for the possibility
+// of satisfying the credential/security configuration through different means
+// in the future (via adding methods to this impl taking different args).
```
arjan-bal: Rustdoc for `impl` blocks is usually unnecessary. Instead we should document the public methods and struct.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+impl<Runtime> ChannelBuilder<MissingOpt, Runtime> {
+    pub fn credentials(
```
arjan-bal: Please add rustdoc for public methods. I would also recommend using [documentation tests](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html) to demonstrate the usage.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+{
+    fn into_config(self) -> Arc<dyn ChannelCredentials> {
+        Arc::new(self)
+    }
+}
+
+pub struct ChannelBuilder<C, R> {
+    // Required values.
+    target: String,
+    credentials: C,
```
arjan-bal: What is the benefit of using the typed builder pattern for credentials compared to passing them directly into the builder's constructor? From my understanding, typed builders help enforce parameter configuration order, but our credentials do not seem to have any interdependencies.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+            runtime: PresentOpt(runtime),
+            channel_authority: self.channel_authority,
+        }
+    }
+}
+
+impl<C, R> ChannelBuilder<C, R> {
+    // TODO(nathanielford) Revist if this is subsumed by the SecurityOpts authority.
+    // May need to change how this is being set on the builder.
+    pub fn channel_authority(mut self, authority: impl Into<String>) -> Self {
```
arjan-bal: This is a user-provided override for the channel authority, which can also be set via the target URI or the name resolver. The user-provided override takes the highest precedence. Consequently, the authority passed to credentials in SecurityOpts is the final resolved value after evaluating all three sources.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+    fn into_config(self) -> Arc<dyn ChannelCredentials> {
+        Arc::new(self)
+    }
+}
+
+pub struct ChannelBuilder<C, R> {
+    // Required values.
+    target: String,
+    credentials: C,
+    runtime: R, // Can be defaulted w/Tokio runtime feature.
```
arjan-bal: Similar to my question above, what is the benefit of using a typed builder for the runtime? A simpler alternative would be to introduce a feature-flag-protected constructor that defaults to the Tokio runtime when enabled:

```rust
impl Channel {
    #[cfg(feature = "_runtime-tokio")]
    pub fn builder(target: impl Into<String>) -> ChannelBuilder {}

    #[cfg(not(feature = "_runtime-tokio"))]
    pub fn builder(target: impl Into<String>, rt: Arc<dyn Runtime>) -> ChannelBuilder {}
}
```

grpc/src/client/channel.rs
```diff
@@ -325,7 +345,8 @@ impl ActiveChannel {
 
         let work_scheduler = Arc::new(ResolverWorkScheduler { wqtx });
         let resolver_opts = name_resolution::ResolverOptions {
-            authority: persistent_channel.authority.clone(),
+            // authority: persistent_channel.security_opts.authority.clone(),
+            authority: "ignored".to_string(), // TODO(nathanielford) currently, this option is always ignored.
```
arjan-bal: This authority will get used by the xDS resolver, we should keep it.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+        // will be finalized with the 'Internal Channel Design' with
+        // consideration for:
+        // - error handling (inc. always-failing resolvers due to invalid targets))
+        // - testing (inc. credential and transport configuration)
+
+        // TODO(nathanielford) Find a better place to set up default registries.
+        setup_registeries();
+
+        let target = Target::from_str(self.target.as_str()).unwrap();
         let resolver_builder = global_registry().get(target.scheme()).unwrap();
```
arjan-bal, line 209: We should keep the TODO about not panicking here.

grpc/src/client/channel.rs
```diff
@@ -222,53 +138,157 @@ impl Invoke for Channel {
...
+    pub fn build(self) -> Channel {
+        // TODO(nathanielford) This construction is currently a rough-cut placeholder.
+        // The design of PersistentChannel and how and where it is initialized
+        // will be finalized with the 'Internal Channel Design' with
+        // consideration for:
+        // - error handling (inc. always-failing resolvers due to invalid targets))
+        // - testing (inc. credential and transport configuration)
+
+        // TODO(nathanielford) Find a better place to set up default registries.
+        setup_registeries();
```
arjan-bal: What's wrong with registering plugins here?

grpc/src/client/channel.rs
nathanielford: There is no indirection that the users are required to know here. The blanket implementations act as a pass-through type:
- If the credentials are passed as an arc, the implementation for CredentialConfig for arcs will leave it as-is. 
- If the credentials are not passed as an arc, that impl will arc it. The credentials are saved to the struct as a `PresentOpt<Arc<dyn ChannelCredentials>>`.
- If anything but an `Arc<dyn ChannelCredentials>` or a `ChannelCredentials` is passed, it throws a compiler error. 

At this point, any type inspection will show that the builder is of type `ChannelBuilder<PresentOpt<Arc<dyn ChannelCredentials>>, MissingOpt>` (modulo that last `MissingOpt` being filled in already).

grpc/src/client/channel.rs
nathanielford: Are we using tracking issues for any of this? I was told we weren't, really.

That said, I do intend to address all of these in near-future followups for optional values and for the ServiceConfig work (which is actively blocked on this). I'm not entirely sure the best way to track the state of work given the uncomfortable tension between a) avoiding TODOs, b) not using issues, c) avoiding huge CLs.

grpc/src/client/channel.rs
nathanielford: There are two primary reasons we discussed in the design doc:
* Type builders to accumulate options is pretty standard in the wider Rust ecosystem.
* This gives us flexibility w/o cruft around what the options are. 

The `credentials()` call is a good example of a place where there is enough uncertainty on how we want to provide credentials in the future that being able to provide paths for satisfying that requirement is advantageous. Right now, `.credentials(my_creds)` is what we provide, but we are free to add any number of functions that yield a `ChannelBuilder<PresentOpt<Arc<dyn ChannelCredentials>>`, T>`. As with the runtime, if we wanted to provide a 'default' option in the future, or one configured via env variables or whatever other method.

(To be clear, we probably _won't_ with credentials, but there are some questions around how this interacts with various concerns that giving ourselves a clear pathway to handle this cleanly is nice. Given this and the fact we are doing it with the runtime means it makes sense to keep it consistent.)

There are other nice to haves (e.g. free ordering of parameters) but the above are the two key reasons we went this route.

grpc/src/client/channel.rs
nathanielford: Idiomatically this isn't how most libraries do their builders. You land yourself in a world where a string of parameters creates craziness. Consider how the example above looks if you have two parameters with feature flag and feature flag-less pathways, and what happens to code that flips a flag. This also lets us handle optional parameters in the same fashion.

grpc/src/client/channel.rs
nathanielford: Replaced! That was an oversight, not sure how it got removed.

grpc/src/client/channel.rs
nathanielford: It might be fine, and we might just leave it. However, this sets up the default registries for the entire library for all channels. It's a bit odd for this to be the responsibility of the channel builder, though it may be the first and only reasonable place for it to happen in actual usage.

grpc/src/client/channel.rs
nathanielford: It sounds like this is safe to leave as a string then, and encapsulate the hierarchy logic in the appropriate place - the builder here shouldn't care. I've updated the comment - lmk if that works!

grpc/src/client/channel.rs
nathanielford: I would like to leave this for a follow up; the `persistent_channel.security_opts.authority` is the wrong type for `ResolverOptions`, but I don't want to make a decision about how to manage this right now since it's not directly related to the ChannelBuilder.

grpc/src/client/channel.rs
nathanielford: (Still working on this.)

grpc/src/client/channel.rs
```diff
@@ -84,71 +81,9 @@ use crate::credentials::client::ClientHandshakeInfo;
 use crate::credentials::common::Authority;
 use crate::rt;
 use crate::rt::GrpcRuntime;
+#[cfg(feature = "_runtime-tokio")]
 use crate::rt::default_runtime;
```
arjan-bal, line 85: We can remove the `default_runtime` function, the `NoOpRuntime` and directly call `TokioRuntime::default()` in the `Channel` builder constructor since it's already guarded by a feature flag now.

grpc/src/client/channel.rs
```diff
@@ -161,33 +96,29 @@ pub struct Channel {
...
-            inner: Arc::new(PersistentChannel::new(
-                target,
-                default_runtime(),
-                options,
-                credentials,
-            )),
+    ) -> ChannelBuilder {
+        ChannelBuilder {
+            target: target.into(),
+            credentials: credentials,
```
arjan-bal: nit: Please fix this clippy warning.

grpc/src/client/channel.rs
arjan-bal: I don't think we need to track these configurations outside of the gRPC Rust project plan for these features. Since these features won't be usable without builder support anyway, I think it makes sense for the feature owners to drive the builder changes directly as part of their scope.

grpc/src/client/channel.rs
arjan-bal: Lucio pointed out that idiomatic Rust avoids side effects on library import, similar to [Go's init functions](https://go.dev/doc/effective_go#init). To ensure registration is explicit yet seamless, integrating it into channel creation is our best option, as it avoids forcing users to call a separate setup function.

grpc/src/client/channel.rs
arjan-bal: The comment should come above the `unwrap` calls.

grpc/src/client/channel.rs
```diff
@@ -222,53 +153,88 @@ impl Invoke for Channel {
...
+        // - testing (inc. credential and transport configuration)
+
+        // TODO(nathanielford) Find a better place to set up default registries.
+        setup_registeries();
+
+        let target = Target::from_str(self.target.as_str()).unwrap();
         let resolver_builder = global_registry().get(target.scheme()).unwrap();
-        let authority = options
+        // TODO(nathanielford): Return errors here instead of panicking.
+        let target = name_resolution::Target::from(target);
```
arjan-bal: Please fix the warning about redundant call to `from`.

grpc/src/client/channel.rs
```diff
@@ -222,53 +153,88 @@ impl Invoke for Channel {
...
+                active_channel: Mutex::default(),
+                target,
+                security_opts,
+                runtime: self.runtime,
+                resolver_builder,
+            }),
         }
     }
 
+    pub fn channel_authority(mut self, authority: impl Into<String>) -> Self {
```
arjan-bal: We should document this. You can take inspiration from Go's dial option: https://pkg.go.dev/google.golang.org/grpc#WithAuthority

arjan-bal: nit: The word `channel` seems redundant and may be omitted since the builder is already a channel builder.

grpc/src/client/channel.rs
arjan-bal: You can call `Authority::host_port_string()` to get the canonical string representation of the authority to align the types.

grpc/src/client/channel.rs
nathanielford: Done!

grpc/src/client/channel.rs
nathanielford: Done!

grpc/src/client/channel.rs
nathanielford: Fixed!

grpc/src/client/channel.rs
nathanielford: After some discussion and experimentation with how the docs look, we're going with the 'standard' builder pattern, at least until there is a compelling reason to do things otherwise! Thanks for pushing on this, @arjan-bal !

grpc/src/client/channel.rs
nathanielford: That makes sense. I wasn't really sure, hence the comment. I'm going to leave the registry calls in their own function just for the sake of being clear they're the setup tasks, unless you see a good reason to do it in-line. I've removed the comment, though, since the call will probably stay near the top of that function.

grpc/src/client/channel.rs
nathanielford: I would like to do this in a follow-up. Right now we're using it over sixteen files so I'd rather those not pollute this change with that, if that's alright?

grpc/src/client/channel.rs
nathanielford: Done! LMK if the language is clear enough.

grpc/src/client/channel.rs
nathanielford: I believe all the doc tests are updated with this, and the public functions in channel.rs all have documentation now.

grpc/src/client/channel.rs
arjan-bal: Sounds good!

arjan-bal (approved):
LGTM!

PR #2677, opened 2026-06-05
fix(grpc): percent-decode target URI path and reject non-UTF-8 characters

arjan-bal:
This PR ensures that non-UTF-8 strings in the target URI path are rejected. 

## Key Changes:
* **`Target::from_str`:** Updated to fail for targets with non-UTF-8 symbols in the percent-decoded path.
* **`Target::path()`:** Modified to return the percent-decoded path.
* **`ResolverBuilder::default_authority()`:** percent-encode path characters disallowed by RFC 3986 section-3.2.


ejona86:
This seems weird to me. We should be assuming UTF-8-encoded URIs in general, so I could understand adding a way to get the un-decoded path or a path-as-bytes. And then only this code would do something different.

But I'd expect we'd decode URIs as UTF-8 in general, as the encoding of the URI can be different than the encoding of the on-disk path. For example, if the system is using the Latin-1 codepage (`LANG=en_US.ISO-8859-1`, not the now-typical `LANG=en_US.UTF-8`), then we'd decode the URI as UTF-8 and then (someone would) convert that to Latin-1 when opening the socket. Historically that was libc that would do that conversion. I see [Java do that conversion](https://github.com/openjdk/jdk/blob/286f7bbc81cad42357f6a7840c0cde44a6103063/src/java.base/unix/classes/sun/nio/fs/UnixPath.java#L129). It looks like [Go assumes the Linux system uses utf-8](https://cs.opensource.google/go/go/+/master:src/cmd/vendor/golang.org/x/sys/unix/zsyscall_linux.go;l=95;drc=bf3d291ea3829cac95b59303d44159ce2f95544d), which is actually fair these days. Go probably converts the UTF-8 to UTF-16 for Windows.

RFC 8089 [encourages UTF-8](https://datatracker.ietf.org/doc/html/rfc8089#section-4) as we do represent them as strings in every other language to my knowledge. https://encoding.spec.whatwg.org/#encoding assumes Unicode, and we'd definitely be using UTF-8.

So, yeah... While it is true that the files don't even need to be any encoding at all (you can store raw binary, except for `\0` and `/`), and rust allows you to do that, I don't think we want to deal with that in general.

arjan-bal:
@ejona86, to clarify, is the recommendation that gRPC should validate that the target URI contains only valid UTF-8 in the percent-decoded path?

ejona86:
Yeah, we should validate it. But either PercentDecode.decode_utf8() or decode_utf8_lossy() might work. [Java uses a replacement character.](https://github.com/openjdk/jdk/blob/21482fa7dba9aa649a37abe640910eda9d505b85/src/java.base/share/classes/java/net/URI.java#L2936). Go ~fails URL parsing~ [doesn't validate](https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/url/url.go;l=172;drc=95470667eb6613f73557a4228e0731f4bacef0ec). Dunno what C++ does.

Edit: [C++ doesn't validate](https://github.com/grpc/grpc/blob/0a8376ecfc0aa1e56a8b77ab29852c3c14b81243/src/core/util/uri.cc#L218). But C++ doesn't care if you store raw binary in strings, and Go is tolerant to invalid UTF-8 in strings. Java doesn't allow that, and neither does Rust to my knowledge.

arjan-bal:
I've reworked the approach to use `PercentDecode.decode_utf8()` during target URL parsing, which will now fail if it encounters non-UTF-8 characters. I prefer this to using replacement characters because it prevents unexpected behaviour and fails fast to ensures users are immediately aware of the invalid target so they can fix it.

grpc/src/client/name_resolution/mod.rs
```diff
@@ -136,7 +134,7 @@ impl Display for Target {
             "{}://{}{}",
             self.scheme(),
             self.authority_host_port(),
-            self.path()
+            self.decoded_path
```
ejona86: This doesn't look right. The path is being represented in a URI, so it needs to be percent-encoded.

ejona86 (commented):
TODO: I need to check callers to understand how this impacts other code.

grpc/src/client/name_resolution/mod.rs
arjan-bal: Oops, fixed it and the test case.

ejona86 (commented):
All callsites except for default_authority look fine. default_authority does need some percent encoding done on it. We can potentially say that things like XDS would need to implement their own, but in that case having a broken default doesn't seem good. If we don't want to do the percent encoding by default we could instead have the channel or whatever calls default_authority check that only permitted characters are present.

arjan-bal:
I overlooked authority encoding while implementing the `default_authority()` method. I've added percent-encoding to match the behavior in [Go](https://github.com/grpc/grpc-go/blob/5c7f93679ec9a831acce68b7368c1cb555484a9c/clientconn.go#L1887-L1942) and [Java](https://github.com/grpc/grpc-java/blob/2b86f8f42ff4906e327003bfc26a1d482c5b9fce/core/src/main/java/io/grpc/internal/GrpcUtil.java#L879-L932).

ejona86 (approved):

PR #2679, opened 2026-06-08
feat(grpc): Add resolver wrapper to handle HTTP CONNECT proxy configuration

arjan-bal:
This PR implements a resolver wrapper that determines proxy routing by checking the `HTTPS_PROXY` and `NO_PROXY` environment variables. By default, `hyper-util` matches cURL's behavior by falling back to `ALL_PROXY` when `HTTPS_PROXY` is unset. However, because Go's [`FromEnvironment`](https://pkg.go.dev/golang.org/x/net/http/httpproxy#FromEnvironment) and gRPC C++ ignore `ALL_PROXY`, gRPC Rust manually configures the `Matcher` to bypass it, ensuring cross-language consistency.

**Resolution Flow**  
When `HttpsProxyResolver::Builder` builds a resolver for a target URI, it uses `hyper_util::client::proxy::Matcher` to evaluate the environment variables:

1. **Direct Connection:** If no proxy is required, it delegates resolution entirely to the wrapped child builder, bypassing the proxy resolver.
2. **Proxied Connection:** If a proxy is required, it creates an `HttpsProxyResolver`, which uses the child DNS resolver to resolve the *proxy server's* hostname instead of the target.
3. **Attribute Injection:** The `HttpsProxyResolver` intercepts resolution updates from the child resolver. It attaches the necessary proxy configuration (the original target authority and basic auth credentials) as attributes to each resolved proxy address.

In follow-up PRs, the subchannel will read these address attributes to wrap the channel credentials and carry out the HTTP `CONNECT` handshake prior to the standard credential handshake.

Internal design doc: [go/grpc-rust-http-connect](http://go/grpc-rust-http-connect)
Transport changes: #2698


grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,641 @@
...
+
+    let builder = Matcher::builder();
+    // Only read NO_PROXY and HTTPS_PROXY. This avoids reading ALL_PROXY,
+    // which is not read by gRPC Go and C++.
+    Some(
+        builder
+            .no(get_first_env(&["NO_PROXY", "no_proxy"]))
+            .https(https_proxy)
+            .build(),
+    )
```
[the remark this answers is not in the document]
arjan-bal: `hyper-util` treats empty string as the env variable being unset. This code is similar to the code in hyper-util's `Matcher::from_env()` constructor.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,641 @@
...
+    fn new_resolver(
+        &self,
+        target: &Target,
+        options: ResolverOptions,
+        matcher: Option<&Matcher>,
+    ) -> Result<Box<dyn Resolver>, (String, ResolverOptions)> {
+        // If HTTPS_PROXY is unset, avoid parsing the target as a DNS hostname.
+        let Some(matcher) = matcher else {
+            return Ok(self.child_builder.build(target, options));
+        };
```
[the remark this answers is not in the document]
arjan-bal: Since [gRPC c-core skips](https://github.com/grpc/grpc/blob/0ed4b66878c5f095827a8778e9e67a6242138386/src/core/handshaker/http_connect/http_proxy_mapper.cc#L201-L208) proxy mapping for unix URLs, followed the same here.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,641 @@
...
+
+        work_rx.recv().await.unwrap();
+
+        let (mut channel_controller, mut update_rx) = TestChannelController::new_pair();
+        resolver.work(&mut channel_controller);
+
+        let update = update_rx.recv().await.unwrap();
+        let err = update.endpoints.unwrap_err();
+        assert!(err.contains("invalid target host in URL"));
+    }
```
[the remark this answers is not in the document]
arjan-bal: Updated the tests to ensure unix URLs skip proxy mapping, but targets with custom schemes don't.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,689 @@
...
+        Self { child_builder }
+    }
+
+    fn new_resolver(
+        &self,
+        target: &Target,
+        options: ResolverOptions,
+        matcher: Option<&Matcher>,
+    ) -> Result<Box<dyn Resolver>, (String, ResolverOptions)> {
+        // Skip proxy lookup for known non-TCP schemes.
```
ejona86: I'd expect to skip it for anything that isn't dns. I think there had been a little discussion around that, but I don't know where it ended.

The host will be passed to the proxy and the proxy then has to do the resolution, but we've discarded the scheme. It seems we are assuming dns formatting of the target string as well. We shouldn't be parsing the target string without knowledge of the specific scheme in use. Things would be better if we delegated to the name resolver to parse the target string into an authority.

arjan-bal: Answering inline:

> I'd expect to skip it for anything that isn't dns. I think there had been a little discussion around that, but I don't know where it ended.
 
The initial intention behind implementing proxy resolution as a separate resolver builder was to allow all custom resolvers to get HTTP CONNECT support for free. For context, the [C++ implementation](https://github.com/grpc/grpc/blob/3e1d845d437633917b7451bae9d226836658c0bb/src/core/handshaker/http_connect/http_proxy_mapper.cc#L201-L208) skips proxying for unix URLs, while in [Go](https://github.com/grpc/grpc-go/blob/8d66cfc36ad17184269e7dfe62da865921f0773a/internal/resolver/delegatingresolver/delegatingresolver.go#L142-L145), target host resolution can only be skipped for `dns` URLs. Java simply implements the proxy logic inside the DNS resolver itself.

I've now updated the Rust implementation to skip proxying for all non-`dns` URLs.

>  Things would be better if we delegated to the name resolver to parse the target string into an authority.

I've updated the code to use the target resolver builder for authority parsing. Based on the previous point, this will inherently be a DNS resolver.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,759 @@
...
+use std::fmt::Debug;
+use std::sync::Arc;
+use std::sync::LazyLock;
+
+use http::HeaderValue;
+use hyper_util::client::proxy::matcher::Matcher;
+use url::Url;
+
+use super::ResolverOptions;
+use super::Target;
```
dfawley: Nit: please try to avoid `use super` except in test mods.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,759 @@
...
+/// the resolved addresses.
+pub(crate) struct Builder {
+    child_builder: Arc<dyn ResolverBuilder>,
+}
+
+impl ResolverBuilder for Builder {
+    fn build(&self, target: &Target, options: ResolverOptions) -> Box<dyn Resolver> {
+        match self.new_resolver(target, options, MATCHER.as_ref()) {
+            Ok(resolver) => resolver,
+            Err((err, options)) => NopResolver::new_with_err(err, options),
```
dfawley: Optional: consider making `new_resolver` infallible and return the `NopResolver` itself instead.  It avoids needing to pass the options back and forth.

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,759 @@
...
+
+        Ok(Box::new(HttpsProxyResolver {
+            child,
+            proxy_options: Arc::new(proxy_options),
+        }))
+    }
+}
+
+struct HttpsProxyResolver {
+    child: Box<dyn Resolver>,
```
dfawley, line 195: Interestingly this is always a `dns::DnsResolver`.  Should we put that type in here directly instead of a `dyn`?

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,759 @@
...
+impl ProxyOptions {
+    /// Returns the value of the `Proxy-Authorization` header, if present.
+    pub(crate) fn proxy_authorization_header(&self) -> Option<&http::HeaderValue> {
+        self.proxy_authorization_header.as_ref()
+    }
+
+    /// Returns the address of the proxy server to connect to (host:port).
+    /// This is Punycode-encoded, i.e., it's a valid URL host:port.
+    pub(crate) fn target_authority(&self) -> &str {
+        &self.connect_addr
```
dfawley: Would it be better if the names matches?

grpc/src/client/name_resolution/proxy_resolver.rs
```diff
@@ -0,0 +1,759 @@
...
+    inner: &'a mut dyn ChannelController,
+    proxy_options: &'a Arc<ProxyOptions>,
+}
+
+impl<'a> ChannelController for InterceptingController<'a> {
+    fn update(&mut self, mut update: ResolverUpdate) -> Result<(), String> {
+        if let Ok(endpoints) = &mut update.endpoints {
+            for endpoint in endpoints {
+                for address in &mut endpoint.addresses {
+                    address.attributes = address.attributes.add(self.proxy_options.clone());
```
dfawley: I think the attribute should be somewhere more central, like in `grpc/client/transport` maybe?

grpc/src/client/name_resolution/proxy_resolver.rs
arjan-bal: Fixed and noted.

grpc/src/client/name_resolution/proxy_resolver.rs
arjan-bal: Done.

grpc/src/client/name_resolution/proxy_resolver.rs
arjan-bal: The DNS resolver builder returns a `NopResolver` if the target is an IP address here: https://github.com/grpc/grpc-rust/blob/829ce1896fa4ac12fe2ff5d283d824239be6c017/grpc/src/client/name_resolution/dns/mod.rs#L212-L217

To return a single concrete type from `dns::Builder`, we could introduce a `dns::Resolver` enum with `NopResolver` and `DnsResolver` variants and expose it via a `pub(crate)` method. 

I'm wondering if this refactoring provides any real benefit. Since `HttpsProxyResolver` only requires the methods already defined on the `Resolver` trait, staying with the trait might be cleaner. What do you think?

grpc/src/client/name_resolution/proxy_resolver.rs
arjan-bal: Renamed both, the field and the getter to `connect_authority`.

grpc/src/client/name_resolution/proxy_resolver.rs
arjan-bal: Moved the ProxyOptions and its helpers into the transport module.

grpc/src/client/name_resolution/proxy_resolver.rs
dfawley: > I'm wondering if this refactoring provides any real benefit.

No, I think this is fine.  If this was always one type I think it would make sense to just name it and have it there, but since it's more, then this way is better.

grpc/src/client/transport/mod.rs
```diff
@@ -124,7 +127,53 @@ impl<T: Transport> DynTransport for T {
 
 #[derive(Clone)]
 pub(crate) struct SecurityOpts {
-    pub(crate) credentials: Arc<dyn DynChannelCredentials>,
+    pub(crate) credentials: Arc<dyn ChannelCredentials>,
     pub(crate) authority: Authority,
     pub(crate) handshake_info: ClientHandshakeInfo,
 }
+
+/// Options for establishing an HTTP CONNECT proxy tunnel.
```
dfawley: Maybe mention that this can be added to addresses' attributes, and transports that support it will read it from there and use the address as a CONNECT proxy instead of a backend.

What about transports that don't support it, though?  Should those error?  (It seems like they should...)  We can deal with this later, but something to think about. Maybe we should have a separate address type for a proxy address or something.

grpc/src/client/transport/mod.rs
```diff
@@ -124,7 +127,53 @@ impl<T: Transport> DynTransport for T {
...
+
+impl ProxyOptions {
+    /// Creates a new `ProxyOptions`.
+    ///
+    /// # Arguments
+    /// * `target_authority` - The address of the target server to connect to (host:port).
+    ///   Must be a valid hostname.
+    /// * `proxy_authorization_header` - The value of the `Proxy-Authorization` header, if present.
+    pub(crate) fn new(
+        target_authority: String,
```
dfawley, line 154: `connect_authority`?

I'm not sure what the best name to use for this is (consider also `target_address` or `backend_address` maybe?), but it should match between field, constructor, and accessor.

dfawley (approved):

grpc/src/client/transport/mod.rs
arjan-bal: Based on the latest revision of the design, the subchannel will wrap the channel credentials in an HTTP CONNECT handshaker, making it transparent to transports. I've added this information in the rustdoc.

grpc/src/client/transport/mod.rs
arjan-bal: Changed to `target_authority` throughout.

PR #2682, opened 2026-06-12
protoc-gen-rust-grpc: Include exe inside rlib

ejona86:
OUT_DIR is a good temporary directory, but downstream users interested in caching won't have OUT_DIR available. The caching systems only preserve rlib, so we should include the important data in our library directly.

Since we're now writing out files during execution, the various methods were changed to return errors. That is then re-used if protobuf compilation was skipped during building.

CC @dfawley

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -55,16 +55,9 @@ jobs:
         if: steps.protoc-binaries-cache.outputs.cache-hit != 'true'
         shell: bash
         run: |
-          mkdir -p protoc-cache
-          # Find where cmake built the binaries and any companion DLLs.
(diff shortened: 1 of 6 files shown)


ejona86:
This didn't do what I hoped, for any system:
 * This doesn't allow us to delete OUT_DIR in our github actions caches. cargo's .d files track the files pulled in via `include_bytes!`, so if you delete those files it invalidates the build
 * This reduces the cache hit rate for sccache, while also not reducing build time. sccache only replaces rustc so cmake is still run locally. Since the c++ binaries will be included in the rust library, sccache would have a cache hit only if those binaries are identical. Today the binaries wouldn't be part of the cache key
 * This doesn't prevent rebuilding when OUT_DIR changes (e.g., if you move where your project checkout is located). cargo's .d files use absolute paths for OUT_DIR (because we must use absolute paths for `include_bytes!`), so it'll notice the old OUT_DIR is gone and rebuild. This change does _cause_ a rebuild instead of just failing in a downstream build when it can't find the old OUT_DIR
 * Ditto for other systems, for similar reasons.

PR #2683, opened 2026-06-15
chore(CI): Cache well-known-type protos with protoc binary

arjan-bal:
**Background:** In #2665, the well-known-type protos were included in the output of the `protoc-gen-rust-grpc` crate, but CI only cached `protoc` and `protoc-gen-rust-grpc`. 

**Changes in this PR:**
* Caches the included protos to easily reference them in the upcoming Google creds example (#2680).
* Removes the redundant `protoc` installation via the `taiki-e/install-action@protoc` action, since `prost-build` now works with the cached binary.
* Ensures the `protoc` plugin is rebuilt whenever the workflow file changes, even if the source code remains untouched.


ejona86:
Ah, so you were looking at this as well. I just sent out https://github.com/grpc/grpc-rust/pull/2685

I think with my change it'd be okay to not include .github/workflows/CI.yml in the cache key, since there is so little logic in CI.yml with my change. I didn't notice/remove `taiki-e/install-action@protoc`. Is that installed protoc earlier in the PATH, so we're only using the cached protoc-gen-rust-grpc today?

arjan-bal:
> Ah, so you were looking at this as well. I just sent out #2685
> 
> I think with my change it'd be okay to not include .github/workflows/CI.yml in the cache key, since there is so little logic in CI.yml with my change. I didn't notice/remove `taiki-e/install-action@protoc`. Is that installed protoc earlier in the PATH, so we're only using the cached protoc-gen-rust-grpc today?

I had asked Doug to remove it in an [earlier PR](https://github.com/grpc/grpc-rust/pull/2654/changes/1d38e508d7ef44809efa6881205fe189cc5d0516#r3309958428), but tonic's prost codegen relied on well-known type protos that were missing in the compiled protoc. Now that we're including those, we don't need the install action.

arjan-bal:
Closing in favour of https://github.com/grpc/grpc-rust/pull/2685.

PR #2684, opened 2026-06-15
protoc-gen-rust-grpc: Make only well-known protos available

ejona86:
The testing protos and internal protos are now excluded from include/.


arjan-bal (approved):
LGTM

PR #2685, opened 2026-06-15
protoc-gen-rust-grpc: Have cmake install to a separate directory

ejona86:
This allows cleaning up the install directory before builds, so no left-over artifacts are around from a previous build. Also, it makes it easier to include the well-known protos in CI.yml's protoc-cache.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,8 +56,8 @@ jobs:
         shell: bash
         run: |
           cp -r target/debug/build/protoc-gen-rust-grpc-*/out/install protoc-cache/
-          echo "Cached binaries (the biggest files):"
-          ls -la protoc-cache/bin/
+          echo "Cached files:"
+          find protoc-cache/ -type f -exec ls -lh {} +
 
   clippy:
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -55,16 +55,9 @@ jobs:
...
-            # Verify it's in the build output directory to avoid copying source or unrelated files.
-            if [[ "$file" == *"/out/bin/"* ]] || [[ "$file" == *"/out/build/bin/"* ]]; then
-              cp "$file" protoc-cache/
-            fi
-          done
-          echo "Cached files:"
-          ls -la protoc-cache/
+          cp -r target/debug/build/protoc-gen-rust-grpc-*/out/install protoc-cache/
+          echo "Cached binaries (the biggest files):"
+          ls -la protoc-cache/bin/
```
arjan-bal: optional: `find protoc-cache -type f -exec ls -lh {} +` can print all the cached files to simplify debugging. The output should not be too long.

arjan-bal (approved):
LGTM with an optional comment.

PR #2687, opened 2026-06-15
ci: Exclude targets from rust-cache

ejona86:
target/ is quite large so this substantially reduces the size of caches, at the cost of additional CPU to recompile dependencies. Pretty much all the caches become ~85 MB as they download approximately the same deps. That might also allow future cache reduction by sharing caches between jobs.

v0-rust-test-Linux is a cache that shrinks quite a bit, from 710 MB. v0-rust-codegen-Linux shrinks less, from 170 MB. The sum of caches from rust-cache would reduce from 6410 MB to 1355 MB, except I opt to continue caching slower Windows jobs so it reduces down to 2425 MB instead. We're now able to fit 3 completely unique build caches within the 10 GB limit (protoc-binaries is too big to fit 4).

To see the speed reductions, here are the wall times for a build, if we disabled targets everywhere:

Job         | Cached | Less Cache
----------- | ------ | ----------
rustfmt     |     6s |     6s
msrv        |    37s | 1m 22s
doc         | 1m 51s | 1m 46s
doc-test    |    42s | 2m 22s
semver      | 1m 38s | 1m 36s
external-types | 42s | 1m 45s
clippy      |    46s | 1m 56s
codegen     |    18s |    30s
udeps       | 5m 59s | 5m 35s
check mac   | 3m 47s | 5m 50s
check lin   | 2m 12s | 4m 59s
check win   | 6m  4s | 11m 2s
interop mac | 1m  6s | 1m 23s
interop lin |    56s | 1m 22s
interop win | 4m  7s | 5m 40s
test mac    | 2m  6s | 3m 38s
test lin    | 1m 58s | 3m 45s
test win    | 5m 25s | 9m 37s

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -160,6 +160,7 @@ jobs:
       run: echo "${{ github.workspace }}/protoc-cache" >> $GITHUB_PATH
     - uses: Swatinem/rust-cache@v2
       with:
+        # Windows is the long pole of a build. Caching it avoids adding ~5 minutes of build time
         cache-targets: ${{ matrix.os == 'windows-latest' }}
     - name: Check features
       run: cargo hack check --workspace --no-private --each-feature --no-dev-deps
@@ -238,6 +239,7 @@ jobs:
       run: echo "${{ github.workspace }}/protoc-cache" >> $GITHUB_PATH
     - uses: Swatinem/rust-cache@v2
       with:
+        # Windows is the long pole of a build. Caching it avoids adding ~4 minutes of build time
         cache-targets: ${{ matrix.os == 'windows-latest' }}
     - run: cargo nextest run --workspace --all-features
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -133,14 +150,17 @@ jobs:
...
       with:
         path: protoc-cache
         key: protoc-binaries-${{ runner.os }}-${{ hashFiles('protoc-gen-rust-grpc/src/cpp_source/**', 'protoc-gen-rust-grpc/build.rs') }}
     - name: Add cached binaries to PATH
       shell: bash
-      run: echo "${{ github.workspace }}/protoc-cache/bin" >> $GITHUB_PATH
+      run: echo "${{ github.workspace }}/protoc-cache" >> $GITHUB_PATH
     - uses: Swatinem/rust-cache@v2
+      with:
+        cache-targets: ${{ matrix.os == 'windows-latest' }}
```
arjan-bal, line 164: It would be helpful to add a comment explaining why caching is only enabled for specific steps on Windows. My understanding is that `check` and `test` are the only Windows steps that take over 6 minutes without it, and using a target cache ensures all workflows finish within that limit.

arjan-bal (approved):
LGTM!

PR #2689, opened 2026-06-16
fix(tonic-xds): consume immediate first tick in OD housekeeping actor

LYZJU2019:
## Summary

A flaky test was reported ([link](https://github.com/grpc/grpc-rust/pull/2619#issuecomment-4711913222)) after #2619 was merged.

The root cause is: `tokio::time::interval(period)` fires immediately on its first poll, so the outlier-detection housekeeping actor was running `run_housekeeping` at t≈0 — before the data path had recorded anything. That early sweep calls `snapshot_and_reset` and zeros whatever counters have landed. Therefore, `test_outlier_detection_reinsert_preserves_state` flakes with `(0, 0) != (3, 0)` when the actor's first tick races the assertion read.

Fix: consume the immediate tick before entering the loop so the first real sweep fires one full `interval` after spawn, matching the gRFC A50 semantics.

## Testing

\`cargo test -p tonic-xds --lib test_outlier_detection_reinsert_preserves_state\` — 10/10 pass under stress.


gu0keno0:
Thanks for the fix. maybe https://docs.rs/tokio/latest/tokio/time/fn.interval_at.html is a better way?

gu0keno0 (approved):

PR #2690, opened 2026-06-16
grpc/metadata: add as_bytes() for binary values

arjan-bal:
Currently, there is no way to get the underlying bytes from a `MetadataValue<Binary>`. This gap wasn't caught by existing tests since they only checked for equality. This commit moves `as_bytes()` from `impl MetadataValue<Ascii>` to the shared `impl MetadataValue<VE>` block to expose this functionality.

### DIFF
--- a/grpc/src/metadata/value.rs
+++ b/grpc/src/metadata/value.rs
@@ -192,6 +192,23 @@ impl<VE> MetadataValue<VE> {
         self.inner.is_sensitive
     }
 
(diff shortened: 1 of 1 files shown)


sauravzg (approved):

sauravzg:
Unit tests maybe or not worth it ?

arjan-bal:
> Unit tests maybe or not worth it ?

The doc comment is a test. It runs as part of cargo test.

PR #2691, opened 2026-06-16
grpc: propagate trailers-only metadata

arjan-bal:
When a Tonic client receives a trailers-only response, the metadata is embedded within the returned `Status`. Previously, the Tonic transport was not reading the metadata from this status. This change fixes that issue. 

Additionally, it updates the Tonic to gRPC metadata map conversion to accept a reference rather than an owned value. This simplifies the transport code by avoiding the need to deconstruct the Tonic `Status` into its underlying code, message, and metadata components.


grpc/src/client/transport/tonic/test.rs
```diff
@@ -782,6 +805,59 @@ async fn tonic_transport_recv_drop_cancels_send() {
...
 }
 
+#[tokio::test]
+async fn trailers_only_metadata() {
+    let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let addr = listener.local_addr().unwrap();
+    let shutdown_notify = Arc::new(Notify::new());
+    let shutdown_notify_copy = shutdown_notify.clone();
+
+    // Prepare custom metadata for the server resposnse.
```
sauravzg: nit: typo

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -292,22 +291,27 @@ impl RecvStream for TonicRecvStream {
                                     StatusCodeError::Internal,
                                     format!("error decoding response: {e}"),
                                 )),
-                                None,
+                                &TonicMeta::default(),
```
sauravzg, line 293: Mostly a nit, since it largely doesn't matter , since `default` for tonic metadata shouldn't result in any heap allocations.

We replaced Option<T> with T& and then construct default instances to imply absence. 

Should we have used the concept of `optional reference` instead to distinguish between presence , emptiness and value ? or should we treat metadata as a container type where traditionally empty and absence tend to mean the same thing?

sauravzg (approved):

grpc/src/client/transport/tonic/test.rs
arjan-bal: Fixed.

grpc/src/client/transport/tonic/mod.rs
arjan-bal: Since the gRPC Trailers contain a non-optional `MetadataMap`, an empty Tonic Metadata and `None` are handled in the same way. Practically,  TonicMetadata is a wrapper around http::HeaderMap, which [guarantees](https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.new) that an empty map doesn't allocate. Not making any changes for now.

PR #2692, opened 2026-06-16
grpc: remove redundant timeout in transport

arjan-bal:
Currently, the subchannel propagates a connect timeout to the transport layer, which applies it only during TCP connection creation. While this timeout should also cover the credentials handshake, passing it to the transport is actually redundant because the subchannel already enforces it here:
https://github.com/grpc/grpc-rust/blob/db5409026a64868d69d93b26ac2efec9495edb0c/grpc/src/client/subchannel.rs#L363-L385

This PR removes the redundant timeout from the transport layer and adds a test to verify that the timeout correctly applies to the security client handshake. Additionally, since the channel constructor currently hardcodes a 20-second connect timeout, the new test utilizes Tokio's virtual time to fast-forward the clock and prevent the test from blocking for 20 seconds.


grpc/src/client/transport/tonic/test.rs
```diff
@@ -782,6 +791,108 @@ async fn tonic_transport_recv_drop_cancels_send() {
...
+        &target,
+        SlowChannelCredentials::new_arc(Duration::from_secs(21)),
+        Default::default(),
+    );
+
+    // Spawn the RPC call because it will block waiting for connection.
+    let rpc_handle = tokio::spawn(async move { perform_unary_echo_failure(&channel).await });
+
+    // Advance time to trigger the timeout in subchannel connect.
+    time::pause();
```
sauravzg: Wouldn't pausing time after spawning result in a race where the spawned task may be using real time before  the time is paused? 


Regardless, Is it a conscious choice to not use `start_paused=true` i.e. do we need the pause, sleep , resume to happen within the test or can we pause time for the entire test ? I'd prefer the latter , since mixing real time and fake time is rather annoying to deal with. 

https://tokio.rs/tokio/topics/testing#pausing-and-resuming-time-in-tests

sauravzg (approved):

grpc/src/client/transport/tonic/test.rs
arjan-bal: Yes, I agree, the time should be paused before calling `perform_unary_echo_failure`. Pausing the time at the beginning is better. I didn't do it to make it clear which part of the test requires time to be skipped. Made the change now.

PR #2693, opened 2026-06-16
examples: gRPC Google credentials example

arjan-bal:
This PR adds an example demonstrating how to call the Cloud Pub/Sub API using Google call credentials and TLS with system root certificates, similar to the [tonic GCP](https://github.com/grpc/grpc-rust/tree/master/examples/src/gcp) example.

The example utilizes protobuf well-known types (WKT) crates. I've also added `From<protobuf_codegen::Dependency> for Dependency` to easily convert dependency types returned by WKT/codegen crates into the format  expected by `grpc_protobuf_build`.

Unlike other gRPC examples, the generated code is not checked-in as it's approximately 60k lines https://github.com/grpc/grpc-rust/commit/15b7f7f97a386a601ea0602c481fc0c24db149df.

### DIFF
--- a/examples/build.rs
+++ b/examples/build.rs
@@ -1,4 +1,5 @@
-use std::{env, path::PathBuf};
+use std::env;
+use std::path::PathBuf;
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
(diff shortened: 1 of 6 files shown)


examples/build.rs
```diff
@@ -82,6 +83,31 @@ fn main() {
             .compile()
             .unwrap();
     }
+
+    if env::var_os("CARGO_FEATURE_GRPC_GCP").is_some() {
+        let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+        let dependencies = protobuf_well_known_types::get_dependency("protobuf_well_known_types")
```
ejona86, line 89: This dependency system is pretty nice. So you did https://github.com/grpc/grpc-rust/pull/2683 _just_ for descriptor.proto. Not for any of the regular well known types. That means we might want to exclude the well known types from protoc-gen-rust-grpc in the future. But worst-case we'll just copy descriptor.proto here.

ejona86 (approved):

PR #2694, opened 2026-06-17
grpc/client: fix trait bounds on channel credentials to avoid double Arcing

arjan-bal:
## Background

The channel constructor currently accepts an `Arc<C>` because the credentials might be reused to create "out of band" (OOB) channels (e.g., to an RLS server). If we accepted raw credentials and wrapped them inside the constructor instead, creating an OOB channel would result in a second, redundant `Arc` wrapper, leading to unnecessary pointer indirection and reference counting.

## Problem

Passing an `Arc<dyn DynChannelCredentials>` to `Channel::new` currently fails to compile, even if we add an `impl ChannelCredentials for dyn DynChannelCredentials`.

```rust
error[E0277]: the size for values of type `dyn dyn_wrapper::DynChannelCredentials` cannot be known at compilation time
   --> grpc/src/credentials/dyn_wrapper.rs:307:38
    |
307 |         let ch = Channel::new("abc", creds, ChannelOptions::default());
    |                  ------------        ^^^^^ doesn't have a size known at compile-time
    |                  |
    |                  required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `dyn dyn_wrapper::DynChannelCredentials`
note: required by an implicit `Sized` bound in `channel::Channel::new`
   --> grpc/src/client/channel.rs:170:16
    |
170 |     pub fn new<C>(target: impl Into<String>, credentials: Arc<C>, options: ChannelOptions) -> Self
    |                ^ required by the implicit `Sized` requirement on this type parameter in `Channel::new`
help: consider relaxing the implicit `Sized` restriction
   --> grpc/src/client/channel.rs:172:30
    |
172 |         C: ChannelCredentials + ?Sized,
    |                               ++++++++

For more information about this error, try `rustc --explain E0277`.

```

This happens because the generic type parameter `C` in `Channel::new` has an implicit `Sized` bound, which trait objects like `dyn DynChannelCredentials` do not satisfy.

## Fix

This PR introduces an opaque wrapper struct, `SharedChannelCredentials`, and updates the channel constructor to accept `impl Into<SharedChannelCredentials>`.

Specifically, this change:

* Adds a `From<Arc<C>>` implementation to convert any valid `Arc<C: ChannelCredentials>` into `SharedChannelCredentials`.
* Implements the `ChannelCredentials` trait directly on `SharedChannelCredentials` so it can be used seamlessly.

This resolves the compilation error while preserving the optimization against double-`Arc`ing. An existing `SharedChannelCredentials` object can be passed directly to an OOB channel; calling `.into()` on it is essentially a no-op, securely reusing the existing reference count.

### DIFF
--- a/grpc/src/client/channel.rs
+++ b/grpc/src/client/channel.rs
@@ -80,12 +80,10 @@ use crate::client::transport::TransportRegistry;
 #[cfg(feature = "_runtime-tokio")]
 use crate::client::transport::tonic as tonic_transport;
 use crate::core::RequestHeaders;
-use crate::credentials::ChannelCredentials;
+use crate::credentials::SharedChannelCredentials;
 use crate::credentials::client::ClientHandshakeInfo;
 use crate::credentials::common::Authority;
-use crate::credentials::dyn_wrapper::DynChannelCredentials;
 use crate::rt;
-use crate::rt::GrpcEndpoint;
 use crate::rt::GrpcRuntime;
 use crate::rt::default_runtime;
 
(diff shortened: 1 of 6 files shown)


arjan-bal:
We discussed this in the maintainers meeting and decided to make the `ChannelCredentials` dyn Compatable, avoiding the need for the `Into` and opaque struct. I'll be raising a PR with that change.

PR #2686, opened 2026-06-15
feat(tonic-xds): compute gRFC A42 request hash from header hash policy

madhurishgupta:
### Summary

Implement gRFC [A42: xDS Ring Hash LB Policy](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md) in tonic-xds. It lands the per-request hash computation.

What this PR does

- Adds a xds::resource::hash_policy module with HashPolicyConfig and the request-hash logic as impl methods:
  - HashPolicyConfig::request_hash(headers, policies) — folds each matching policy's hash into one value using XXH64 (seed 0, per A42) and Envoy's reference rotate-XOR combination (combined = rotl(combined, 1) ^ new), honoring the terminal flag; returns None if no policy produced a hash (picker then uses a random hash).
- Adds RouteDecision.request_hash: Option<u64>, carried from the routing layer to the LB picker via request extensions.
- resolve_route computes request_hash (currently with an empty policy list — TODO to populate from RDS RouteAction.hash_policy in a later PR).

### Tests
Added UTs. 
Currently this path won't be called in production.

### Plan (A42 series)

1. (this PR) Request-hash computation + plumbing.
2. Ring-hash picker + member tracking — builds an A42-conformant ring (min/max ring size, XXH64-keyed vnodes) over the cluster's healthy endpoints and routes each request by its hash; falls through to the next ready endpoint on a miss.
3. CDS wiring — parse lb_policy: RING_HASH and ring_hash_lb_config (validating hash_function == XX_HASH), and select the ring-hash picker from the cluster's LB policy.
4. RDS wiring — parse RouteAction.hash_policy to populate the policy list (replacing the empty scaffold here).

PR #2695, opened 2026-06-19
feat(tonic-xds): add gRFC A42 ring-hash picker + member tracking

madhurishgupta:
## Summary

Adds the [gRFC A42](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md) ring-hash load-balancing picker to the loadbalance stack, giving consistent-hash request affinity: requests carrying the same hash key are routed to the same backend. 

## What's included

- RingHashPicker (pickers/ring_hash.rs)
  - Builds a hash ring over the cluster's full healthy-EDS membership with uniform per-member weighting. 
  - Entries keyed xxh64("{addr}_{i}", 0) (XXH64, seed 0); the ring is held lock-free behind an ArcSwap and rebuilt only on membership change.
  - pick() reads RouteDecision.request_hash (per-request random fallback when absent), finds the ring position closest to that hash (first entry with hash ≥ request, wrapping), and walks clockwise to the first ready host — returning Unavailable if no ring host is ready.
- LoadBalancer member tracking: tracks the full healthy-EDS member set (independent of connection/ejection state), rebuilding the picker's ring once per discovery drain.

## Behavior notes

- Outlier detection composes with no ring-hash-specific code. The ring is built over members; picks resolve against the ready set. An ejected host stays in the ring but is absent from ready, so its keys fall through clockwise to the next ready host.

## Testing
Added UTs.
cargo fmt, clippy, and cargo test -p tonic-xds all clean.

##nPlan (A42 series)

- Last PR: Request-hash computation + plumbing. (https://github.com/grpc/grpc-rust/pull/2686)
- This PR: Add gRFC A42 ring-hash picker + member tracking
- Third PR: CDS wiring — parse lb_policy: RING_HASH and ring_hash_lb_config (validating hash_function == XX_HASH), and select the ring-hash picker from the cluster's LB policy.
- Fourth PR: RDS wiring — parse RouteAction.hash_policy to populate the policy list (replacing the empty scaffold here).

### DIFF
--- a/tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
+++ b/tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
@@ -50,29 +50,12 @@ struct RingEntry {
     addr: EndpointAddress,
 }
 
-/// A ring-hash picker. The ring is held behind an [`ArcSwap`] so the hot path
-/// reads it lock-free while membership rebuilds publish a new ring atomically.
-pub(crate) struct RingHashPicker {
-    config: RingHashConfig,
-    ring: ArcSwap<Vec<RingEntry>>,
-}
+/// A built hash ring: entries kept sorted by hash. [`Ring::new`] is the only
+/// way to construct one, so a `Ring` is always sorted and the binary search in
+/// [`Ring::pick`] is sound.
+struct Ring(Vec<RingEntry>);
 
-impl RingHashPicker {
-    /// Create a picker with an empty ring. The ring is populated on the first
-    /// [`ChannelPicker::on_members_changed`] call.
-    pub(crate) fn new(config: RingHashConfig) -> Self {
-        Self {
-            config,
-            ring: ArcSwap::from_pointee(Vec::new()),
-        }
-    }
-
-    /// Rebuild the ring over the given members and publish it atomically.
-    pub(crate) fn rebuild(&self, members: &IndexSet<EndpointAddress>) {
-        let ring = Self::build_ring(&self.config, members);
-        self.ring.store(Arc::new(ring));
-    }
-
+impl Ring {
     /// Build the ring over `members` with uniform weights. The ring size
     /// is the smallest multiple of `N` that is ≥ `min_ring_size` — the
     /// uniform-weight case of A42's "smallest number giving the smallest-weight
@@ -83,10 +66,10 @@ impl RingHashPicker {
     /// entries. Each entry is keyed
     /// `xxh64("{addr}_{i}", 0)`, `i` being the member's previous appearance
     /// count, and entries are then sorted by hash.
-    fn build_ring(config: &RingHashConfig, members: &IndexSet<EndpointAddress>) -> Vec<RingEntry> {
+    fn new(config: &RingHashConfig, members: &IndexSet<EndpointAddress>) -> Self {
         let n = members.len() as u64;
         if n == 0 {
-            return Vec::new();
+            return Ring(Vec::new());
         }
         // Smallest multiple of N ≥ min_ring_size (so the smallest-weight host
         // gets a whole number of entries), hard-clamped to max_ring_size.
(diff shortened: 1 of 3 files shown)


tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
```diff
@@ -0,0 +1,443 @@
...
+use arc_swap::ArcSwap;
+use indexmap::{IndexMap, IndexSet};
+
+use crate::client::endpoint::EndpointAddress;
+use crate::client::loadbalance::pickers::ChannelPicker;
+use crate::client::route::RouteDecision;
+
+/// Ring-hash LB configuration (gRFC A42 `ring_hash_lb_config`).
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct RingHashConfig {
```
LYZJU2019, line 30: Maybe add a validation to `RingHashConfig`? This can prevent massive vector allocation for `ring` object when the config value is invalid.

YutaoMa:
Is the plan to only support uniform weighting after the 4 PRs? A42 requires EDS and locality weighting. It's ok to defer it after this PR but to declare proper A42 support adding weight support is needed.

tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
```diff
@@ -0,0 +1,443 @@
...
+        // member k owns entries until the running total reaches
+        // ceil(ring_size * (k + 1) / N). Each member therefore gets floor or
+        // ceil of ring_size / N, the total is exactly ring_size, and when
+        // N > ring_size the trailing members get zero entries.
+        let mut ring = Vec::with_capacity(ring_size as usize);
+        let mut emitted: u64 = 0;
+        for (k, addr) in members.iter().enumerate() {
+            let target = (ring_size as u128 * (k as u128 + 1)).div_ceil(n as u128) as u64;
+            for i in 0..(target - emitted) {
+                let key = format!("{addr}_{i}");
```
YutaoMa: can reuse the same String buffer in each iteration instead of allocating every time

tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
madhurishgupta: This validation is part of third PR in which CDS wiring will be implemented.

tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
```diff
@@ -0,0 +1,443 @@
...
+    /// is the smallest multiple of `N` that is ≥ `min_ring_size` — the
+    /// uniform-weight case of A42's "smallest number giving the smallest-weight
+    /// host a whole number of entries" — hard-clamped to `max_ring_size`. Those
+    /// entries are spread across members by cumulative rounding, so each gets
+    /// floor or ceil of `ring_size / N` and the total is exactly `ring_size`;
+    /// when members outnumber the clamped size the trailing ones get zero
+    /// entries. Each entry is keyed
+    /// `xxh64("{addr}_{i}", 0)`, `i` being the member's previous appearance
+    /// count, and entries are then sorted by hash.
+    fn build_ring(config: &RingHashConfig, members: &IndexSet<EndpointAddress>) -> Vec<RingEntry> {
```
YutaoMa: one Rust idiomatic style suggestion: use a new type pattern for `Ring(Vec<RingEntry>)` so that only valid rings can be represented. Can then move most of the build and pick logic into method on the Ring type, and the picker does minimal delegation between config and the ring.

tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
madhurishgupta: Makes sense, made the change

tonic-xds/src/client/loadbalance/pickers/ring_hash.rs
madhurishgupta: Thanks for mentioning this, it makes sense.
Create a new struct `Ring` with `new` and `pick` functions.

madhurishgupta:
> Is the plan to only support uniform weighting after the 4 PRs? A42 requires EDS and locality weighting. It's ok to defer it after this PR but to declare proper A42 support adding weight support is needed.

Correct. Uniform weighting is an intentional initial scope for this PR.
Full A42 conformance needs weight-proportional rings and that's planned as a dedicated follow-up.

YutaoMa (approved):

PR #2697, opened 2026-06-21
feat(transport): allow setting a custom rustls CryptoProvider on TLS configs

### LINKED ISSUE #2696: Allow configuring a custom rustls CryptoProvider on ClientTlsConfig and ServerTlsConfig
#### Motivation

`ClientTlsConfig` and `ServerTlsConfig` build their rustls configuration from
`CryptoProvider::get_default()` (in `channel/service/tls.rs` and
`server/service/tls.rs`), falling back to the provider selected by the
`tls-ring` or `tls-aws-lc` feature. There is no way to set the `CryptoProvider`
for a specific channel or server.

So an application that wants a non-default provider, for example a FIPS build or
a provider that constrains the cipher suites and key-exchange groups (such as a
post-quantum profile), can only do so by installing it process-wide with
`CryptoProvider::install_default`. That is all or nothing: it forces every
rustls user in the process onto the same provider, which is often undesirable
when only some connections should use the custom posture.

#### Proposal

Add a `with_provider` builder method to both configs:

```rust
impl ClientTlsConfig {
    pub fn with_provider(self, provider: Arc<rustls::crypto::CryptoProvider>) -> Self;
}
impl ServerTlsConfig {
    pub fn with_provider(self, provider: Arc<rustls::crypto::CryptoProvider>) -> Self;
}
```

When set, tonic builds the rustls config with `builder_with_provider(provider)`
instead of `get_default()`; roots, identity, ALPN, and client auth are layered
on top unchanged. The field is optional and defaults to current behavior, so it
is backward compatible.

This mirrors the existing escape hatch for custom certificate verifiers (#2612),
and composes with tonic's existing configuration rather than replacing it, which
avoids the conflicts a full `ClientConfig`/`ServerConfig` override would
introduce with roots, identity, and ALPN.

I have an implementation ready and will open a PR.

jgowdy:
## Motivation

`ClientTlsConfig` and `ServerTlsConfig` build their rustls configuration from
`CryptoProvider::get_default()`, so a non-default provider can only be used by
installing it process-wide with `install_default`. That forces every rustls user
in the process onto the same provider. Some applications want a non-default
provider, for example a FIPS build or a provider that constrains the cipher
suites and key-exchange groups (such as a post-quantum profile), on specific
channels or servers only.

Closes #2696.

## Solution

Add a `with_provider(Arc<CryptoProvider>)` builder method to both
`ClientTlsConfig` and `ServerTlsConfig`. When set, tonic builds the rustls config
with `builder_with_provider` instead of `get_default()`; roots, identity, ALPN,
and client auth are layered on top unchanged. The field is optional and defaults
to the existing behavior, so this is backward compatible.

This follows the pattern of the custom `ServerCertVerifier` escape hatch (#2612).

A unit test covers that a client connector built with an explicit provider
succeeds without a process-wide default installed. The server path is symmetric.
There is no integration-test harness for `tonic::transport` TLS in the repo, and
the example certs live outside the `tonic` crate, so I kept the test
self-contained. Happy to add an integration test if you would prefer one.

### DIFF
--- a/tonic/src/transport/channel/service/tls.rs
+++ b/tonic/src/transport/channel/service/tls.rs
@@ -35,6 +35,7 @@ impl TlsConnector {
         trust_anchors: Vec<TrustAnchor<'static>>,
         identity: Option<Identity>,
         server_cert_verifier: Option<Arc<dyn ServerCertVerifier>>,
+        provider: Option<Arc<crypto::CryptoProvider>>,
         domain: &str,
         assume_http2: bool,
         use_key_log: bool,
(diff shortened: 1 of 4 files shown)


nathanielford:
Hey there! Thanks for the contribution. We're not going to land this because we are winding down Tonic development and the new grpc-rust library will have a modular credential architecture that will allow this provider to be specified without leaking the core channel abstraction. If there are other concerns not addressed by this, or you'd like us to reconsider merging this into Tonic (which may never get another release at this point), let us know! Thanks again!

PR #2698, opened 2026-06-22
grpc/client: HTTP CONNECT proxy support in transport

arjan-bal:
This PR introduces an `HttpConnectHandshaker` that wraps the `ChannelCredentials` to complete the HTTP CONNECT handshake based on the `Address` attributes introduced in #2679. This allows the subchannel to configure proxying transparently without requiring changes to the Transport API.


grpc/src/client/channel.rs
```diff
@@ -248,6 +249,7 @@ impl PersistentChannel {
         // TODO(nathanielford): Return errors here instead of panicking.
         let target = Target::from_str(&target.into()).unwrap();
         let resolver_builder = global_registry().get(target.scheme()).unwrap();
+        let resolver_builder = Arc::new(proxy_resolver::Builder::new(resolver_builder));
```
dfawley: Wondering if we should have a `proxy_resolver::Builder::new_arc` instead that returns `child_builder` directly if `child_builder.scheme() != "dns"`?

grpc/src/client/subchannel.rs
```diff
@@ -306,11 +308,18 @@ impl InternalSubchannel {
         transport: Arc<dyn DynTransport>,
         backoff: Arc<dyn Backoff>,
         runtime: GrpcRuntime,
-        security_opts: SecurityOpts,
+        mut security_opts: SecurityOpts,
         work_queue: WorkQueueTx,
     ) -> Arc<dyn Subchannel> {
         let on_drop = Arc::new(Notify::new());
         let address_string = address.address.to_string();
+        let transport_options = TransportOptions::default();
```
dfawley: This is unused -- remove or change line 335 to use it (and transfer that comment here)?

grpc/src/rt/mod.rs
```diff
@@ -249,6 +249,92 @@ impl<T: GrpcEndpoint> AsyncWrite for AsyncIoAdapter<T> {
     }
 }
 
+/// A wrapper around an asynchronous I/O stream that implements
+/// [GrpcEndpoint].
+pub(crate) struct TokioIoStream<T> {
```
dfawley: Should this still mention `Tokio` if it's not tokio-specific?

grpc/src/client/transport/http_connect/rewind.rs
```diff
@@ -0,0 +1,114 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+// Note: This file contains modifications by the gRPC authors; see revision
+// history for details.
```
dfawley: I don't think the revision history is going to help here, is it?  Because our first version will have the edits.

grpc/src/client/transport/http_connect/mod.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// A credential wrapper that performs an HTTP CONNECT handshake before
+/// delegating to an inner security credential (like TLS).
+pub(crate) struct HttpConnectHandshaker {
+    inner: Arc<dyn ChannelCredentials>,
+    options: ProxyOptions,
+}
+
+impl HttpConnectHandshaker {
+    /// Constructs a new `ProxyChannelCredentials` wrapping the inner credentials.
```
dfawley: `ProxyChannelCredentials` -> `HttpConnectHandshaker`?

dfawley (commented):
I still need to look at the tests, but here's a first round of comments (all pretty minor I think).

grpc/src/client/transport/http_connect/mod.rs
arjan-bal: Fixed.

grpc/src/client/transport/http_connect/rewind.rs
arjan-bal: To address this, I've listed the initial modifications in the file header and noted that the git history will track all subsequent changes.

grpc/src/client/subchannel.rs
arjan-bal: Good catch, removed. This was a leftover artifact from reverting an earlier implementation that passed proxy configurations via `TransportOptions` to the credentials.

grpc/src/rt/mod.rs
arjan-bal: There are now two similar structs in this module: one that implements `GrpcEndpoint` for an async I/O stream, and another that implements `AsyncRead + AsyncWrite` for a `GrpcEndpoint`. To make this distinction clear, I've renamed them to be symmetrical:
* `AsyncIoAdapter` is now `EndpointIoStream`.
* `TokioIoStream` is now `StreamEndpoint`.

grpc/src/client/channel.rs
arjan-bal: Done.

grpc/src/client/transport/http_connect/test.rs
```diff
@@ -0,0 +1,590 @@
...
+    let server_key = fs::read(certs_path.join("server.key")).expect("failed to read server.key");
+    let ca_cert = fs::read(certs_path.join("ca.pem")).expect("failed to read ca.pem");
+
+    let identity = Identity::from_pem(server_cert, server_key);
+    let identity_provider = StaticProvider::new(vec![identity]);
+    let server_tls_config = ServerTlsConfig::new(identity_provider);
+    let server_creds = RustlsServerCredendials::new(server_tls_config).unwrap();
+
+    let server_listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+    let server_addr = server_listener.local_addr().unwrap();
```
dfawley: Maybe we can make some helpers to avoid some of the duplication here?  Even if it's broken up into a few smaller and reusable parts, like one function that creates the credentials, that could be helpful.

dfawley (approved):

grpc/src/client/transport/http_connect/test.rs
arjan-bal: Created helper functions to construct client/server credentials, perform the CONNECT handshake, and initialize the server.

Issue #2666, opened 2026-05-29
Add grpc-rust to the official benchmark


BijanVan:
## Feature Request
Add grpc-rust to the official benchmark.
https://grafana-dot-grpc-testing.appspot.com/
https://grpc.io/docs/guides/benchmarking/

dfawley:
Thanks for filing this issue.  We full intend to do this; hopefully within the next month or so.

PR #2699, opened 2026-06-22
grpc-benchmark: add generated code and TLS certificates

arjan-bal:
Contributes to: #2666 

This PR add the proto files and TLS certificates required to run the [gRPC benchmarks](https://grpc.io/docs/guides/benchmarking/).

The proto files are copied from the [grpc-proto repository](https://github.com/grpc/grpc-proto/tree/master/grpc/testing).

The TLS certs are copied from [the gRPC Go repo](https://github.com/grpc/grpc-go/tree/8d66cfc36ad17184269e7dfe62da865921f0773a/testdata) which seem to be copied from the [core repo](https://github.com/grpc/grpc/tree/3e1d845d437633917b7451bae9d226836658c0bb/src/core/tsi/test_creds).


grpc-benchmark/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "grpc-benchmark"
+version = "0.0.0-alpha"
```
ejona86: Are we planning on publishing this crate in the future? Should we just drop the version field? I think that would mean we then wouldn't need `publish = false`

grpc-benchmark/proto/grpc/testing/benchmark_service.proto
```diff
@@ -0,0 +1,48 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// An integration test service that covers all the method signature permutations
+// of unary/streaming requests/responses.
+syntax = "proto3";
+
+package grpc.testing;
+
+import "messages.proto";
```
ejona86: This file has been modified from what is in grpc-proto? Can we not make them identical?

I noticed because build.rs's include() was wrong; it needs to point to the protobuf root.

grpc-benchmark/build.rs
```diff
@@ -0,0 +1,65 @@
...
+    let behchmark_service_grpc = out_dir.join("benchmark_service/grpc");
+    let behchmark_service_tonic = out_dir.join("benchmark_service/tonic");
+
+    grpc_protobuf_build::CodeGen::new()
+        .include("proto/grpc/testing")
+        .inputs(["benchmark_service.proto", "messages.proto"])
+        .output_dir(behchmark_service_grpc)
+        .client_only()
+        .compile()
+        .unwrap();
```
ejona86: Why do two separate invocations of grpc_protobuf_build?

ejona86 (commented):
It seems like files are missing; I don't see the actual benchmark workers.

grpc-benchmark/Cargo.toml
arjan-bal: No, this crate doesn't need to be published. Removed the `version` and `publish` fields.

grpc-benchmark/build.rs
arjan-bal: I used two tonic-build invocations when benchmarking Tonic in 2024, though I don't recall why. A single gRPC protobuf build invocation works perfectly, so I've combined them.

grpc-benchmark/proto/grpc/testing/benchmark_service.proto
arjan-bal: Copied the exact file from the `grpc-protos` repo. I changed the imports to quickly work around some build errors I was seeing when benchmarking Tonic 1.5 years ago.

arjan-bal:
> It seems like files are missing; I don't see the actual benchmark workers.

Now I've copied all the files over from https://github.com/grpc/grpc-proto/blob/master/grpc/testing, except `test.proto` which seems to be used for xDS tests. The worker service is defined in `grpc-benchmark/proto/grpc/testing/worker_service.proto`.

arjan-bal:
While working on the worker implementation, I realized that we need a worker server (not a client), so I switched to generating the worker service code for Tonic.

ejona86 (commented):
I still don't understand this PR. How does this fix #2666? I don't see any client/server code that implements any functionality.

arjan-bal:
This is the first in a series of PRs addressing #2666. I intentionally used "Addresses" instead of "Fixes" so GitHub links the PR without automatically closing the issue.

I originally drafted the benchmarking binary using `tonic` clients and servers last year (see https://github.com/arjan-bal/tonic/pull/1/). To make the code easier to review, I'm splitting the changes into self-contained parts. The next PR in the chain, which contains the worker and benchmark server implementation, is #2706.

grpc-benchmark/data/tls/server1.key
ejona86, line 1: What is `server.key`? Should this be `server1.key`? I don't see how someone would know that these are specific shared files, especially if they are renamed. We need some documentation saying this is a copy from grpc/src/core/tsi .

ejona86:
But "addresses" means the same as "fixes" in this context, as it implies completion. So this specific PR is _not_ addressing the issue. I typically see language like "contributes to" or "as part of addressing" for this sort of thing. (The README was similarly misleading in that it said it contained the implementations.)

ejona86 (approved):
Fix the "addresses" language to make clear this is only the first piece.

grpc-benchmark/data/tls/server1.key
arjan-bal: Added READMEs in both the `/data/tls` and the `/proto/` directories mentioning the source of the files.

PR #2700, opened 2026-06-23
feat(tonic-xds): parse A32 circuit breaking config

mingley:
## Motivation

This is the first scaffold PR for gRFC A32 circuit breaking under #2444. It starts carrying the CDS `circuit_breakers.thresholds[].max_requests` value through the validated `ClusterResource` without changing request routing or load-balancing behavior yet.

## Solution

Adds a small `CircuitBreakingConfig` resource helper that:

- defaults `max_requests` to 1024 when CDS circuit breakers are absent or no DEFAULT threshold is configured
- selects the first `RoutingPriority::Default` threshold, matching Envoy/gRFC behavior
- parses only `max_requests`; all other circuit-breaker fields remain ignored for this scaffold

Enforcement is intentionally left to a follow-up PR so permit lifetime can be handled correctly for streaming RPC response bodies.

## Testing Done

- `cargo fmt --all --check`
- `cargo test -p tonic-xds xds::resource`
- `cargo check -p tonic-xds --all-features`

### DIFF
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -564,6 +564,7 @@ mod tests {
             eds_service_name: None,
             lb_policy: LbPolicy::RoundRobin,
             security: None,
+            circuit_breaking: Default::default(),
         })
     }
 
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -564,7 +564,6 @@ mod tests {
             eds_service_name: None,
             lb_policy: LbPolicy::RoundRobin,
             security: None,
-            circuit_breaking: Default::default(),
         })
     }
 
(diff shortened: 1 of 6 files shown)


YutaoMa:
Same comment as in https://github.com/grpc/grpc-rust/pull/2701, let's avoid validating config end to end until the feature support is added: you can do this either by detaching the wiring from Cluster config layer, or implement logic first and then come back to this PR.

mingley:
> Same comment as in #2701, let's avoid validating config end to end until the feature support is added: you can do this either by detaching the wiring from Cluster config layer, or implement logic first and then come back to this PR.

addressed in [9afebba](https://github.com/grpc/grpc-rust/pull/2700/commits/9afebba99a96f5071b7093bdb95a0c83284fab98)

YutaoMa (approved):

PR #2703, opened 2026-06-25
grpc/credentials/client: Make ChannelCredentials object safe

arjan-bal:
This update makes the `ChannelCredentials` trait object-safe and removes the `DynChannelCredentials` trait, which simplifies the channel constructor by dropping its generic parameters. It also renames `ClientConnectionSecurityInfo` to `ChannelSecurityInfo`.

### DIFF
--- a/grpc-protobuf-build/Cargo.toml
+++ b/grpc-protobuf-build/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "grpc-protobuf-build"
-version = "0.9.0"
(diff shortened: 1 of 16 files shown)


arjan-bal:
FYI @nathanielford

dfawley (approved):

ejona86 (approved):

ejona86:
FWIW, the `BoxEndpoint` is more significant in this change than boxing the security context, as it is called significantly more often.

PR #2704, opened 2026-06-25
feat(tonic-xds): implement OTEL metrics recorder

YutaoMa:
## Motivation

Ref: #2444 

gRFC A78 defines the XdsClient OpenTelemetry metrics, and #2669 added the framework-agnostic `MetricsRecorder` extension point in `xds-client` — but shipped no bundled backend. Every consumer therefore had to implement the
 trait by hand just to get the metrics into OpenTelemetry. This PR provides that backend and exposes it through `tonic-xds`.

## Solution

Now an `OtelMetricsRecorder` is bundled when `otel` feature flag is enabled. User can provide their own OTEL meter and the five A78 XdsClient metrics (`connected`, `server_failure`, `resource_update_valid/invalid`, `resources`) are emitted through that meter. This PR marks the completion of metrics support in `tonic-xds` based on the currently supported xDS features.

## Out of Scope

A few gaps exist as compared to A78, all because of other unsupported gRFCs:

1. `nached_but_cached` and `cache_state` are not emitted, pending A88 data-error caching.
2.`grpc.xds.authority` uses `#old` sentinel for now as xDS Federation is not yet supported.


xds-client/Cargo.toml
```diff
@@ -26,6 +26,9 @@ http = { version = "1", optional = true }
 envoy-types = { version = "0.7", optional = true }
 prost = { version = "0.14", optional = true }
 
+# Optional dependency for the bundled OpenTelemetry metrics recorder
+opentelemetry = { version = "0.32", optional = true, default-features = false, features = ["metrics"] }
```
ankurmittal: are we going to have direct dep on otel version? if google planning the same?

xds-client-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn record_gauge_i64(&self, instrument: &'static Instrument, value: i64, attrs: &[KeyValue]) {
+        if let Some(CachedInstrument::Gauge(g)) = self.get(instrument) {
+            g.record(value, &to_otel_attrs(attrs));
+        }
+    }
+}
+
+/// Convert the crate's attribute slice into OpenTelemetry key/value pairs.
+fn to_otel_attrs(attrs: &[KeyValue]) -> Vec<opentelemetry::KeyValue> {
+    attrs.iter().map(to_otel_key_value).collect()
```
ankurmittal, line 155: is this hot path? maybe use fixed size (if known) small vec?

tonic-xds/src/lib.rs
```diff
@@ -150,5 +150,10 @@ pub use client::channel::{
 pub use xds::bootstrap::{BootstrapConfig, BootstrapError};
 pub use xds::uri::{XdsUri, XdsUriError};
 
+pub use xds_client::MetricsRecorder;
```
mingley: Can we re-export the recorder argument types from `tonic-xds` as well? With only `MetricsRecorder` exported, a crate that depends only on `tonic-xds` cannot implement the trait because its methods require `Instrument` and `KeyValue`, which are not available as `tonic_xds::Instrument`/`tonic_xds::KeyValue` (I confirmed with a tiny downstream compile check). Re-exporting `Instrument`, `InstrumentKind`, `KeyValue`, `StringValue`, and `Value` keeps this public hook usable without forcing users to add a direct `xds-client` dependency.

xds-client-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,234 @@
...
+        InstrumentKind::Counter => CachedInstrument::Counter(
+            meter
+                .u64_counter(instrument.name)
+                .with_description(instrument.description)
+                .with_unit(instrument.unit)
+                .build(),
+        ),
+        InstrumentKind::UpDownCounter => CachedInstrument::UpDownCounter(
+            meter
+                .i64_up_down_counter(instrument.name)
```
mingley, line 95: This generic mapping makes `grpc.xds_client.resources` an OpenTelemetry UpDownCounter/Sum, because its descriptor is `InstrumentKind::UpDownCounter`, but A78 defines that metric as a Gauge of the current resource count. Since the worker emits +/-1 deltas for state transitions, OTel consumers will observe a sum instrument instead of the current cached count. Please accumulate the per-attribute totals in the recorder and publish them through an `i64_gauge`, or change the producer to call `record_gauge_i64` with current counts.

xds-client/src/metrics/otel.rs
```diff
@@ -0,0 +1,234 @@
...
+            assert!(
+                recorder.get(instrument).is_some(),
+                "{} not registered",
+                instrument.name
+            );
+        }
+    }
+
+    #[test]
+    fn recording_through_noop_meter_does_not_panic() {
```
mingley: This test only checks that a no-op meter does not panic, so it would still pass if values/attributes were dropped or if `grpc.xds_client.resources` exported as the wrong OTel instrument type. Please add an SDK/manual-reader test that records through `OtelMetricsRecorder` and asserts the exported metric type, value, and key attributes for at least one counter and the resource gauge.

mingley (commented):
A few issues to address.

xds-client-opentelemetry/src/lib.rs
YutaoMa: Valid point, updated to use otel gauge. One design decision highlight: instead of the async gauge used in `grpc-go` for this metric, the new commit uses sync gauge as moving to async gauge requires some significant refactoring of the worker cache to be able to share the resource count. For now I'll KISS even though it means some extra cost for performing the diff on each resource update. The assumption is xDS resource updates is relatively infrequent as compared to gRPC request hot path so this is acceptable.

xds-client/src/metrics/otel.rs
YutaoMa: Updated test to use `InMemoryMetricExporter` to assert on meter output

tonic-xds/src/lib.rs
YutaoMa: Updated.

xds-client-opentelemetry/src/lib.rs
YutaoMa: xDS resource updates are not on data plane hot path so I'll KISS for now. Later if benchmark shows inefficiencies due to labeling cost I'll consider impl optimizations here.

xds-client/Cargo.toml
YutaoMa: Good question, support for OTEL metrics in `grpc` isn't fully defined yet, for now the best plan forward I see is to isolate the concrete otel coupling point (the meter intake) to a separate crate (like the `tracing-opentelemetry` setup) `xds-client-opentelemetry`. So that if user need older (or newer in case our crate hasn't udpated) version of otel support, they are free to implement their own Recorder trait impls and use it.

xds-client-opentelemetry/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "xds-client-opentelemetry"
+description = "OpenTelemetry metrics recorder for xds-client (gRFC A78 XdsClient metrics)"
+version = "0.1.0-alpha.1"
+edition = "2024"
+homepage = "https://github.com/hyperium/tonic"
+repository = "https://github.com/hyperium/tonic"
```
gu0keno0: Shall we change this to grpc-rust?

tonic-xds/src/client/channel.rs
```diff
@@ -151,9 +153,25 @@ const _: fn() = || {
...
+impl Debug for XdsChannelBuilder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("XdsChannelBuilder")
+            .field("config", &self.config)
+            .field(
+                "recorder",
+                &self
+                    .recorder
+                    .as_ref()
+                    .map_or("None", |_| "Some(Arc<dyn MetricsRecorder>)"),
```
gu0keno0: Hmm... do we really want to do it this way? Maybe use std::any::type_name_of_val ?

xds-client-opentelemetry/src/lib.rs
```diff
@@ -0,0 +1,306 @@
+//! OpenTelemetry [`MetricsRecorder`] implementation for `xds-client`.
+//!
+//! [`OtelMetricsRecorder`] adapts the framework-agnostic [`MetricsRecorder`]
+//! trait onto an [`opentelemetry::metrics::Meter`], so the gRFC A78 xDS client
+//! metrics flow into whatever OpenTelemetry SDK the application has configured.
+//!
+//! This lives in a dedicated crate (rather than behind a feature on
+//! `xds-client`) so the OpenTelemetry version it depends on can move
+//! independently of the core `xds-client` release cadence, and so applications
+//! that use a different telemetry framework never compile OpenTelemetry at all.
```
gu0keno0, line 10: Can you walk me through how this can make otel version management easier? I imagine caller of XdsChannelBuilder will need to wire up OTEL, so unless it provides its own MetricsRecorder implementation, they will use this implementation anyways? This library feels like a default impl of MetricsRecorder, so seems like we can just make it a feature of xds-client?

xds-client/src/client/worker.rs
```diff
@@ -146,49 +152,46 @@ impl RecorderHandle {
...
-        type_url: &Arc<str>,
-        prev: Option<&ResourceState>,
-        new: &ResourceState,
-    ) {
+    /// Reconcile the `grpc.xds_client.resources` gauge for `type_url` against an
+    /// authoritative cache snapshot (`cache_state` label -> current count).
+    ///
+    /// The worker's resource cache is the single source of truth; this only
+    /// diffs the snapshot against the values last emitted for `type_url` and
+    /// pushes the buckets that changed. Buckets that dropped out of the snapshot
```
gu0keno0, line 160: Is "bucket" a special terminology in otel / a78?

xds-client/src/client/worker.rs
```diff
@@ -1059,16 +1071,13 @@ where
...
                         resource_name,
                         CachedResource::received(Arc::new(resource.clone())),
                     );
-                    self.recorder.record_resource_transition(
-                        &s.type_url,
-                        prev.as_ref().map(|c| &c.state),
-                        &ResourceState::Received,
-                    );
                 }
+                let counts = s.resource_state_counts();
```
gu0keno0, line 1079: This is O(resources) right? Might be expensive if the client is used for many big resources. E.g. rolling update of a big EDS cluster will trigger this lots of times?

Alternatively just write a reconcile loop to reconcile the counter values with their cache SOT?

xds-client-opentelemetry/src/lib.rs
YutaoMa: This way it's possible to use a older version of `xds-client-opentelemetry` (thus a older version of OTEL) with a newer version of `tonic-xds`, giving people time to update their provider side of code. (e.g. some corporate ecosystem that needs OTEL upgrade); or allow us to release a new version of `xds-client-opentelemetry` without releasing a breaking version in `tonic-xds` immediately. We still include the latest version of `xds-client-opentelemetry` in `tonic-xds` as a feature for ease of use.

xds-client/src/client/worker.rs
YutaoMa: Not really, here the bucket is just referring to the cache entries (hash bucket)

xds-client/src/client/worker.rs
YutaoMa: Two reasons that this won't be a bottleneck:  `sync_resource_counts`  emits only changed state buckets → so the metrics call is bounded by O(states)≈5, not per resource; and for `dispatch_resources` already does O(N) resource clone and inserts, so the `sync_resource_counts` overhead is on the same order. One optimization that can be done is to keep track of per state counters on TypeState for O(1) rescan, although that complicates the insert logic a bit. Let me know what you think.

xds-client-opentelemetry/Cargo.toml
YutaoMa: Sure, updated

tonic-xds/src/client/channel.rs
YutaoMa: Good point, updated.

xds-client/src/client/worker.rs
gu0keno0: resource_state_counts() still loops through all the entries under each state label. At least for deleted resources this seems to be an unecessary overhead.

I guess it might be fine though for now. We are not using Delta, CLA is the finest granularity of endpoint updates, which is the most frequent one practically. I don't worry much about LDS, RDS and CDS because they are not very dynamic.

xds-client-opentelemetry/src/lib.rs
gu0keno0: I see, xds-client-otel is a way to decouple oss otel version bump and xds-client version bump.

Then my only question is, does tonic-xds need to be aware of xds-client-otel (for now)?

I can see that in the near future we might want to instrument tonic-xds by otel as well, yet for now seems like the metrics are solely owned by xds-client?

gu0keno0 (approved):

Issue #2702, opened 2026-06-24
Support channel call-credential hook for xDS transport


choonkijang:
### Crates
* `xds-client`
* `tonic-xds`

### Motivation
There is no way to attach per-stream call credentials (e.g. a bearer token) to the ADS stream. This blocks connecting to credential-gated control planes; GCP Traffic Director, whose bootstrap uses `channel_creds: [{ "type": "google_default" }]` and requires an Application Default Credentials (ADC) bearer token on the stream, over system-root TLS.

Related specification: [gRFC A27 # XdsClient and Bootstrap File](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md#xdsclient-and-bootstrap-file) 

### Proposal

Briefly, mirror grpc-go's implementation
* https://github.com/grpc/grpc-go/blob/master/xds/bootstrap/credentials.go
* https://github.com/grpc/grpc-go/blob/master/xds/bootstrap/bootstrap.go

More specifically,
1. Define `CallCredentials` trait type, and let `TonicTransportBuilder` can take it (eg, `with_call_credentials(Arc<dyn CallCredentials>)`
2. Extend [ChannelCredentialType](https://github.com/grpc/grpc-rust/blob/618389f8ce23a07137a97dac21a9e9c5ddb34f23/tonic-xds/src/xds/bootstrap.rs#L92) to have `google_default` entry to follow [gRFC A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md#xdsclient-and-bootstrap-file).

### Alternatives
* To be fully contained for gRFC A27, bring `google_cloud_auth` dependency just like [grpc-google crate](https://github.com/grpc/grpc-rust/blob/master/grpc-google/src/lib.rs#L33).

choonkijang:
This is part of [overall xDS support](https://github.com/grpc/grpc-rust/issues/2444).

choonkijang:
@gu0keno0 @YutaoMa would you please take a look at this issue and https://github.com/grpc/grpc-rust/pull/2705? Thanks in advance!

YutaoMa:
Thank you for creating the issue and the PR! The described use case makes sense and the PR looks mostly good now after the previous round of edits, just left a few more stylistic comments, ptal!

choonkijang:
> Thank you for creating the issue and the PR! The described use case makes sense and the PR looks mostly good now after the previous round of edits, just left a few more stylistic comments, ptal!

[...]

PR #2705, opened 2026-06-25
feat(xds): per-stream call-credentials hook for the xDS transport

choonkijang:
## Motivation

Per https://github.com/grpc/grpc-rust/issues/2702, credential-gated xDS control planes (e.g. GCP Traffic Director / `google_default`) require two things the transport did not provide: per-stream call credentials on the ADS stream, and a secure channel to carry them. Their bootstrap `server_uri` is often scheme-less (e.g. `trafficdirector.googleapis.com:443`), which parses with no scheme — so `Endpoint` has no signal to negotiate TLS, the ADS connection never establishes a secure session, and routing times out with "route config not yet available". 

## Solution

xds-client:
- Add a minimal `TonicCallCredentials` trait mirroring grpc-go's `PerRPCCredentials`. The tonic transport attaches it on each (re)connect, only over a secure channel.
- Call credentials are accepted only via `TonicTransportBuilder::with_call_credentials`, where channel security is derived from `with_tls_config` and validated at build time to avoid sending credentials to insecure channel. 
- On the secure path, prepend `https://` to `server_uri` in `TonicTransportBuilder::build` when it's missing the scheme.

tonic-xds:
- Promote `google_default` to a first-class `ChannelCredentialType::GoogleDefault` (gRFC A27).
- Add `XdsChannelConfig::with_call_credentials(...)`, so that applications can give a token source without relying on `grpc` or `grpc-google` crates at this time.

examples:
- Add the `channel_with_google_default` tonic-xds example (reusing the testutil greeter), demonstrating two ADC token sources.

BREAKING CHANGE: 
- adds a CallCredentials variant to the Error enum and marks it #[non_exhaustive]

### DIFF
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -68,6 +68,9 @@ tonic-prost-build = "0.14"
 async-stream = "0.3"
 tempfile = "3.27.0"
 rcgen = "0.14"
+grpc = { path = "../grpc" }
+grpc-google = { path = "../grpc-google" }
+google-cloud-auth = { version = "1.9", default-features = false }
 
 [features]
 testutil = ["dep:tonic-prost"]
@@ -80,6 +83,10 @@ tls-aws-lc = ["_tls-any", "tonic/tls-aws-lc", "xds-client/tonic-tls-aws-lc"]
 [[example]]
 name = "channel"
 required-features = ["testutil"]
+
+[[example]]
+name = "channel_with_google_default"
+required-features = ["testutil", "tls-ring"]
 
 [[example]]
 name = "greeter_server"
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -68,8 +68,6 @@ tonic-prost-build = "0.14"
 async-stream = "0.3"
 tempfile = "3.27.0"
 rcgen = "0.14"
-grpc = { path = "../grpc" }
-grpc-google = { path = "../grpc-google" }
 google-cloud-auth = { version = "1.9", default-features = false }
 
 [features]
--- a/tonic-xds/examples/channel_with_google_default.rs
+++ b/tonic-xds/examples/channel_with_google_default.rs
@@ -1,7 +1,7 @@
 //! Call an xDS-fronted service through GCP Traffic Director (`google_default`).
 //!
 //! Supplies the Application Default Credentials (ADC) token as xDS call
-//! credentials by implementing `CallCredentials` directly against
+//! credentials by implementing `TonicCallCredentials` directly against
 //! `google-cloud-auth`.
 //!
 //! Needs a `google_default` bootstrap + ADC:
@@ -15,18 +15,17 @@ use std::sync::Arc;
 
 use google_cloud_auth::credentials::{AccessTokenCredentials, Builder};
 use tonic_xds::testutil::proto::helloworld::{HelloRequest, greeter_client::GreeterClient};
-use tonic_xds::{XdsChannelBuilder, XdsChannelConfig, XdsUri};
-use xds_client::CallCredentials;
+use tonic_xds::{TonicCallCredentials, XdsChannelBuilder, XdsChannelConfig, XdsUri};
 
 const CLOUD_PLATFORM_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
 
 /// Fetches ADC tokens directly from `google-cloud-auth`.
 #[derive(Debug)]
-struct AdcCallCredentials {
+struct AdcTonicCallCredentials {
     creds: AccessTokenCredentials,
 }
 
-impl AdcCallCredentials {
+impl AdcTonicCallCredentials {
     fn new() -> std::result::Result<Self, Box<dyn std::error::Error>> {
         let creds = Builder::default()
             .with_scopes([CLOUD_PLATFORM_SCOPE])
@@ -36,7 +35,7 @@ impl AdcCallCredentials {
 }
 
 #[tonic::async_trait]
-impl CallCredentials for AdcCallCredentials {
+impl TonicCallCredentials for AdcTonicCallCredentials {
     async fn get_request_metadata(
         &self,
         metadata: &mut tonic::metadata::MetadataMap,
@@ -65,7 +64,7 @@ async fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
     let target_str = std::env::var("XDS_TARGET").unwrap_or_else(|_| "xds:///my-service".into());
     let target = XdsUri::parse(&target_str)?;
 
-    let creds: Arc<dyn CallCredentials> = Arc::new(AdcCallCredentials::new()?);
+    let creds: Arc<dyn TonicCallCredentials> = Arc::new(AdcTonicCallCredentials::new()?);
 
     let channel =
         XdsChannelBuilder::new(XdsChannelConfig::new(target).with_call_credentials(creds))
--- a/tonic-xds/examples/channel_with_google_default.rs
+++ b/tonic-xds/examples/channel_with_google_default.rs
@@ -1,0 +1,148 @@
+//! Call an xDS-fronted service through GCP Traffic Director (`google_default`).
+//!
+//! Shows two ways to supply the Application Default Credentials (ADC) token as
+//! xDS call credentials, selected by `XDS_CRED_SOURCE`:
+//!
+//! - `adc` (default): implement `CallCredentials` directly against
+//!   `google-cloud-auth`. No dependency on the `grpc` / `grpc-google` crates,
+//!   which are a preview and "not recommended for any production use".
+//! - `grpc-google`: bridge `grpc_google::GcpCallCredentials` into the seam (a
+//!   convenience wrapper, but it pulls in the preview `grpc` crate).
+//!
+//! Needs a `google_default` bootstrap + ADC:
+//!
+//! ```sh
+//! GRPC_XDS_BOOTSTRAP=/path/to/bootstrap.json \
+//!     cargo run -p tonic-xds --example channel_with_google_default --features "testutil tls-ring"
+//! ```
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use google_cloud_auth::credentials::{AccessTokenCredentials, Builder};
+use grpc::attributes::Attributes;
+use grpc::credentials::SecurityLevel;
+// `GrpcCallCredentials`: trait in scope so its methods resolve on `inner`.
+use grpc::credentials::call::{
+    CallCredentials as GrpcCallCredentials, CallDetails, ClientConnectionSecurityInfo,
+};
+use grpc::metadata::{KeyAndValueRef, MetadataMap};
+use grpc_google::{GcpCallCredentials, TokenProvider};
+use tonic_xds::testutil::proto::helloworld::{HelloRequest, greeter_client::GreeterClient};
+use tonic_xds::{XdsChannelBuilder, XdsChannelConfig, XdsUri};
+use xds_client::{CallCredentials, Error, Result};
+
+const ADS_METHOD: &str =
+    "envoy.service.discovery.v3.AggregatedDiscoveryService/StreamAggregatedResources";
+const CLOUD_PLATFORM_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
+
+/// Fetches ADC tokens directly from `google-cloud-auth`.
+#[derive(Debug)]
+struct AdcCallCredentials {
+    creds: AccessTokenCredentials,
+}
+
+impl AdcCallCredentials {
+    fn new() -> std::result::Result<Self, Box<dyn std::error::Error>> {
+        let creds = Builder::default()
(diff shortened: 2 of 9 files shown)


tonic-xds/Cargo.toml
```diff
@@ -68,6 +68,9 @@ tonic-prost-build = "0.14"
 async-stream = "0.3"
 tempfile = "3.27.0"
 rcgen = "0.14"
+grpc = { path = "../grpc" }
```
YutaoMa: Let's avoid coupling `tonic-xds` to `grpc` and `grpc-google`.

xds-client/src/transport/tonic.rs
```diff
@@ -283,6 +382,9 @@ mod tests {
             &self,
             request: Request<tonic::Streaming<DiscoveryRequest>>,
         ) -> std::result::Result<Response<Self::StreamAggregatedResourcesStream>, Status> {
+            if let Some(v) = request.metadata().get("authorization") {
```
YutaoMa: nit: I recommend creating a separate mock server that returns different response based on the authorization header so no `Arc<Mutex>` setup is needed.

xds-client/src/transport/tonic.rs
```diff
@@ -9,13 +9,31 @@ use crate::error::{Error, Result};
...
 use tonic::transport::{Channel, Endpoint};
 use tonic::{Status, Streaming};
 
+/// Per-stream call credentials for the ADS stream (e.g. a bearer token).
+///
+/// Attached on each (re)connect, only when the channel is secure.
+#[tonic::async_trait]
+pub trait CallCredentials: Send + Sync + std::fmt::Debug + 'static {
+    /// Header key / value map to attach to the ADS stream.
+    async fn get_request_metadata(&self) -> Result<HashMap<String, String>>;
```
YutaoMa: This API warrants a bit more design discussion: we can either hoist this trait to a transport agnostic one, or keep this as a tonic only API for now, which I'd then recommend changing the return type to tonic's MetadataMap to get its semantic invariants (allow duplicate keys and multi-values, supports sensitivity to block accidental logging of credentials, etc.). I think either one has its pros and cons, but both requires some change to the API shape.

tonic-xds/Cargo.toml
choonkijang: Sure, I'll drop related example.

xds-client/src/transport/tonic.rs
choonkijang: Nice point. To be consistent with `grpc` crate, let me apply in a similar way with `grpc` crate.

https://github.com/grpc/grpc-rust/blob/master/grpc/src/credentials/call.rs#L127-L132

xds-client/src/transport/tonic.rs
choonkijang: Sure.

choonkijang:
@YutaoMa PTAL!

xds-client/src/error.rs
```diff
@@ -14,6 +14,11 @@ pub enum Error {
     #[error("stream error: {0}")]
     Stream(#[from] tonic::Status),
 
+    /// Call credentials failed, or require a secure transport.
```
YutaoMa, line 18: Let's remove the feature gating on this error variant, non-additive features in Rust can be a footgun: implicit feature enablement with other dependencies can break compilation.

While we are at it, let's also add `#[non_exhaustive]` to the Error, it's a miss on my end to not mark it so previously. That'll help with future extensions like this.

xds-client/src/transport/tonic.rs
```diff
@@ -9,13 +9,33 @@ use crate::error::{Error, Result};
...
 
+/// Per-stream call credentials for the ADS stream (e.g. a bearer token).
+///
+/// Attached on each (re)connect, only when the channel is secure.
+#[tonic::async_trait]
+pub trait CallCredentials: Send + Sync + std::fmt::Debug + 'static {
+    /// Generates the authentication metadata for a specific call.
+    async fn get_request_metadata(
+        &self,
+        metadata: &mut tonic::metadata::MetadataMap,
```
YutaoMa, line 28: Is there any specific reason to prefer out-param style API? I'd recommend keeping this simple and idiomatic by returning an owned map.

YutaoMa: Ah saw the `grpc` link comment above, seems like it's to keep the style consistent with `grpc`. In that case yeah it's good to keep it as is.

xds-client/src/transport/tonic.rs
```diff
@@ -9,13 +9,33 @@ use crate::error::{Error, Result};
...
 use tonic::client::Grpc;
 use tonic::codec::{Codec, DecodeBuf, Decoder, EncodeBuf, Encoder};
 use tonic::transport::{Channel, Endpoint};
 use tonic::{Status, Streaming};
 
+/// Per-stream call credentials for the ADS stream (e.g. a bearer token).
+///
+/// Attached on each (re)connect, only when the channel is secure.
+#[tonic::async_trait]
+pub trait CallCredentials: Send + Sync + std::fmt::Debug + 'static {
```
YutaoMa: Could you re-export this type in `tonic-xds` so user of `tonic-xds` doesn't need to explicitly depend on `xds-client` to use this feature? (`xds-client` is intended to stay as a low-level ADS transport library so gRPC service developers don't have to depend on it explicitly). Also I suggest naming it `TonicCallCredentials` here: once we support `grpc` transport in `xds-client`, similar API will need to be added for `GrpcCallCredentials`, keeping the name distinguishable helps with future extensions.

xds-client/src/error.rs
choonkijang: Done. Dropped the `#[cfg(feature = "transport-tonic")]` on `CallCredentials` and added `#[non_exhaustive]` to `Error`. See https://github.com/grpc/grpc-rust/pull/2705/changes/447e8490e38c085dc494f0494cec117b115ae563.

xds-client/src/transport/tonic.rs
choonkijang: Yeah, thanks for confirmation!

xds-client/src/transport/tonic.rs
choonkijang: Done. Renamed and re-exported. See https://github.com/grpc/grpc-rust/pull/2705/changes/447e8490e38c085dc494f0494cec117b115ae563.

choonkijang:
@YutaoMa PTAL!

tonic-xds/src/xds/bootstrap.rs
```diff
@@ -619,6 +624,44 @@ mod tests {
...
+        );
+        assert!(config.use_tls());
+        assert_eq!(
+            config.selected_credential(),
+            Some(&ChannelCredentialType::GoogleDefault)
+        );
+    }
+
+    #[test]
+    fn selected_credential_unknown_type_skipped() {
```
gu0keno0: Looks like we can combine this test with "selected_credential_none_supported()"

choonkijang: Nice point! Addressed at https://github.com/grpc/grpc-rust/pull/2705/changes/31419699bc9f0ef073b819f4e87d1e768c71ff05.

choonkijang:
@YutaoMa @gu0keno0 PTAL.

xds-client/src/transport/tonic.rs
```diff
@@ -168,15 +175,79 @@ impl TonicTransportBuilder {
...
+    /// use xds_client::TonicTransportBuilder;
+    ///
+    /// let channel = Channel::from_static("https://xds.example.com:443")
+    ///     .tls_config(ClientTlsConfig::new())?
+    ///     .connect()
+    ///     .await?;
+    ///
+    /// let builder = TonicTransportBuilder::new().with_channel(channel, true);
+    /// ```
+    pub fn with_channel(mut self, channel: Channel, secure: bool) -> Self {
```
gu0keno0: Is this only needed for test? Maybe gate it by #[cfg(test)]

The reason is that "secure" is not tied to "channel". I wish Tonic had channel.is_secure() -> bool, yet since it doesn't have it, better not to expose ambituity.

xds-client/src/transport/tonic.rs
```diff
@@ -192,14 +263,28 @@ impl TransportBuilder for TonicTransportBuilder {
...
 }
 
 impl Transport for TonicTransport {
     type Stream = TonicAdsStream;
 
     async fn new_stream(&self, initial_requests: Vec<Bytes>) -> Result<Self::Stream> {
+        // Guard before connecting: never send credentials over an insecure channel.
+        if let Some(creds) = &self.call_creds
+            && creds.requires_secure_transport()
+            && !self.secure
```
gu0keno0: It's probably better to check this during TonicTransport build time. If secure == false, always set creds to None.

xds-client/src/transport/tonic.rs
```diff
@@ -80,36 +100,15 @@ impl Decoder for BytesDecoder {
 #[derive(Clone, Debug)]
 pub struct TonicTransport {
     channel: Channel,
+    secure: bool,
```
gu0keno0: I think we can drop this if we do the check during TonicTransport build time.

xds-client/src/transport/tonic.rs
```diff
@@ -168,15 +175,79 @@ impl TonicTransportBuilder {
...
+        raw.to_string()
+    }
 }
 
 impl TransportBuilder for TonicTransportBuilder {
     type Transport = TonicTransport;
 
     async fn build(&self, server: &ServerConfig) -> Result<Self::Transport> {
+        // Use a pre-built channel as-is; the server URI is ignored.
+        if let Some((channel, secure)) = &self.channel {
```
gu0keno0: Gate by #[cfg(test)], we shouldn't need to allow (channel, secure) to be passed from outside in prod

gu0keno0 (commented):
PTAL. Thanks for the contribution. I'd like to simplify the states of "secure", "channel" and "per call creds" and avoid invalid states from builder directly.

xds-client/src/transport/tonic.rs
choonkijang: It was originally `TonicTransport::from_channel` function ([link](https://github.com/grpc/grpc-rust/pull/2705/changes#diff-3f85c9d5da3fac5af271b2602d7565d240fd7c6d296990694ff2e3fa375df7faL86-L109)), and I brought it here to make it work with `with_call_credentials`; thinking again upon all your comments, it'd be better to move back to `TonicTransport::from_channel` to to avoid ambiguous cases rather than gating by `#[cfg(test)]`. When call-credential is required, the users will be able to make the channel using `TonicChannelBuilder` instead.

choonkijang:
> PTAL. Thanks for the contribution. I'd like to simplify the states of "secure", "channel" and "per call creds" and avoid invalid states from builder directly.

Reflected to https://github.com/grpc/grpc-rust/pull/2705/commits/f44998e530fc52d619659e30c73c813175b66526. As said [reply](https://github.com/grpc/grpc-rust/pull/2705#discussion_r3547646068), `TonicTransportBuilder::with_channel` moved back to `TonicTransport::from_channel`. Updated PR description accordingly.

xds-client/src/transport/tonic.rs
choonkijang: Good point. Addressed at https://github.com/grpc/grpc-rust/commit/f44998e530fc52d619659e30c73c813175b66526.

xds-client/src/transport/tonic.rs
choonkijang: Addressed at https://github.com/grpc/grpc-rust/commit/f44998e530fc52d619659e30c73c813175b66526.

tonic-xds/examples/channel_with_google_default.rs
```diff
@@ -0,0 +1,82 @@
...
+
+use google_cloud_auth::credentials::{AccessTokenCredentials, Builder};
+use tonic_xds::testutil::proto::helloworld::{HelloRequest, greeter_client::GreeterClient};
+use tonic_xds::{TonicCallCredentials, XdsChannelBuilder, XdsChannelConfig, XdsUri};
+
+const CLOUD_PLATFORM_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
+
+/// Fetches ADC tokens directly from `google-cloud-auth`.
+#[derive(Debug)]
+struct AdcTonicCallCredentials {
```
gu0keno0, line 24: I think we might want to support google credentials out-of-box in tonic-xds' xds bootstrap handling in the future, yet that's out of scope for this PR, so no need to do it here.

gu0keno0 (approved):

choonkijang:
Thanks for review! @gu0keno0 and @YutaoMa !

choonkijang:
Well, I don't have sufficient permission to merge this PR to mater. Can I ask merge? @gu0keno0 @YutaoMa Thanks in advance!

gu0keno0:
> Well, I don't have sufficient permission to merge this PR to mater. Can I ask merge? @gu0keno0 @YutaoMa Thanks in advance!

Merged. Thanks again for the feature.

PR #2706, opened 2026-06-26
grpc-benchmark: Worker and server implementation

arjan-bal:
Contributes to: #2666

This PR adds the worker and the benchmarking server implementations. Both components use `tonic` and `prost` for now, as the gRPC server is still under development. The benchmarking client that uses gRPC will be added in a follow-up PR.


grpc-benchmark/src/main.rs
```diff
@@ -0,0 +1,91 @@
...
+    use std::time::Duration;
+
+    use grpc_benchmark::generated::services::grpc::testing::worker_service_server::WorkerServiceServer;
+    use grpc_benchmark::worker::WorkerServer;
+    use tokio::sync::mpsc;
+    use tokio::time;
+    use tonic::transport::Server;
+
+    pub async fn run_worker(worker_port: u16) -> Result<(), Box<dyn std::error::Error>> {
+        let addr = format!("0.0.0.0:{}", worker_port).parse().unwrap();
```
ejona86: Can we specify the type here, since it isn't clear?

grpc-benchmark/src/server.rs
```diff
@@ -0,0 +1,240 @@
...
+            })?,
+            shutdown_notify,
+            port,
+        })
+    }
+
+    pub(crate) fn get_stats(&mut self, reset: bool) -> Result<ServerStats, Status> {
+        let now = Instant::now();
+        let wall_time_elapsed = now.duration_since(self.last_reset_time);
+        let latest_rusage = getrusage(UsageWho::RUSAGE_SELF).map_err(|err| {
```
ejona86: grpc-java just leaves [time_user as zero for other platforms](https://github.com/grpc/grpc-java/blob/71a10dc9c7db97d95567fe9d6655f41096b914bb/benchmarks/src/main/java/io/grpc/benchmarks/driver/LoadServer.java#L183).

Can we just define a non-linux function that returns the usage as 0, instead of making everything conditional, including the generated code?

ejona86 (commented):
Didn't get all the way through; sending what I have.

grpc-benchmark/src/main.rs
arjan-bal: This is a `std::net::SocketAddr`. I changed the code to use an infallible constructor instead of relying on string parsing.

grpc-benchmark/src/server.rs
arjan-bal: I've added an `rusage` module that has platform specific code to return empty resource usage on non-unix platforms.

grpc-benchmark/src/worker.rs
```diff
@@ -0,0 +1,155 @@
...
+
+                let argtype = request.argtype
+                    .ok_or_else(|| Status::invalid_argument("missing request.argtype"))?;
+
+                match argtype {
+                    Argtype::Setup(server_config) => {
+                        println!("Server creation requested.");
+
+                        if benchmark_server.take().is_some() {
+                            eprintln!("Server setup received when server already exists, shutting down the existing server");
```
ejona86: Does replacing the server actually work? I'd expect start() to fail because the port is already in use. I think we need to set `benchmark_server = None` here. And that might still be racy.

grpc-benchmark/src/server.rs
```diff
@@ -0,0 +1,241 @@
...
+    port: u16,
+}
+
+impl BenchmarkServer {
+    pub(crate) fn start(config: ServerConfig) -> Result<Self, Status> {
+        println!("Starting benchmark server with config: {:?}", config);
+
+        let mut server_builder = Server::builder();
+        // Parse security config.
+        if let Some(securit_params) = config.security_params {
```
ejona86: s/securit/security/

ejona86 (approved):

grpc-benchmark/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+    let (tx, mut rx) = mpsc::channel(1);
+
+    let svc = WorkerServiceServer::new(WorkerServer::new(tx));
+
+    Server::builder()
+        .add_service(svc)
+        .serve_with_shutdown(addr, async {
+            rx.recv().await;
+            // Wait for the quit_worker response to be sent.
+            time::sleep(Duration::from_secs(1)).await;
```
dfawley: Why do we sleep 1 second here?  Will the client fail if the server hard shuts down before the call competes successfully or something?

grpc-benchmark/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use grpc_benchmark::generated::services::grpc::testing::worker_service_server::WorkerServiceServer;
+use grpc_benchmark::worker::WorkerServer;
+use tokio::sync::mpsc;
+use tokio::time;
+use tonic::transport::Server;
+
+pub async fn run_worker(worker_port: u16) -> Result<(), Box<dyn std::error::Error>> {
+    let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), worker_port);
+    let (tx, mut rx) = mpsc::channel(1);
```
dfawley: Nit: `quit_tx, mut quit_rx`?

grpc-benchmark/src/rusage.rs
```diff
@@ -0,0 +1,62 @@
...
+    user_time_ns: i64,
+    system_time_ns: i64,
+}
+
+impl Rusage {
+    #[cfg(unix)]
+    pub(crate) fn now() -> Result<Self, String> {
+        use nix::sys::resource::UsageWho;
+        use nix::sys::resource::getrusage;
+        use nix::sys::time::TimeValLike;
```
dfawley, line 36: What's the philosophy for this?  When do we `use` at a local scope like this vs. at the file level?  It would probably be good to have some kind of guideline if it's not a hard rule like "always `use` at the file level".

grpc-benchmark/src/worker.rs
dfawley: I think if you abort the task in drop, instead of using the `Notify`, then it _should_ be fine, but there might still be stuff going on at the OS level that prevents it from being reused immediately.  I would expect that should be up to the driver to not do that, though, right?

grpc-benchmark/src/main.rs
arjan-bal: Renamed and used `Notify` instead of an mpsc channel.

grpc-benchmark/src/main.rs
arjan-bal: The 1-second wait is unnecessary because the server performs a graceful shutdown when the future resolves. This workaround was originally copied from [the Go implementation](https://github.com/grpc/grpc-go/blob/1a80fca960d39ae4d7d6f2d9323ca2d243fd44bb/benchmark/worker/main.go#L221-L227), which appears to have been written before Go supported graceful shutdown.

I have removed the sleep.

grpc-benchmark/src/server.rs
arjan-bal: Fixed.

grpc-benchmark/src/worker.rs
arjan-bal: Requesting server creation when a server is already running should be treated as a bad request. The previous logic was adapted from the [Go implementation](https://github.com/grpc/grpc-go/blob/1a80fca960d39ae4d7d6f2d9323ca2d243fd44bb/benchmark/worker/main.go#L105-L111), which restarts the server using a non-graceful shutdown. However, [Java's behavior](https://github.com/grpc/grpc-java/blob/c8079eed98cc0c66673b75694b511b8e89844bef/benchmarks/src/main/java/io/grpc/benchmarks/driver/LoadWorker.java#L154-L159) is to fail the stream with an ALREADY_EXISTS status.

Explicitly failing the request is a cleaner approach than attempting to recover from an invalid state transition, so I have updated our implementation to match Java's behavior.

grpc-benchmark/src/rusage.rs
arjan-bal: Placing the imports inside the function body keeps them scoped, as this function is conditionally compiled only for Unix platforms.

A few alternatives include:
* Guarding imports at the file level with individual `#[cfg(unix)]` attributes.
* Using fully qualified paths directly in the code to bypass imports.
* Moving the conditional code into a dedicated module and keeping the imports at the module level.

Inline imports felt like the cleanest approach for a single function, but let me know if you prefer one of the other styles.

dfawley: I think the argument of being inside a function that's conditionally compiled (and the deps also want to be conditional) is probably good enough.  In general I guess we want most symbols to be imported/defined at a global scope, and just leave it to the judgement of the author whether to define helper types either globally or inside of smaller scopes?

PR #2708, opened 2026-06-26
feat(transport): expose http2_max_local_error_reset_streams on Endpoint

gautamg795:
Adds `Endpoint::http2_max_local_error_reset_streams`, the client-side counterpart to the existing `Server::http2_max_local_error_reset_streams`.

**Why this is needed.** h2 caps the number of local error resets a connection may accumulate over its lifetime (`max_local_error_reset_streams`, default 1024); once past it, h2 closes the connection with `GOAWAY(ENHANCE_YOUR_CALM, "too_many_internal_resets")`, failing every in-flight RPC on it. A long-lived client `Channel` that legitimately cancels many server-streaming RPCs — e.g. dropping the response stream on client disconnect or timeout — steadily accrues these resets: an abandoned stream yields a *counted* `STREAM_CLOSED` reset once h2 has forgotten the stream and the peer's in-flight DATA arrives. Under sustained load this trips the limit and tears the connection down.

`Server` already exposes this knob (for the symmetric server-side case), but `Endpoint` does not — so today there is no way to raise or disable the limit on an outbound connection without patching tonic.

**Behavior.** Mirrors the server method's `Option<usize>` signature. Non-breaking: when unset, hyper's default (currently 1024) is unchanged; only an explicit call overrides it, and `None` disables the limit (advisable only for trusted peers, e.g. internal services).

Opening as a draft for maintainer feedback on the approach.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tonic/src/transport/channel/endpoint.rs
+++ b/tonic/src/transport/channel/endpoint.rs
@@ -52,6 +52,7 @@ pub struct Endpoint {
     pub(crate) http2_keep_alive_while_idle: Option<bool>,
     pub(crate) http2_header_table_size: Option<u32>,
     pub(crate) http2_max_header_list_size: Option<u32>,
+    pub(crate) http2_max_local_error_reset_streams: Option<Option<usize>>,
     pub(crate) connect_timeout: Option<Duration>,
     pub(crate) http2_adaptive_window: Option<bool>,
     pub(crate) local_address: Option<IpAddr>,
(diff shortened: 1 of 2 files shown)


gautamg795:
Closing for now — going to validate this against our internal services first before bringing it upstream. Will reopen once the approach is confirmed. Thanks!

PR #2712, opened 2026-06-30
Wire A32 circuit breaking into xDS channels

mingley:
## Summary

This PR connects gRFC A32 circuit breaking to the xDS request path in `tonic-xds`. The circuit breaker works like a shared pool of slots: before an endpoint attempt begins, the channel reserves one slot for the selected `{cluster, EDS service}`. The CDS `max_requests` value sets the pool size, with the A32 default of 1024 when the field is omitted. If no slot is available, the client returns a local, non-retryable gRPC `UNAVAILABLE` response without contacting an endpoint.

- Circuit-breaking limits follow live CDS updates. A permit stays reserved until the response body completes, errors, or is dropped, so the process-wide counter represents active RPC attempts across channels.
- Local circuit-breaker responses are identified as pre-endpoint outcomes. Retry handling does not retry them, and outlier detection does not count them as endpoint failures.
- The xDS cache preserves resource removals with generation and revision tracking. Removing CDS or EDS state clears obsolete breaker state and endpoints; re-added endpoints wait for a connector from the matching generation instead of reusing stale limits, connectors, or endpoint state.

When CDS omits `max_requests`, `tonic-xds` applies 1024 concurrent in-flight requests, matching [gRFC A32](https://github.com/grpc/proposal/blob/master/A32-xds-circuit-breaking.md), Envoy, Java, and Go. A control plane can effectively disable the breaker by sending an explicitly high `max_requests` value.

Part of #2444.

## Testing Done

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p tonic-xds circuit_breaking` — 33 passed, 0 failed, 378 filtered out.

### DIFF
--- a/tonic-xds/src/client/loadbalance/loadbalancer.rs
+++ b/tonic-xds/src/client/loadbalance/loadbalancer.rs
@@ -62,6 +62,7 @@ use tower::discover::{Change, Discover};
 
 use arc_swap::ArcSwap;
 
+use crate::client::circuit_breaking::is_local_pre_endpoint_response;
 use crate::client::endpoint::{Connector, EndpointAddress};
 use crate::client::loadbalance::channel_state::{
     EjectionConfig, IdleChannel, ReadyChannel, UnejectedChannel,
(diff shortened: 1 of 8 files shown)


tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -72,6 +73,26 @@ use crate::client::loadbalance::outlier_detection::{OutlierDetector, OutlierStat
 use crate::client::loadbalance::pickers::ChannelPicker;
 use crate::xds::resource::outlier_detection::OutlierDetectionConfig;
 
+trait LbResponseOutcome {
```
YutaoMa, line 79: Need some comments on this trait abstraction and its functions.

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -72,6 +73,26 @@ use crate::client::loadbalance::outlier_detection::{OutlierDetector, OutlierStat
...
+    }
+}
+
+impl<B> LbResponseOutcome for http::Response<B> {
+    fn is_local_pre_endpoint_response(&self) -> bool {
+        is_local_pre_endpoint_response(self)
+    }
+}
+
+fn outlier_outcome<Resp: LbResponseOutcome, Error>(result: &Result<Resp, Error>) -> Option<bool> {
```
YutaoMa, line 98: not clear from the signature alone what the returned boolean signals

tonic-xds/src/client/loadbalance/loadbalancer.rs
mingley: Fixed in 221f2219. Documented why the response-outcome trait exists and how its classifier is used.

tonic-xds/src/client/loadbalance/loadbalancer.rs
mingley: Fixed in 221f2219. Documented the `Some(true)`, `Some(false)`, and `None` semantics.

PR #2714, opened 2026-06-30
CI: Use best practices in release-protoc-gen-rust-grpc

ejona86:
Reducing the scope of write permissions is nice and makes it more obvious where the permissions are used. ${{}} in inline shell scripts is just a textual replacement, so it is better to be in the habit of using environment variables and letting the shell handle the replacement. I left ${{ matrix }} cases, since several of those are intended as textual replacements.


dfawley (approved):
Would you mind doing a release on your fork with the right tag name to confirm this is working correctly?

ejona86:
Looks like it still works.

image

PR #2716, opened 2026-07-01
grpc: Implement grpc-status-details-bin rich error details

ejona86:
The existing grpc status APIs have been duplicated in grpc-protobuf, but with the addition of status payloads.  The expectation is these will be removed when absl status is available.  Since we now know the existing grpc status APIs will _not_ be be converted to absl status, they have been changed to use the rust-typical Result.

StatusCodeError is redefined in grpc-protobuf rather than aliased from the grpc crate to allow a future 1:1 migration to absl status.

This is API breaking for existing status usages; grpc-protobuf now uses a different status type and the grpc crate no longer has Status/StatusOr.

This also adds a new public API grpc::StatusError::into_parts() for more efficient conversion. This could have been into_message() or message_mut(), but into_parts() seemed nicer for this struct that is unlikely to have fields added in the future.

CC @arjan-bal, @sauravzg

### DIFF
--- a/grpc-protobuf/build.rs
+++ b/grpc-protobuf/build.rs
@@ -1,0 +1,24 @@
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+
+    println!("cargo:rerun-if-env-changed=GRPC_RUST_REGENERATE_PROTO");
+    if env::var_os("GRPC_RUST_REGENERATE_PROTO").is_some() {
+        let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+        let dependencies = protobuf_well_known_types::get_dependency("protobuf_well_known_types")
+            .into_iter()
+            .map(|d| d.into())
+            .collect();
+
+        grpc_protobuf_build::CodeGen::new()
+            .output_dir(manifest_dir.join("generated"))
+            .include(manifest_dir.join("third_party/googleapis"))
+            .inputs(["google/rpc/status.proto"])
+            .dependencies(dependencies)
+            .client_only()
+            .compile()
+            .unwrap();
+    }
+}
--- a/grpc-protobuf/src/client/client_streaming.rs
+++ b/grpc-protobuf/src/client/client_streaming.rs
@@ -25,8 +25,9 @@
 use std::marker::PhantomData;
 use std::pin::Pin;
 
-use grpc::Status;
-use grpc::StatusOr;
+use crate::Status;
+use crate::StatusOr;
+use crate::trailers_conv::status_from_trailers;
 use grpc::client::CallOptions;
 use grpc::client::InvokeOnce;
 use grpc::client::RecvStream;
@@ -169,7 +170,7 @@ where
         loop {
             let i = rx.recv(&mut res).await;
             if let ResponseStreamItem::Trailers(t) = i {
-                return t.into_status();
+                return status_from_trailers(t);
             }
         }
     }
--- a/grpc-protobuf/src/client/client_streaming.rs
+++ b/grpc-protobuf/src/client/client_streaming.rs
@@ -25,9 +25,6 @@
 use std::marker::PhantomData;
 use std::pin::Pin;
 
-use crate::Status;
-use crate::StatusOr;
-use crate::trailers_conv::status_from_trailers;
 use grpc::client::CallOptions;
 use grpc::client::InvokeOnce;
 use grpc::client::RecvStream;
@@ -46,7 +43,10 @@ use protobuf::MessageView;
 use crate::CallBuilder;
 use crate::ProtoRecvMessage;
 use crate::ProtoSendMessage;
+use crate::Status;
+use crate::StatusOr;
 use crate::client::Internal;
+use crate::trailers_conv::status_from_trailers;
 
 /// Configures a client-streaming call for gRPC Protobuf.  Implements
 /// [`IntoFuture`] which begins the call and resolves to a
--- a/grpc-protobuf/src/status.rs
+++ b/grpc-protobuf/src/status.rs
@@ -1,0 +1,261 @@
+/*
+ *
+ * Copyright 2026 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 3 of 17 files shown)


joshuatants:
Haven't tried it out yet but the changes in `trailers_from_tonic_status` should fix redirects for me. I'll sound off if it doesn't. Thanks!

grpc-protobuf/src/status.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    /// Get the value for `type_url`.
+    pub fn get_payload<'a>(&'a self, type_url: &[u8]) -> Option<&'a [u8]> {
+        self.payloads.get(type_url).map(|v| v.as_slice())
+    }
+
+    /// Set the value for `type_url`.
+    pub fn set_payload(&mut self, type_url: &[u8], payload: &[u8]) {
+        self.payloads.insert(type_url.to_vec(), payload.to_vec());
+    }
```
dfawley, line 203: It might be nice to have examples of how we think usage of this would look in actual user code.  Mainly: how do you get the `type_url` of a proto message type as a byte slice -- assuming there is a way besides hard-coding the string.  If it's just hard coding the string, then I guess we don't really need an example.

grpc-protobuf/src/trailers_conv.rs
```diff
@@ -0,0 +1,428 @@
...
+            let expected_code = StatusCodeError::from(code as i32);
+            match bin_val {
+                None => Err(StatusError::new(expected_code, message)),
+                Some(meta_val) => {
+                    let rpc_status_err = parse_rpc_status(meta_val.as_bytes())?;
+                    if rpc_status_err.code() != expected_code {
+                        return Err(StatusError::new(
+                            StatusCodeError::Internal,
+                            format!(
+                                "RPC status code mismatch: gRPC code {:?}, RPC code {:?}",
```
dfawley: `gRPC code {:?}, RPC status details code {:?}`?  Or `grpc-status-details-bin status code {:?}`?

With gRPC vs. RPC it might not even be very obvious what this is referring to.

grpc-protobuf/src/trailers_conv.rs
```diff
@@ -0,0 +1,428 @@
...
+                        ));
+                    }
+                    Err(rpc_status_err)
+                }
+            }
+        }
+    }
+}
+
+fn parse_rpc_status(buf: &[u8]) -> StatusOr<StatusError> {
```
dfawley, line 78: > `StatusOr<StatusError>`

👀

grpc-protobuf/src/trailers_conv.rs
```diff
@@ -0,0 +1,428 @@
...
+    for any in rpc_status.details() {
+        status_err.set_payload(any.type_url().as_bytes(), any.value());
+    }
+
+    Ok(status_err)
+}
+
+/// Converts the status to trailers and inserts grpc-status-details-bin into the metadata.
+#[allow(dead_code)]
+pub(crate) fn trailers_from_status(s: Status) -> Trailers {
```
dfawley, line 105: This is for server-side?  So this wants to be some kind of `ServerStatus` newtype version of `Status`, right?

grpc-protobuf/src/client/client_streaming.rs
```diff
@@ -25,8 +25,9 @@
 use std::marker::PhantomData;
 use std::pin::Pin;
 
-use grpc::Status;
-use grpc::StatusOr;
+use crate::Status;
+use crate::StatusOr;
+use crate::trailers_conv::status_from_trailers;
```
dfawley: Nit: please format using:

```
rustfmt --edition=2021 --config imports_granularity=Item --config group_imports=StdExternalCrate $(find . -name '*.rs')
```

Which helps organize the imports a bit better than the default behavior.

grpc-protobuf/build.rs
```diff
@@ -0,0 +1,24 @@
...
+
+    println!("cargo:rerun-if-env-changed=GRPC_RUST_REGENERATE_PROTO");
+    if env::var_os("GRPC_RUST_REGENERATE_PROTO").is_some() {
+        let manifest_dir = PathBuf::from(env::var_os("CARGO_MANIFEST_DIR").unwrap());
+        let dependencies = protobuf_well_known_types::get_dependency("protobuf_well_known_types")
+            .into_iter()
+            .map(|d| d.into())
+            .collect();
+
+        grpc_protobuf_build::CodeGen::new()
```
dfawley, line 15: Would it make sense to use the protobuf_build crate directly instead, since there are no services?

grpc-protobuf/src/client/client_streaming.rs
ejona86: Is there a reason we don't put that in a [rustfmt.toml](https://github.com/rust-lang/rustfmt#configuring-rustfmt)?... Because these are both unstable features and unstable features should be disallowed on the command line, but aren't. https://github.com/rust-lang/rustfmt/issues/6257

grpc-protobuf/src/trailers_conv.rs
ejona86: Yeah, when the server-side stuff lands it may want to change this. This API probably would be fine to use as-is though, as the ServerStatus would still include a absl::Status inside.

grpc-protobuf/src/status.rs
ejona86: I think it is simply "a PITA." Even if it were encoded as an Any it looks like you have to do everything yourself: hard-code the name and manually serialize.

grpc-protobuf/src/trailers_conv.rs
ejona86: "Beautiful" ain't it? Error-parsing code makes success produce the parsed error... All seems correct, but definitely not something I originally thought I'd see; I did try to remove it, but the alternatives seemed worse.

grpc-protobuf/build.rs
ejona86: If there were services (and we weren't using them), then I would have. _Because_ there are no services, it didn't seem to matter which was used. grpc_protobuf_build has the protoc-gen-rust-grpc integration and such, so it seemed nicer for now to reuse it. Maybe once we upstream more parts of it it'd make more sense to use protobuf_build directly.

grpc-protobuf/src/trailers_conv.rs
ejona86: I've changed it to `gRPC code {:?}, google.rpc.Code {:?}` which is the enum type.

grpc-protobuf/src/status.rs
dfawley: Maybe this is something we can bring up with the proto team, too.

dfawley (approved):

grpc-protobuf/src/status.rs
ejona86: Any is a standard feature, so I have to believe the proto team is very well aware. I've started discussing with them to get an idea of the timing.

PR #2718, opened 2026-07-04
tonic: document Server::timeout semantics

### LINKED ISSUE #936: Better documentation on tonic::transport::Server::.timeout function
Hello,

I have spent some time digging through the implementation to discover the following:

1. What is the default timeout if no timeout is set (I see that the default value is None, but what does that mean (a fixed timeout, no timeout, ... ?)
2. What does this timeout mean? My understanding is that this regulates the amount of time a given request has to executer, after which it will be dropped. Is this true?

I ask this specifically because by looking for help on this topic I read elsewhere that Tonic server wouldn't have a timeout functionality because this could be controlled by clients, but now I see that this functionality does exist.

btw: muy understanding is that once a graceful shutdown future completes, in-flight requests will complete (respecting this timeout setting) and then the server will stop. Is this assumption correct?

It would be nice to document these behaviors so that others can find this information more quickly.

Thanks,
Marlon

thsfranca:
## Summary
- Expand `Server::timeout` rustdoc to answer the questions in #936: default behavior, client `grpc-timeout` precedence, error mapping, streaming caveat, and graceful shutdown interaction.
- Cross-reference timeout behavior in `serve_with_incoming_shutdown`.

## Test plan
- [x] `cargo test -p tonic --doc`

Fixes: #936

### DIFF
--- a/tonic/src/transport/server/mod.rs
+++ b/tonic/src/transport/server/mod.rs
@@ -209,7 +209,42 @@ impl<L> Server<L> {
         Server { load_shed, ..self }
     }
 
-    /// Set a timeout on for all request handlers.
+    /// Set a maximum duration for each request handler.
+    ///
+    /// If [`timeout`] is not called, the server does not impose a per-request deadline.
+    /// Clients may still send a [`grpc-timeout`][spec] metadata header (for example via
+    /// [`Endpoint::timeout`]), which is enforced even when no server timeout is configured.
+    ///
+    /// # Precedence
+    ///
+    /// When both the server and the client set a deadline, the shorter duration is used.
+    ///
+    /// # Error behavior
+    ///
+    /// The timeout applies to the wall-clock time for the request handler future to
+    /// complete. If the handler does not finish within the budget, the request fails with
+    /// [`TimeoutExpired`], which surfaces to clients as a [`Status`] with code
+    /// [`Code::Cancelled`] and message `"Timeout expired"`.
+    ///
+    /// # Streaming note
+    ///
+    /// The deadline applies to the handler future returned by the service, not to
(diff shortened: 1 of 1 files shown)


tonic/src/transport/server/mod.rs
```diff
@@ -209,7 +209,42 @@ impl<L> Server<L> {
         Server { load_shed, ..self }
     }
 
-    /// Set a timeout on for all request handlers.
+    /// Set a maximum duration for each request handler.
+    ///
+    /// If [`timeout`] is not called, the server does not impose a per-request deadline.
+    /// Clients may still send a [`grpc-timeout`][spec] metadata header (for example via
+    /// [`Endpoint::timeout`]), which is enforced even when no server timeout is configured.
```
nathanielford, line 216: > Clients may still send a [`grpc-timeout`][spec] metadata header (for example via [`Endpoint::timeout`]), which is enforced even when no server timeout is configured.

I think this might be incorrect. `Endpoint::timeout` does not set the `grpc-timeout` header on requests; it only applies a client-side Tower timeout layer (see `Endpoint::timeout` docs: *"This does not set the timeout metadata (grpc-timeout header) on the request, meaning the server will not be informed of this timeout, for that use [`Request::set_timeout`]"*).

Please update this citation and the doc link on line 260 to reference `Request::set_timeout` instead:

```rust
/// Clients may still send a [`grpc-timeout`][spec] metadata header (for example via
/// [`Request::set_timeout`]), which is enforced even when no server timeout is configured.

nathanielford (changes requested):
Looks pretty good! There is one change I think we need, and then we can get this merged.

PR #2722, opened 2026-07-06
tonic/xds: wire up lbv2 as a feature

gu0keno0:
Add tonic-xds' own LB module as a feature

## Motivation

Tower's LB is limited. We cannot support advanced xds features such as customized load balancer and outlier detection on top of it.

We have been developing tonic-xds' own LB for a while, and this PR make it ready for future consumptions.
We will still need to wire up xds configs for OD and RingHash, those can be done in future PRs.

## Solution
Added the wire-up modules to make the tonic-xds' LB usable behind a feature flag.


tonic-xds/src/client/loadbalance/service.rs
```diff
@@ -0,0 +1,156 @@
...
+        let discover = discover_endpoints(&self.cache, cluster_name);
+        let connector = Arc::new(XdsLbConnector::new(
+            self.cache.clone(),
+            cluster_name.to_string(),
+            #[cfg(feature = "_tls-any")]
+            self.cert_provider_registry.clone(),
+        ));
+        let picker: Arc<
+            dyn ChannelPicker<ReadyChannel<EndpointChannel<Channel>>, LbRequest> + Send + Sync,
+        > = Arc::new(P2cPicker);
```
LYZJU2019, line 92: Can you help me understand what's the reason to hardcode `P2cPicker` here?

gu0keno0: This should be replaced by dynamic xds resolution in the near future. The wire up here is just to unblock those clean up works.

gu0keno0: I've added a few TODOs for wiring up ringhash and OD.

YutaoMa:
Found one potential issue with the new LB's readiness logic that needs verification:
1. [`LoadBalancer::poll_ready`](https://github.com/grpc/grpc-rust/blob/39cda4e88d2e532581c3b9749c7150bb726c513f/tonic-xds/src/client/loadbalance/loadbalancer.rs#L328-L375) treats non-empty `self.ready` as sufficient readiness check, but when a cluster is dropped it doesn't evict from `self.ready`, leaving them stale forever. If those endpoints are down `call` will fail forever and even if new CDS config comes in using the same config name, because the code will reuse the existing registry entry, it would continue to reuse the broken set of `self.ready` instead of rebuild one.
2. The old `tower::Balance` path is less problematic because it re-polls readiness on every call and evict broken endpoints.

I think to fix it we either need to do eviction in `call` and/or rebuild entry in `ClusterLbRegistry`.

gu0keno0:
> Found one potential issue with the new LB's readiness logic that needs verification:
> 
> 1. [`LoadBalancer::poll_ready`](https://github.com/grpc/grpc-rust/blob/39cda4e88d2e532581c3b9749c7150bb726c513f/tonic-xds/src/client/loadbalance/loadbalancer.rs#L328-L375) treats non-empty `self.ready` as sufficient readiness check, but when a cluster is dropped it doesn't evict from `self.ready`, leaving them stale forever. If those endpoints are down `call` will fail forever and even if new CDS config comes in using the same config name, because the code will reuse the existing registry entry, it would continue to reuse the broken set of `self.ready` instead of rebuild one.
> 2. The old `tower::Balance` path is less problematic because it re-polls readiness on every call and evict broken endpoints.
> 
> I think to fix it we either need to do eviction in `call` and/or rebuild entry in `ClusterLbRegistry`.

poll_ready calls poll_discovery, so IMO it is poll_discovery's responsibility to handle cluster removal / adding back. Currently this is not working because cluster discovery is not re-established. I have a TODO here: https://github.com/gu0keno0/tonic/blob/39cda4e88d2e532581c3b9749c7150bb726c513f/tonic-xds/src/xds/lb_discovery.rs#L17-L19 .

This is also an issue for tower-lb as well: https://github.com/grpc/grpc-rust/blob/master/tonic-xds/src/xds/cluster_discovery.rs#L179 . If cluster_watch is closed (I think cluster removal triggers it today), cluster discovery dies.

So both tower-lb and lbv2 are under the same issue.

PR #2725, opened 2026-07-08
feat(grpc): add server-side Listener and Transport traits

sauravzg:
Server-owned connection lifecycle and graceful shutdown:

- Server accepts connections from a Listener and manages their lifecycle. Each accepted Transport is bound to a handler via Transport::serve(), which returns a GracefulConnection future that the Server polls and can shut down cooperatively.

- Graceful shutdown follows a pattern combining the tonic Server API with hyper's graceful shutdown guide (https://hyper.rs/guides/0.14/server/graceful-shutdown/).

- ListenerAddress provides a type-safe, downcastable address for listeners, analogous to Go's net.Addr.

Runtime API changes:

- Runtime trait gains tcp_listener() and unix_listener() methods (doc(hidden)) that return EndpointListener (pub(crate)).
- EndpointListener trait (pub(crate)) wraps accept() and local_addr() for server-side listening sockets.
- Tokio implementations: TokioTcpListener, TokioUnixListener.
- New address module (rt::address) with ListenerAddress trait and impls for SocketAddr and UnixListenerAddress.

Breaking changes:

- Server::serve() now takes &impl Listener (sealed) instead of the previous ServerListener. The old ServerListener returned a pre-assembled ServerCall with constructed streams; the new Listener yields raw Transport objects and stream construction moves into Transport::serve().

- serve_with_shutdown() is pub(crate) pending a design decision on whether the shutdown signal should be per-listener or per-server.

New public API surface:

- Dyn versions of handler, recvstream (pub) - Same as client so should be okay
- ListenerAddress trait (pub) — network(), as_any() - Needs to be public because it'll be exposed eventually
- Listener trait (pub, sealed) — accept(), local_addr()
- InMemoryServingConnection (pub) 
- ListenerAddress impl for std::net::SocketAddr
- Runtime::tcp_listener(), Runtime::unix_listener() (doc(hidden))
- Transport trait (pub, doc(hidden), private token) — serve()
- GracefulConnection trait (pub, doc(hidden), private token) — graceful_shutdown()
- UnixListenerAddress (pub(crate))
- InMemoryListenerAddress (crate-private)
- EndpointListener trait (pub(crate))

Open Questions:
- `local_addrs` may need to be updated to be errorable(contrary to our previous discussion), tonic's API is errorable and we currently do an expect.
- `shutdown` per listener vs `shutdown` on server
- The use of associated types creates annoying issues of sealing and private token since dependent types need to be made public.

### DIFF
--- a/grpc/src/rt/address.rs
+++ b/grpc/src/rt/address.rs
@@ -1,0 +1,159 @@
+/*
+ *
+ * Copyright 2026 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use std::any::Any;
+use std::fmt;
+
+/// The address a [`Listener`](super::Listener) is bound to.
+///
+/// All implementations provide a human-readable string via [`Display`].
+/// Callers needing structured data (e.g. `TcpAddress` or `std::net::SocketAddr`) can
+/// downcast via stable trait upcasting coercion (`let any: &dyn Any = addr;`)
+/// and [`downcast_ref`](Any::downcast_ref).
+///
+/// This is analogous to Go's `net.Addr` interface.
+pub trait ListenerAddress: fmt::Display + fmt::Debug + Any + Send + Sync + 'static {
+    /// Returns the network type (e.g. `"tcp"`, `"unix"`, `"inmemory"`).
+    fn network(&self) -> &str;
+}
+
+// ---------------------------------------------------------------------------
+// TcpAddress (Scheme-Prefixed TCP Address Wrapper)
+// ---------------------------------------------------------------------------
+
+/// A scheme-prefixed TCP address wrapper around [`std::net::SocketAddr`].
+/// Formats via `Display` as `tcp://<ip>:<port>`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct TcpAddress(pub std::net::SocketAddr);
+
+impl ListenerAddress for TcpAddress {
+    fn network(&self) -> &str {
+        "tcp"
+    }
+}
+
+impl fmt::Display for TcpAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "tcp://{}", self.0)
+    }
+}
+
+// ---------------------------------------------------------------------------
+// UnixListenerAddress
+// ---------------------------------------------------------------------------
+
+/// Address for a Unix socket listener.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub(crate) struct UnixListenerAddress {
+    path: String,
+}
+
+impl UnixListenerAddress {
+    /// Creates a new Unix listener address.
+    pub(crate) fn new(path: String) -> Self {
+        Self { path }
+    }
+
+    /// Returns the socket path.
+    pub(crate) fn path(&self) -> &str {
+        &self.path
+    }
+}
+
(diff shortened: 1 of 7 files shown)


grpc/src/rt/address.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// The address a [`Listener`](super::Listener) is bound to.
+///
+/// All implementations provide a human-readable string via [`Display`].
+/// Callers needing structured data (e.g. `std::net::SocketAddr`) can
+/// downcast via [`as_any`](ListenerAddress::as_any) and
+/// [`downcast_ref`](Any::downcast_ref).
+///
+/// This is analogous to Go's `net.Addr` interface.
+pub trait ListenerAddress: fmt::Display + Send + Sync + 'static {
```
dfawley: Add `TODO: consider Debug instead` or whatever?

grpc/src/rt/address.rs
```diff
@@ -0,0 +1,153 @@
...
+    /// Returns the network type (e.g. `"tcp"`, `"unix"`, `"inmemory"`).
+    fn network(&self) -> &str;
+
+    /// Enables downcasting to a concrete address type.
+    fn as_any(&self) -> &dyn Any;
```
dfawley: Add `Any` as supertrait instead?

grpc/src/rt/address.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+}
+
+impl fmt::Display for UnixListenerAddress {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.path)
```
dfawley: This should probably say `unix:`?  Or `inmemory` can skip it?  I'm not sure.  But we should be consistent.  I think it probably _does_ belong here, though.

Note that if we did the former, then `SocketAddr` would be wrong, in which case maybe we need a newtype for `SocketAddr` so that we can define `Display` how we want.

grpc/src/server/mod.rs
```diff
@@ -33,53 +34,190 @@ use crate::core::RequestHeaders;
...
+    // TODO(sauravz): The shutdown signal is currently per-listener like tonic
+    // Other implementations have signal on the server level.
+    // We need to evaluate what's the correct behavior here.
+    // Per Listener is more flexible and easy to implement, but
+    // doesn't allign with other implementations.
+    pub(crate) async fn serve_with_shutdown(
+        &self,
+        listener: &impl Listener,
+        signal: impl Future<Output = ()>,
+    ) {
```
dfawley, line 224: I think there's another interesting question here, which is: who runs the task that polls the serving future?

It could be the application, like this is (and your example in the design doc shows).  But instead it could be the grpc library that uses the runtime to fork a task for it.  In the case of the latter it would have to return an object with a `graceful_shutdown` method and would need a `Drop` impl to cancel the task.

I'm not sure of all the pros/cons for these two options, but it's probably worth spending some time thinking about.

grpc/src/rt/address.rs
sauravzg: I've updated the trait to derive debug and also inherit Display, mostly along the lines of `debug` for swe debugging and `display` for end user display.

LMK if you'd like me to just keep debug

grpc/src/rt/address.rs
sauravzg: Done. Also, Removed the `as_any` . It initially stemmed from not realizing our msrv had trait upcasting (1.86) present in our msrv.

grpc/src/rt/address.rs
sauravzg: Everything should now be consistent. Used newtype pattern for SocketAddr .  PTAL if it looks okay.

grpc/src/server/mod.rs
sauravzg: Are there any benefits to grpc polling the future? The extra task is the additional con for me right now.

I haven't thought about it either but the approach is consistent with what tonic, hyper and axum do. I can't immediately think of a pro right away for creating a new task. 

I can take a look at other grpc implementations to see what they do , but IIUC there are probably two patterns `start + wait`(likely java, cpp , python ) and `serve`(go) . At a high level our approach to have the application drive it seems close-ish to them from rust.

Is there any particular advantage of having grpc drive things?


Additionally, updated to take listener by value as discussed in the design.

dfawley: > The extra task is the additional con for me right now.

Isn't it just a matter of _who_ makes the task? Either way there's a task per listener?  Or, I guess if the application owns it, then it's a future, and maybe one task can poll multiple futures?  Is that worth worrying about though?

The advantage of grpc owning it would be: if a library wants to start a grpc server, it would need to figure out what runtime to use to spawn the task.  grpc is already configured with the runtime by the application, so it already knows how to do this.

I don't know if this is a good enough reason to change it, but I think it's worth thinking about the implications of the different paths.

sauravzg: > I guess if the application owns it, then it's a future, and maybe one task can poll multiple futures? Is that worth worrying about though?

Us spawning takes this decision away from the user on how they want to poll the future and forcibly spawn a task instead of maybe doing a join(or any other way of polling multiple futures).

Currently, the additional overhead of a new task without absolutely needing it, doesn't seem like the optimal choice.

Are there any precedences around  peculiar situations that we are trying to unblock here, which could help me think about the implications around both cases?

grpc/src/rt/address.rs
dfawley: I think we should generally prefer `Debug` over `Display`.  We don't often deal with "end user display" since we are a pretty low-level library.

dfawley (commented):
LGTM but I would prefer if you could change to implement `Debug` instead of `Display` for these types before merging.

Also a side note: please in the future avoid force pushing like the plague, _unless_ nobody has done a code review yet, _or_ if the code review is done and you're just trying to get things sorted out on your end before committing.

To update yourself to master you can do `git merge upstream/master` and it will not require a force push like `git rebase upstream/master` would.

But even if you need a rebase, please don't replace commits.  With multiple commits I can more easily see the exact changes made since the last time I reviewed the PR.

Thanks!

dfawley (approved):

sauravzg:
Removed display.

PR #2728, opened 2026-07-09
grpc: expose unstable APIs behind __unstable feature

dfawley:
We'd like to implement our own features outside of the grpc crate that require unstable APIs in the grpc crate.  If we don't implement them in foreign crates, then the situation quickly devolves into "all things must be inside the grpc crate behind feature flags".

This PR introduces an `__unstable` feature flag that publicly re-exports the runtime, LB policy, name resolver, and service config modules, but under a top-level `__unstable` module path to make it as obvious as possible that these features are unstable and should not be relied upon by end users.  It also doc-hides them for two purposes:

1. to prevent them from rendering on docs.rs.
2. to allow for some unstable things to use foreign external types, at least for now (mainly: serde symbols for LB policy config).  (The `check-external-types` checker we run automatically ignores doc-hidden symbols.)

This PR then makes public many symbols that were previously private or pub(crate), so that they can be exported via the `__unstable` module.

cc @gu0keno0 - I think your xds protos change (#2723) should just go ahead and put the protos into a new `grpc-xds` crate which will subsequently be free to use these unstable APIs.

### DIFF
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -48,6 +48,7 @@ tls-rustls = [
     "dep:rustls-platform-verifier",
     "dep:rustls-webpki",
 ]
+__unstable = []
 
 [dependencies]
 base64 = "0.22"
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -40,23 +40,23 @@ use crate::core::RequestHeaders;
 use crate::metadata::MetadataMap;
 use crate::rt::GrpcRuntime;
 
-pub(crate) mod child_manager;
-pub(crate) mod graceful_switch;
-pub(crate) mod lazy;
-pub(crate) mod pick_first;
-pub(crate) mod round_robin;
-pub(crate) mod subchannel;
 pub(crate) mod subchannel_sharing;
+
+pub mod child_manager;
+pub mod graceful_switch;
+pub mod lazy;
+pub mod pick_first;
+pub mod registry;
+pub mod round_robin;
+pub mod subchannel;
+pub use registry::GLOBAL_LB_REGISTRY;
 
 #[cfg(test)]
 pub(crate) mod test_utils;
 
-pub(crate) mod registry;
-pub(crate) use registry::GLOBAL_LB_REGISTRY;
-
 /// An LB policy factory that produces LbPolicy instances used by the channel
 /// to manage connections and pick connections for RPCs.
-pub(crate) trait LbPolicyBuilder: Send + Sync + Debug + 'static {
+pub trait LbPolicyBuilder: Send + Sync + Debug + 'static {
     type LbPolicy: LbPolicy;
 
     /// Builds and returns a new LB policy instance.
@@ -88,7 +88,7 @@ pub(crate) trait LbPolicyBuilder: Send + Sync + Debug + 'static {
 /// LB policies are responsible for creating connections (modeled as
 /// Subchannels) and producing Picker instances for picking connections for
 /// RPCs.
-pub(crate) trait LbPolicy: Send + Sync + Debug + 'static {
+pub trait LbPolicy: Send + Sync + Debug + 'static {
     type LbConfig: Any + Send + Sync + Debug + 'static;
 
     /// Called by the channel when the name resolver produces a new set of
@@ -121,7 +121,7 @@ pub(crate) trait LbPolicy: Send + Sync + Debug + 'static {
 /// A collection of data configured on the channel that is constructing this
 /// LbPolicy.
 #[derive(Debug)]
-pub(crate) struct LbPolicyOptions {
+pub struct LbPolicyOptions {
     /// A hook into the channel's work scheduler that allows the LbPolicy to
     /// request the ability to perform operations on the ChannelController.
     pub work_scheduler: Arc<dyn WorkScheduler>,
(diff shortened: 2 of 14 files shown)


grpc/src/client/service_config.rs
```diff
@@ -26,7 +26,7 @@
 /// a JSON object.
 // TODO: this shouldn't be public; users should set with JSON instead.
```
gu0keno0: May need to clean up this comment?

grpc/src/client/load_balancing/subchannel.rs
```diff
@@ -131,9 +131,7 @@ pub(crate) mod private {
 ///
 /// When a Subchannel is dropped, it is disconnected automatically, and no
 /// subsequent state updates will be provided for it to the LB policy.
-pub(crate) trait Subchannel:
-    private::Sealed + DynHash + DynPartialEq + Any + Send + Sync
-{
+pub trait Subchannel: private::Sealed + DynHash + DynPartialEq + Any + Send + Sync {
```
gu0keno0, line 134: There doesn't seems to be a public / __unstable constructor for Subchannel?

I'm just thinking about how can grpc-xds build Subchannels.

grpc/src/lib.rs
```diff
@@ -55,20 +55,37 @@ pub mod core;
...
 pub use status::StatusCodeError;
 pub use status::StatusError;
 pub use status::StatusOr;
 
-mod byte_str;
-mod rt;
-mod send_future;
+#[cfg(feature = "__unstable")]
+#[doc(hidden)]
+pub mod __unstable {
```
gu0keno0, line 73: Maybe add some comments about the path for a __unstable API to graduate to be stable?

grpc/src/client/load_balancing/mod.rs
```diff
@@ -195,7 +195,7 @@ impl ParsedJsonLbConfig {
         }
     }
 
-    pub(crate) fn from_value(value: serde_json::Value) -> Self {
+    pub fn from_value(value: serde_json::Value) -> Self {
```
gu0keno0, line 198: This would require grpc-xds to update together with grpc, if there's major version bump of serde. But I guess it might be fine? Both crates will be owned by the same group of people anyways.

gu0keno0 (approved):
Some nits / clarification questions.

gu0keno0:
Thanks, it will be great if we can make grpc-xds cleanly separated. My intuition is that it might make 3rd party customization (e.g. custom LB) as capable as the built-in LBs -- great!

grpc/src/client/load_balancing/mod.rs
ejona86: In an earlier conversation with Doug, we were thinking we'd need to pin the version of grpc that grpc-xds uses, because of the unstable API. So that part probably doesn't matter. But, yeah, we should keep an eye out here if we can avoid exposing the dep. (C uses serialized json in cases like this. Java uses JSON object model with regular Java objects (list, map, string, double).) Can be done later.

grpc/src/client/service_config.rs
ejona86: I think the idea would be that'd we'd still hide this, or make it opaque. But I assume some other things will need to land first. "set with JSON" may be referring to what I mentioned in the other comment, about passing JSON as an encoded string.

ejona86 (approved):
This doesn't seem too hard to reason about, as there are clear groupings that are pretty easy to remember ("the things that are unstable in any language that exposes them"; name resolver, lb, lower-level transport details aka runtime)

grpc/src/client/service_config.rs
dfawley: This particular one needs more design behind it.  I think it's still a TODO.

Basically, the name resolver needs to produce something that represents a parsed service config.  But it should be opaque.  I made the field `pub(crate)` for now to ensure it stays opaque.

grpc/src/client/load_balancing/subchannel.rs
dfawley: This is by design.  To wrap a `Subchannel` you implement `ForwardingSubchannel` instead (which implements `Subchannel`).  You implement the `ForwardingSubchannel::delegate` method plus any other methods you want to override from `Subchannel`.  This pattern allows us to add methods to `Subchannel` in the future if needed.

grpc/src/client/load_balancing/mod.rs
dfawley: Correct.  Removing serde from our API at this point is going to be a lot of work, but something we'd probably have to do if we ever decided we must have a stable LB policy API.  For now I thought I'd just leave it as-is so we can make progress on grpc-xds (and health checking and ORCA and other features), and we'll figure out the details here some time in the future, since it's all unstable and meant for our own use anyway (from other crates).

grpc/src/lib.rs
dfawley: Hmm, what's the best way to document "hidden" feature flags?  Should I just put a comment in our `Cargo.toml` that says "for internal gRPC use only; users should never set this or depend on APIs it exposes"?  I think for now I don't want to set any expectations that any of these things will become public any time soon.

Pretty soon we'll need to make the runtime traits public, but sealed to prevent implementation.

grpc/src/client/load_balancing/subchannel.rs
gu0keno0: Got it, so xds can call ChannelController.new_subchannel to make it.

grpc/src/lib.rs
gu0keno0: Sounds good, agree that soon we will know it better.

gu0keno0 (approved):

grpc/Cargo.toml
```diff
@@ -48,6 +48,7 @@ tls-rustls = [
     "dep:rustls-platform-verifier",
     "dep:rustls-webpki",
 ]
+__unstable = []
```
LucioFranco, line 51: with additive features this could be enabled when users don't expect, this is a reason tokio uses cfg features over cargo features like this. I believe though that a large reason for that is because the flag can change behavior in the implementation. So I think if we stick to only public exports that are additive then this is fine for now, but be aware of this.

LucioFranco (approved):

PR #2729, opened 2026-07-10
fix(tonic-xds): reject a zero cert refresh_interval

W4lspirit:
## Context
The file-watcher certificate provider's duration deserializer already rejects negative refresh intervals, but a refresh_interval of "0s" was accepted and parsed to `Duration::ZERO`.

`tokio::time::interval(ZERO)` then panics inside the spawned refresh task, silently killing certificate rotation while provider construction still succeeds.

## Motivation

Improve runtime error message when user miss-configured the refresh duration. 

## Solution
Extend the existing guard to reject zero as well as negative durations, so the misconfiguration surfaces as a clear bootstrap error (matching grpc-java's `refreshInterval > 0` [check](https://github.com/grpc/grpc-java/blob/v1.83.x/xds/src/main/java/io/grpc/xds/internal/security/certprovider/FileWatcherCertificateProviderProvider.java#L127)).

### DIFF
--- a/tonic-xds/src/xds/cert_provider/file_watcher.rs
+++ b/tonic-xds/src/xds/cert_provider/file_watcher.rs
@@ -69,12 +69,14 @@ where
     let secs: f64 = num
         .parse()
         .map_err(|_| serde::de::Error::custom(format!("invalid duration number: '{num}'")))?;
-    if secs <= 0.0 {
+    let duration = Duration::try_from_secs_f64(secs)
+        .map_err(|e| serde::de::Error::custom(format!("invalid duration '{s}': {e}")))?;
+    if duration.is_zero() {
         return Err(serde::de::Error::custom(format!(
             "invalid duration '{s}': must be greater than 0"
         )));
     }
-    Ok(Some(Duration::from_secs_f64(secs)))
+    Ok(Some(duration))
 }
 
 /// A certificate provider that reads PEM files from disk.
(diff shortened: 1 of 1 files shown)


dfawley:
I just noticed this is still a draft after assigning @YutaoMa as a review -- is this still supposed to be a draft @W4lspirit?

W4lspirit:
@dfawley  Sorry I was responding to an incident, I was waiting for the cla check to  pass before moving it to ready

tonic-xds/src/xds/cert_provider/file_watcher.rs
```diff
@@ -69,9 +69,9 @@ where
...
         .parse()
         .map_err(|_| serde::de::Error::custom(format!("invalid duration number: '{num}'")))?;
-    if secs < 0.0 {
+    if secs <= 0.0 {
         return Err(serde::de::Error::custom(format!(
-            "invalid duration '{s}': must not be negative"
+            "invalid duration '{s}': must be greater than 0"
         )));
     }
     Ok(Some(Duration::from_secs_f64(secs)))
```
YutaoMa: Could you help make this use `try_from_secs_f64` too? Realized that other f64 values like NaN could be causing panics.

W4lspirit: done in https://github.com/grpc/grpc-rust/pull/2729/commits/ae924efdc24ade300d61ccbd18282c5a883cfc4d

YutaoMa (approved):

PR #2730, opened 2026-07-11
fix(reflection): preserve custom options in served descriptors

### LINKED ISSUE #2719: tonic-reflection strips custom options (e.g. google.api.http) by re-encoding decoded FileDescriptorProtos
## Summary

`tonic-reflection` silently drops **custom options** — extensions on `MethodOptions` / `FieldOptions` / etc., e.g. `google.api.http` — from the descriptors it serves. Standard definitions (services, methods, messages, enums) round-trip fine, but any custom option is removed, so reflection consumers that depend on them (REST transcoding / grpc-gateway-style gateways, `buf`, grpcui, validation extensions) can't see them.

## Root cause

`ReflectionServiceState` stores each file as a **decoded** `prost_types::FileDescriptorProto` and **re-encodes** it on every response, so the round-trip is `raw bytes → decode → encode`, and the decode is lossy for custom options.

In `tonic-reflection/src/server/mod.rs`:

- **Register (lossy decode)** — `ReflectionServiceState::new`:
  ```rust
  for encoded in encoded_file_descriptor_sets {
      file_descriptor_sets.push(FileDescriptorSet::decode(encoded)?);   // ~line 129
  }
  // stored as the decoded form; raw bytes discarded:
  files:   HashMap<String, Arc<FileDescriptorProto>>,   // ~line 117
  symbols: HashMap<String, Arc<FileDescriptorProto>>,
  ```
  `prost_types::MethodOptions` has no field for the `google.api.http` extension (field `72295728`, in the `extensions >= 1000` range), and prost drops unknown fields on decode — so the option is lost right here. The original bytes (`encoded_file_descriptor_sets: Vec<&[u8]>`) are only used to decode and are not retained.

- **Serve (re-encode the stripped form)** — `symbol_by_name` (`file_containing_symbol`) and `file_by_filename`:
  ```rust
  Some(fd) => {
      let mut encoded_fd = Vec::new();
      fd.clone().encode(&mut encoded_fd) ... ;   // ~lines 257 / 271
      Ok(encoded_fd)
  }
  ```
  This re-encodes the already-stripped `FileDescriptorProto`, so the served bytes never contain the option.

## Reproduction

1. Compile a proto whose method carries `option (google.api.http)` (importing `google/api/annotations.proto`) and emit a `FileDescriptorSet` via `file_descriptor_set_path`.
2. `Builder::configure().register_encoded_file_descriptor_set(FDS).build_v1()`, serve it.
3. Fetch it back over `ServerReflectionInfo` → `file_containing_symbol` and decode the returned `FileDescriptorProto`.

The returned method's options no longer contain the `google.api.http` extension.

Concretely: in my setup the raw `echo.proto` in the `FileDescriptorSet` is **2744 bytes** (a `DescriptorPool` built directly from that same set resolves the `google.api.http` binding), but the bytes reflection returns are **2639** — exactly a prost decode→encode round-trip, with the option gone. Standard `+json` transcoding works (message types survive); only the custom option is missing.

## Impact

Every reflection consumer that reads custom options is affected over tonic-reflection: `google.api.http` REST transcoding, `google.api.method_signature`, `buf.validate` / protoc-gen-validate rules, etc. Go / C++ / Java reflection servers preserve these because they serve the original descriptor bytes.

## Proposed fix

Serve the **original per-file bytes** instead of re-encoding the decoded `FileDescriptorProto`. The raw `FileDescriptorSet` bytes are already in hand at registration; keep a `name → raw file bytes` map (split the set's `repeated file = 1` entries; decode only to read `name` and build the symbol index) and return those raw bytes from `file_containing_symbol` / `file_by_filename`. No public API change, no behavior change beyond faithfully preserving options. Happy to open a PR if this direction is agreeable.

## Environment

- `tonic-reflection` 0.14.6, `prost-types` 0.14.4

debdattabasu:
Fixes #2719

## Motivation

`tonic-reflection` silently dropped **custom options** — extensions on `MethodOptions` / `FieldOptions` / etc., e.g. `google.api.http` — from the descriptors it served. `ReflectionServiceState` stored each file as a decoded `prost_types::FileDescriptorProto` and re-encoded it on every response, so the round-trip was `raw bytes → decode → encode`. prost drops unknown fields on decode, so any extension in the `>= 1000` range was lost right at registration.

This breaks every reflection consumer that reads custom options: `google.api.http` REST transcoding, `google.api.method_signature`, `buf.validate` / protoc-gen-validate rules, etc. The Go / C++ / Java reflection servers preserve these because they serve the original descriptor bytes.

## Solution

Serve the **original per-file bytes** instead of re-encoding the decoded form.

- For encoded descriptor sets (`register_encoded_file_descriptor_set`), split the set into its per-file byte ranges at the wire level (field 1) **without** decoding the inner message, and store those raw bytes. The `FileDescriptorProto` is still decoded — but only to build the symbol index; the decoded value is discarded and the original bytes are what gets served.
- Sets registered already-decoded (`register_file_descriptor_set`) have no original bytes, so they are re-encoded as before (unavoidable for that entry point).

The internal representation for both `files` and `symbols` becomes `Arc<[u8]>` (per-file encoded bytes). No public API change; the processing order (and therefore dedup precedence and `service_names` ordering) is preserved. The only behavior change is faithfully preserving options.

## Tests

Adds `test_custom_options_are_preserved`, which registers a `FileDescriptorProto` carrying an extension-range field (`72295728`, standing in for `google.api.http`) and asserts the served bytes equal the original for **both** `FileByFilename` and `FileContainingSymbol` lookups. It includes a sanity check (`assert_ne!`) proving a naive prost decode/re-encode round-trip would strip the field — so the test genuinely guards the regression rather than being a tautology. All existing reflection tests continue to pass.

### DIFF
--- a/tonic-reflection/src/server/mod.rs
+++ b/tonic-reflection/src/server/mod.rs
@@ -137,22 +137,35 @@ impl ReflectionServiceState {
             symbols: HashMap::new(),
         };
 
-        // Sets provided already-decoded have no original bytes to preserve, so
-        // re-encode each file (custom options may already have been dropped when
-        // the caller decoded the set into `prost_types`).
-        for fds in file_descriptor_sets {
-            for fd in fds.file {
-                let raw = fd.encode_to_vec();
+        // The builder accepts descriptor sets in two forms: raw bytes (via
+        // `register_encoded_file_descriptor_set`) and already-decoded
+        // `prost_types::FileDescriptorSet`s (via `register_file_descriptor_set`).
+        //
+        // The encoded sets are registered first, on purpose: `register_file`
+        // keeps the first entry seen for a given file name (see below), and the
+        // encoded form is the one that faithfully preserves custom options. So
+        // if the same file arrives in both forms, the lossless bytes win.
+        //
+        // Each encoded set is split into its per-file byte ranges without
+        // decoding the inner message, so the original bytes (including custom
+        // options such as `google.api.http`) are retained and served unchanged.
+        for encoded in encoded_file_descriptor_sets {
+            for raw in split_file_descriptor_set(encoded)? {
+                let fd = FileDescriptorProto::decode(raw.as_slice())?;
                 state.register_file(&fd, Arc::from(raw), use_all_service_names)?;
             }
         }
 
-        // Encoded sets are split into their per-file byte ranges without
-        // decoding the inner message, so the original bytes (including custom
-        // options) are retained and served unchanged.
-        for encoded in encoded_file_descriptor_sets {
-            for raw in split_file_descriptor_set(encoded)? {
-                let fd = FileDescriptorProto::decode(raw.as_slice())?;
+        // Already-decoded sets carry no original bytes, so each file has to be
+        // re-encoded. This is lossy for custom options, but unavoidable: those
+        // options were already dropped when the caller decoded the set into
+        // `prost_types` (which has no field to hold them). Callers that need
+        // custom options preserved should use `register_encoded_file_descriptor_set`
+        // instead. Registered second, as a fallback for files not already
+        // supplied in encoded form above.
+        for fds in file_descriptor_sets {
+            for fd in fds.file {
+                let raw = fd.encode_to_vec();
                 state.register_file(&fd, Arc::from(raw), use_all_service_names)?;
             }
         }
@@ -173,6 +186,9 @@ impl ReflectionServiceState {
             Some(n) => n,
         };
 
+        // First registration for a file name wins; later duplicates are ignored.
+        // `new` registers the option-preserving encoded sets before the
+        // re-encoded ones, so a file present in both keeps its original bytes.
         if self.files.contains_key(&name) {
             return Ok(());
         }
--- a/tonic-reflection/src/server/mod.rs
+++ b/tonic-reflection/src/server/mod.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::fmt::{Display, Formatter};
 use std::sync::Arc;
 
+use prost::encoding::{WireType, decode_key, decode_varint};
 use prost::{DecodeError, Message};
 use prost_types::{
     DescriptorProto, EnumDescriptorProto, FieldDescriptorProto, FileDescriptorProto,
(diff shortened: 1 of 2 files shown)


tonic-reflection/src/server/mod.rs
```diff
@@ -114,74 +115,98 @@ impl<'b> Builder<'b> {
...
-
         let mut state = ReflectionServiceState {
             service_names,
             files: HashMap::new(),
             symbols: HashMap::new(),
         };
 
+        // Sets provided already-decoded have no original bytes to preserve, so
+        // re-encode each file (custom options may already have been dropped when
+        // the caller decoded the set into `prost_types`).
```
dfawley: I'm not able to parse this sentence.  "The FileDescriptorSets provided are already-decoded, so they have no ..." maybe?

tonic-reflection/src/server/mod.rs
```diff
@@ -114,74 +115,98 @@ impl<'b> Builder<'b> {
...
+            }
+        }
 
-                let fd = Arc::new(fd);
-                state.files.insert(name, fd.clone());
-                state.process_file(fd, use_all_service_names)?;
+        // Encoded sets are split into their per-file byte ranges without
+        // decoding the inner message, so the original bytes (including custom
+        // options) are retained and served unchanged.
+        for encoded in encoded_file_descriptor_sets {
```
dfawley, line 152: I don't really understand the code here (the original).  Why are both decoded and encoded FD sets supported?  Either way maybe the comment should mention this is done second so that any duplicates will result in the raw bytes version being used instead?

tonic-reflection/src/server/mod.rs
```diff
@@ -114,74 +115,98 @@ impl<'b> Builder<'b> {
...
+        let name = match fd.name.clone() {
+            None => {
+                return Err(Error::InvalidFileDescriptorSet("missing name".to_string()));
+            }
+            Some(n) => n,
+        };
+
+        if self.files.contains_key(&name) {
+            return Ok(());
+        }
```
dfawley, line 194: We don't overwrite the value in the map?  Then should the order in `new` be reversed?

dfawley (commented):
Looks good to me with one comment nit and one question.  @LucioFranco should probably also take a quick pass since I'm not really familiar with this code.

tonic-reflection/src/server/mod.rs
debdattabasu: Good catch. In the case where the same file name is registered via both prost and raw bytes, this would keep the prost round tripped copy instead of the raw bytes. It should be the other way round. Usually people would use one or the other but this should be fixed regardless. Will fix. 

Edit: Done

tonic-reflection/src/server/mod.rs
debdattabasu: There are 2 pre existing methods to register.
1. register_file_descriptor_set: takes prost_types::FileDescriptorSet.
2. register_encoded_file_descriptor_set: takes raw bytes. 

The previous code was coercing both into prost and now we coerce both into raw thats the change. 

The updated code in this pull request still supports both pathways and as per your comment on R176 to R178, I would make sure 2 is given precedence in my next commit.

Edit: Done

tonic-reflection/src/server/mod.rs
debdattabasu: I mean if someone has already decoded raw bytes into prost and they pass that in, we can't get the missing fields out because we simply don't have access to them so we have to re encode them exactly same as the previous version and there's not much to be done about it. If someone cares about http fields they should pass in as register_encoded_file_descriptor_set not register_file_descriptor_set. Will update the comment. 

Edit: Done

dfawley (approved):
LGTM, but @LucioFranco should ideally also review before merging.

PR #2731, opened 2026-07-13
grpc: add service config types and parsing

nathanielford:
## Motivation

ServiceConfig requires an internal representation that we can de/serialize. 

## Solution

This CL implements the ServiceConfig struct and necessary plumbing for de/serialization. In a follow-on CL(s) we will be adding the plumbing to fully utilize the ServiceConfig, including adding the option to the `Channel` builde, passing it to the load balancer tree for utilization, and interactions with the Name Resolver. 

## Notes

* Serde-specific data transfer objects were created so that we weren't exporting foreign types, and giving us options for de/serialization without breaking users going forward.
* Initially this PR included deserializing stringified u32 types, in concordance with C++ and Java. We may re-implement this going forward, but for now we are forcing the json to have valid raw integers. The concern here is making sure bugs don't arise because macro annotations get missed on edits. We may decide this is stable enough that we don't care, or we may have an internal type (i.e. GrpcU32) like we do with GrpcDuration.


grpc/src/client/service_config/json.rs
```diff
@@ -0,0 +1,220 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use serde::Deserialize;
+use std::time::Duration;
```
dfawley: Nit: please try to follow import groupings: std, external, internal.

grpc/src/client/service_config/json.rs
```diff
@@ -0,0 +1,220 @@
...
+
+// See [service config format notes](https://protobuf.dev/reference/protobuf/google.protobuf/#duration)
```
dfawley: The link text here may be misleading.  The link itself is to the proto duration parsing page, not service config notes.

Also, without triple slashes, I don't think this ever renders as a link anyway?  (Probably should triple-slash this?)

grpc/src/client/service_config/json.rs
```diff
@@ -0,0 +1,220 @@
...
+        assert_eq!(parse_duration("0.000000001s").unwrap(), Duration::new(0, 1));
+        assert_eq!(parse_duration("1.0000001s").unwrap(), Duration::new(1, 100));
+        assert_eq!(
+            parse_duration("1.0001s").unwrap(),
+            Duration::new(1, 100_000)
+        );
+
+        assert!(parse_duration("").is_err());
+        assert!(parse_duration("1").is_err());
+        assert!(parse_duration("1.s").is_err());
```
dfawley: Is `".5s"` valid or an error?  Please add a test for it.

grpc/src/client/service_config/json.rs
```diff
@@ -0,0 +1,220 @@
...
+) -> Result<Option<Duration>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let opt = Option::<String>::deserialize(deserializer)?;
+    opt.map(|s| parse_duration(&s).map_err(serde::de::Error::custom))
+        .transpose()
+}
+
+// This included to allow deserializing both u32 and string (or "stringified")
```
dfawley:
```suggestion
// This is included to allow deserializing both u32 and string (or "stringified")
```
?

grpc/src/client/service_config/json.rs
```diff
@@ -0,0 +1,220 @@
...
+// representations of u32, as well as null values. This is used for the
+// `max_concurrent_streams` field in the service config.
```
dfawley: We don't have a `max_concurrent_streams` in service config, do we?

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
```
dfawley: As above.  Also please avoid use clases with braces.  rust-analyzer can be configured to not do this, and to do the grouping, too.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+/// An in-memory representation of a service config, provided to gRPC as a JSON object.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ServiceConfig {
+    /// Ordered list of LB policies. The first supported one is used.
+    pub load_balancing_config: Option<Vec<LoadBalancingConfig>>,
+    /// Per-method configuration overrides.
+    pub method_config: Option<Vec<MethodConfig>>,
+    /// Global retry throttling parameters.
+    pub retry_throttling: Option<RetryThrottlingPolicy>,
```
dfawley: Service config should be opaque for everything but grpc itself, so these fields should all be `pub(crate)`, and all the related ones below.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+ *
+ */
+
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+
+pub(crate) mod json;
+
+/// An in-memory representation of a service config, provided to gRPC as a JSON object.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
```
dfawley: How can we avoid deriving Serialize/Deserialize on a pub type?  We want to avoid these foreign (serde) symbols in our crate.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+    #[serde(default, deserialize_with = "json::deserialize_uint32_opt")]
+    pub max_request_message_bytes: Option<u32>,
+    #[serde(default, deserialize_with = "json::deserialize_uint32_opt")]
+    pub max_response_message_bytes: Option<u32>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct MethodName {
+    /// e.g., "grpc.examples.echo.Echo" or "grpc.examples.echo.Echo/Echo".
+    /// The service name (fully qualified).
```
dfawley: The slash would be the separator between service and method.  So I wouldn't expect that we'd allow this to have a slash in it.

https://github.com/grpc/grpc-proto/blob/99135b19189588fcc787acb84cff27991787473d/grpc/service_config/service_config.proto#L45

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct MethodConfig {
+    /// List of methods this config applies to.
+    pub name: Vec<MethodName>,
+    #[serde(default, deserialize_with = "json::deserialize_duration_opt")]
+    pub timeout: Option<Duration>,
+    pub retry_policy: Option<RetryPolicy>,
+    #[serde(default, deserialize_with = "json::deserialize_uint32_opt")]
```
dfawley: If we're not supporting strings (are you sure we have to?), then `#[serde(default)]` handles `Option<u32>` just fine.

grpc/src/client/service_config/json.rs
dfawley: Why are we supporting stringified numbers?  Is that a requirement?

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        use serde::ser::SerializeMap;
+        let mut map = serializer.serialize_map(Some(1))?;
+        map.serialize_entry(&self.name, &self.config)?;
+        map.end()
+    }
+}
+
+impl ServiceConfig {
+    /// Parses a service configuration from a JSON string.
+    pub fn parse(config_json: &str) -> Result<Self, String> {
```
dfawley: I believe we might want a `service_config::ParseResult` type here that is `Result<ServiceConfig, String>`, for newtype reasons.  The name resolve will be passing the full `ParseResult` to the channel.

(Go reference: https://pkg.go.dev/google.golang.org/grpc@v1.82.1/serviceconfig#ParseResult + passed to channel with https://pkg.go.dev/google.golang.org/grpc/resolver#State)

grpc/src/client/channel.rs
```diff
@@ -464,9 +463,9 @@ impl name_resolution::ChannelController for ResolverChannelController {
     fn update(&mut self, update: ResolverUpdate) -> Result<(), String> {
         let json_config = if let Ok(Some(service_config)) = update.service_config.as_ref()
             && service_config
-                .load_balancing_policy
+                .load_balancing_config
                 .as_ref()
-                .is_some_and(|p| *p == LbPolicyType::RoundRobin)
+                .is_some_and(|lbc| lbc.iter().any(|c| c.name == "round_robin"))
         {
             json!([{round_robin::POLICY_NAME: {}}])
```
dfawley: I think we can delete the first part of this `if` and just use pick_first if there isn't an LB policy in the config at all.  Unless you're saving that for the next PR because it cases even more changes?

grpc/src/client/channel.rs
dfawley: This can be deleted if you do the above?

grpc/src/inmemory/mod.rs
```diff
@@ -428,9 +428,14 @@ impl Resolver for InMemoryResolver {
...
-                    crate::client::service_config::LbPolicyType::RoundRobin,
-                ),
+                load_balancing_config: Some(vec![
+                    crate::client::service_config::LoadBalancingConfig {
+                        name: "round_robin".to_string(),
+                        config: serde_json::Value::Object(serde_json::Map::new()),
+                    },
+                ]),
+                method_config: None,
+                retry_throttling: None,
```
dfawley: Can this use the `channel_controller.parse_service_config` on a string instead?

grpc/src/client/service_config/json.rs
nathanielford: Done. And fixed my IDE to handle this now.

grpc/src/client/service_config/json.rs
nathanielford: Fixed the language here - LMK what you think.

grpc/src/client/service_config/json.rs
nathanielford: An error. I've added a test for this.

grpc/src/client/service_config/json.rs
nathanielford: Done!

nathanielford: I read the spec as requiring it, but re-reading it I think there are multiple ways to parse it. In any event, it appears that Golang does _not_ support stringification, but Java and C++ do:

Java:
* Call site: https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ManagedChannelServiceConfig.java#L278
* Deserialization function: https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/ServiceConfigUtil.java#L125-L258

C++:
* Call site:  https://github.com/grpc/grpc/blob/master/src/core/lib/channel/message_size_filter.cc#L74-L78
* Deserialization function: https://github.com/grpc/grpc/blob/master/src/core/util/json/json_object_loader.cc#L29-L41

I think in the original design it deferred to the C++ implementation. Let me know if you want to pull this, however, and we can only accept pure numbers.

grpc/src/client/channel.rs
nathanielford: That'll be in the next PR. I just wanted this to compile.

grpc/src/client/service_config/json.rs
nathanielford: Yeah, I crossed my wire there. Fixed it to what I meant.

grpc/src/inmemory/mod.rs
nathanielford: Done.

grpc/src/client/service_config/mod.rs
nathanielford: I've gone ahead and implemented this. We didn't talk about it in the design stage, but it seems like a good idea. That said, take a look and let me know if that fits properly.

grpc/src/client/service_config/mod.rs
```diff
@@ -2,142 +2,177 @@
  *
  * Copyright 2026 gRPC authors.
  *
- * Permission is hereby granted, free of charge, to any person obtaining a copy
```
dfawley: I don't believe this wanted to be rewrapped.  It should match all the other source files licenses and this line was 80 columns before, which is fine.

grpc/src/client/service_config/mod.rs
dfawley: Let's do a simple `pub type ParseResult = Result<ServiceConfig, String>` to avoid all the reimplementations.

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,249 @@
...
+    pub(crate) max_tokens: u32,
+    pub(crate) token_ratio: f32,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct RetryPolicySerDe {
+    pub(crate) max_attempts: u32,
+    #[serde(deserialize_with = "json::deserialize_duration")]
+    pub(crate) initial_backoff: Duration,
```
dfawley: Can we make our own type here instead of using duration (ideally `type JsonDuration = Duration`) and then make that parse specially without needing the `#[deserialize_with=]` on every usage inside the parent struct?  We could do the same thing with `Option<u32>` (if we have to support stringified numbers which is TBD).

grpc/src/client/service_config/json.rs
dfawley: I'd rather avoid it, but let me ask others.

grpc/src/client/service_config/mod.rs
nathanielford: Done!

grpc/src/client/service_config/mod.rs
nathanielford: Fixed!

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,310 @@
...
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct MethodConfigSerDe {
+    pub(crate) name: Vec<MethodNameSerDe>,
+    pub(crate) wait_for_ready: Option<bool>,
+    pub(crate) timeout: Option<GrpcDuration>,
+    pub(crate) retry_policy: Option<RetryPolicySerDe>,
+    pub(crate) hedging_policy: Option<HedgingPolicySerDe>,
+    #[serde(default, deserialize_with = "deserialize_uint32_opt")]
+    pub(crate) max_request_message_bytes: Option<u32>,
```
dfawley: Per the other language leads, we should support stringification.  But we theoretically should be doing it for every number field.  So maybe we need our own SerdeU32 type or something like that?

grpc/src/client/service_config/serde_bindings.rs
```diff
@@ -0,0 +1,310 @@
...
+    pub(crate) hedging_policy: Option<HedgingPolicySerDe>,
+    #[serde(default, deserialize_with = "deserialize_uint32_opt")]
+    pub(crate) max_request_message_bytes: Option<u32>,
+    #[serde(default, deserialize_with = "deserialize_uint32_opt")]
+    pub(crate) max_response_message_bytes: Option<u32>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub(crate) struct MethodNameSerDe {
+    pub(crate) service: String,
```
dfawley, line 62: Should literally every field be in an `Option<>`?  Or what is the failure mode supposed to be if a single field fails to parse?

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+}
+
+impl From<MethodConfigSerDe> for MethodConfig {
+    fn from(dto: MethodConfigSerDe) -> Self {
+        Self {
+            name: dto.name.into_iter().map(Into::into).collect(),
+            wait_for_ready: dto.wait_for_ready,
+            timeout: dto.timeout.map(Into::into),
```
dfawley: Both of these are GrpcDurations? (Clippy is unhappy)

dfawley (commented):
Thanks for the updates.  The Duration stuff LGTM; glad we can do this without needing to annotate every instance of the field individually.

grpc/src/client/service_config/serde_bindings.rs
nathanielford: If deserialization fails, it should return an Error, which will be bubbled up appropriately. Probably better to fail here directly I think?

grpc/src/client/service_config/serde.rs
nathanielford: I've introduced a newtype for that.

grpc/src/client/service_config/serde.rs
nathanielford: This is fixed!

grpc/src/client/service_config/serde.rs
nathanielford: (See GrpcDuration and SerdeU32)

grpc/src/client/service_config/json.rs
nathanielford: To close the loop: we determined we need to support stringified u32s and have introduced a SerdeU32 newtype to handle the deserialization w/o needing to resort to annotations.

grpc/src/client/service_config/serde_bindings.rs
dfawley: I think I looked at everything, and I think this the only problem.  It is valid to have a fully empty method_config.name.  See this example:

https://github.com/grpc/grpc-proto/blob/99135b19189588fcc787acb84cff27991787473d/grpc/service_config/service_config.proto#L58

However, if `service` is unset, then we must validate that `method` is also unset.

grpc/src/client/service_config/mod.rs
```diff
@@ -409,4 +427,83 @@ mod test {
...
+    fn test_default_method_config_empty_name() {
+        let json_data = json!({
+            "methodConfig": [
+                {
+                    "name": [],
+                    "timeout": "5s",
+                    "waitForReady": true
+                },
+                {
+                    "name": [{}],
```
dfawley, line 251: So....this doesn't error?  I was assuming it would, given how `MethodName` is defined without `service` being `Option<>`.  But since this passes, that means it allows `service` to be unset, and it uses its default value?

In that case, are we sure we are employing `Option` only where it's needed?  Like we probably don't need it on `method` either; we can just consider unset and `""` as equivalent.

grpc/src/client/service_config/mod.rs
dfawley: This seems to still be changed?

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,344 @@
...
+    #[serde(default)]
+    pub(crate) service: String,
+    pub(crate) method: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct RetryThrottlingPolicySerde {
+    pub(crate) max_tokens: SerdeU32,
+    pub(crate) token_ratio: f32,
```
dfawley: Do we need to support stringified numbers for floats, too?

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,344 @@
...
+    pub(crate) max_attempts: SerdeU32,
+    pub(crate) hedging_delay: GrpcDuration,
+    #[serde(default)]
+    pub(crate) non_fatal_status_codes: Option<Vec<String>>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct HealthCheckConfigSerde {
+    #[serde(rename = "serviceName", alias = "ServiceName")]
```
dfawley: Why the `alias` for this field?

grpc/src/client/service_config/serde_bindings.rs
```diff
@@ -0,0 +1,344 @@
...
+#[serde(rename_all = "camelCase")]
+pub(crate) struct HealthCheckConfigSerde {
+    #[serde(rename = "serviceName", alias = "ServiceName")]
+    pub(crate) service_name: Option<String>,
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct ConnectionScalingSerde {
+    #[serde(default = "default_max_connections_per_subchannel")]
```
dfawley, line 83: Why is the field named in this `default` statement, but none of the others?

nathanielford: This is defering the defaulting to the `default_max_connections_per_subchannel` function. (Which defaults it to 10.)

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// Global retry throttling parameters.
+    pub(crate) retry_throttling: Option<RetryThrottlingPolicy>,
+    /// Health check configuration.
+    pub(crate) health_check_config: Option<HealthCheckConfig>,
+    /// Connection scaling configuration.
+    pub(crate) connection_scaling: Option<ConnectionScaling>,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct MethodConfig {
```
dfawley: pub(crate) (and below)

But maybe we don't need these parallel types at all?

nathanielford: Parallel types removed. I think it'll be fine. 
image

grpc/src/client/service_config/serde.rs
nathanielford: This was the result of some confusion in my original research and it snuck through. At some point Go was supporting PascalCase for this field, but we're explicitly not doing that any more in any language. I've removed this.

grpc/src/client/service_config/serde.rs
nathanielford: This has been implemented.

grpc/src/client/service_config/mod.rs
nathanielford: I've switched it to be a naked string, defaulting to "" as well. :)

grpc/src/client/service_config/serde_bindings.rs
nathanielford: This validation is in place.

grpc/src/client/service_config/mod.rs
nathanielford: __Now__ it's fixed. Now.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+            "shuffleAddressList".to_string(),
+            serde_json::Value::Bool(true),
+        );
+        map.insert(
+            "pick_first".to_string(),
+            serde_json::Value::Object(pf_config),
+        );
+        ParsedJsonLbConfig::from_value(serde_json::Value::Array(vec![serde_json::Value::Object(
+            map,
+        )]))
```
dfawley: This seems out of place.  IIRC, the service config parsing shouldn't implement this.  Instead, the channel should use pick first if nothing (that exists) is specified.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,463 @@
...
+        if let Some(ref configs) = self.inner.load_balancing_config
+            && !configs.is_empty()
+        {
+            let vec: Vec<serde_json::Value> = configs
+                .iter()
+                .map(|c| {
+                    let mut map = serde_json::Map::new();
+                    map.insert(c.name.clone(), c.config.clone());
+                    serde_json::Value::Object(map)
+                })
```
dfawley: I believe we want this part to do a bit more -- and during parsing, not afterwards.  Here's what we do in Go per config entry:

https://github.com/grpc/grpc-go/blob/4c226daff88f54441d70f710815e07b81fb162b2/internal/serviceconfig/serviceconfig.go#L82-L120

- Lookup the balancer in the registry; skip entry if doesn't exist
- Call it to parse the config; error all service config parsing if it fails
- Collapse to a single selected policy and its parsed config, not a vec.

It's nice to do this here instead of in the channel, because we have LB policies that have similar "child" fields that will want to do this exact same logic.

grpc/src/client/service_config/mod.rs
nathanielford: Fixed!

grpc/src/client/service_config/mod.rs
nathanielford: I've added a TODO around this. After looking at it, I'd rather do the 'mechanical' and plumbing stuff in the followup or in a smaller PR and keep this more bare bones. LMK if you think that's ok or if we keep good context by putting it here. It seems straightforward enough but would add additional files to the surface of this PR.

dfawley: I don't mind if you want to put the implementation in another PR, but then some of these abstractions I think are just wrong in light of the above, and should be deleted if they aren't updated.

1. `ServiceConfig` shouldn't have a `pub fn lb_config()`.  `ServiceConfig` should be completely opaque externally (only parsed and passed around).

2. The `LoadBalancingConfigSerde` field should ideally (I think) parse a JSON array, but collapse it down to a single value: the first supported policy - it can also be empty if either the field is absent or if it contains no entries that are in the registry.

nathanielford: I think this should be addressed now. Let me know what you think!

grpc/src/client/service_config/mod.rs
```diff
@@ -55,29 +56,18 @@ impl ServiceConfig {
...
-                    let mut map = serde_json::Map::new();
-                    map.insert(c.name.clone(), c.config.clone());
-                    serde_json::Value::Object(map)
-                })
-                .collect();
+    pub(crate) fn lb_config(&self) -> Option<ParsedJsonLbConfig> {
+        if let Some(ref config) = self.inner.load_balancing_config {
+            let mut map = serde_json::Map::new();
+            map.insert(config.name.clone(), config.config.clone());
             return Some(ParsedJsonLbConfig::from_value(serde_json::Value::Array(
```
dfawley: This appears to be misusing the `ParsedJsonLbConfig` type.  It is supposed to be just a single lb policy's config in json.  It's what we pass to `LbPolicyBuilder::parse_config`:

https://github.com/grpc/grpc-rust/blob/0a67722863f90392c3b50fddf147c1dc9e33576e/grpc/src/client/load_balancing/mod.rs#L78-L81


I think this method should return `Option<(String, <a *parsed* LB config>)>` - the effective policy we're using and its parsed config.  Where "a parsed LB config" is:

https://github.com/grpc/grpc-rust/blob/0a67722863f90392c3b50fddf147c1dc9e33576e/grpc/src/client/load_balancing/mod.rs#L92

So maybe we want a type alias like `type LbConfig = Any + Send + Sync + Debug + 'static` so that it can be named here and there and match?

grpc/src/client/service_config/serde.rs
```diff
@@ -107,14 +112,32 @@ pub(crate) struct LoadBalancingConfigSerde {
...
+    {
+        use serde::ser::SerializeMap;
+        let mut map = serializer.serialize_map(Some(1))?;
+        map.serialize_entry(&self.name, &self.config)?;
+        map.end()
+    }
+}
+
+#[derive(Debug, Clone)]
+struct RawLbConfig {
```
dfawley: Should this get moved inside `mod load_balancing_config_serde`, even?  It's only used from there.

grpc/src/client/service_config/serde.rs
```diff
@@ -107,14 +112,32 @@ pub(crate) struct LoadBalancingConfigSerde {
     pub(crate) config: serde_json::Value,
 }
 
-impl<'de> Deserialize<'de> for LoadBalancingConfigSerde {
+impl Serialize for LoadBalancingConfigSerde {
```
dfawley: Must we implement `Serialize`?  I don't think we ever produce these?

grpc/src/client/service_config/serde.rs
```diff
@@ -31,7 +31,12 @@ use super::duration::GrpcDuration;
 #[serde(rename_all = "camelCase")]
 pub(crate) struct ServiceConfigSerde {
     pub(crate) load_balancing_policy: Option<String>,
-    pub(crate) load_balancing_config: Option<Vec<LoadBalancingConfigSerde>>,
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        with = "load_balancing_config_serde"
```
dfawley: Is it not possible to implement Deserialize for `LoadBalancingConfigSerde` to avoid the need for this `with`?

grpc/src/client/service_config/serde.rs
```diff
@@ -139,15 +162,79 @@ impl<'de> Deserialize<'de> for LoadBalancingConfigSerde {
...
+            fn visit_seq<S>(self, mut seq: S) -> Result<Self::Value, S::Error>
+            where
+                S: serde::de::SeqAccess<'de>,
+            {
+                let mut selected = None;
+                while let Some(candidate) = seq.next_element::<RawLbConfig>()? {
+                    if selected.is_none()
+                        && GLOBAL_LB_REGISTRY.get_policy(&candidate.name).is_some()
+                    {
+                        selected = Some(LoadBalancingConfigSerde {
```
dfawley: Can we not just `return` this value here as soon as we find it?  Or do we have to consume the whole seq or something?

grpc/src/client/service_config/serde.rs
```diff
@@ -139,15 +162,79 @@ impl<'de> Deserialize<'de> for LoadBalancingConfigSerde {
...
+                            name: candidate.name,
+                            config: candidate.config,
```
dfawley: We need to parse the config, though, and error if it fails.

grpc/src/client/service_config/serde.rs
nathanielford: We can't using this visitor pattern, or at least I couldn't figure out a way how and it didn't seem worth digging more - if something is selected we just zip through the rest of the elements in the sequence without doing anything. (`if selected.is_none()`). I figure this is probably fast enough?

grpc/src/client/service_config/serde.rs
nathanielford: Removed!

grpc/src/client/service_config/serde.rs
nathanielford: Yup! Done.

grpc/src/client/service_config/serde.rs
nathanielford: I can't do that directly because the incoming value is an (possible) array. I did put a deserialize implementation over that below, though it's not packaged the same. Not super happy with it, but it does remove the annotation.

grpc/src/client/service_config/serde.rs
nathanielford: This parsing is done. I had to touch files external to the ServiceConfig, unfortunately. LMK if that resolves things.

grpc/src/client/service_config/mod.rs
nathanielford: I think I have these types straightened out, but take a look.

grpc/src/client/service_config/duration.rs
```diff
@@ -0,0 +1,226 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+// TODO(nathanielford) Move this to the internal crate once we have one.
+
```
dfawley: You can remove this. I don't think we're going to end up with a separate crate for internal APIs as it seems to make versioning a mess.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -42,10 +39,20 @@ use crate::client::load_balancing::child_manager::ChildUpdate;
 use crate::client::name_resolution::ResolverUpdate;
 use crate::rt::GrpcRuntime;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct GracefulSwitchLbConfig {
-    child_builder: Arc<DynLbPolicyBuilder>,
-    child_config: Option<DynLbConfig>,
+    pub child_builder: Arc<DynLbPolicyBuilder>,
+    pub child_config: Option<DynLbConfig>,
```
dfawley: It seems these changes are not necessary, as the fields are not ever accessed directly anyway?

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -57,8 +64,10 @@ pub struct GracefulSwitchLbConfig {
 /// to active and tear down the previously active policy.
 #[derive(Debug)]
 pub struct GracefulSwitchPolicy {
-    child_manager: ChildManager<()>, // Child ID empty - only the name of the child LB policy matters.
-    last_update: Option<LbState>, // Saves the last output LbState to determine if an update is needed.
+    child_manager: ChildManager<()>, /* Child ID empty - only the name of the child LB policy
+                                      * matters. */
+    last_update: Option<LbState>, /* Saves the last output LbState to determine if an update is
+                                   * needed. */
```
dfawley: Please revert these changes to keep this PR as small as possible.

grpc/src/client/service_config/duration.rs
```diff
@@ -0,0 +1,226 @@
...
+use std::time::Duration;
+
+use serde::Deserialize;
+use serde::Serialize;
+use serde::Serializer;
+
+// Wraps std::time::Duration to provide custom serialization and deserialization
```
dfawley, line 29: Same comment as before: I don't think we need/want serialization; please remove support for serialization.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+use crate::client::load_balancing::ParsedJsonLbConfig;
+use crate::client::load_balancing::pick_first;
+
+pub type ParseResult = Result<ServiceConfig, String>;
+
+/// An in-memory representation of a service config, provided to gRPC as a JSON
+/// object.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ServiceConfig {
+    pub(crate) inner: serde::ServiceConfigSerde,
```
dfawley: Remove "pub(crate)"?

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+}
+
+impl ServiceConfig {
+    /// Parses a service configuration from a JSON string.
+    pub fn parse(config_json: &str) -> ParseResult {
```
dfawley: I don't think we want this to be pub, at least not for now.  The resolver should call the `ChannelController::parse_service_config` to parse a config.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+        let config_serde: serde::ServiceConfigSerde = match serde_json::from_str(config_json) {
+            Ok(c) => c,
+            Err(e) => {
+                return Err(format!("failed to deserialize service config JSON: {e}"));
+            }
+        };
```
dfawley: This might be more readable if it works?

```suggestion
        let config_serde: serde::ServiceConfigSerde =
            serde_json::from_str(config_json)
                .map_err(|e| format!("failed to deserialize service config JSON: {e}")?;
```

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+/// object.
+#[derive(Debug, Clone, PartialEq)]
+pub struct ServiceConfig {
+    pub(crate) inner: serde::ServiceConfigSerde,
+}
+
+impl ServiceConfig {
+    /// Parses a service configuration from a JSON string.
+    pub fn parse(config_json: &str) -> ParseResult {
+        let config_serde: serde::ServiceConfigSerde = match serde_json::from_str(config_json) {
```
dfawley: The `serde::` here makes it look the same as our external dependency of the same name.  I think it would help to `use crate::....::ServiceConfigSerde` instead.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+            inner: config_serde,
+        })
+    }
+
+    /// Extracts the load balancing configuration per gRPC specification rules:
+    /// 1. First supported entry in `loadBalancingConfig`
+    /// 2. Supported policy in `loadBalancingPolicy`
+    /// 3. Default `pick_first` policy
+    pub(crate) fn lb_config(&self) -> (Arc<DynLbPolicyBuilder>, Option<DynLbConfig>) {
+        if let Some(config) = self.inner.load_balancing_config.as_ref()
```
dfawley: When we store the config, can we also store the `Arc<Builder>` next to it?  I don't like looking it up again in the global registry here, as it could have changed or have different behavior when/if we add local registries for the channel.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+        }
+
+        Self::default_lb_policy()
+    }
+
+    /// Returns the default load balancing policy (`pick_first`).
+    pub(crate) fn default_lb_policy() -> (Arc<DynLbPolicyBuilder>, Option<DynLbConfig>) {
+        let builder = GLOBAL_LB_REGISTRY
+            .get_policy(pick_first::POLICY_NAME)
+            .expect("pick_first policy must be registered");
```
dfawley, line 81: Should this just get it at `pick_first::PickFirstBuilder` instead?  We'd probably need to make `DynAdapter` `pub(crate)` but that should be OK.

grpc/src/client/service_config/mod.rs
```diff
@@ -0,0 +1,479 @@
...
+        let default_json = ParsedJsonLbConfig::from_value(serde_json::json!({
+            "shuffleAddressList": true
```
dfawley: This should not be the default behavior.  It should be fine to parse an empty config and get the proper default.

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,662 @@
...
+            && match (&self.config, &other.config) {
+                (Some(a), Some(b)) => Arc::ptr_eq(a, b),
+                (None, None) => true,
+                _ => false,
+            }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Default)]
+pub(crate) struct SelectedLbConfig(pub(crate) Option<LoadBalancingConfigSerde>);
```
dfawley: Nit:

`LbConfigArray` would better describe the field perhaps?

dfawley: Given the oddity of this type, I think it would be better to keep the inner field private and make an accessor for it instead.  E.g.: `pub(crate) fn selected_policy(&self) -> (Arc<dyn DynLbPolicyBuilder>, Arc<dyn LbPolicyConfig>)`

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,662 @@
...
+        SelectedLbConfig(Some(cfg))
+    }
+}
+
+impl<'de> Deserialize<'de> for SelectedLbConfig {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        struct Visitor;
```
dfawley: I still think you should not need this visitor pattern, which would help simplify if so.  You should be able to deserialize directly into:

`Vec<HashMap<String, serde_json::Value>>` and then loop through the resulting vector.  Like what I was doing in the graceful switch policy before.

For u32/f32 it made sense, because those could be either JSON numbers _or_ strings.  For this field there is only one valid type.

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,662 @@
...
+    #[serde(default = "default_max_connections_per_subchannel")]
+    pub(crate) max_connections_per_subchannel: SerdeU32,
+}
+
+fn default_max_connections_per_subchannel() -> SerdeU32 {
+    SerdeU32(10)
+}
+
+#[derive(Debug, Clone)]
+pub(crate) struct LoadBalancingConfigSerde {
```
dfawley: This type doesn't appear to be "the Serde version of some other type so that we can trivially parse into this and then convert"

 Maybe *this* should be `SelectedLbConfig` and contain `(Arc<dyn LbPolicyBuilder>, Option<Arc<dyn DynLbConfig>>)`?

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,662 @@
...
+}
+
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct RetryPolicySerde {
+    pub(crate) max_attempts: SerdeU32,
+    pub(crate) initial_backoff: GrpcDuration,
+    pub(crate) max_backoff: GrpcDuration,
+    pub(crate) backoff_multiplier: SerdeF32,
+    pub(crate) retryable_status_codes: Vec<String>,
```
dfawley: We aren't supporting retry yet, so I think it would be best to remove all of the related code from this PR to keep it smaller and allow it to be finished sooner.

When we do support it, we'll need something that can give us `Vec<StatusCodeError>`, not raw strings that can be either code numbers or strings and have to be interpreted separately.

grpc/src/client/load_balancing/graceful_switch.rs
```diff
@@ -331,7 +310,8 @@ mod test {
         subchannel_list: TestSubchannelList,
     }
 
-    // Defines the functions resolver_update and subchannel_update to test graceful switch
+    // Defines the functions resolver_update and subchannel_update to test graceful
+    // switch
```
dfawley: It would be good to not change these in this PR, but if you do, then they should be wrapped at 80 columns.  With `graceful` still there it's 84.

grpc/src/client/service_config/serde.rs
nathanielford: I've updated the types on this to better reflect the situation.

grpc/src/client/service_config/serde.rs
nathanielford: I just made this private, since the `ServiceConfig.lb_config()` method effectively serves as the accessor.

nathanielford: It will only ever return a single object, so I changed it to `LbConfigSerde`. LMK if that makes sense in the new type regime.

grpc/src/client/service_config/mod.rs
nathanielford: I've updated the internal struct to store the builder and parsed config side by side so we can just surface it.

grpc/src/client/service_config/mod.rs
nathanielford: I do have a question here: what is the expected behavior if someone re-registered the pick first builder?

grpc/src/client/service_config/serde.rs
```diff
@@ -0,0 +1,489 @@
+/*
+ *
+ * Copyright 2025 gRPC authors.
```
dfawley: 2026

grpc/src/client/service_config/mod.rs
dfawley: This is fine - we'll get it from the registry.

dfawley:
As discussed offline, we should fail if we get through a (non-empty) list of lb configs and find no registered policy.

grpc/src/client/service_config/serde.rs
nathanielford: Fixed!

grpc/src/client/service_config/duration.rs
```diff
@@ -0,0 +1,145 @@
...
+    type Target = Duration;
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+/// Service Config uses the [protobuf Duration format](https://protobuf.dev/reference/protobuf/google.protobuf/#duration)
+/// when parsing. The format is a string with a number followed
+/// by an optional fraction and the letter 's' for seconds. For example, "1s",
+/// "1.5s", "0.000000001s" are valid durations.
```
dfawley: Nit: rewrap this comment please.

grpc/src/client/service_config/serde_bindings.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub(crate) fn is_none(&self) -> bool {
+        self.0.is_none()
+    }
+
+    pub(crate) fn is_some(&self) -> bool {
+        self.0.is_some()
+    }
+}
+
+impl PartialEq for LbConfigSerde {
```
dfawley: Why do we need equality for lb configs?  Is it just for testing?  The `Arc::ptr_eq()` on the configs below is a little dodgy for this to be a global equality check is why I'm asking.  If we need this for anything other than testing then we'd want to put an Eq requirement on the config, I'd say.  And otherwise, maybe we can move this to testing code and not a global `PartialEq` implementation, to avoid any future confusion.

grpc/src/inmemory/mod.rs
```diff
@@ -483,13 +482,13 @@ impl Resolver for InMemoryResolver {
             })
             .collect();
 
+        let service_config = channel_controller
+            .parse_service_config(r#"{"loadBalancingConfig": [{"round_robin": {}}]}"#)
+            .map(Some);
```
dfawley, line 487: Hmm, I'm pretty sure the ResolverUpdate type actually wants to be:

```rust
pub service_config: Option<ParseResult>
```

But I'm not sure if it makes sense to try to change that in this PR or wait until later.

dfawley (commented):
LGTM except for the mild concern about the `PartialEq` impl!

grpc/src/inmemory/mod.rs
nathanielford: Probably wait until later? I don't really want to mess with the resolver update signature on this pass.

grpc/src/client/service_config/serde_bindings.rs
nathanielford: This is an artifact of the visitor thing I think. Regardless, it's not needed here, nor do the other fields need it so I've removed it. We can add it back if necessary.

dfawley (approved):

PR #2732, opened 2026-07-14
grpc/xds: add test utility crate

gu0keno0:
## Motivation
Add a xds-test-util crate that contains fake control plane implementation and a few xds config helpers to enable writing e2e tests for xds implementations.

Also, used the new crate to add a few tests to tonic-xds. We can definitely add more in the future.

## Solution
New test utility library and new tests added.


gu0keno0:
@YutaoMa @ejona86 @dfawley this is the test util crate that I mentioned.

xds-test-util/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "xds-test-util"
+version = "0.1.0"
```
ejona86: If we never plan to publish this, then we can drop the version instead of using `publish = false`?

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+#[derive(Debug)]
+struct Inner {
+    state: Mutex<State>,
+    next_stream_id: AtomicU64,
+}
+
+impl Inner {
+    /// Handles a single inbound `DiscoveryRequest` for a stream.
+    fn handle_request(&self, stream_id: u64, tx: &ResponseSender, req: DiscoveryRequest) {
+        let mut state = self.state.lock().expect("control plane state poisoned");
```
ejona86: nit: s/poisoned/not poisoned/. You expect the state to not be poisoned.

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+            );
+            return;
+        }
+
+        let type_url: AdsTypeUrl = match req.type_url.parse() {
+            Ok(url) => url,
+            Err(_) => {
+                tracing::error!(
+                    "Received request with empty or invalid type_url for stream {}",
+                    stream_id
```
ejona86: Not that it really matters, but I'm surprised the `req.type_url` isn't in the message. That seems more useful than the stream_id.

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+                );
+                return;
+            }
+        };
+
+        // Nonce check: if the request carries a response nonce, it must match
+        // the last nonce we sent on this (type, stream); otherwise ignore it.
```
ejona86: The "must" makes it seem this is a failure of the client or something. Say "otherwise it is out-of-date so ignore it"? Or reword something like: "if the request's nonce doesn't match what we last sent, then the request is out-of-date; ignore it and wait for the future ack."

(If the response_nonce is empty, I think we should really be comparing it as well, but I think that the Java code was lazy for the initial request. Which is fine; the code is trying to be as lazy as possible. It just means the comment needs to be careful.)

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+            }),
+        }
+    }
+
+    /// Sets the full set of resources for `type_url`, replacing any previous
+    /// resources of that type, and pushes an update to every current subscriber.
+    ///
+    /// Each resource is packed into a protobuf `Any` tagged with `type_url`.
+    /// Resource types are treated as state-of-the-world.
+    pub fn set_xds_config<M: Message>(&self, type_url: &AdsTypeUrl, resources: HashMap<String, M>) {
```
ejona86: nit: it would be best to have this call another function as soon as `packed` is computed in order to drop the generics ASAP.

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+        let packed: HashMap<String, Any> = resources
+            .into_iter()
+            .map(|(name, msg)| {
+                (
+                    name,
+                    Any {
+                        type_url: type_url.to_string(),
```
ejona86:
```suggestion
                        type_url: msg.type_url(),
```

It's possible for different entries to have different types, and for the type to not match the AdsTypeUrl. In particular, with `envoy.service.discovery.v3.Resource`. Also, if they don't match, it will be _much_ better to get an unexpected type url than to decode the protobuf bytes as the wrong type.

YutaoMa (approved):

xds-test-util/src/control_plane.rs
```diff
@@ -0,0 +1,506 @@
...
+            .state
+            .lock()
+            .expect("control plane state poisoned");
+        state.resources.insert(type_url.clone(), packed);
+
+        // getAndIncrement: the pushed response carries the pre-increment version.
+        let version = {
+            let counter = state.versions.entry(type_url.clone()).or_insert(1);
+            let current = *counter;
+            *counter += 1;
```
YutaoMa: This test server currently returns version `1` and empty resources list if a client subscribes before any `set_xds_config` call. And then after the first `set_xds_config` it sends version `1` again before of read-then-increment. Currently not an issue with `xds-client` implementation because it doesn't dedup on version number, but some other clients may be stricter. Worth changing into increment-then-read, and can also set the initial state to `0` for clarity.

YutaoMa:
Added the nit comment above on version semantics, otherwise lgtm.

ejona86 (approved):
We should change the type url thing, but otherwise looks good.

xds-test-util/src/control_plane.rs
gu0keno0: Thanks. This issue was what I worried about at the beginning. Now that you've caught it for real, I refactored the  internal state and added a VersionedResourceBundle type, whose version is updated whenever its content is updated. This way, the version of the resources is managed consistently everywhere.

xds-test-util/Cargo.toml
gu0keno0: Done.

xds-test-util/src/control_plane.rs
gu0keno0: Done.

xds-test-util/src/control_plane.rs
gu0keno0: True, I updated a few other places as well.

xds-test-util/src/control_plane.rs
gu0keno0: Reworded the comment; now it's just re-stating the xDS doc directly and added a ref to the original xDS doc.

xds-test-util/src/control_plane.rs
gu0keno0: Good point. I've updated it to use prost::Name and type_url associative method.

ejona86 (approved):

PR #2738, opened 2026-07-17
xds-client/retry: add jitter to connection backoff

W4lspirit:
## Motivation
The reconnect backoff was fully deterministic: every client that lost its xDS stream at the same time (control-plane restart or a network blip) computed the identical delay schedule and retried in lockstep. That synchronized retry storm hammers the management server right when it's least able to absorb it.

## Solution
Randomize each delay by a configurable factor per gRFC A6 so clients spread their reconnects across a window instead of colliding. Default jitter is 0.2 (±20%) ; `with_jitter` validates the factor is in [0.0, 1.0], and 0.0 restores the old deterministic behavior.

## Random
I hesitated between `rand` crate and `fastrand`, in the end I fallback to the same dep grpc is already using for the jitter in [named_resolution](https://github.com/grpc/grpc-rust/blob/master/grpc/src/client/name_resolution/backoff.rs#L27)

### DIFF
--- a/xds-client/src/client/retry.rs
+++ b/xds-client/src/client/retry.rs
@@ -27,7 +27,7 @@ pub struct RetryPolicy {
     /// Initial backoff duration for the first retry attempt.
     ///
     /// Default: 1 second.
-    pub initial_backoff: Duration,
+    initial_backoff: Duration,
 
     /// Maximum backoff duration.
     ///
@@ -35,7 +35,7 @@ pub struct RetryPolicy {
     /// retry attempts have been made.
     ///
     /// Default: 30 seconds.
-    pub max_backoff: Duration,
+    max_backoff: Duration,
 
     /// Multiplier for exponential backoff.
     ///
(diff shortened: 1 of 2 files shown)


xds-client/src/client/retry.rs
```diff
@@ -259,9 +287,10 @@ impl Backoff {
     ///
     /// Returns `None` if `max_attempts` is set and has been exceeded.
     pub fn next_backoff(&mut self) -> Option<Duration> {
-        let duration = self.policy.backoff_duration(self.attempt)?;
+        let base = self.policy.backoff_duration(self.attempt)?;
         self.attempt += 1;
-        Some(duration)
+        let factor = 1.0 + self.policy.jitter * rand::rng().random_range(-1.0..1.0);
+        Some(base.mul_f64(factor))
```
YutaoMa, line 293: negative value would cause panic here, and because currently `RetryPolicy` fields are public, the `[0.0, 1.0]` range isn't a type level invariant. `grpc-go` clamps backoff to be non-negative, but I'm good with privatize `RetryPolicy` and enforce invariant there if you want.

YutaoMa:
Thanks for the PR! All logic and tests looks great to me, my only comment is regarding the negative delay issue above. Feel free to address it and ping me again for approval.

xds-client/src/client/retry.rs
W4lspirit: @YutaoMa Addressed in https://github.com/grpc/grpc-rust/pull/2738/commits/eeca7f2decc6d2f76bae9e0b83597172dd30a7c6 (all fields are private)

Let me know what you think .

YutaoMa (approved):

PR #2746, opened 2026-07-22
tonic-xds/xds-client: client permanently stuck on stale endpoints after backend pod rotation — no keepalives or connect timeouts anywhere in the xDS path

aryehlev:
The client never detects dead connections. The ADS channel is built with no keepalives and the worker just calls recv(), so a half-open connection after an xDS server restart pends forever and the client keeps serving stale endpoints. Endpoint channels are connect_lazy() with no timeout, and lazy channels always look ready to the balancer — so requests picked onto deleted pod IPs (which Kubernetes black-holes, no RST) hang until the caller's deadline.

Net effect: a rolling restart of the backend is a permanent outage. Requests hang on dead IPs, and the EDS update that would fix it never arrives because the ADS stream is also dead.

This PR adds connect timeouts and HTTP/2 keepalives to both the ADS channel and per-endpoint channels. Defaults: 10s connect, 30s ping interval, 10s ack timeout. Configurable on the transport builder.

Scoped to keepalive/timeout settings only. The creds issue will be a separate PR. The unused connectivity-aware LB in client/loadbalance/ is also untouched — fail-fast on dead endpoints is enough to recover from pod churn.

### DIFF
--- a/tonic-xds/src/xds/cluster_discovery.rs
+++ b/tonic-xds/src/xds/cluster_discovery.rs
@@ -14,6 +14,7 @@
 //! connector is kept and a warning is logged.
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use arc_swap::ArcSwap;
 use tokio::sync::mpsc;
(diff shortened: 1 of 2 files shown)


YutaoMa:
the reasoning looks legit to me, the same keep-alive configs were set up in LinkedIn's internal tonic-xds implementation after we open sourced the previous state of code, thus was missed in this tonic-xds crate. Approving and could you help update the issue and PR description to explain this PR is limited to the keepalive settings and the creds issue will be in another PR (also looks legit to me, feel free to open that follow-up), and I'll help merge the PR after that.

YutaoMa (approved):

aryehlev:
@YutaoMa thank you so much!!

PR #2747, opened 2026-07-22
feat(tonic-xds): pre-route header-mutation interceptor

YutaoMa:
## Motivation

`tonic-xds` currently offers no way to customize routing per request: a request is matched to a route exactly as received, with no hook to inspect or transform it first. The xDS mechanism for this is the HTTP filter framework (gRFC A39) — filters running as interceptors around routing/LB, configured via `http_filters` / `typed_per_filter_config` — but that framework is not implemented in `tonic-xds` today, and isn't currently planned.

Some use cases only need a small slice of that capability: transforming a request (e.g. rewriting or injecting a header) **before** route selection, so the standard route table can then match on the result. With no extension point at all, these are impossible to express against `tonic-xds`.

## Solution

Add a minimal, opt-in **pre-route interceptor** hook that unblocks header-driven route customization without committing to the full filter framework:

- **`PreRouteInterceptor`** trait + **`XdsChannelBuilder::with_pre_route_interceptor(...)`** — runs before route selection and may mutate request headers.
- `RouteConfigMetadata` now parsed, passed into the interceptor as a customization point.

To keep the interceptor and router consistent, the routing stack is restructured:

- A `RouteConfigSelector` layer is extracted out of the routing layer, to own the async config swaps. This is done so the config snapshot can be inserted into request extension and reused by both the interceptor layer and the routing layer, avoid double and potentially inconsistent subscriptions.

When no interceptor is registered, behavior is unchanged. This is deliberately minimal; a future A39-style filter framework can build on the same watcher/selector/metadata pieces.

## Testing

- Unit tests for the watcher / selector / pure-matcher split.
- End-to-end test: an interceptor injects a header derived from route-config metadata, and the router selects the matching cluster.

### DIFF
--- a/tonic-xds/src/client/route.rs
+++ b/tonic-xds/src/client/route.rs
@@ -369,11 +369,10 @@ mod tests {
         // Config carries filter_metadata that the interceptor is expected to see.
         let mut filter_metadata = std::collections::HashMap::new();
         filter_metadata.insert("partitioning".to_string(), Struct::default());
-        let metadata = Metadata {
+        let metadata = RouteConfigMetadata::from_proto(Metadata {
             filter_metadata,
             ..Default::default()
-        }
-        .into();
+        });
 
         let rc = Arc::new(RouteConfigResource {
             name: "rc".into(),
--- a/tonic-xds/src/client/route.rs
+++ b/tonic-xds/src/client/route.rs
@@ -1,5 +1,6 @@
 use crate::common::async_util::BoxFuture;
-use crate::xds::routing::RoutingError;
+use crate::xds::resource::route_config::{RouteConfigMetadata, RouteConfigResource};
+use crate::xds::routing::{RouteConfigWatcher, RoutingError};
 use http::Request;
 use std::sync::Arc;
 use std::task::{Context, Poll};
@@ -12,6 +13,12 @@ pub(crate) struct RouteInput<'a> {
     pub authority: &'a str,
     /// The HTTP headers of the request. These can be used for header-based routing decisions.
     pub headers: &'a http::HeaderMap,
+    /// Route configuration already bound to this request by an upstream
+    /// [`RouteConfigSelectorService`]. When `Some`, the router matches against
+    /// it directly (the single-pass path shared with a pre-route interceptor);
+    /// when `None`, routing fails with `RoutingError::NotReady` (the selector
+    /// layer always sets it in the assembled channel stack).
+    pub config: Option<&'a RouteConfigResource>,
 }
 
 /// Represents the routing decision made by the routing layer.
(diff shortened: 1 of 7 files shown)


tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -13,11 +14,60 @@ use xds_client::{Error, Resource};
...
+    /// Returns `true` when no `filter_metadata` namespaces are present.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.filter_metadata.is_empty()
+    }
+
+    /// Builds the view from an xDS `Metadata`, pre-encoding each namespace's
+    /// `Struct` to bytes.
+    pub(crate) fn from_proto(metadata: Metadata) -> Self {
+        let filter_metadata = metadata
```
gu0keno0, line 73: Is it possible to use typed_filter_metadata?

tonic-xds/src/client/route.rs
```diff
@@ -112,6 +150,125 @@ impl<S> Layer<S> for XdsRoutingLayer {
     }
 }
 
+/// Tower service that snapshots the active route configuration once and binds it
```
gu0keno0: We will be adding two more layers (and their associated BoxFutures) into the stack, and I'm thinking, does this have to be implemented as 2 tower services or shall we just add it as a new capability into the routing layer?

YutaoMa: It could, I only kept the interceptor part out for better modularity. If we think the associated perf and complexity cost is not worth it, I'm also good with adding the hook in-layer.

tonic-xds/src/xds/resource/route_config.rs
YutaoMa: good point, I'll expose both fields in the metadata config

tonic-xds/src/client/route.rs
gu0keno0: Will be great if we can simplify it and use less layers. I just feel that the type complexity, compile time complexity and the runtime memory allocations overhead will gradually add up if we start doing the "one feature one layer" pattern. And therefore, we might also want to keep the stack simple in the future.

YutaoMa: 👍 , changed to in-layer design

YutaoMa:
Rebased on latest

tonic-xds/src/client/route.rs
```diff
@@ -177,4 +215,120 @@ mod tests {
...
+        struct PartitionInterceptor;
+        impl PreRouteInterceptor for PartitionInterceptor {
+            fn on_request(&self, headers: &mut http::HeaderMap, metadata: &RouteConfigMetadata) {
+                assert!(
+                    metadata.filter_metadata("partitioning").is_some(),
+                    "interceptor must see the untyped metadata",
+                );
+                let typed = metadata
+                    .typed_filter_metadata("partitioning-typed")
+                    .expect("interceptor must see the typed metadata");
```
gu0keno0, line 294: Nit: same format for assertions

gu0keno0 (approved):

PR #2748, opened 2026-07-22
LICENSE and check for grpc-rust.

nathanielford:
## Motivation

Ensure that licenses are being added and not improperly modified to all code files, in compliance with CNCF requirements.

## Solution

This is a python script that runs in the CI and can be configured to run in a pre-commit. It also adds a LICENSE file to the grpc directory to conform with the MIT license being used throughout that part of the code-base. This script only checks the grpc directory at this time.


dfawley (commented):
I think we could simplify this with a few minor changes.
 
- We don't need to worry about files with apache licenses at all.  We should just be enforcing that our source files have the MIT license.  The Apache licensed code that is grpc-authored is all the protos that we just make copies of here.

- The ALLOWED_AUTHORS thing can be removed and we can assert that the copyright directly states "gRPC authors".

- I don't think we need a separate LICENSE in the grpc directory.  We should update the LICENSE in the root, and also have @LucioFranco update all the copyright notices to have the same author ("gRPC authors") - or send him a PR that does this for him to approve.  [EDIT: by "all" I mean the ones that are currently "(c) Lucio Franco"]

- The `.pre-commit-config.yaml` and surrounding tooling feels like overkill to me at this point.

LucioFranco:
If someone can send me a PR that would be best otherwise i think what @dfawley said makes sense.

nathanielford:
I'll send you a PR @LucioFranco  - I didn't want to overstep, but if you're good switching the licenses at this juncture I'll go ahead and do that for the entire repo. I won't merge this PR until that one is in.

Should I move the tools/ directory then to the top level and have this check the entire repo for the MIT/gRPC authors license or leave it across the grpc/ directory only? Also note this doesn't cover protos really, but should be adaptable to do so if we need it.

dfawley:
> Should I move the tools/ directory then to the top level

SGTM.

> and have this check the entire repo for the MIT/gRPC authors license or leave it across the grpc/ directory only?

If it's not too painful, that would be great.  If there are a ton of problems then we can expand the scope over time.

nathanielford:
It's probably easy enough to put this across the entire repo. I was being conservative in scope just because I didn't want to overstep.

One last question: should I also try and cover proto files? These are largely under Apache licenses in the larger repo. Ignoring third_party and special cases (rls) is easy, but I'm not wholly sure what to do with the protos, some of which date back to 2015 or so and are licensed to Google LLC. For instance: examples/proto/googleapis/google/api/http.proto

dfawley:
We don't need an automated checker for those.  However we want to ignore them should be fine.  The main thing we need the checker for is to ensure that new files get the proper license clause in them.  AFAIK we don't intend to author new proto files in this repo so if you were wanting to exclude anything ending in `.proto`, that should be fine.

nathanielford:
Ok, I think this is ready to go.

tools/check_license.py
```diff
@@ -0,0 +1,200 @@
+#!/usr/bin/env python3
+"""
+Checks source files (*.rs, *.proto) for exact match against MIT or Apache 2.0 license boilerplate.
+Performs piecewise validation and unified diffing against expected boilerplate templates.
```
dfawley: Nit: still mentions apache

tools/check_license.py
```diff
@@ -0,0 +1,200 @@
...
+import sys
+from pathlib import Path
+
+IGNORE_DIRS = {"target", ".git", ".gemini", "generated", "node_modules"}
+IGNORE_FILES = {"src/credentials/rustls/key_log.rs"}  # Third-party Apache 2.0 file
+TARGET_EXTENSIONS = (".rs",)
+
+MIN_YEAR = 2025
+
+# Regex to capture: (prefix) Copyright [(c)] (year_spec)
```
dfawley: Why is there an optional `(c)`?  Are we inconsistent?

tools/check_license.py
```diff
@@ -0,0 +1,200 @@
...
+            if line.strip() == "*/":
+                break
+        return header, None
+    elif first_line.startswith("//"):
+        for line in lines:
+            if line.strip().startswith("//"):
+                header.append(line.rstrip())
+            else:
+                break
+        return header, None
```
dfawley: Can this bit be removed now?  Or do we support `//`?  Also: related things mentioning `//` above/below.

tools/check_license.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return (
+        None,
+        f"Expected comment header block (/* or //) on line 1, found: {repr(first_line)}",
+    )
+
+
+def validate_copyright_years(year_spec: str) -> tuple[bool, str | None]:
+    current_year = datetime.date.today().year
+    years = [int(y) for y in re.findall(r"\b\d{4}\b", year_spec)]
```
dfawley: I think this is more complex than it needs to be since we removed the range support?

dfawley (commented):
Thanks for jumping on this!  Just a couple more minor nits.

tools/check_license.py
nathanielford: Removed extraneous stuff!

tools/check_license.py
nathanielford: Removed it. It was only there because of key_log.rs, really. Standardized on blockstar quotes.

tools/check_license.py
nathanielford: The root level LICENSE had it, but nothing else did. We also don't need it, legally, so have standardized everything to not have it.

tools/check_license.py
nathanielford: Fixed!

nathanielford:
All comments have been handled. I had to make some additional changes to make clippy happy with the key_log.rs file.

grpc-benchmark/proto/grpc/testing/messages.proto
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015-2016 gRPC authors.
+// Copyright 2015 gRPC authors.
```
dfawley: We should not modify any existing licenses.  This was copied as-is from the grpc-proto repo, and should remain intact.  IIUC we're skipping proto files anyway, so this should be unnecessary.  Also below.

grpc/src/credentials/rustls/key_log.rs
```diff
@@ -37,7 +46,7 @@ impl KeyLogFileInner {
         let file = match OpenOptions::new().append(true).create(true).open(path) {
             Ok(f) => Some(f),
             Err(e) => {
-                eprintln!("unable to create key log file {path:?}: {e}");
+                eprintln!("unable to create key log file {}: {e}", path.display());
```
dfawley: Could you please revert the changes here/below from this file?  This should be focused on copyrights only and I don't want to mix copyright changes in with other modifications (even if they are not significant).

tools/check_license.py
```diff
@@ -0,0 +1,197 @@
...
+    valid_copy, err = validate_copyright(header_lines)
+    if not valid_copy:
+        return False, err
+
+    norm_header = normalize_header(header_lines)
+    return check_diff(norm_header, filepath)
+
+
+def main():
+    targets = sys.argv[1:] if len(sys.argv) > 1 else ["."]
```
dfawley: If this only supports one argument (the path to check) then the below gets simpler (removes redundancy).  Just

```py
target = sys.argv[1] if len(sys.argv) > 1 else "."
files_to_check = []

for dirpath, dirnames, filenames in os.walk(target):
...
```

And optionally before the loop above:

```py
if !os.path.isdir(target): # error
```

nathanielford: It's meant to take a list of paths to check, or list of files to check, to allow a list of specific files (from, for example, the currently staged files).

grpc/src/credentials/rustls/key_log.rs
nathanielford: Sure! Clippy was complaining, so I changed it but you're right - should be a separate PR for that cleanup.

grpc-benchmark/proto/grpc/testing/messages.proto
nathanielford: Done!

grpc/src/credentials/rustls/key_log.rs
```diff
@@ -55,11 +64,11 @@ impl KeyLogFileInner {
 
         self.buf.truncate(0);
         write!(self.buf, "{label} ")?;
-        for b in client_random.iter() {
+        for b in client_random {
             write!(self.buf, "{b:02x}")?;
         }
         write!(self.buf, " ")?;
-        for b in secret.iter() {
+        for b in secret {
```
dfawley: Please revert these, too.

tools/check_license.py
dfawley: Are you planning on using it that way?  If not, then "KISS" is best.

tools/check_license.py
```diff
@@ -0,0 +1,197 @@
...
+                files_to_check.append(target)
+        elif os.path.isdir(target):
+            for dirpath, dirnames, filenames in os.walk(target):
+                dirnames[:] = [d for d in dirnames if d not in IGNORE_DIRS]
+                for f in filenames:
+                    if f.endswith(TARGET_EXTENSIONS):
+                        full_path = os.path.join(dirpath, f)
+                        if not should_ignore(full_path):
+                            files_to_check.append(full_path)
+
```
dfawley:
```suggestion
                            files_to_check.append(full_path)
        else:
            # Error?
```

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2025-2026 gRPC authors.
+Copyright 2025 gRPC authors.
```
dfawley: My best guess is this should be 2019 since that's when the repo was created and progress began.

grpc/src/credentials/rustls/key_log.rs
nathanielford: Done.

tools/check_license.py
nathanielford: My thinking in skipping this is that it's something particularly weird and it shouldn't halt the license check. Any such file system object being in the non-directory, non-file category should be the responsibility of another check. Rust/proto/other files with licenses should always succeed at `isfile()`. Hence no error - but I could easily be missing a case I should be thinking about!

LICENSE
nathanielford: Done!

tools/check_license.py
dfawley: This would be: user specified a command-line parameter that wasn't a file or a directory.  That's improper use of the tool, right?  Otherwise I believe if you run `check_license.py jklfdsllj` then it will report "SUCCESS".

tools/check_license.py
nathanielford: The use case I was imagining was checking your local un/staged commits for problems, particularly if CI reported something and you wanted to clean it up:

```
$ git diff --name-only | xargs python3 tools/check_license.py                                                1.91.1 nathanielford@nford  16:37:46
Checked 1 code files for exact license boilerplate match.
FAILED: Found 1 files with boilerplate discrepancies:

--- grpc/src/client/channel.rs ---
Boilerplate mismatch:
--- Expected MIT License (Block)
+++ grpc/src/client/channel.rs
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright <YEAR> gRPC authors.
+ * Copyright <YEAR> gRPC wumpus.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to
  ```
  
 But I'm happy to pull the loop if we want to hew to a spartan approach!

File: tools/check_license.py (hunk 9 above)
Comment (nathanielford) [reply tools/check_license.py]: That's true - it'll report success with 0 files checked (since `jklfdsllj` isn't a valid target). If the question this check is asking is, "Do all existing files that should have a boilerplate have the correct boilerplate?" that would be correct behavior. I think it gets hairy if we special case what happens if the user asks for a specific file which doesn't fall under the 'existing file that should have boilerplate' that doesn't conform to some part of that category ('exists'). I don't care deeply, though, if you think it benefits from the error here.

File: tools/check_license.py (hunk 7 above)
Comment (dfawley) [reply tools/check_license.py]: I'm not sure anyone would ever run it that way, since it needs to pass in CI by running on all files.  And it should run ~very quickly even doing a full scan, so I don't think there's much need for a targeted run-style.

File: tools/check_license.py [hunk 11]
```diff
@@ -0,0 +1,197 @@
...
+"""
+
+import datetime
+import difflib
+import os
+import re
+import sys
+from pathlib import Path
+
+IGNORE_DIRS = {"target", ".git", ".gemini", "generated", "node_modules"}
```
dfawley: I have some doubts about this:

- What if we add a `src` sub-directory named `target`?  This would skip it, right?  I guess we need to skip `target` at the top-level, and the top-level of crates.  So maybe....adjacent to `Cargo.toml` files or something?
- What's the significance of skipping the `.gemini` directory?
- What's `node_modules` here for?

nathanielford: All of these were included as standard areas where tools might be generating something we want to skip. That said, when I was thinking about this it was also as a precommit tool. 

Now... I think maybe we just get rid of this. The filesystem on the repo in github should have none of these anyway. `.gitignore` will prevent `target` as it stands now. I've seen `.gemini` pollute directories but I think modern versions are just putting scratch information in the home directory, so it's not a concern.

dfawley: It seems like supporting a local run is useful, just "whole repo at a time" seems to make the most sense there.  For that use case, pruning `target` directories is actually a big deal, since they can end up at hundreds of gigabytes sometimes.  Unless you want to use `git ls-files '*.rs'` (e.g.) in this script to list the files that should be checked?  That could work.

dfawley (approved):
Looks great, thank you!

PR #2749, opened 2026-07-23
License update

nathanielford:
## Motivation

Licenses need to be updated and standardized to 'gRPC authors'. 

## Solution

Search-replace in most cases, moving Lucio copyrights to gRPC authors. Update root license.


codegen/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "codegen"
-authors = ["Lucio Franco <luciofranco14@gmail.com>"]
+authors = ["gRPC authors"]
```
LucioFranco, line 3: is there an email we want to associate with it?

nathanielford: Added the grpc-io@googlegroups.com email, per @dfawley's suggestion. LMK if we need a different email address, though.

nathanielford:
I think this covers everything now - sorry for the explosion of files. This now also introduces the required license to all the files that don't have it, and adds the email everywhere.

nathanielford:
On further reflection based on internal guidance we decided to omit the email address. Let us know if you think that might be a problem for some reason we didn't consider. Presumably contact info should be held in the README or CONTRIBUTORS or the like.

LucioFranco (approved):

LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2025 Lucio Franco
+Copyright (c) 2025-2026 gRPC authors.
```
dfawley, line 1: I believe we refrain from using ranges in any of our copyright dates.  We leave it at the first copyright date.

dfawley (commented):
LGTM.  Maybe remove the range in the next PR?

PR #2750, opened 2026-07-24
refactor(tonic-xds): extract transport-agnostic RetryClassifier seam

LYZJU2019:
## Summary

Refactors the `tonic-xds` client retry layer to be **transport-agnostic** so non-gRPC transports (e.g. plain HTTP) can reuse the same retry engine without duplicating a retry state machine. This is a **non-breaking, standalone** change: all items stay crate-internal (`pub(crate)`) and the existing gRPC behavior is preserved.

## Why

First step of generalizing the xDS retry layer so it can serve both gRPC and HTTP transports. Keeping the gRPC-specific logic (retryable codes + `grpc-previous-rpc-attempts` header) behind the classifier keeps `tonic-xds` transport-agnostic while leaving gRPC the default.

## Testing

- `cargo build -p tonic-xds` — clean (no warnings)
- `cargo test -p tonic-xds` — 340 passed, 0 failed (incl. retry unit tests + `test_retry_once_on_unavailable`)
- `cargo clippy -p tonic-xds --all-targets` — no new warnings in changed files
- `cargo doc -p tonic-xds --no-deps` (with `RUSTDOCFLAGS=-D warnings`) — clean

Behavior is unchanged; the diff is behavior-preserving.

### DIFF
--- a/tonic-xds/src/client/retry.rs
+++ b/tonic-xds/src/client/retry.rs
@@ -22,7 +22,13 @@
  *
  */
 
-//! gRPC retry utilities.
+//! Transport-agnostic retry utilities.
+//!
+//! The retry *decision* state (attempt cap, backoff, body cloning) lives in the
+//! generic [`RetryPolicy`], while transport-specific decisions (which responses
+//! are retryable, and any per-retry request mutation) live behind the
+//! [`RetryClassifier`] seam. [`GrpcRetryClassifier`] is the default gRPC
+//! implementation; [`GrpcRetryPolicy`] is the gRPC policy alias.
 
 use std::fmt::Debug;
 use std::io;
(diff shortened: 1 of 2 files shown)


tonic-xds/src/client/retry.rs
```diff
@@ -219,45 +242,53 @@ fn make_backoff(config: &GrpcRetryBackoffConfig) -> backoff::ExponentialBackoff
...
+/// retry *state machine* stays entirely in tower's `Retry`/`ResponseFuture`;
+/// this type only implements the `Policy` trait, so no state machine is
+/// reimplemented here.
 #[derive(Clone, Debug)]
-pub(crate) struct GrpcRetryPolicy {
-    config: Arc<ArcSwap<GrpcRetryPolicyConfig>>,
+pub(crate) struct RetryPolicy<C> {
+    config: Arc<ArcSwap<RetryConfig>>,
+    /// Decides retryability and per-retry request mutation for the transport.
+    classifier: C,
```
YutaoMa, line 262: while review this PR I noticed both before and after this PR that the classifier part isn't updated according to xDS `retry_on` config. It is a bug needs fixing. I'm good with merging this refactoring for now and address that existing issue in a follow-up PR, but heads up that will likely involve a shape change here as well since `classifier` may need to be `Arc<ArcSwap>` as well.

YutaoMa (approved):

tonic-xds/src/client/retry.rs
LYZJU2019: Good catch, @YutaoMa. The `RetryLayer` is ready but retry config is not wired yet. This is a feature to be added later. I will keep this comment open until I add the control plane code.

PR #2775, opened 2026-07-28
feat(grpc-xds): gRFC A74 model validated xDS resources and XdsConfig snapshot

YutaoMa:
## Motivation

Ref: #2754

Per [gRFC A74](https://github.com/grpc/proposal/blob/master/A74-xds-config-tears.md), to avoid config tear across xDS layers, all watches (LDS, RDS, CDS, EDS) are moved into a single resolver behind an `XdsDependencyManager` that emits a single complete `XdsConfig`.

## Solution

This PR implements the `XdsConfig` modeling, along with its xDS resource types(the validated types corresponding to the Envoy proto types) and validation logic. This PR is a pre-req to the `XdsDependencyManager`.

The shape of `XdsConfig` closely mirrors gRFC A74's C++ sketch, as well as other gRPC implementations such as grpc-go. A few design choices were made for Rust ergonomics, such as using an index to store selected virtual host rather than a reference to avoid self-referential borrow.


YutaoMa:
Hi @ejona86 @dfawley looking for reviews on this `XdsConfig` PR. Thanks!

grpc-xds/src/xds_config.rs
```diff
@@ -0,0 +1,250 @@
...
+
+use crate::resource::{
+    ClusterResource, EndpointAddress, EndpointsResource, ListenerResource, RouteConfigResource,
+    RouteSource, VirtualHost,
+};
+
+/// The atomic xDS configuration snapshot for a channel.
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+pub(crate) struct XdsConfig {
```
dfawley, line 47: I am assuming this would, itself, end up in an `Arc`, so that it can be shared between the name resolver and all the LB policies?  If so do we need Arcs on the individual fields as well?

grpc-xds/src/resource/route.rs
```diff
@@ -0,0 +1,928 @@
...
+                if name.is_empty() {
+                    return Err(Error::Validation("weighted cluster name is empty".into()));
+                }
+                total_weight += u64::from(weight);
+                if total_weight > u64::from(u32::MAX) {
+                    return Err(Error::Validation(format!(
+                        "sum of weighted cluster weights exceeds {}",
+                        u32::MAX
+                    )));
+                }
```
dfawley, line 475: I think there should be a validation of the total_weight field here:


```suggestion
                }
                if let Some(tw) = wc.total_weight_opt() {
                    let expected_total = tw.value();
                    if expected_total == 0 || u64::from(expected_total) != total_weight {
                        return Err(Error::Validation(format!(
                            "sum of weighted cluster weights ({total_weight}) does not match explicit total_weight ({expected_total})"
                        )));
                    }
                }

```

per route_components.proto:

> ```
>  // Specifies the total weight across all clusters. The sum of all cluster weights must equal this
>  // value, which must be greater than 0. Defaults to 100.
>  google.protobuf.UInt32Value total_weight = 3 [(validate.rules).uint32 = {gte: 1}];
> ```

and A28:
> - Can be Weighted_clusters
>   - The sum of weights must add up to the total_weight.

grpc-xds/src/xds_config.rs
```diff
@@ -0,0 +1,250 @@
...
+        route_config: Arc<RouteConfigResource>,
+        virtual_host_index: usize,
+        clusters: HashMap<String, ClusterResult>,
+    ) -> Option<Self> {
+        route_config.virtual_hosts.get(virtual_host_index)?;
+        if let RouteSource::Inline(inline) = &listener.route_source
+            && !Arc::ptr_eq(inline, &route_config)
+        {
+            return None;
+        }
```
dfawley, line 85: Should we also validate that if route_source is RouteSource::Rds, that the rds name matches the route_config name?

grpc-xds/src/resource/listener.rs
```diff
@@ -0,0 +1,313 @@
...
+        let api_listener = message.api_listener();
+
+        if !api_listener.has_api_listener() {
+            return Err(Error::Validation(
+                "api_listener missing inner api_listener Any field".into(),
+            ));
+        }
+        let any: Any = api_listener.api_listener().to_owned();
+
+        let hcm = HttpConnectionManager::parse(any.value()).map_err(|e| {
```
dfawley, line 108: We should validate `any`'s `type_url` before parsing this to ensure the message type is what we expect.

grpc-xds/src/resource/route.rs
```diff
@@ -0,0 +1,928 @@
...
+    Ok(RouteMatch {
+        path_specifier,
+        headers,
+        case_sensitive,
+        match_fraction,
+    })
+}
+
+fn validate_header_matcher(hm: HeaderMatcherView<'_>) -> xds_client::Result<HeaderMatcher> {
+    let name = hm.name().to_str().unwrap_or_default().to_string();
```
dfawley, line 372: Should we validate `!name.is_empty()`?

grpc-xds/src/resource/listener.rs
```diff
@@ -0,0 +1,313 @@
...
+        Listener::parse(&bytes)
+            .map_err(|e| Error::Validation(format!("failed to decode Listener: {e}")))
+    }
+
+    fn name(message: &Self::Message) -> &str {
+        message.name().to_str().unwrap_or_default()
+    }
+
+    fn validate(message: Self::Message) -> xds_client::Result<Self> {
+        let name = message.name().to_str().unwrap_or_default().to_string();
```
dfawley, line 82: Should we validate `!name.is_empty()` here as well, matching cluster/endpoint validation?

grpc-xds/src/resource/endpoint.rs
```diff
@@ -0,0 +1,573 @@
...
+impl From<EnvoyHealthStatus> for HealthStatus {
+    fn from(value: EnvoyHealthStatus) -> Self {
+        match value {
+            EnvoyHealthStatus::Healthy => Self::Healthy,
+            // Envoy's TIMEOUT is documented as "interpreted by Envoy as
+            // UNHEALTHY". Per gRFC A27, only HEALTHY and UNKNOWN are usable;
+            // DEGRADED and anything unrecognized fall back to Unknown below.
+            EnvoyHealthStatus::Unhealthy | EnvoyHealthStatus::Timeout => Self::Unhealthy,
+            EnvoyHealthStatus::Draining => Self::Draining,
+            _ => Self::Unknown,
```
dfawley: This catch-all maps `Degraded` to `Unknown`, and then we will end up using the endpoint since we are supposed to use either HEALTHY or UNKNOWN endpoints per Envoy: https://github.com/envoyproxy/envoy/blob/8f77c4a301bd96d2c7231915e6c3d1f85ea88ae1/api/envoy/config/core/v3/health_check.proto#L36

grpc-xds/src/resource/route.rs
```diff
@@ -0,0 +1,928 @@
...
+        .headers()
+        .iter()
+        .map(validate_header_matcher)
+        .collect::<xds_client::Result<Vec<_>>>()?;
+
+    // Per A28: use runtime_fraction.default_value, normalize to numerator out
+    // of 1,000,000. runtime_key is ignored (gRPC has no runtime config).
+    let match_fraction = rm.runtime_fraction_opt().and_then(|rf| {
+        if !rf.has_default_value() {
+            return None;
```
dfawley: If there is no default value set, this should probably error.  (The field is required per the protos: https://github.com/envoyproxy/envoy/blob/b67c14052c49890a7e3afe614d50979c346c024b/api/envoy/api/v2/core/base.proto#L368-L369, so this probably won't ever happen in practice)

grpc-xds/src/resource/cluster.rs
```diff
@@ -0,0 +1,475 @@
...
+fn validate_aggregate_discovery(
+    custom: crate::generated::envoy::config::cluster::v3::cluster::CustomClusterTypeView<'_>,
+) -> xds_client::Result<ClusterDiscovery> {
+    if !custom.has_typed_config() {
+        return Err(Error::Validation(
+            "aggregate cluster missing typed_config".into(),
+        ));
+    }
+    let any = custom.typed_config();
+    let cluster_config = AggregateClusterConfig::parse(any.value()).map_err(|e| {
```
dfawley, line 244: This should also validate any.type_url

dfawley (commented):
Looks great overall, thanks!  I've found a handful of things that should be fixed, but it should all be pretty small.

grpc-xds/src/xds_config.rs
YutaoMa: The individual fields `Arc` would still be needed to share configs across snapshots, like a CDS-only update should keep reusing the LDS/RDS `Arc`s.

grpc-xds/src/resource/route.rs
YutaoMa: In the v3 proto the `total_weight` has been marked as `deprecated`: https://github.com/envoyproxy/envoy/blob/a282c557f6b18241f90bd20c2d00c985b4c3c70b/api/envoy/config/route/v3/route_components.proto#L539-L540, and `the client will use the sum of all cluster weights`. `grpc-go` also ignores this field. I assume we'd like to target v3 APIs only in gRPC Rust?

grpc-xds/src/resource/listener.rs
YutaoMa: Good point, adding the `type_url` check.

grpc-xds/src/resource/route.rs
YutaoMa: Valid, adding the check.

grpc-xds/src/resource/listener.rs
YutaoMa: Valid, adding the check.

grpc-xds/src/resource/cluster.rs
YutaoMa: 👍 , adding the check here as well.

grpc-xds/src/resource/endpoint.rs
YutaoMa: Noted, I'll keep it aligned with other grpc impl to add in the few other variants and default to a different variant like `Other(i32)` instead.

grpc-xds/src/resource/route.rs
YutaoMa: Makes sense, changing it to validation failure.

grpc-xds/src/xds_config.rs
YutaoMa: Valid, adding the same name check.

YutaoMa:
@dfawley all prev comments addressed, ready for another round of review. Thanks!

grpc-xds/src/resource/route.rs
dfawley: Yes, we're not supporting v2.  You are correct that we should ignore this.

grpc-xds/src/resource/endpoint.rs
```diff
@@ -86,18 +86,28 @@ pub(crate) enum HealthStatus {
...
-            // DEGRADED and anything unrecognized fall back to Unknown below.
+            // UNHEALTHY".
             EnvoyHealthStatus::Unhealthy | EnvoyHealthStatus::Timeout => Self::Unhealthy,
             EnvoyHealthStatus::Draining => Self::Draining,
-            _ => Self::Unknown,
+            EnvoyHealthStatus::Degraded => Self::Degraded,
+            // `HealthStatus` is an open enum, so a newer control plane can send
+            // a status this build does not know. Per gRFC A27 only HEALTHY and
+            // UNKNOWN are usable, so an unrecognized status fails closed.
+            other => Self::Other(i32::from(other)),
```
dfawley, line 110: I assume our other languages are treating anything outside this range as UNKNOWN, which would then cause us to use it?  That seems wrong to me, but then this is a behavior difference.  But also I think if envoy ever added values to this enum, it would cause some trouble.

dfawley (approved):
I'm not sure about this health enum, but aside from that everything LGTM.  Thanks!

grpc-xds/src/resource/endpoint.rs
YutaoMa: No, for clarification the other languages also handles a out of range status as a special error case. `UNKNOWN` is genuinely only used for when status equals `UKNOWN`. The previous commit of this PR made the mistake of converting out of range to unknown, the current commit matches other langs.

dfawley: Yes, it sounds like you're right. Go explicitly checks for UNKNOWN or HEALTHY, and per [the proto spec](https://protobuf.dev/programming-guides/enum/), this is an **open** enum that is expected to represent out of range values directly in the field itself.  I didn't realize that had changed from proto2 behavior.  Thanks for clarifying.

PR #2776, opened 2026-07-28
feat(tonic-xds): make cluster discovery generic over the connector factory

LYZJU2019:
## Motivation

Continues making the `tonic-xds` transport pluggable. Today `XdsClusterDiscovery` is hard-wired to the built-in gRPC connector. This makes it generic over a `MakeConnector` factory so a consumer can supply its own transport connector (e.g. an environment-specific TLS/identity story) without touching the discovery machinery.

The default type parameter preserves the current gRPC plaintext/TLS behavior exactly, so this is **additive and non-breaking**.

## Public surface (kept minimal, per the earlier connector-API review)

- Change the (previously unused) `MakeConnector` trait to take a per-cluster `ClusterConfig` view and return a **type-erased** `Result<Arc<dyn Connector<Service = Self::Service> + Send + Sync>, BoxError>`, matching the fallible, rebuilt-per-CDS-update behavior of the built-in gRPC factory. Returning `dyn Connector` (rather than an associated connector type) means an implementor can keep its concrete connector type(s) private and hand back different connectors per cluster without a wrapper enum. `Service` is bound `Send + 'static` (required by the discovery task and the endpoint-change stream).
- Add an opaque `ClusterConfig<'_>` view that currently exposes only `name()`. Read access to the parsed security settings is deferred to a later change.

## Internals (all `pub(crate)`)

- Add `GrpcMakeConnector` (the default factory); refactor `build_connector` to take the parsed security config and return `Arc<dyn Connector>`.
- `XdsClusterDiscovery` and `GrpcMakeConnector` stay `pub(crate)`; only `MakeConnector` (changed) and `ClusterConfig` (new) are public.

## Validation

- `cargo build` under default and `tls-ring,testutil`, with `-D warnings`.
- `cargo test` (`tls-ring,testutil`): all lib + doctests pass.
- `cargo doc --no-deps` under default and `tls-ring`, with `-D warnings`.
- `cargo check-external-types --all-features`: clean (no allowlist change; `BoxError` was already allowed and `ClusterConfig` is crate-owned).

### DIFF
--- a/tonic-xds/src/xds/cluster_discovery.rs
+++ b/tonic-xds/src/xds/cluster_discovery.rs
@@ -45,8 +45,11 @@ use tokio::sync::mpsc;
 use tokio_stream::StreamExt as _;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::transport::{Channel, Endpoint};
+use tower::BoxError;
 
-use crate::client::endpoint::{Connector, EndpointAddress, EndpointChannel};
+use crate::client::endpoint::{
+    ClusterConfig, Connector, EndpointAddress, EndpointChannel, MakeConnector,
+};
 use crate::client::lb::{BoxDiscover, ClusterDiscovery};
 use crate::common::async_util::BoxFuture;
 use crate::xds::cache::XdsCache;
@@ -55,8 +58,6 @@ use crate::xds::cert_provider::verifier::XdsServerCertVerifier;
 #[cfg(feature = "_tls-any")]
 use crate::xds::cert_provider::{CertProviderRegistry, CertificateProvider};
 use crate::xds::endpoint_manager::{ConnectorSwap, EndpointManager};
-use crate::xds::resource::ClusterResource;
-#[cfg(feature = "_tls-any")]
 use crate::xds::resource::security::ClusterSecurityConfig;
 
 /// Buffer capacity for the discovery channel between the spawned task and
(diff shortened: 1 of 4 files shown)


tonic-xds/src/xds/cluster_discovery.rs
```diff
@@ -185,26 +174,89 @@ impl ClusterDiscovery<EndpointAddress, EndpointChannel<Channel>> for XdsClusterD
...
+
+    #[cfg(not(feature = "_tls-any"))]
+    pub(crate) fn new() -> Self {
+        Self {}
+    }
+}
+
+impl MakeConnector for GrpcMakeConnector {
+    type Service = EndpointChannel<Channel>;
+    type Connector = GrpcConnector;
```
YutaoMa: Looking at the new type wrapping of `GrpcConnector` over the enum made me realize we might want to type erase this Connector type instead: have `make_connector` return `Arc<dyn Connector>` instead of associated type:
1. the pub trait forces the Connector types to be public, which forces user to expose their transport implementation detail.
2. the user will also need this `WrapperType(TLSVariantEnum)` scaffolding.

YutaoMa (approved):

tonic-xds/src/xds/cluster_discovery.rs
LYZJU2019: Resolved in [a24a5df](https://github.com/grpc/grpc-rust/pull/2776/commits/a24a5dfce98c79a6de6c86980171a1a2abfa5809)

PR #2777, opened 2026-07-29
try fixing deadlock in xds when more than 65 services on the same port.

aryehlev:
we have in our environment alot of versions of services all on the same port, this causes a deadlock in our xDS client at startup.
The shared route config for the port references more clusters than the client's internal queue can hold (64), and the two tasks that need each other to drain it end up waiting on each other.

### DIFF
--- a/tonic-xds/src/xds/resource_manager.rs
+++ b/tonic-xds/src/xds/resource_manager.rs
@@ -617,4 +617,162 @@ mod tests {
         assert!(state.rds_watcher.is_some());
         assert_eq!(state.rds_name.as_deref(), Some("rc"));
     }
+
+    /// End-to-end replica of the production Istio hang: an RDS update that
+    /// references far more clusters than the xds-client worker's command
+    /// channel buffers (64), reconciled through the real cascade path —
+    /// `handle_rds` holds the event's `ProcessingDone` token while
+    /// `reconcile_clusters` awaits one `watch()` per cluster.
+    ///
+    /// Before xds-client gained ADS flow control, the worker sat inside
+    /// `handle_response` awaiting that token and stopped draining commands;
+    /// once the command channel filled, `reconcile_clusters` blocked on the
+    /// 65th watch and the client deadlocked. This test times out under that
+    /// behavior and completes under the fixed worker.
+    #[tokio::test]
+    async fn rds_referencing_many_clusters_reconciles_without_deadlock() {
+        use envoy_types::pb::envoy::config::route::v3::{
+            Route, RouteAction, RouteConfiguration, RouteMatch, VirtualHost, route::Action,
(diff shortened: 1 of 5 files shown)


tonic-xds/src/xds/resource_manager.rs
```diff
@@ -617,4 +617,162 @@ mod tests {
...
+        // Well past the worker's 64-slot command channel.
+        const CLUSTER_COUNT: usize = 100;
+
+        /// ADS server that answers the first RDS subscription with one
+        /// RouteConfiguration referencing CLUSTER_COUNT clusters and ignores
+        /// everything else (ACKs, CDS subscriptions).
+        struct ManyClustersServer;
+
+        #[tonic::async_trait]
+        impl AggregatedDiscoveryService for ManyClustersServer {
```
YutaoMa: would be better to reuse the existing `xds-test-util::XdsTestControlPlaneServic` utility we created in-repo recently.

xds-client/src/client/worker.rs
```diff
@@ -791,6 +823,29 @@ where
         }
     }
 
+    /// Build the in-flight delivery future for one response's staged watcher
+    /// notifications, or `None` when there is nothing to deliver.
+    fn dispatch_pending(deliveries: Vec<Delivery>) -> Option<PendingDispatch> {
```
YutaoMa: optimization suggestion: instead of sending `Vec<Delivery>` around, use refcount around one token, i.e. `Arc<Option<Oneshot>>` in `ProcessingDone`. I found out all the other grpc implementations use refcounts to track the updates dispatch.

YutaoMa:
@aryehlev , thanks for the PR! It's indeed a crucial bug and exactly the kind of report we appreciate seeing from real production usage. The solution makes sense to me, especially that it matches the dispatching design of other grpc implementations mostly. I added two comments above.

xds-client/src/client/worker.rs
aryehlev: im not sure about this one, you mean to avoid this vec in line 833?
let mut done_rxs = Vec::with_capacity(deliveries.len());

YutaoMa: Sorry for the confusion, the comment is specifically about the oneshot channel in `ProcessingDone`. `dispatch_resources` need to do N(resources) * W(watchers) oneshot chanel allocations for each response. This is a pre-existing issue with the `ProcessingDone` design, not introduced by this PR. I recommend refactor `ProcessingDone` to use one `Arc` to track a single response now to both eliminate the allocations as well as simplifying `Delivery` shape here, it can drop the `Receiver` field.

YutaoMa (approved):
Current fix lgtm after the optimization refactoring. Merging it as is. There is another similar deadlock case identified that I'll follow up in the Issue.

PR #2784, opened 2026-07-30
(refactor) Pick First LB state machine

nathanielford:
## Motivation

The initial implementation of the PickFirst LB was a little tangled in how it implemented the Happy Eyeballs requirements. This meant it was handling some phases with special state and others without. Overall it made it hard to reason about. 

Further, there was abehavior gaps between the original implementation and the spec: the original discards unselected addresses upon entering READY, causing subsequent reconnections from IDLE to only attempt the single failed address instead of sweeping all resolved backends.

## Solution

This implementation is fairly different, using a state machine to hold state for each of the four states of Happy Eyeballs (Idle, FirstPass, SteadyState and Ready), and being clear about state transitions between them. The construction ensures the retention of appropriate information (i.e. addresses), and makes things easier to reason about.

As a bonus, this pass also eliminates a bunch of allocations that were unneeded. Gemini estimates about a 70% smaller memory footprint. As a double plus bonus, it now passes linting checks.


grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+                let current = std::mem::take(&mut self.state);
+                self.state = current.resolver_update(&mut ctx, addresses);
+                Ok(())
+            }
+            Err(error) => {
+                let has_addresses = match &self.state {
+                    PickFirstState::Idle(s) => !s.addresses.is_empty(),
+                    PickFirstState::FirstPass(s) => !s.addresses.is_empty(),
+                    PickFirstState::SteadyState(s) => !s.addresses.is_empty(),
+                    PickFirstState::Ready(s) => !s.addresses.is_empty(),
```
arjan-bal: I believe in all these cases, addresses must be present and the value can directly be `true`.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+    #[allow(clippy::unused_self)]
+    fn resolver_update(
+        self,
+        ctx: &mut PickFirstContext<'_>,
+        addresses: Vec<Address>,
+    ) -> PickFirstState {
+        FirstPassState::fresh_enter(ctx, addresses)
+    }
+
+    #[allow(clippy::unused_self)]
```
arjan-bal: `self` is used in this method, so `#[allow(clippy::unused_self)]` can be avoided.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+        self,
+        _ctx: &mut PickFirstContext<'_>,
+        _subchannel: Arc<dyn Subchannel>,
+        _state: &SubchannelState,
+    ) -> PickFirstState {
+        PickFirstState::Idle(self)
+    }
+
+    fn work(self, ctx: &mut PickFirstContext<'_>) -> PickFirstState {
+        self.exit_idle(ctx)
```
arjan-bal, line 322: There was [a TODO](https://github.com/grpc/grpc-rust/blob/816ea83308f741fd365250bb365dcf60d1b11688/grpc/src/client/load_balancing/pick_first.rs#L592-L593) about checking if it's safe to assume every call to `work` should trigger `exit_idle`. Has this been confirmed?

Also, all the TODOs seem to be removed, have they been addressed?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -333,287 +435,351 @@ impl PickFirstPolicy {
...
+            Self::Ready(r) => f
+                .debug_struct("Ready")
+                .field("selected", &r.selected.address())
+                .finish(),
+        }
+    }
+}
+
+#[cfg(test)]
+impl PickFirstPolicy {
```
arjan-bal: Can we move this to the test mod below?

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+    }
+
+    #[allow(clippy::unused_self)]
+    fn resolver_update(
+        self,
+        ctx: &mut PickFirstContext<'_>,
+        addresses: Vec<Address>,
+    ) -> PickFirstState {
+        FirstPassState::fresh_enter(ctx, addresses)
+    }
```
arjan-bal, line 310: Pickfirst should remain in idle state unless the idle picker is used or the channel calls `exit_idle`.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+        };
+
+        match update.endpoints {
+            Ok(endpoints) => {
+                let addresses = match self.compile_address(endpoints, config) {
+                    Ok(addrs) => addrs,
+                    Err(e) => {
+                        self.state = PickFirstState::Idle(IdleState {
+                            addresses: Vec::new(),
+                        });
```
arjan-bal: Pick-first should probably enter one of the Transient Failure (TF) states here.

When a valid resolver update arrives while the balancer is in TF, it should start using the new address list immediately. However, if the balancer is in IDLE, it should wait until `exit_idle()` is called. We should add tests to verify this.

grpc/src/client/load_balancing/pick_first.rs
```diff
@@ -98,233 +92,341 @@ impl LbPolicyBuilder for PickFirstBuilder {
...
+    }
+
+    fn work(self, ctx: &mut PickFirstContext<'_>) -> PickFirstState {
+        self.exit_idle(ctx)
+    }
+
+    fn exit_idle(self, ctx: &mut PickFirstContext<'_>) -> PickFirstState {
+        if self.addresses.is_empty() {
+            ctx.controller.request_resolution();
+            return PickFirstState::Idle(self);
```
arjan-bal: The balancer must transition the channel out of IDLE here. The IDLE picker only triggers `exit_idle()` once. If the PF balancer remained in IDLE on receiving resolver updates, it will remain permanently stuck in the IDLE state.

arjan-bal (commented):
Leaving initial comments, haven't reviewed the entire PR.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Done.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Done.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I think this misbehavior also existed in the original - good catch!
```
    if let Some(ready_subchannel) = self.rebuild_subchannels(new_addresses, channel_controller) {
        self.subchannel_activate(ready_subchannel, channel_controller);
    } else {
        self.start_connection_pass(channel_controller);
    }
```

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I've had this switch to SteadyState with empty values, which should handle this properly, I think, but please double check. There is also a new test around this.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: Correct me if I'm wrong, but if a newly built LB (in PickFirstState::Idle), with an empty address list, gets a resolver update that errors we could get here without having addresses. 

I also changed the handling if we are idle, have no addresses, or are already in transient failure to enter SteadyState, to avoid being in the Idle state during transient failure.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I've had it enter a fresh first pass regardless here. I think that covers the cases necessary.

grpc/src/client/load_balancing/pick_first.rs
nathanielford: I think all of these have been addressed now:
  1. Line 125 (TODO: should steady_state be a "mode" selector enum...):
      • This was the primary motivation for this PR, replacing mutable fields with the PickFirstState enum (Idle, FirstPass, SteadyState, Ready). 
  2. Line 217 (TODO: prevent redundant IDLE updates?):
      • This is now being handled by state node typestates; IdleState::enter() publishes picker state only on entry and update.
  3. Line 258 (TODO: avoid this update if we are in TF (i.e. sticky TF)?):
      • Sticky TRANSIENT_FAILURE is now isolated inside `SteadyState`, which maintains the FailingPicker without emitting CONNECTING updates until a subchannel reaches READY. So I think this is handled.
  4. Line 266 (TODO: set the last connection error?):
      • SteadyState::enter explicitly accepts last_error: String and passes the most recent error to set_failing_picker.
  5. Line 461 (TODO: make error mandatory.):
      • ctx.set_failing_picker(&error) takes &str instead of Option<String>.
  6. Line 592 (TODO: is it safe to assume any call to work() while idle means we should connect?):
      • I think this should be safe because IdlePicker is the sole scheduler in IDLE. This is mapped in IdleState::work via self.exit_idle(ctx). But if you can think of another scenario let me know!

nathanielford:
I think I've addressed everything so far. Let me know if you have any further feedback!

PR #2751, opened 2026-07-27
feat(tonic-xds): expose the transport (connector) API publicly

LYZJU2019:
## Summary

Makes the `tonic-xds` **transport (connector) seam** and endpoint types public and re-exports them from the crate root.

This is part of the work to make the underlying transport of tonic-xds channel to be pluggable, allowing for custom transport connector use cases.

## Testing

Toolchain builds/tests run with default and `tls-ring` features under `-D warnings`:

- `cargo build -p tonic-xds` (default and `--features tls-ring,testutil`) — clean, no warnings
- `cargo test -p tonic-xds --features tls-ring,testutil` — 377 passed, 0 failed (+ 4 doctests)
- `cargo doc -p tonic-xds --no-deps` (default and `--features tls-ring`, `RUSTDOCFLAGS=-D warnings`) — clean
- `cargo clippy -p tonic-xds` — no new warnings in the changed files

PR #2785, opened 2026-07-30
feat(tonic-xds): add transport-generic build_transport_channel

LYZJU2019:
## What

Adds `XdsChannelBuilder::build_transport_channel`, a transport-generic build path that wires the xDS routing → retry → load-balancing stack over a caller-provided `MakeConnector` and returns a type-erased, cloneable `Send + Sync` `tower::Service`. The existing gRPC channel becomes the special case of this shared stack.

This is an additive, non-breaking change — no existing public API changes behavior, and the gRPC connect path and all gRPC tests are untouched.

## Why

Follow-up to the xDS transport-generalization work (#2750, #2751, #2776). Downstream users that speak a non-gRPC transport (e.g. plain HTTP) need to drive the same xDS routing/retry/LB machinery with their own per-endpoint connector, without reimplementing the stack.

## Details

- **Shared core:** extracted a private `build_stack` that both the gRPC and generic paths call; the gRPC path (`build_from_cache`) is now a thin caller. Also extracted `load_bootstrap` / `build_xds_runtime` helpers. The cert-provider registry stays gRPC-only (built before the bootstrap node is consumed, so the generic path never pays for it).
- **Body flow:** the retry engine converts `Request<B>` → `Request<SharedBody<B>>` so a request can be cloned for replay; a `post_retry_map` bridges back to the body the per-endpoint service expects. gRPC rewraps into a `tonic` body (unchanged); the generic path's connector accepts `SharedBody<B>` directly (identity map).
- **Public retry seam:** made `RetryClassifier`, `RetryConfig`, `RetryBackoffConfig`, and `is_retryable_connection_error` public so non-gRPC transports can plug their own retry logic. Retry behavior is threaded into `build_transport_channel` as a `classifier` + `RetryConfig`; the internal `RetryPolicy` (and its `tower::retry::Policy` impl) stays private to keep the public surface minimal.
- **check-external-types:** allowlisted `shared_http_body` — the caller's connector accepts a `SharedBody`-wrapped body.

## Testing

- New e2e test (`xds_channel_e2e_transport_generic_routes_to_backend`) drives the generic path end-to-end through the mock control plane, using a custom `RetryClassifier` + a `SharedBody`-accepting connector written entirely against the public API.
- Build/test/doc with `-D warnings` (400 tests + 4 doctests), clippy, rustfmt, and `check-external-types --all-features` all pass.

### DIFF
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -114,6 +114,7 @@ allowed_external_types = [
     # not major released
     "prost::*",
     "opentelemetry::*",
+    "shared_http_body::*",
 
     "tower_service::Service",
     "tower::BoxError",
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -22,8 +22,8 @@
  *
  */
 
-use crate::client::cluster::ClusterClientRegistryGrpc;
-use crate::client::endpoint::{EndpointAddress, EndpointChannel};
+use crate::client::cluster::{ClusterClientRegistry, ClusterClientRegistryGrpc};
+use crate::client::endpoint::{EndpointAddress, EndpointChannel, MakeConnector};
 use crate::client::lb::{ClusterDiscovery, XdsLbService};
 use crate::client::route::{PreRouteInterceptor, Router, XdsRoutingLayer};
 use crate::xds::bootstrap::{BootstrapConfig, BootstrapError};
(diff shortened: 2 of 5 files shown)


LYZJU2019:
Superseded by #2786, which reworks the retry configuration to be control-plane-driven: instead of a static `RetryConfig` parameter, `build_transport_channel` now takes a `Fn(&RouteConfigMetadata) -> Option<RetryConfig>` closure and a background task hot-swaps the config from `RouteConfiguration` updates.

PR #2786, opened 2026-07-31
feat(tonic-xds): drive gRPC retry config from RDS RouteAction.retry_policy

LYZJU2019:
## Summary

Drives the gRPC channel's retry configuration from the control plane so retry behavior tracks **RDS** (`RouteConfiguration`) updates without rebuilding the channel.

The retry settings come from the standard Envoy `RouteAction.retry_policy` (gRFC A44), so `tonic-xds` parses them **natively** — no caller-supplied extractor closure is required. This is additive and non-breaking; there is no public API change.

Built on top of #2794. Retry is **per route**: each request retries according to the exact route it matched. The Envoy retry policy is validated when the `RouteConfiguration` is validated; the gRPC retry config is compiled **once per RDS update** and shared behind an `Arc`, so the request hot path just looks it up and clones that pointer, doing no parsing or allocation.

## What changed

- **Validate the retry policy at RDS-validation time; keep gRPC types out of the resource layer.** `RouteRetryConfig::from_proto` validates the Envoy `RetryPolicy` (gRFC A44 — NACK on `numRetries == 0`, non-positive `baseInterval` / `maxInterval`, or out-of-range `google.protobuf.Duration` values) and stores the validated, transport-neutral values on the matched route as `RouteConfig.retry_config: Option<Arc<RouteRetryConfig>>`. A route uses its own `RouteAction.retry_policy` when set, otherwise it inherits the enclosing `VirtualHost.retry_policy` (gRFC A44: a route-level policy completely overrides the virtual host's — values are not merged); `None` when neither specifies retry, and routes inheriting the vhost policy share one `Arc`. The xDS resource types hold no gRPC-specific types.
- **Compile and select the gRPC config per route via the routing decision.** A `RoutingSnapshot` (client layer) bundles the validated `RouteConfigResource` with a map of `GrpcRetrySharedConfig`s compiled from it **once per RDS update** (the `XdsRouter` watch task), keyed by route-config identity. The routing layer resolves the matched route, looks up its compiled retry config in that same snapshot, and stamps it into the request's `RouteDecision`. Because routing and retry read one snapshot, they always act on the same RDS version (no cross-layer skew), and retry runs inside routing so the decision is fixed across a request's retry attempts. Bundling both in one `Arc` keeps the request hot path to a map lookup plus a pointer clone.
- **Separate the shared config from per-request state.** `RetrySharedConfig<C>` holds the immutable config (attempt cap, backoff, retryable code set) behind an `Arc`; `RetryPolicy<C>` holds a pointer to it plus the per-request retry state (backoff cursor, attempt count). Instantiating a policy for a request (`RetryPolicy::from_shared`) is an `Arc` pointer clone plus a zero-field state init — the request hot path does no parsing or allocation. Requests with no `RouteDecision` (non-xDS callers), whose route carries no retry policy, or whose `retry_on` maps to no gRPC status code use the layer's fallback config.
- **Map `retry_on` to gRPC status codes.** `grpc_retry_on_codes` maps Envoy `retry_on` conditions to gRPC status codes (gRFC A44); non-gRPC tokens are ignored (connection-level retries are handled separately). When the mapped set is empty, the route installs no status-code policy — it falls back to the layer default rather than masking connection retries. Envoy `numRetries` maps directly to `RetryConfig.num_retries` (retries, not attempts); unset Envoy fields fall back to `RetryConfig` defaults.

The retry engine (`RetryPolicy`, `RetrySharedConfig`, and the `RetryClassifier` seam) stays transport-agnostic; only the retry layer is gRPC-specific, because it reads the concrete `RouteDecision` extension. The xDS resource types stay free of gRPC business logic: the gRPC retry config is compiled in the client/routing layer, once per RDS update, which is what keeps the request path allocation-free.

## Example RDS consumed

```json
{
  "@type": "type.googleapis.com/envoy.config.route.v3.RouteConfiguration",
  "name": "AccessControlApi",
  "virtualHosts": [{
    "name": "AccessControlApi",
    "domains": ["*"],
    "routes": [{
      "match": { "prefix": "" },
      "route": {
        "cluster": "AgentLifecycleGrpc|0",
        "timeout": "60s",
        "retryPolicy": {
          "retryOn": "unavailable",
          "numRetries": 2,
          "retryBackOff": { "baseInterval": "0.100s", "maxInterval": "1s" }
        }
      }
    }]
  }]
}
```

The `retryPolicy` may equivalently be set at the `virtualHosts[*]` level, in which case every route in the virtual host that doesn't set its own inherits it.

## Testing

`cargo +1.97.0` with `--features tls-ring,testutil`:
- `fmt --check`, `check` (lib + tests), `test` (414 unit + 4 doc), `doc` with `RUSTDOCFLAGS="-D warnings"` (default + `tls-ring`), `clippy` (no new warnings)
- `cargo +nightly-2025-10-18 check-external-types --all-features` (no allowlist change — no public API change)

New unit tests cover the mapping, the split config/state model, the RDS-validation rules, and the virtual-host/route precedence: `from_route_retry` field and default mapping onto `GrpcRetrySharedConfig`, and its `None` result when the mapped code set is empty; `from_shared` instantiating zeroed per-request state while sharing the config by pointer; per-request state independence across cloned policies; RDS validation NACKing invalid retry policies (zero retries, non-positive or out-of-range intervals); and a route overriding the virtual-host policy while sibling routes with no policy share the inherited config `Arc` (plus the no-policy-anywhere case yielding `None`). `test_retry_once_on_unavailable` exercises a real retry with `retry_on = "unavailable"`.

### DIFF
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -367,6 +367,18 @@ impl XdsChannelBuilder {
         resource_manager: XdsResourceManager,
     ) -> XdsChannelGrpc {
         let router: Arc<dyn Router> = Arc::new(XdsRouter::new(&cache));
+
+        // Retry config is control-plane-driven from RDS, per route. It is parsed
+        // and validated once, when the `RouteConfiguration` is validated. The
+        // routing layer (outer) stamps the matched route's shared retry config
+        // into the request's `RouteDecision`; the retry layer reads the shared
+        // config `Arc` and instantiates a per-request policy from it, so the
+        // request hot path does no parsing or allocation. The default below is
+        // the fallback used when a request carries no route retry config (non-xDS
+        // callers, or a route with no retry policy). See
+        // [`RetryLayer`](crate::client::retry::RetryLayer).
+        let retry_layer = RetryLayer::new(GrpcRetryPolicy::default());
+
         #[cfg(feature = "_tls-any")]
         let discovery: Arc<
             dyn ClusterDiscovery<EndpointAddress, EndpointChannel<Channel>>,
@@ -378,7 +390,6 @@ impl XdsChannelBuilder {
         let discovery: Arc<
             dyn ClusterDiscovery<EndpointAddress, EndpointChannel<Channel>>,
         > = Arc::new(XdsClusterDiscovery::new(cache, GrpcMakeConnector::new()));
-        let retry_policy = GrpcRetryPolicy::default();
 
         let resources = Arc::new(XdsChannelResources {
             _resource_manager: resource_manager,
@@ -386,7 +397,6 @@ impl XdsChannelBuilder {
         });
 
         let routing_layer = XdsRoutingLayer::new(router, self.pre_route.clone(), self.authority());
-        let retry_layer = RetryLayer::new(retry_policy);
         let cluster_registry = Arc::new(ClusterClientRegistryGrpc::new());
         let lb_service = XdsLbService::new(cluster_registry, discovery);
         let inner = ServiceBuilder::new()
@@ -527,6 +537,7 @@ mod tests {
             Ok(RouteDecision {
                 cluster: "test-cluster".to_string(),
                 request_hash: None,
+                retry_config: None,
             })
         }
     }
@@ -680,9 +691,7 @@ mod tests {
 
         let retry_policy = GrpcRetryPolicy::new(
             RetryConfig::new().num_retries(1),
-            GrpcRetryClassifier {
-                retry_on: vec![tonic::Code::Unavailable],
-            },
+            GrpcRetryClassifier::new(vec![tonic::Code::Unavailable]),
         );
 
         let xds_channel = XdsChannelBuilder::new(test_config()).build_grpc_channel_from_parts(
@@ -734,6 +743,7 @@ mod tests {
                         match_fraction: None,
                     },
                     action: RouteConfigAction::Cluster(cluster_name.to_string()),
+                    retry_config: None,
                 }],
             }],
             metadata: Default::default(),
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -370,13 +370,9 @@ impl XdsChannelBuilder {
     ) -> XdsChannelGrpc {
         let router: Arc<dyn Router> = Arc::new(XdsRouter::new(&cache));
 
-        // Retry config is control-plane-driven from RDS, per route, compiled once
-        // per RDS update in the routing layer and carried on the request's
-        // `RouteDecision`; the retry layer reads that shared config `Arc` and
-        // instantiates a per-request policy from it, so the request hot path does
-        // no parsing or allocation. The default below is the fallback used when a
-        // request carries no route retry config (non-xDS callers, or a route with
-        // no retry policy). See [`RetryLayer`](crate::client::retry::RetryLayer).
+        // Fallback retry config for requests that carry no per-route config
+        // (non-xDS callers, or a route with no retry policy). Per-route configs
+        // come from RDS via the request's `RouteDecision`; see `RetryLayer`.
         let retry_layer = RetryLayer::new(Arc::new(GrpcRetrySharedConfig::default()));
 
         #[cfg(feature = "_tls-any")]
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -46,8 +46,6 @@ use xds_client::{
     ClientConfig, MetricsRecorder, Node, ProstCodec, TokioRuntime, TonicTransportBuilder, XdsClient,
 };
 
-#[cfg(test)]
-use crate::client::retry::GrpcRetryPolicy;
 use crate::client::retry::{GrpcRetrySharedConfig, RetryLayer};
 
 /// Configuration for building [`XdsChannel`] / [`XdsChannelGrpc`].
@@ -418,17 +416,17 @@ impl XdsChannelBuilder {
     }
 
     /// Builds an `XdsChannelGrpc` from the given router, cluster discovery, retry
-    /// policy, and optional pre-route interceptor.
+    /// config, and optional pre-route interceptor.
     #[cfg(test)]
     pub(crate) fn build_grpc_channel_from_parts(
         &self,
         router: Arc<dyn Router>,
         discovery: Arc<dyn ClusterDiscovery<EndpointAddress, EndpointChannel<Channel>>>,
-        retry_policy: GrpcRetryPolicy,
+        retry: Arc<GrpcRetrySharedConfig>,
         interceptor: Option<Arc<dyn PreRouteInterceptor>>,
     ) -> XdsChannelGrpc {
         let routing_layer = XdsRoutingLayer::new(router, interceptor, self.authority());
-        let retry_layer = RetryLayer::new(retry_policy.into_shared());
+        let retry_layer = RetryLayer::new(retry);
         let cluster_registry = Arc::new(ClusterClientRegistryGrpc::new());
         let lb_service = XdsLbService::new(cluster_registry, discovery);
         let inner = ServiceBuilder::new()
@@ -464,7 +462,7 @@ mod tests {
         XdsChannelConfig::new(XdsUri::parse("xds:///test-service").unwrap())
     }
     use crate::client::lb::{BoxDiscover, ClusterDiscovery};
-    use crate::client::retry::GrpcRetryPolicy;
+    use crate::client::retry::GrpcRetrySharedConfig;
     use crate::client::route::RouteDecision;
     use crate::client::route::RouteInput;
     use crate::client::route::Router;
@@ -616,7 +614,7 @@ mod tests {
         let xds_channel = xds_channel_builder.build_grpc_channel_from_parts(
             xds_manager.clone(),
             xds_manager.clone(),
-            GrpcRetryPolicy::default(),
+            Arc::new(GrpcRetrySharedConfig::default()),
             None,
         );
 
@@ -674,7 +672,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_retry_once_on_unavailable() {
-        use crate::client::retry::{GrpcRetryClassifier, GrpcRetryPolicy, RetryConfig};
+        use crate::client::retry::{GrpcRetryClassifier, RetryConfig};
         use crate::testutil::grpc::spawn_fail_first_n_server;
 
         // Server fails the first request with UNAVAILABLE, succeeds on retry.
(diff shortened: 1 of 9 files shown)


YutaoMa:
A few issues with the current code that require fundamental design change:
1. I see the newer commits reverted back to a dedicated watch task in the retry layer. It is still more ideal to preserve config consistency across the two layers through request extension passing.
2. The retry config field in `VirtualHost` level isn't read. According to A44, both VH and Route level retry config are respected, with the Route level taking priority.
3. A recommendation to keep the design both completes A44 and efficient on the request path: extract and validate retry config at resource validation time. And separate out the per-request retry state (current backoff, etc.) from the shared config (limit, retryOn, etc.). Pass down shared config through `Arc` pointers, and instantiate the state in retry layer per request, which would be simplified to only a pointer clone and some zero-field init.

Let me know if you have any questions re: the design proposal.

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -149,6 +152,52 @@ pub(crate) struct RouteConfigResource {
...
+    /// `retry_back_off.max_interval` if set.
+    pub max_interval: Option<Duration>,
+}
+
+impl RouteRetryConfig {
+    /// Parse an Envoy `RetryPolicy` proto into a [`RouteRetryConfig`].
+    fn from_proto(rp: &RetryPolicy) -> Self {
+        let (base_interval, max_interval) = match rp.retry_back_off.as_ref() {
+            Some(backoff) => (
+                backoff.base_interval.as_ref().and_then(proto_duration),
```
YutaoMa: In [A44](https://github.com/grpc/proposal/blob/master/A44-xds-retry.md#converting-envoy-retrypolicy-to-grpc-retrypolicy), some limits are given for these fields, such as "base_interval and max_interval must be greater than zero", the resource should fail validation for those conditions, so the XDS client can properly NACK according to the spec.

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -306,11 +391,25 @@ fn validate_route(
...
+    let retry_config;
+    match action {
+        route::Action::Route(route_action) => {
+            // Parse and validate the route's retry policy before `route_action`
+            // is consumed. Deriving the gRPC config here — once, at RDS-validation
+            // time — is what keeps the request hot path free of parsing and
+            // allocation (see `RouteConfig::retry_config`). A route-level policy
+            // takes precedence over the virtual host's (gRFC A44); the vhost
+            // fallback is applied below.
+            retry_config = route_action.retry_policy.as_ref().map(|rp| {
```
YutaoMa: you can early return on None here to save a unnecessary parse.

tonic-xds/src/client/retry.rs
```diff
@@ -303,10 +360,61 @@ impl<C> RetryPolicy<C> {
...
+    /// wrapped in an [`Arc`] and carried on the matched route, so instantiating a
+    /// per-request policy is just a pointer clone (see
+    /// [`RetryPolicy::from_shared`]).
+    pub(crate) fn from_route_retry(retry: &RouteRetryConfig) -> Self {
+        let mut config = RetryConfig::new();
+        if let Some(num_retries) = retry.num_retries {
+            config = config.num_retries(num_retries);
+        }
+        if let Some(base_interval) = retry.base_interval {
+            let mut backoff = RetryBackoffConfig::new(base_interval);
```
YutaoMa, line 367: `proto_duration` has no range check  and`RetryBackoffConfig::new`  does base_interval * 10 , so a very large `base_interval` can panic. I recommend adding a range check in `proto_duration` to reject invalid duration and also make `RetryBackoffConfig::new` use `checked_mul`. (For context, the documented protobuf.Duration type max should be `315_576_000_000`, just that the Rust `envoy_types` type itself doesn't enforce that invariant. It's safe to NACK the config if the provided value exceeds that)

tonic-xds/src/xds/resource/route_config.rs
```diff
@@ -25,18 +25,21 @@
...
-    RouteConfiguration, RouteMatch, route, route_action, route_match,
+    RetryPolicy, RouteConfiguration, RouteMatch, route, route_action, route_match,
 };
 use prost::Message;
 use regex::Regex;
 use xds_client::resource::TypeUrl;
 use xds_client::{Error, Resource};
 
 use super::string_matcher::StringMatcher;
+use crate::client::retry::GrpcRetrySharedConfig;
```
YutaoMa: actually although I mentioned previously that we should minimize request path allocation and compute, it'd be too aggressive to put grpc business logic types into the xds resource types, it violates module boundaries. Let's still keep the validated proto -> actual grpc config type in the retry layer, it will be done only once per RDS update anyway.

tonic-xds/src/client/retry.rs
```diff
@@ -303,10 +360,61 @@ impl<C> RetryPolicy<C> {
...
+    }
+}
+
+/// Map Envoy `retry_on` conditions (comma-separated) to gRPC [`tonic::Code`]s.
+///
+/// Only the gRPC-status conditions from gRFC A44 are recognized; non-gRPC tokens
+/// (e.g. `5xx`, `gateway-error`, `reset`, `connect-failure`) are ignored because
+/// connection-level retries are handled separately by
+/// [`is_retryable_connection_error`].
+pub(crate) fn grpc_retry_on_codes(retry_on: &str) -> Vec<tonic::Code> {
```
YutaoMa, line 383: A44 has a minor behavior detail: `If the resulting retryableStatusCodes is empty, gRPC xDS client will not include a retry policy in the corresponding route action`, which this code currently doesn't do, it instead sets an empty policy. This matters because connection errors are also dependent on the retry policy, so right now an empty `retryableStatusCodes` = empty policy = disable connection retry. I checked other gRPC implementations and confirmed connection retry should not be masked by empty `retryableStatusCodes`. To keep it aligned with gRPC retry behavior we should return a `None` at policy level when `retry_on` is empty.

tonic-xds/src/client/retry.rs
```diff
@@ -347,50 +453,92 @@ where
...
-impl<P> RetryLayer<P> {
-    /// Create a new retry layer with the given policy.
-    pub(crate) fn new(policy: P) -> Self {
-        Self { policy }
+impl RetryLayer {
+    /// Create a retry layer whose `fallback` policy is used when a request
+    /// carries no per-route retry config. Only the policy's shared config is
+    /// kept; its per-request state is discarded. Route-specific configs are read
+    /// from the request's [`RouteDecision`] (see [`RetryLayer`]).
+    pub(crate) fn new(fallback: GrpcRetryPolicy) -> Self {
```
YutaoMa: since we only need the shared config here, maybe take the `Arc<SharedConfig>` as the param?

tonic-xds/src/xds/resource/route_config.rs
LYZJU2019: Done. `RouteRetryConfig::from_proto` now NACKs (returns `Error::Validation`) when `num_retries == 0`, or when `base_interval` / a set `max_interval` are non-positive, matching the grpc-go `generateRetryConfig` behavior (gRFC A44). Tests cover each case.

tonic-xds/src/xds/resource/route_config.rs
LYZJU2019: Done. `validate_route` now takes the retry policy off the action and runs `validate_route_action` first (which may return `None` to drop the route); the retry policy is parsed only when the route is kept, so dropped routes cost no retry parsing.

tonic-xds/src/client/retry.rs
LYZJU2019: Done. `proto_duration` now rejects values outside the documented `google.protobuf.Duration` range (seconds not in [0, 315576000000] or nanos not in [0, 999999999]) and returns `None`, which NACKs. `RetryBackoffConfig::new` now uses `checked_mul(10)` with a fallback to `base_interval`, so the range-checked base cannot overflow the default max.

tonic-xds/src/xds/resource/route_config.rs
LYZJU2019: Done. The xDS resource layer now carries only the validated `RouteRetryConfig` (no gRPC types). A new `RoutingSnapshot` in the client layer bundles the `RouteConfigResource` with a map of gRPC retry configs compiled once per RDS update (the `XdsRouter` watch task). Routing and retry read the same snapshot, so the request hot path stays a map lookup plus an `Arc` clone while the resource/client boundary stays clean.

tonic-xds/src/client/retry.rs
LYZJU2019: Done. `GrpcRetrySharedConfig::from_route_retry` now returns `Option<Self>` and yields `None` when the parsed `retry_on` set is empty, so the route falls back to the layer default instead of installing a status-code policy that would mask connection-level retries (gRFC A44). Added a test.

tonic-xds/src/client/retry.rs
LYZJU2019: Done. `RetryLayer::new` now takes `Arc<GrpcRetrySharedConfig>` directly. Added `RetryPolicy::into_shared(self) -> Arc<RetrySharedConfig<C>>` for the call sites that had a `GrpcRetryPolicy` in hand.

tonic-xds/src/client/channel.rs
YutaoMa: should just take shared here as well instead of converting

tonic-xds/src/client/channel.rs
```diff
@@ -46,7 +46,9 @@ use xds_client::{
     ClientConfig, MetricsRecorder, Node, ProstCodec, TokioRuntime, TonicTransportBuilder, XdsClient,
 };
 
-use crate::client::retry::{GrpcRetryPolicy, RetryLayer};
+#[cfg(test)]
+use crate::client::retry::GrpcRetryPolicy;
```
YutaoMa: move test imports to tests module

tonic-xds/src/client/channel.rs
LYZJU2019: Done. `build_grpc_channel_from_parts` now takes `Arc<GrpcRetrySharedConfig>` directly; the call sites pass the shared config (`Arc::new(GrpcRetrySharedConfig::default())` / `Arc::new(GrpcRetrySharedConfig::new(...))`), so there is no `into_shared()` conversion in the helper.

tonic-xds/src/client/channel.rs
LYZJU2019: Done. Removed the module-scope `#[cfg(test)] use ... GrpcRetryPolicy;`. The retry-config import now lives in the `tests` module (`GrpcRetrySharedConfig`), and `GrpcRetryPolicy` is no longer referenced in this file.

YutaoMa (approved):
LGTM

PR #2789, opened 2026-08-04
feat(grpc): add ServerBuilder, registration, interceptors, and routing

sauravzg:
Introduce the server-side handle/router API for building a gRPC `Server` from a fluent builder.

- ServerBuilder: fluent construction via `Server::builder()`, with `add_service`, `interceptor`and `build`.
- Service + ServiceExt: `Service` trait for method registration, plus `with_interceptor` to wrap all of a service's methods (InterceptedService).
- Interceptors: `Intercept` trait, no-op `Identity`, and `InterceptExt::chain` for composing interceptors into an `InterceptorChain` (first added runs outermost).
- Descriptors: `ServiceDescriptor`, `MethodDescriptor`, and `MethodType`.
- Routing: `RouterBuilder` maps method paths to `DynHandle`s.
- Options: `ServerOptions` currently empty, but a kitchen sink for options.


grpc/src/server/builder.rs
```diff
@@ -0,0 +1,308 @@
...
+/// A fluent builder for constructing a [`Server`].
+///
+/// Register services with [`add_service()`](ServerBuilder::add_service) and add
+/// global interceptors with [`interceptor()`](ServerBuilder::interceptor), then
+/// finish with [`build()`](ServerBuilder::build).
+///
+/// # Examples
+///
+/// ```ignore
+/// let server = Server::builder()
```
dfawley: We have both `ServerBuilder::new()` and `Server::builder()`?  Why?

grpc/src/server/descriptor.rs
```diff
@@ -0,0 +1,181 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+/// The type (cardinality) of a gRPC method.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MethodType {
```
dfawley, line 27: I thought we would hide this from grpc itself and let it treat everything as bidi streaming?

grpc/src/server/interceptor.rs
```diff
@@ -89,6 +89,114 @@ pub trait HandleExt: Handle + Sized {
...
+impl Intercept for Identity {
+    async fn intercept(
+        &self,
+        headers: RequestHeaders,
+        options: CallOptions,
+        tx: &mut impl SendStream,
+        rx: impl RecvStream + 'static,
+        next: &impl Handle,
+    ) -> Trailers {
+        next.handle(headers, options, tx, rx).await
```
dfawley, line 108: With optimizations on, does this cause literally no overhead?

grpc/src/server/mod.rs
```diff
@@ -173,20 +181,40 @@ impl GracefulCoordinator {
...
+    /// let server = Server::builder()
+    ///     .add_service(greeter_service)
+    ///     .build();
+    /// ```
+    pub fn builder() -> builder::ServerBuilder<interceptor::Identity> {
+        builder::ServerBuilder::new()
+    }
+
+    /// Creates a new server with the given handler, runtime, and options.
+    pub(crate) fn new(
```
dfawley: Shouldn't we delete `new` and require the use of the builder?

grpc/src/server/mod.rs
```diff
@@ -173,20 +181,40 @@ impl GracefulCoordinator {
...
-        H: Handle + Send + Sync + 'static,
-    {
-        self.handler = Some(Arc::new(h))
+    /// Returns the runtime used by this server.
+    pub fn runtime(&self) -> &GrpcRuntime {
+        &self.runtime
+    }
+
+    /// Returns the server options.
+    pub fn options(&self) -> &ServerOptions {
```
dfawley: What is this for?

grpc/src/server/router.rs
```diff
@@ -0,0 +1,733 @@
...
+use crate::server::Trailers;
+use crate::server::descriptor::{MethodDescriptor, ServiceDescriptor};
+use crate::server::interceptor::Identity;
+use crate::server::interceptor::Intercept;
+use crate::server::interceptor::InterceptExt;
+use crate::server::interceptor::InterceptorChain;
+use crate::server::service::Service;
+
+/// A builder for constructing an immutable [`Router`].
+pub(crate) struct RouterBuilder<I = Identity> {
```
dfawley: Hmm, why not use a default type for `ServerBuilder`, too?  I assume this is here so that you can `impl RouterBuilder` instead of needing to `impl RouterBuilder<Identity>`?

grpc/src/server/router.rs
```diff
@@ -0,0 +1,733 @@
...
+/// A `Router` is immutable once built; use [`RouterBuilder`] to construct one.
+///
+/// # Example
+///
+/// ```ignore
+/// let server = Server::builder()
+///     .add_service(my_service)
+///     .build();
+/// ```
+pub struct Router<I = Identity> {
```
dfawley: This one I can't see why there's a default type?

grpc/src/server/builder.rs
```diff
@@ -0,0 +1,308 @@
...
+
+/// A fluent builder for constructing a [`Server`].
+///
+/// Register services with [`add_service()`](ServerBuilder::add_service) and add
+/// global interceptors with [`interceptor()`](ServerBuilder::interceptor), then
+/// finish with [`build()`](ServerBuilder::build).
+///
+/// # Examples
+///
+/// ```ignore
```
dfawley: I would prefer if we could avoid `ignore` for our examples.  Especially as we're heavily iterating on things, there's a high likelihood that they will become broken this way.  Please use other tags instead like `no_run` or use leading `#` to hide things from the rendered documentation.

grpc/src/server/builder.rs
```diff
@@ -0,0 +1,308 @@
...
+    use crate::server::DynHandle;
+    use crate::server::Handle;
+    use crate::server::RecvStream;
+    use crate::server::RequestHeaders;
+    use crate::server::ResponseStreamItem;
+    use crate::server::SendOptions;
+    use crate::server::SendStream;
+    use crate::server::Server;
+    use crate::server::Trailers;
+    use crate::server::descriptor::{MethodDescriptor, MethodType, ServiceDescriptor};
```
dfawley: Please `rustfmt` to ungroup imports:

`rustfmt --edition=2024 --config imports_granularity=Item --config group_imports=StdExternalCrate <files>`

or inside the grpc crate, this will get everything:

`cargo fmt -- --config imports_granularity=Item --config group_imports=StdExternalCrate`

grpc/src/server/builder.rs
sauravzg: `new()` is now `pub (crate)` .

grpc/src/server/interceptor.rs
sauravzg: I can't confirm this , but I don't see a reason why.  It's a zero sized type with no allocations and I'd assume the compiler can inline this to make it no-op.

Should we add a microbenchmark for this?

grpc/src/server/mod.rs
sauravzg: new is `pub(crate)` , the only public way to construct is via the builder which on build calls this new. 

Are we looking for something different?

grpc/src/server/mod.rs
sauravzg: I kept it as a kitchen sink config for server options. It initially had graceful timeout until we decided to move it to the application layer.

sauravzg: Deleted this since it's unused.

grpc/src/server/router.rs
sauravzg: Removed this. Mostly consistency since other things had it.

grpc/src/server/builder.rs
sauravzg: Done

grpc/src/server/builder.rs
sauravzg: I've decided to remove most of the doc tests. Didn't see a huge value in the ones with `no_run` . 
Fixing them right now would make a bunch of symbols public which I would want to avoid until it's needed.

grpc/src/server/router.rs
sauravzg: ServerBuilder is already default typed. The chain of default types was keeping everything similar .

Anyways, followed the call chain and realized that it's not needed, since the way it's modelled right now is to chain interceptors and the only public entrypoint is `Server::builder()`, having default type on ServerBuilder combined with `impl RouterBuilder<Identity>` should be good enough.

grpc/src/server/descriptor.rs
sauravzg: Sorry, I missed addressing this yesterday. I agree in principle but I think we have precedent at least in java and very likely c++(but not c-core), where the descriptors are a part of interceptor API and support a notion of method type.

Slighly confused about go , but you probably know better about it.

Given, that this is very useful information to have about an RPC, I am leaning towards the Java-ish approach which would essentially be adding another enum named "unknown" here and be done with it.

grpc/src/server/builder.rs
dfawley: IIUC `no_run` means it will still be compiled which can be useful.

grpc/src/server/descriptor.rs
dfawley: Hmm, @ejona86 what do you think about this?  Java _seems_ to use it in the stub code only, which does care about the cardinality, but I'm not sure.

I don't think we would need this for protocol enforcement or for optimizations like more quickly requesting the request message.  I think we discussed this being potentially useful for interceptors?  But interceptors could theoretically use reflection based on the method name, instead, I'd think.

What were you planning on using this for in Rust, @sauravzg?

grpc/src/server/interceptor.rs
dfawley: Can we inspect the code generation to confirm one-time?  I think it should all be optimized away completely (with compiler optimizations enabled), but this will be in every RPC so it would be worth checking our assumptions.

grpc/src/server/builder.rs
sauravzg: So, it was more painful to have a long example compile and needed some symbols to be exposed publicly, which I'd like to avoid for  now. I think unit tests should be good enough to address the compilation needs.

grpc/src/server/descriptor.rs
sauravzg: I did not have any plans to use it. It was based on the precedent that I recalled Java interceptor API exposing `MethodDescriptors`  and it having an enum for method type.

I can happily erase it for now and add things later if we end up needing it.

grpc/src/server/interceptor.rs
sauravzg: Do you mean investigating the assembly code, llvm IR or some benchmark? I have no prior experience at reading IR or assembly, but I can happily point gemini at it if you are okay with it.

sauravzg: Here's what I got out of it.  gpaste/4854736038264832

All symbols seemed to belong `drop_glue` which gemini thinks is not relevant.

grpc/src/server/builder.rs
dfawley: That's fine, too.  I'd honestly rather not have an example than have one that isn't tested, _especially_ while we're unstable.

grpc/src/server/interceptor.rs
dfawley: I did some exploring on https://godbolt.org/.  Gemini argued with me for a long time that it has zero overhead, but it does appear there is a cost.

https://godbolt.org/z/aajnjnnhv

Using `Identity` results in 32 bytes of extra stack allocations and 4 more store instructions.

Is this significant? I have no idea.  But it's not zero-cost and it's happening on every call.

Can we make `Identity` _not_ implement `Intercept` and have two different implementations of the ServerBuilder and the Server: one with interceptors and one without any (generic on Identity)?  The one with interceptors would avoid the extra interceptor at the end and the one without interceptors would just...not have any interception happening and do direct dispatch to handlers?

I'm not sure how much extra effort this would be.  Maybe it's not worth it?

sauravzg: Great catch!! That was insightful.

But this has me question other supposedly "false zero cost abstractions" like interceptor chainging. I'll generating assembly to see if that needs to be optimised as well. (Although not in this PR). 

So, the approach about specialization for `NoInterceptor` for server (and deps) sounds reasonable and shouldn't be too complex. I'll make the change and update the comment.

dfawley: I tried something else:
```rust
impl Intercept for Identity {
    fn intercept(
        &self,
        headers: RequestHeaders,
        options: CallOptions,
        tx: &mut impl SendStream,
        rx: impl RecvStream + 'static,
        next: &impl Handle,
    ) -> impl Future<Output = Trailers> {
        next.handle(headers, options, tx, rx)
    }
}
```

And now it seems the costs go away...  So that's another option (a much simpler one, given where we are at currently), I think.

sauravzg: That was a bummer :( . Anyways, done.

grpc/src/server/interceptor.rs
```diff
@@ -63,16 +63,15 @@ where
...
 {
-    async fn handle(
+    fn handle(
         &self,
         headers: RequestHeaders,
         options: CallOptions,
         tx: &mut impl SendStream,
         rx: impl RecvStream + 'static,
-    ) -> Trailers {
+    ) -> impl std::future::Future<Output = Trailers> + Send {
```
dfawley, line 72: Nit: can you `use` and just have `Future` here?

grpc/src/server/interceptor.rs
```diff
@@ -89,6 +88,103 @@ pub trait HandleExt: Handle + Sized {
...
+        tx: &mut impl SendStream,
+        rx: impl RecvStream + 'static,
+        next: &impl Handle,
+    ) -> Trailers {
+        // Build a temporary Handle that runs `second` then delegates to `next`.
+        let inner = SecondThenNext {
+            second: &self.second,
+            next,
+        };
+        self.first.intercept(headers, options, tx, rx, &inner).await
```
dfawley, line 158: Can this one also turn into `-> impl Future`?

dfawley (approved):
LGTM except for the two suggestions.

grpc/src/server/interceptor.rs
sauravzg: fixed.

grpc/src/server/descriptor.rs
sauravzg: Discussed offline. I've removed it.

PR #2791, opened 2026-08-04
tonic/client: support immediate cancellation of streams without sending End-of-Stream

arjan-bal:
## Motivation

Tonic's client streaming and bidirectional (bidi) streaming APIs do not currently support immediate request cancellation. Because outbound request streams are infallible, completing them automatically transmits a `DATA` frame with an `End-of-Stream (EOS)` flag, which signals a normal, graceful termination. An `RST_STREAM` frame is only sent by the client after the request stream terminates and the corresponding response stream is subsequently dropped. 

This delayed teardown creates a problem: the server may interpret the `EOS` flag as a graceful finish and process or commit incomplete payloads before it ever receives the `RST_STREAM` frame. To prevent potential data corruption, Tonic requires a mechanism to abort and tear down the HTTP/2 stream immediately without sending a misleading `EOS` flag.

## Solution

Introduce a paired `CancellationHandle` to coordinate immediate, client-side cancellation of outbound request streams.

### Core Changes

*   `CancellationHandle` (User-Facing): Users can create a handle by calling `request.cancellation_handle()`. Invoking `.cancel()` on this handle triggers an immediate, non-graceful abort.
*   The `CancellationHandle` wraps a `CancellationToken` which is extracted from the request extensions by Tonic's client dispatcher to actively track the cancellation state of the active stream.
* Immediate Abort in Encoder : `EncodeBody::poll_frame` actively checks the `CancellationListener`. If cancellation is flagged, it immediately returns an error mapped with `H2Reason::CANCEL`. This bypasses the empty `DATA` frame with `EOS` and immediately triggers an HTTP/2 level `RST_STREAM`.

## Alternative
Use the Encoder API to inject an error causing hyper to cancel the stream: https://github.com/grpc/grpc-rust/pull/2788

This seems like a mis-use of the codec API and causes the RST_STREAM code to be `internal`.


examples/Cargo.toml
dfawley, line 1: What is the significance of the diffs in this file?  I think these are essentially equivalent before/after?  Isn't tonic's "h2" feature independent from this one?

tonic/src/codec/encode.rs
```diff
@@ -340,6 +364,21 @@ where
...
+        if let Some(cancellation_fut) = self_proj.cancellation_fut.as_pin_mut()
+            && let Poll::Ready(()) = cancellation_fut.poll(cx)
+        {
+            let status = Status::cancelled("client cancelled");
+
+            #[cfg(feature = "h2")]
+            let mut status = status;
+            #[cfg(feature = "h2")]
+            {
+                status.set_source(std::sync::Arc::new(H2Error::from(H2Reason::CANCEL)));
```
dfawley, line 376: What's the purpose of this source having an H2 error in it?  Is this needed?  It only happens if the (newly added) h2 feature is set, or if "server" is set...why doesn't client-side get it always?

It seems like we could just eliminate it since we already set status to cancelled, but maybe I'm missing an important use case?

dfawley (commented):
Looks good, but I'm curious about the feature changes and also you had mentioned needing to manage the waker before, but I don't see anything about that in this PR.  Thanks!

examples/Cargo.toml
arjan-bal: The `cargo-udeps` check in CI fails when a feature shares the same name as an optional dependency (see this [failure example](https://github.com/grpc/grpc-rust/actions/runs/31167727266/job/92832368310)). 

I previously encountered a similar issue when introducing a feature flag for the optional `tower` dependency. We resolved it by explicitly declaring a feature flag with the same name as the optional dependency (see [PR #2432](https://github.com/grpc/grpc-rust/pull/2432/changes)). While the exact root cause is not fully clear, it likely stems from how Cargo implicitly defines feature flags for optional dependencies, which confuses `cargo-udeps` during unused dependency detection.

arjan-bal: I've made `h2` a required dependency in `tonic` now, making the changes unecessary.

tonic/src/codec/encode.rs
arjan-bal: > What's the purpose of this source having an H2 error in it? 

When [poll_frame returns an error](https://github.com/hyperium/hyper/blob/99f243450268cfc8125ff232e0b7de016a1dce5b/src/proto/h2/mod.rs#L231), hyper calls [on_user_error](https://github.com/hyperium/hyper/blob/99f243450268cfc8125ff232e0b7de016a1dce5b/src/proto/h2/mod.rs#L251-L259) to send a `RST_STREAM`. This, in turn, calls [h2_reason](https://github.com/hyperium/hyper/blob/8a98d13c83b1c833301d76121bb445b561dc6f2f/src/error.rs#L353-L359), which inspects the error's cause chain for an `h2::Reason`, defaulting to `INTERNAL` if none is found. By explicitly setting `h2::Reason::CANCEL` as the source in the Tonic status here, we guarantee that the RST code received by the server is `CANCEL` rather than `INTERNAL`.

I wasn't sure initially why the `h2` dependency was optional and gated behind the `server` feature flag, so I kept the new client-side usages behind a feature flag as well to maintain that behavior. However, upon closer consideration, I suspect it was made optional to prevent `cargo-udeps` from flagging it as unused when it was only referenced in server code. I changed `h2` to be a a required dependency now to ensure the correct RST code will be sent every time.

examples/Cargo.toml
dfawley: I'm still not sure I follow why `examples` needed to change, though?  Didn't you need to set the h2 flag in tonic, not change this?

tonic/src/codec/encode.rs
dfawley: Thanks for the explanation.  So if you aren't using hyper then this does nothing?  So should this be predicated on "channel"?  But also if you aren't using hyper then how do we express stream cancellation to the transport  properly (whatever that would be)?  Something here seems like it needs to be understood a little better, I think.

LucioFranco: This feels correct to me (for how we designed this code), h2/hyper is optional but the one that has the best support. So if you are not using hyper then you have to figure it out yourself. We need to pass the source through as a "flag" for hyper/h2 so it knows how to respond but that has to tie into that crate somehow. So this is a way for us to support it.

dfawley (approved):

examples/Cargo.toml
arjan-bal: If you revert `examples/Cargo.toml` and run the following, you'll see a similar failure to the udeps CI:
```sh
$ git revert examples/Cargo.toml
$ cargo udeps --no-default-features --features h2 -p tonic
error: package `examples v0.0.0 (HOME/grpc-rust/examples)` does not have feature `h2`

help: an optional dependency with that name exists, but the `features` table includes it with the "dep:" syntax so it does not have an implicit feature with that name
Dependency `h2` would be enabled by these features:
        - `streaming`
```

Adding a feature in the `tonic` crate interacts with the `examples` crate which uses the `dep:` syntax to specify an optional dependency on `h2`. I couldn't figure out why this happens, but adding a feature named `h2` in the examples crate works around the issue.

arjan-bal: After discussing this with Gemini, the issue seems to be due to the behavior of `cargo metadata` (which `cargo-udeps` uses internally) in workspaces.

```bash
cargo metadata --no-default-features --features X
```

The above command will:

* Succeed if a feature named `X` is either present or entirely absent in a crate.
* Fail if a feature named `X` is absent, but an optional dependency `dep:X` exists.

So `cargo metadata --no-default-features --features h2` fails, while `cargo metadata --no-default-features --features server` succeeds.

examples/Cargo.toml
```diff
@@ -276,7 +276,8 @@ health = ["dep:tonic-health"]
 grpc-web = ["dep:tonic-web", "dep:bytes", "dep:http", "dep:hyper", "dep:hyper-util", "dep:tracing-subscriber", "dep:tower", "dep:tower-http", "tower-http?/cors"]
 tracing = ["dep:tracing", "dep:tracing-subscriber"]
 uds = ["dep:tokio-stream", "tokio-stream?/net", "dep:tower", "dep:hyper", "dep:hyper-util"]
-streaming = ["dep:tokio-stream", "dep:h2"]
+h2 = ["dep:h2"]
```
LucioFranco: prob dont need this feature in examples?

tonic/src/codec/encode.rs
```diff
@@ -340,6 +364,23 @@ where
...
+            let status = Status::cancelled("client cancelled");
+
+            #[cfg(feature = "h2")]
+            let mut status = status;
+            #[cfg(feature = "h2")]
+            {
+                // h2 inspects the error's source chain to determine the RST
+                // code, so we set it here.
+                status.set_source(std::sync::Arc::new(H2Error::from(H2Reason::CANCEL)));
+            }
```
LucioFranco: I wonder if using something like https://doc.rust-lang.org/stable/std/macro.cfg_select.html would make this cleaner to read

LucioFranco (approved):
LGTM left you to suggestions

tonic/src/codec/encode.rs
arjan-bal: Yes, it would. However, `cfg_select` was stabilized in 1.95, while our MSRV is presently 1.88. 

I've changed the variable assignment to use two mutually exclusive `#[cfg]` attributes to make things clearer.

examples/Cargo.toml
arjan-bal: This was required because of a limitation of `cargo udeps`. udeps [resolves the entire workspace](https://github.com/est31/cargo-udeps/blob/f9d686686ae8b4ef8513acb51d3c547dc3c6072f/src/lib.rs#L279-L288) with the given feature flag in one go. This fails when one crate has a feature flag (`h2` in tonic here) while another crate uses the `dep:` syntax for the same name (`dep:h2` in examples here):
```sh
$ cargo udeps --no-default-features --features h2 -p tonic
error: package `examples v0.0.0 (HOME/grpc-rust/examples)` does not have feature `h2`

help: an optional dependency with that name exists, but the `features` table includes it with the "dep:" syntax so it does not have an implicit feature with that name
Dependency `h2` would be enabled by these features:
        - `streaming`
```

arjan-bal: I've raised a PR to fix udeps: https://github.com/est31/cargo-udeps/pull/338

arjan-bal: Added a TODO to remove the unnecessary feature in the examples crate once the udeps bug is released.

PR #2792, opened 2026-08-05
grpc: fork headers/trailers to be separate between client/server

dfawley:
When looking at adding peer info (address, security info) to headers/trailers, it became awkward dealing with different behaviors around shared types.  Splitting them will allow us to add accessor/helper methods that are more focused for the side we're dealing with, and not to have to use `Option<>` as much.  (E.g. client-side response headers should _require_ peer info and server-side request headers should, too.)


grpc/src/inmemory/mod.rs
```diff
@@ -241,15 +244,20 @@ impl Invoke for InMemoryConnection {
...
         let (trailer_tx, trailer_rx) = oneshot::channel();
 
+        let (method_name, metadata) = headers.into_parts();
+        let server_headers = ServerRequestHeaders::new()
+            .with_method_name(method_name)
+            .with_metadata(metadata);
+
         let call = InMemoryServerCall {
-            headers,
+            headers: server_headers,
```
ejona86, line 260: For my benefit: Is the `headers:` prefixed name important here?

ejona86 (approved):
Yeah, for extensibility having them separate will probably be better. And there's really very little cost to having them separate.

grpc/src/inmemory/mod.rs
dfawley: Discussed offline

grpc/src/client/mod.rs
```diff
@@ -363,3 +365,139 @@ impl<'a> RecvStream for Box<dyn DynRecvStream + 'a> {
...
+    }
+
+    /// Replaces the metadata of self with `metadata`.
+    pub fn with_metadata(mut self, metadata: MetadataMap) -> Self {
+        self.metadata = metadata;
+        self
+    }
+
+    /// Returns the full (e.g. "/Service/Method") method name for these headers.
+    pub fn method_name(&self) -> &String {
```
sauravzg: Probably &str

sauravzg (approved):

grpc/src/client/mod.rs
dfawley: Changed, thanks for catching that.

PR #2793, opened 2026-08-05
Move Address to core module; change Transport::connect to accept &Address

dfawley:
[This contains the 2 commits from #2792; please skip them.  The third and fourth commits represent the actual content of this PR.]

This is part 2 of preparation for adding the peer information to headers/trailers.  Here I'd like to use `Address` in a common `PeerInfo` for both client and server, so moving it to `core` seemed like the best idea for that.  Also, changing `Transport::connect` to accept an `Address` instead of just a string was helpful for populating the `PeerInfo`, so I've made that change here as well -- as a separate commit to make it easier to review.  Since both are pretty simple and related, I combined them into one PR.  If you prefer them to be committed separately I can do that, instead.


grpc/src/core/mod.rs
```diff
@@ -112,3 +119,34 @@ impl dyn RecvMessage + '_ {
         }
     }
 }
+
+/// An Address is an identifier that indicates how to connect to a server.
+#[non_exhaustive]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Ord, PartialOrd)]
+pub struct Address {
```
sauravzg, line 126: So, we discussed a bit about the address type on server side in https://github.com/grpc/grpc-rust/pull/2725 , using an address trait similar to go-ish probably.

This'll likely also be a part of public API, so should we converge on a single type or do these serve different purposes and one should be renamed-ish?

sauravzg: Also, the stuff about debug /display.

dfawley: Actually I would like this to be a trait instead of a struct.  And I think when I tried to do this a long long time ago, it didn't work out well, IIRC in part because I wasn't able to make `Any` a supertrait.  Now that that's stable, I think it's worth revisiting this.  We would need to Box (or Arc?) the trait if we do that, but that's probably OK, too.  This would be getting copied/cloned on every RPC to put it into the peer, so we'd want to be careful.

I'll take a look at this more today if I can.

grpc/src/core/mod.rs
```diff
@@ -112,3 +119,34 @@ impl dyn RecvMessage + '_ {
...
+    /// when connecting to this address.  Typically TCP_IP_ADDRESS_TYPE.
+    pub network_type: &'static str,
+
+    /// The address itself is passed to the transport in order to create a
+    /// connection to it.
+    pub address: ByteStr,
+
+    /// Attributes contains arbitrary data about this address intended for
+    /// consumption by the subchannel.
+    pub attributes: Attributes,
```
ejona86, line 137: We should consider removing attributes from here and moving them to somewhere else in the NR API. But 1:1 move is appropriate in this PR.

ejona86 (approved):

grpc/src/core/mod.rs
dfawley: There are two main options I see here:

1. Just delete it and use endpoint attributes instead
2. Remove it from here and wrap this in a tuple that contains Address+Attributes

But yes I don't think attributes are appropriate in this struct since this will eventually be shared between client and server, and server doesn't want attributes.  (Aside: this might really want to become a trait.)

PR #2794, opened 2026-08-06
fix(tonic-xds): reduce hot-path allocs and config tearing in the routing layer

YutaoMa:
## Motivation

This PR includes a few optimizations I identified while trying to use the routing layer interceptor APIs.

The routing layer loaded the config twice per request with the header-mutation hook in between, so the hook and the matching that consumes its output could see different versions and the request would silently take the default route. The same split also skipped the hook until the first config arrived.

Matching was async only to host that wait, so every request paid a nested boxed future and the owned copies a `'static` future forces; the second load handed the hook an owned metadata clone, and the config `Arc` stayed alive for the whole request rather than just the decision.

## Solution

One required trait method returning either an available config or a future for the first one. Matching borrows it and turns synchronous, dropping a nested boxed future, an authority `String`, a `HeaderMap` clone and a metadata deep clone; block-scoped so the `Arc` releases before the inner call.


tonic-xds/src/client/route.rs
```diff
@@ -115,17 +140,19 @@ where
...
-                interceptor.on_request(request.headers_mut(), &metadata);
-            }
-            let headers = &request.headers();
-            let route_input = RouteInput {
-                authority: &authority,
-                headers,
+            // Scoped so the config `Arc` is dropped before the inner call:
+            // to avoid holding that config version for the life of a long-running stream.
+            let route_decision = {
+                let config = router.acquire().get().await?;
```
mingley, line 146: This snapshot can outlive the CDS/EDS entries for the cluster it selects. During an A -> B route update, `reconcile_clusters` removes A immediately; if this is the first RPC to A, `XdsLbService` then lazily creates a fresh `watch_cluster("A")` after removal. That watch has no producer, so `channel.ready()` can wait indefinitely. I reproduced this with a barrier test: the PR head stalls, while `master` routes the same schedule through B. Please keep the route generation dependencies alive through LB resolution, or otherwise defer retirement until no in-flight snapshot can select them.

tonic-xds/src/xds/routing.rs
```diff
@@ -1122,15 +1126,15 @@ mod tests {
...
-        };
-        // The router now blocks waiting for config; verify it returns
-        // NotReady after the timeout elapses.
-        let result =
-            tokio::time::timeout(std::time::Duration::from_millis(100), router.route(&input)).await;
+        // With no config yet, `acquire` yields the waiting variant, which
+        // blocks and then reports NotReady.
+        assert!(router.snapshot().is_none());
+        assert!(matches!(router.acquire(), AcquiredConfig::Pending(_)));
+        let result = tokio::time::timeout(
```
mingley: This test cannot exercise the inner `RoutingError::NotReady`: the outer timeout is 100 ms while `DEFAULT_READY_TIMEOUT` is 30 seconds, and the test explicitly accepts the outer `Elapsed`. It would still pass if the pending future never returned. Please use paused Tokio time, advance by `DEFAULT_READY_TIMEOUT`, and assert exactly `Err(RoutingError::NotReady)`.

tonic-xds/src/client/route.rs
YutaoMa: This is a known issue because of lack of support for proper cluster removal logic, see: https://github.com/grpc/grpc-rust/pull/2722#issuecomment-5098838006. It will be addressed in the separate PR.

tonic-xds/src/xds/routing.rs
YutaoMa: Good catch, I've removed the outer timeout and updated the assertion to be on the `NotReady` error.

ankurmittal (approved):

PR #2799, opened 2026-08-12
grpc/attributes: relax ordering requirement for values

arjan-bal:
Attributes only need to be comparable for equality; the ordering requirement was unnecessary and unused. This PR removes it.


dfawley (approved):
LGTM thanks!

PR #2802, opened 2026-08-13
grpc: add ConnectionInfo to client's ResponseHeaders and Trailers

dfawley:
Also: return `ConnectionInfo` from `Transport::connect` instead of `SecurityInfo` _(yes, I know I just changed it to produce `SecurityInfo`)_ so that the subchannel can set the full `ConnectionInfo` in Trailers when producing errors locally, since the address of the subchannel may be relevant to those errors.

FWIW I wanted to make the subchannel apply an interceptor that could always set `ConnectionInfo` in response streams (either `ResponseHeaders` or in `Trailers`-only responses), but because the field is mandatory in `ResponseHeaders`, the transport would need to set it to some default/empty value, which felt wrong.  Anything producing a `ResponseHeaders` or `Trailers` should ensure the `ConnectionInfo` is set correctly.

cc @joshuatants as FYI - this is the change that will provide you with the ability to determine the server's address.

### DIFF
--- a/grpc/src/client/mod.rs
+++ b/grpc/src/client/mod.rs
@@ -443,7 +443,7 @@ impl RequestHeaders {
     }
 
     /// Returns the full (e.g. "/Service/Method") method name for these headers.
-    pub fn method_name(&self) -> &String {
+    pub fn method_name(&self) -> &str {
         &self.method_name
     }
 
--- a/grpc/src/client/mod.rs
+++ b/grpc/src/client/mod.rs
@@ -53,6 +53,7 @@ use std::time::Instant;
 
 use tonic::async_trait;
 
+use crate::core::PeerInfo;
 use crate::core::RecvMessage;
 use crate::core::SendMessage;
 use crate::metadata::MetadataMap;
@@ -367,15 +368,19 @@ impl<'a> RecvStream for Box<dyn DynRecvStream + 'a> {
 }
 
 /// Contains all information transmitted in the response headers of an RPC.
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub struct ResponseHeaders {
     metadata: MetadataMap,
+    peer_info: PeerInfo,
 }
 
 impl ResponseHeaders {
     /// Returns a default ResponseHeaders instance.
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(peer_info: PeerInfo) -> Self {
+        Self {
+            metadata: MetadataMap::default(),
+            peer_info,
+        }
     }
 
     /// Replaces the metadata of self with `metadata`.
@@ -396,6 +401,17 @@ impl ResponseHeaders {
 
     pub(crate) fn into_metadata(self) -> MetadataMap {
         self.metadata
+    }
+
+    /// Replaces the peer_info of self with `peer_info`.
+    pub fn with_peer_info(mut self, peer_info: PeerInfo) -> Self {
+        self.peer_info = peer_info;
+        self
+    }
+
+    /// Replaces the peer_info of self with `peer_info`.
+    pub fn peer_info(&self) -> &PeerInfo {
+        &self.peer_info
     }
 }
 
@@ -427,7 +443,7 @@ impl RequestHeaders {
     }
 
     /// Returns the full (e.g. "/Service/Method") method name for these headers.
-    pub fn method_name(&self) -> &str {
+    pub fn method_name(&self) -> &String {
         &self.method_name
     }
 
@@ -454,6 +470,7 @@ impl RequestHeaders {
 pub struct Trailers {
     status: crate::Result<()>,
     metadata: MetadataMap,
+    peer_info: Option<PeerInfo>,
 }
 
 impl Trailers {
@@ -462,6 +479,7 @@ impl Trailers {
         Self {
             status,
             metadata: MetadataMap::default(),
+            peer_info: None,
         }
     }
 
@@ -495,6 +513,23 @@ impl Trailers {
     /// Returns the status in the [`Trailers`], consuming the entire status.
     pub fn into_status(self) -> crate::Result<()> {
         self.status
+    }
+
+    /// Replaces the peer info in self with `peer_info`.
+    pub fn with_peer_info(mut self, peer_info: Option<PeerInfo>) -> Self {
+        self.peer_info = peer_info;
+        self
+    }
+
+    /// Returns the peer info in the trailers, if present.  Peer information
+    /// will not be available in trailers in any the following circumstances:
+    ///
+    /// 1. A ResponseHeaders was already present on the response stream.
+    ///
+    /// 2. The error was generated locally on the client before a connection was
+    ///    chosen for the RPC.
+    pub fn peer_info(&self) -> &Option<PeerInfo> {
+        &self.peer_info
     }
 
     pub(crate) fn into_parts(self) -> (crate::Result<()>, MetadataMap) {
--- a/grpc/src/client/mod.rs
+++ b/grpc/src/client/mod.rs
@@ -53,7 +53,7 @@ use std::time::Instant;
 
 use tonic::async_trait;
 
-use crate::core::PeerInfo;
+use crate::core::ConnectionInfo;
 use crate::core::RecvMessage;
 use crate::core::SendMessage;
 use crate::metadata::MetadataMap;
@@ -371,12 +371,12 @@ impl<'a> RecvStream for Box<dyn DynRecvStream + 'a> {
 #[derive(Debug, Clone)]
 pub struct ResponseHeaders {
     metadata: MetadataMap,
-    peer_info: PeerInfo,
+    peer_info: ConnectionInfo,
 }
 
 impl ResponseHeaders {
     /// Returns a default ResponseHeaders instance.
-    pub fn new(peer_info: PeerInfo) -> Self {
+    pub fn new(peer_info: ConnectionInfo) -> Self {
         Self {
             metadata: MetadataMap::default(),
             peer_info,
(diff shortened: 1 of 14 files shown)


grpc/src/client/transport/tonic/mod.rs
```diff
@@ -188,39 +192,24 @@ impl Invoke for TonicTransport {
             TonicRecvStream {
                 state: StreamState::AwaitingHeaders(resp_rx),
                 cancel_tx: Some(cancel_tx),
+                peer_info: Some(self.peer_info.clone()),
```
ejona86: Do we really want to clone the entire thing? I see we're cloning the entire thing for CallCredentials already. Are we going to Arc things inside Attributes instead, like the client's certificate?

grpc/src/core/mod.rs
```diff
@@ -150,3 +151,59 @@ impl Display for Address {
         write!(f, "{}:{}", self.network_type, self.address.to_string())
     }
 }
+
+/// Information about the connection to the RPC's peer (from the client/server
+/// pair).
+#[derive(Debug, Clone)]
+pub struct PeerInfo {
+    local_address: Address,
```
ejona86, line 159: Local address is in the peer's information?

ejona86 (commented):
Didn't have time to look at it deeply.

grpc/src/core/mod.rs
dfawley: I did :eyes: at this as well.  `PeerInfo`->`ConnectionInfo`?  We call it `Peer` in Go and have the local address there (just FYI, not as justification; it happened that we added it later there): https://pkg.go.dev/google.golang.org/grpc/peer#Peer

grpc/src/client/transport/tonic/mod.rs
dfawley: `Attributes` is a linked list with `Arc`'d nodes, so a clone of `PeerInfo` is _fairly_ small:

- Two `Address`es, each with `&'static`, `ByteStr` (contains `Bytes` whose data is `Arc`'d), `Attributes` (`Arc`'d)
- One `SecurityInfo` with `&'static`, `SecurityLevel` (3 value enum with no data), and `Attributes` (`Arc`'d)

I do want to delete the `Attributes` from `Address`, so it would be down to two `Arc`s: the `ByteStr` and `SecurityInfo`.

The tradeoff would be simplicity for performance.  If `PeerInfo` needed to be mutated in the stack somewhere, it might be worse if it was `Arc`'d.  But, I wouldn't expect that to happen, so maybe it's best in its own `Arc`?

I think we could change this in a hidden/backward compatible way if we wanted to convert it to hold `Arc<InnerPeerInfo>`.  The accessors just hand out references, which could be references to the value inside the `Arc`, and `PeerInfo` is immutable.

grpc/src/server/mod.rs
```diff
@@ -535,6 +542,17 @@ impl RequestHeaders {
         &mut self.metadata
     }
 
+    /// Replaces the peer_info of self with `peer_info`.
+    pub fn with_peer_info(mut self, peer_info: PeerInfo) -> Self {
+        self.peer_info = peer_info;
+        self
+    }
```
arjan-bal: Under what circumstances would we need to update peer info after creating the response headers?

grpc/src/server/mod.rs
```diff
@@ -521,7 +528,7 @@ impl RequestHeaders {
     }
 
     /// Returns the full (e.g. "/Service/Method") method name for these headers.
-    pub fn method_name(&self) -> &str {
+    pub fn method_name(&self) -> &String {
```
arjan-bal: Why change the return type? It's idiomatic to return `&str` since `&Str` can automatically coerce into a `&str`. `&str` can also represent a `&'static str` to avoid heap allocations.

grpc/src/server/mod.rs
```diff
@@ -494,18 +495,24 @@ impl ResponseHeaders {
...
     metadata: MetadataMap,
+    /// Information about the client.
+    peer_info: PeerInfo,
 }
 
 impl RequestHeaders {
     /// Returns a default RequestHeaders instance.
-    pub fn new() -> Self {
-        Self::default()
+    pub fn new(method_name: impl Into<String>, peer_info: PeerInfo) -> Self {
```
arjan-bal: Do we need the headers to contain peer info as well? Since trailers are always present, wouldn't adding it there be sufficient?

grpc/src/inmemory/mod.rs
```diff
@@ -386,9 +392,10 @@ impl ClientRecvStream for InMemoryClientRecvStream {
                     match trailer_rx.await {
                         Ok(trailers) => {
                             let (status, metadata) = trailers.into_parts();
-                            return ResponseStreamItem::Trailers(
-                                ClientTrailers::new(status).with_metadata(metadata),
-                            );
+                            let mut client_trailers =
+                                ClientTrailers::new(status).with_metadata(metadata);
+                            client_trailers = client_trailers.with_peer_info(self.peer_info.take());
```
arjan-bal: nit: We can chain the `with_peer_info` call in the statement above and avoid the `mut`.

grpc/src/client/mod.rs
```diff
@@ -497,6 +515,23 @@ impl Trailers {
         self.status
     }
 
+    /// Replaces the peer info in self with `peer_info`.
+    pub fn with_peer_info(mut self, peer_info: Option<PeerInfo>) -> Self {
```
arjan-bal: Does this accept an `Option` just for ergonomics, or is there a use-case for unsetting the peer info?

grpc/src/inmemory/mod.rs
```diff
@@ -369,13 +373,15 @@ impl Drop for InMemoryClientSendStream {
...
     trailer_rx: Option<oneshot::Receiver<ServerTrailers>>,
+    peer_info: Option<PeerInfo>,
 }
 
 impl ClientRecvStream for InMemoryClientRecvStream {
     async fn recv(&mut self, msg: &mut dyn RecvMessage) -> ResponseStreamItem {
         match self.rx.recv().await {
             Some(InMemoryResponseStreamItem::Headers(h)) => ResponseStreamItem::Headers(
-                ClientResponseHeaders::new().with_metadata(h.into_metadata()),
+                ClientResponseHeaders::new(self.peer_info.take().unwrap())
```
arjan-bal: nit: Maybe add a comment explaining why calling `unwrap` is safe here?

grpc/src/client/transport/tonic/mod.rs
```diff
@@ -278,32 +315,40 @@ impl RecvStream for TonicRecvStream {
...
-                                &TonicMeta::default(),
-                            )
+                            let Some(peer_info) = self.peer_info.take() else {
+                                return self.trailers_from_grpc_result(
+                                    Err(StatusError::new(
+                                        StatusCodeError::Internal,
+                                        "required peer info missing",
+                                    )),
+                                    None,
+                                );
```
arjan-bal, line 325: Is this possible in the absence of a bug? I believe the in-memory transport is panicking in this case.

grpc/src/server/mod.rs
dfawley: Probably none, but this is here for interceptors that need to do this and for testing.

grpc/src/server/mod.rs
dfawley: I don't understand how I keep messing this up.  I'm just bad at git merges, sorry.  I fixed the other similar instance of &String in the client, too.

grpc/src/server/mod.rs
dfawley: Yes; the application will need access to it during the RPC in some circumstances.

grpc/src/client/mod.rs
dfawley: I did it purely for ergonomics reasons (there are several places I was otherwise doing `if let Some(peer) = ... { t.with_peer_info(peer) }`), though if you're doing interception or testing you might want to clear it.

grpc/src/inmemory/mod.rs
dfawley: Added a comment.  I could easily see wanting to return an error here instead of panicking, but the recv stream here doesn't track its state, and probably wants a bit of a larger refactor.

grpc/src/inmemory/mod.rs
dfawley: Thanks; done.

grpc/src/core/mod.rs
dfawley: Renamed

grpc/src/client/transport/tonic/mod.rs
ejona86: I wasn't just concerned with Arcs; I had just mentioned it because Attributes can hold some quite large values. I didn't realize Attributes is already an Arc; that makes this mostly fine.

Being able to make it Arc later is a good point, as long as the API remains like it is. If Address becomes a trait, we'll probably want it to be Arc'ed to avoid the useless allocations.

grpc/src/core/mod.rs
```diff
@@ -150,3 +151,59 @@ impl Display for Address {
...
+/// pair).
+#[derive(Debug, Clone)]
+pub struct ConnectionInfo {
+    local_address: Address,
+    remote_address: Address,
+    security_info: SecurityInfo,
+}
+
+impl ConnectionInfo {
+    /// Constructs a new PeerInfo with the given fields.
```
arjan-bal: nit: The comment still mentions `PeerInfo`.

arjan-bal (approved):
Some private fields and a comment still use "peer" instead of "connection". Otherwise, LGTM.

grpc/src/core/mod.rs
dfawley: Fixed

dfawley (commented):
Fixed _all_ the lingering "peer" references.

arjan-bal (approved):
LGTM

PR #2804, opened 2026-08-13
fix(tonic-xds): match routes on request path and anchor safe_regex

YutaoMa:
## Motivation

Two route-matching bugs were discovered while testing `tonic-xds::RoutingLayer`. Previously because the routing config wasn't wired in, the buggy path wouldn't be exercised in production. Now they are surfaced and needs fixing:

1. `RoutingLayer` used to fetch `:path` header from the `HeaderMap`, but that pseudo header isn't actually inserted there. (Rust `http::HeaderMap` does not allow storing pseudo headers).
2. The `StringMatcher::SafeRegex` does partial match, while the spec mandates full match (see https://github.com/envoyproxy/envoy/blob/743baafef00f4f8d5fc234e6ee5cb7ed87cba148/api/envoy/type/matcher/v3/regex.proto#L66)

## Solution

- Change to extract the path from the request URI, with the query string excluded as the matcher requires.
- `safe_regex` patterns are now anchored as `\A(?:{})\z` in the new type `SafeRegex`. We enforce full match by only exposing that matching function from the type. For debuggability the `Debug` implementation does print the original unanchored pattern.

### DIFF
--- a/tonic-xds/src/xds/resource/safe_regex.rs
+++ b/tonic-xds/src/xds/resource/safe_regex.rs
@@ -1,0 +1,128 @@
+/*
+ *
+ * Copyright 2025 gRPC authors.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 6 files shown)


tonic-xds/src/xds/resource/safe_regex.rs
```diff
@@ -0,0 +1,137 @@
...
+
+impl SafeRegex {
+    /// Compile `pattern` to match only the entire input.
+    ///
+    /// The non-capturing group keeps a top-level alternation from escaping the
+    /// anchors and confines any inline flags the pattern sets; `\A`/`\z` hold
+    /// regardless of those flags, whereas `^`/`$` become line anchors under
+    /// `(?m)`.
+    pub(crate) fn new(pattern: &str) -> Result<Self, regex::Error> {
+        Regex::new(&format!("{ANCHOR_PREFIX}{pattern}{ANCHOR_SUFFIX}")).map(Self)
```
gu0keno0, line 58: Validate the regex first so that injection such as "foo)|bar(?:" will not cause loopholes

tonic-xds/src/xds/resource/safe_regex.rs
```diff
@@ -0,0 +1,137 @@
...
+    #[test]
+    fn inline_flags_cannot_widen_the_anchors() {
+        // A control plane could set `(?m)`; the anchors must still bind to the
+        // whole haystack rather than to a line within it.
+        let re = SafeRegex::new("(?m)/a").unwrap();
+        assert!(!re.is_match("x\n/a\ny"));
+    }
+
+    #[test]
+    fn an_invalid_pattern_is_rejected() {
```
gu0keno0, line 126: Can add to this test as well.

gu0keno0 (commented):
One comment about regex safety, otherwise LGTM

tonic-xds/src/xds/resource/safe_regex.rs
YutaoMa: Good catch, added a pre-anchoring compile so patterns with unbalanced parens get rejected.

tonic-xds/src/xds/resource/safe_regex.rs
YutaoMa: Added tests for the above cases.

gu0keno0 (approved):

Issue #2790, opened 2026-08-04
Single Threaded Tonic Client/Server


artsiomkaltovich:
### Motivation

Tonic requires Send + Sync for the client code, that make it hard to introduce to single-threaded async code. Not every service should be multi-threaded and now adding tonic in single threaded code could be painful.
This topic was raised several times:

Single threaded server example with async_trait(?Send) (open): [https://github.com/grpc/grpc-rust/issues/2171](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
how to generate trait with #[async_trait(?Send)] (closed): [https://github.com/grpc/grpc-rust/issues/844](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
feat: Support thread-local functionality (closed PR): [https://github.com/grpc/grpc-rust/pull/1505](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)
add Single-threaded runtime (closed): [https://github.com/grpc/grpc-rust/issues/784](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

however, it wasn't fully implemented yet.

### Proposal

I think ability to run tonic natively in single-threaded without send+sync addition would benefit the community a lot.

To be honest I haven't check the src yet. There were suggestions to replace `async_trait` with `async_trait(?Send)` another approach is code split and introducing local versions for code (with/or without feature flag).

sauravzg:
Hi, This is not something that we have on our current grpc-rust roadmap or bandwidth to explore/prioritise this.

@LucioFranco  It seems like something that's been brought up a few times for tonic. Is this something we want to address?

artsiomkaltovich:
I've just checking if this is an option or just strict no.

I could try to implement it, however I am not very familiar with a project, so I can't guarantee anything. :)
Any insights welcomed.

sauravzg:
Feasibility , design, tradeoff analysis and implementation would require some effort, that we currently don't have bandwidth to put into for the near future.

Issue #2171, opened 2025-02-01
Single threaded server example with `async_trait(?Send)`


cavemanloverboy:
There are zero examples of anything using `#[tonic::async_trait(?Send)]`. Can we fix this?

harry0349:
It seems single-threaded server not implemented.
https://github.com/hyperium/tonic/blob/78be69e92f297929d8f711d1927e4bdda8c324da/tonic-build/src/server.rs#L224

Eclextic:
I also am interested. Using tonic is so horrible when you want to use a different async framework like monoio that doesn't require Send.

Issue #844, opened 2021-11-23
how to generate trait with #[async_trait(?Send)]


zojw:
## Feature Request

Hi, I'm investigating how to use [tokio-uring](https://github.com/tokio-rs/tokio-uring) to read/write file system in a GRPC server.

tokio-uring [return future always `!Send`](https://github.com/tokio-rs/tokio-uring/pull/1/files#diff-3dc5dd454e080eb849ee5efaf79df2585fbe0a06804d69249b4c81da05a63875R138)

so the first question, I meet is #[async_trait] need change to [#[async_trait(?Send)]](https://docs.rs/async-trait/0.1.51/async_trait/#non-threadsafe-futures) and current_thread runtime.

but it seems tonic generate code always be #[async_trait]

https://github.com/hyperium/tonic/blob/master/tonic-build/src/server.rs#L189

do we have any advice about this~? or do we have the advice to integrate tonic with tokio-uring? 😄 

### Crates

- tonic-build
- not sure whether "current_thread runtime" can works well with other tonic crate too - - 

### Motivation

read_at a file by tokio-uring(or other io-uring libs) in tonic server-side

### Proposal

n/a

### Alternatives

n/a

davidpdrsn:
Tonic currently doesn't support `!Send` futures so generating the traits with `#[async_trait(?Send)]` isn't supported either. Not quite sure if its feasible to support both `Send` and `!Send` futures 🤔

As far as workarounds, honestly not sure. Maybe someone in `#tokio-users` in the tokio discord is able to give advice?

zojw:
for my situation, it seems I can use more simple network protocol and raw TCP to workaround, so close this first :)

PR #2806, opened 2026-08-14
Single-threaded (`!Send`) client & server: `tonic::local`

artsiomkaltovich:
Prototype for #2790 (see also #2171, #844, #1505, #784 — the same request keeps coming back).

## Motivation

Tonic's generated servers require `Send + Sync + 'static` handlers, and the client path boxes
futures/bodies as `Send`. That makes tonic hard to adopt in single-threaded async code: anything
holding `Rc`, `RefCell`, or a non-`Send` third-party handle has to be redesigned or wrapped, even
though the service will only ever run on one thread. Not every service needs a multi-threaded
runtime — a `current_thread` runtime is often the right tool (lower latency variance, no
cross-thread synchronisation, simpler state).

This PR makes that case work natively, without touching the multi-threaded path.

## What this adds

- `tonic::local::` — a `!Send` client, server, router, body, streaming and a minimal HTTP/2
  transport, all driven by `tokio::task::spawn_local`.
- Codegen support: `tonic_prost_build::Builder::local(true)` (and `CodeGenBuilder::local`) emits
  `#[async_trait(?Send)]` service traits, `Rc<T>` handlers and non-`Send` futures.
- Works on either a `tokio::task::LocalSet` (any tokio 1.x) or `tokio::runtime::LocalRuntime`
  (stable since tokio 1.51). No tokio version bump: the library only uses `spawn_local`.

## Design

**Feature-gated, additive, zero changes to existing public API.**

- Runtime: `local` (pure — no new dependencies, wasm-friendly) and `local-transport`
  (adds the hyper/tokio bits for the provided server + channel).
- Codegen: a `local` feature on `tonic-build` / `tonic-prost-build` gates the `local()` builder
  method, mirroring the existing `transport`-feature precedent.
- **No existing `pub` item changed signature or behaviour.** Everything new lives under
  `tonic::local::`; the default build is byte-identical in API terms.
- Duplication is kept out of the way by sharing the heavy logic through *private* generics:
  `BodyKind<B>`, `StreamingImpl<T, DecoderStore, B>`, `classify_response`,
  `set_grpc_response_headers`, `ServerGrpcConfig`. The `Send` and `!Send` types are thin wrappers
  over the same cores, so the decode state machine, framing, compression and codec handling exist
  once. Decode benches are at parity with the core path.
- Errors stay `Send + Sync` (`Status`, `BoxError`) — required by hyper's HTTP/2 signatures, and
  harmless: `Send` on an error *value* never constrains handlers, futures or state.

## Trade-offs of this approach

**Pros**
- Zero risk to existing users: nothing on the multi-threaded path changes, no semver impact.
- Opt-in at both layers (cargo feature + codegen flag).
- The `!Send` types are honestly `!Send` — misuse is a compile error, not a runtime surprise.
- Heavy logic is shared, so bug fixes land in one place.

**Cons**
- The `Grpc<T>` method bodies exist twice (client ~230 lines, server ~290): identical code whose
  *only* difference is `Send`/`'static` tokens in `where` clauses. See below for why.
- Two parallel type families to learn (`tonic::Streaming` vs `tonic::local::Streaming`, …).
- The local transport is deliberately minimal: HTTP/2 only, no TLS, no graceful shutdown, no
  reconnect/balancing. Enough for a prototype, not feature parity. (Good enough for prototype).

## Alternatives considered

1. **Full duplication (a parallel `local` implementation).** Simplest to write and to reason about
   in isolation, but every fix has to be applied twice and the two copies drift. Rejected as the
   primary strategy — though see the point below: for the `Grpc` method bodies it is what remains.
2. **Unification via traits ("traitification"): one implementation generic over Send-ness.**
   The attractive option, and the one we probed hardest. It does not work on stable Rust: a shared
   trait seam must fix its method bounds once for all impls, so the core impl's internal `Send`
   requirements (`Body::new`, `Streaming::new_response`) would leak into the trait signature and
   force `Send` back onto the local call sites — defeating the feature. There is no way to make a
   bound conditionally present per instantiation. A `macro_rules!` template could cut ~150 lines,
   but at the cost of tt-munged `where` clauses and materially worse compiler errors: a bad trade
   for code that is otherwise plainly readable and diffable. What *could* be shared has been, as
   private generics (above).
3. **Make the existing types generic over their storage** (e.g. `Streaming<T, Store = …>`).
   Removes most duplication, but changes tonic's core public types and their inference/rustdoc
   identity — too invasive for a prototype, and a call for maintainers rather than a contributor. (To be honest, I prefer this, but it is you to decide)
4. **A defaulted streaming parameter on the two core service traits** —
   `ClientStreamingService<R, St = Streaming<R>>` (same for `StreamingService`). Backward
   compatible (the default keeps every existing impl, including generated ones, compiling
   unchanged) and would delete ~58 lines plus a codegen branch, letting `local` reuse the core
   traits. **Deliberately not applied here** — it modifies public traits, so it is offered as a
   ready follow-up if you want it.

## Example / tests

End-to-end echo service on a `current_thread` runtime:
`tests/local/` — [`proto/echo.proto`](https://github.com/grpc/grpc-rust/pull/2806/changes#diff-b39ddf99715ccb4c1f5c832b0b3c95cadd59231174721ff203762c6569d5b8b8),
[`tests/local.rs`](https://github.com/grpc/grpc-rust/pull/2806/changes#diff-cbd3d8ca056eb85b1512073d280c505ec119da41d8d84921e70978de685e5d80).

It covers all four RPC shapes with `Rc<RefCell<_>>` handler state and non-`Send` streams in both
directions, an interceptor round-trip, an unimplemented-path check, and the same flow on
`LocalRuntime` with no `LocalSet`. Static assertions pin the guarantee that the local types are
`!Send`, so `tokio::spawn` of them cannot compile.

## Status

All existing tests pass unchanged; new unit + e2e + codegen tests added; rustdoc clean under
`-D warnings`; feature combinations (`local` alone, `local-transport` alone, no-default-features)
all build. Happy to split this into smaller PRs, or to adopt alternatives above, if that suits
review better.


artsiomkaltovich:
Hello @nathanielford. Thank you for your reply.

Could please clarify what you mean by "We are halting development on tonic, in favor of the grpc-rust library being built."?

nathanielford:
Sure, @artsiomkaltovich! There is more information on the [grpc.io blog](https://grpc.io/blog/grpc-welcomes-tonic/) but Tonic has joined gRPC and we are building an official gRPC-compliant library, including a bevy of features that were previously unavailable. This can be found in the `grpc/` subdirectory of this repo. This does mean is that most of our development effort is going to the `grpc-rust` crate, and while Tonic will remain available it's not clear if we will have another release. We will attempt to triage outstanding issues and PRs in flight on a best-effort basis, but its for that reason that it may be a minute before we can attend to this one, particularly due to the impact and size.

PR #2807, opened 2026-08-14
fix/silence all clippy warnings

dfawley:
cc @YutaoMa in this PR I deleted the tonic-xds tests that were assertions of constants, as that also causes clippy to be unhappy, and it felt unnecessary to keep the tests and silence the warnings.  Let me know if you feel strongly otherwise and I can bring them back with a `#[allow(clippy::...)]`.

### DIFF
--- a/examples/src/grpc-gcp/client.rs
+++ b/examples/src/grpc-gcp/client.rs
@@ -22,7 +22,7 @@
  *
  */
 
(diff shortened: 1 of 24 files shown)


YutaoMa (approved):

arjan-bal (approved):
LGTM

PR #2810, opened 2026-08-17
grpc-protobuf: remove redundant stream validator for server streaming RPCs

arjan-bal:
`GrpcStreamingResponse::new()` already wraps the `RecvStream` in a `RecvStreamValidator`.

https://github.com/grpc/grpc-rust/blob/e6ccfc41b642880454ab3e1cb2003e66af8b8724/grpc-protobuf/src/client/mod.rs#L122-L128


dfawley (approved):
Good catch thanks!

PR #2811, opened 2026-08-17
feat(tonic-xds): add transport-generic build_transport_channel

LYZJU2019:
## Motivation

The xDS channel was hard-wired to the gRPC transport, so an out-of-tree client cannot reuse its routing/retry/load-balancing stack over a different transport (e.g. HTTP).

## Solution

Adds `XdsChannelBuilder::build_transport_channel`, generic over the request/response body, which wires a caller-supplied `MakeConnector` whose endpoint service consumes the retry-buffered `Request<SharedBody<B>>`; it also adds the `XdsTransportChannel<B, ResBody>` alias and re-exports `SharedBody`. The gRPC and transport-generic paths now funnel through a single generic `build_channel` helper, with the gRPC entry point a thin specialization that supplies the tonic connector and the `SharedBody -> TonicBody` body reconstruction, keeping it behavior-preserving. Building on the now-merged retry-classifier generalization (#2812), a non-gRPC transport can install its own `RetryClassifierFactory` via `with_retry_classifier_factory` for per-route status-code retries, otherwise falling back to connection-error retry.

### DIFF
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -110,6 +110,7 @@ allowed_external_types = [
     "bytes::*",
     "http::*",
     "http_body::*",
+    "shared_http_body::*",
 
     # not major released
     "prost::*",
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -114,7 +114,7 @@ allowed_external_types = [
     # not major released
     "prost::*",
     "opentelemetry::*",
-    "shared_http_body::SharedBody",
+    "shared_http_body::shared_body::SharedBody",
 
     "tower_service::Service",
     "tower::BoxError",
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -114,7 +114,7 @@ allowed_external_types = [
     # not major released
     "prost::*",
     "opentelemetry::*",
-    "shared_http_body::*",
+    "shared_http_body::SharedBody",
 
     "tower_service::Service",
     "tower::BoxError",
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -110,11 +110,11 @@ allowed_external_types = [
     "bytes::*",
     "http::*",
     "http_body::*",
-    "shared_http_body::*",
 
     # not major released
     "prost::*",
     "opentelemetry::*",
+    "shared_http_body::*",
 
     "tower_service::Service",
     "tower::BoxError",
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -417,13 +417,19 @@ impl XdsChannelBuilder {
     /// Wires the shared routing / retry / load-balancing stack from a pre-built
     /// [`XdsRuntimeParts`] and a caller-chosen connector, type-erasing it into an
     /// [`XdsTransportChannel`]. Both the gRPC and transport-generic entry points
-    /// funnel through here; only the connector and the `to_endpoint_req` body
-    /// reconstruction differ per transport.
+    /// funnel through here; the connector, the `to_endpoint_req` body
+    /// reconstruction, and the `fallback_retry` config differ per transport.
+    ///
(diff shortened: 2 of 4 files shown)


tonic-xds/src/client/channel.rs
```diff
@@ -371,104 +405,192 @@ impl XdsChannelBuilder {
...
+    /// # Retry semantics
+    ///
+    /// Per-route retry policies are compiled by the [`RetryClassifierFactory`]
+    /// set via
+    /// [`with_retry_classifier_factory`](Self::with_retry_classifier_factory).
+    /// The default gRPC factory only produces status-code retries for gRPC
+    /// responses, so a non-gRPC transport supplies its own factory to interpret
+    /// `retry_on` (e.g. `5xx`, `gateway-error`) for that transport. Without a
+    /// custom factory only connection-error retries take effect.
+    pub fn build_transport_channel<MC, B, ResBody>(
```
YutaoMa, line 580: Add a TODO comment explaining that because we haven't exposed cert registry yet, user may not be able to implement transport channel with proper mtls support yet.

tonic-xds/src/client/channel.rs
```diff
@@ -371,104 +405,192 @@ impl XdsChannelBuilder {
...
+    where
+        MC: MakeConnector,
+        MC::Service: Service<
+                Request<ReqBody>,
+                Response = Response<ResBody>,
+                Error: Into<BoxError>,
+                Future: Send + 'static,
+            > + Load<Metric: Debug>
+            + Send
+            + 'static,
```
YutaoMa: `MakeConnector::Service` already includes Send + 'static bound, can remove it here.

tonic-xds/Cargo.toml
```diff
@@ -110,6 +110,7 @@ allowed_external_types = [
     "bytes::*",
     "http::*",
     "http_body::*",
+    "shared_http_body::*",
```
YutaoMa: move to the section below, this is "not major released"

YutaoMa: also only allow the `SharedBody` type

tonic-xds/src/client/channel.rs
```diff
@@ -371,104 +405,192 @@ impl XdsChannelBuilder {
...
-            Some(factory) => XdsRouter::with_retry_factory(&cache, factory),
-            None => XdsRouter::new(&cache),
+            Some(factory) => XdsRouter::with_retry_factory(&parts.cache, factory),
+            None => XdsRouter::new(&parts.cache),
         });
 
         // Fallback retry config for requests that carry no per-route config
         // (non-xDS callers, or a route with no retry policy). Per-route configs
         // come from RDS via the request's `RouteDecision`; see `RetryLayer`.
         let retry_layer = RetryLayer::new(Arc::new(RetrySharedConfig::grpc_default()));
```
YutaoMa: Need a non-grpc-specific default retry classifier here: the default empty retry_on and default retry of connection errors are fine for other transport (e.g. HTTP traffic) but the grpc classifier inserts "grpc-previous-rpc-attempts" header on those connection retry which can be confusing to user. Better to make the default just do the connection level retry, and grpc layer has its own default that includes the header.

tonic-xds/src/client/channel.rs
LYZJU2019: Addressed in 95d30a5f07151a9ee751f6ee2f53c9a502d43e85

tonic-xds/src/client/channel.rs
LYZJU2019: Addressed in 95d30a5f07151a9ee751f6ee2f53c9a502d43e85

tonic-xds/Cargo.toml
LYZJU2019: Addressed in 95d30a5f07151a9ee751f6ee2f53c9a502d43e85

tonic-xds/src/client/channel.rs
LYZJU2019: Addressed in 95d30a5f07151a9ee751f6ee2f53c9a502d43e85

YutaoMa (approved):

PR #2813, opened 2026-08-18
grpc/server: fork client call options

arjan-bal:
Currently, the `server::Handle::handle()` method accepts `client::CallOptions`. However, server-side options are distinct from client-side options and should not share the same struct. This PR introduces a dedicated `server::CallOptions` struct for the server APIs.


sauravzg (approved):

dfawley (commented):
Sorry, I missed this one yesterday..  On the client, call options allow the application to configure the call.  On the server, though, this doesn't really make sense, and the current struct is empty.  Should this concept just be removed entirely instead?  I think we were going to pass all the out-of-band stuff related to the call (e.g. the deadline) via a task-local context?

PR #2817, opened 2026-08-19
grpc-protobuf: move client exports to a submodule 

arjan-bal:
## Why
If both client and server symbols are exported directly from the crate root, it can lead to name collisions. Client symbols must now be imported from `grpc_protobuf::client` and server symbols will be exported from `grpc_protobuf::server`.


grpc-protobuf/src/lib.rs
```diff
@@ -33,12 +33,12 @@
 //! There are four basic RPC (Remote Procedure Call) types and a corresponding
 //! call builder for each.
```
dfawley: This is all assuming client-side stuff only and will want to be rewritten at least when we add the server symbols, or now if you want to do it.  We can either document all the client and server stuff together in here, or we can move this into the client module and leave something else useful behind for the top level.

dfawley (approved):

grpc-protobuf/src/lib.rs
arjan-bal: Moved the docs to the `client` module and added a `Modules` section in its place, [similar to the grpc crate](https://docs.rs/grpc/latest/grpc/#modules).

grpc-protobuf/src/client/mod.rs
```diff
@@ -22,6 +22,23 @@
  *
  */
 
+//! Types and call builders for client-side RPCs.
```
dfawley: Super-nit: there isn't really a thing that is a "client-side RPC".  I'd probably do:

```rust
//! Client-side types and call builders for RPCs (Remote Procedure Calls).
```

grpc-protobuf/src/client/mod.rs
```diff
@@ -22,6 +22,23 @@
...
+//!
+//! # Basic usage
+//!
+//! There are four basic RPC (Remote Procedure Call) types and a corresponding
```
dfawley: Nit: move the definition of the term to the first use.

grpc-protobuf/src/lib.rs
```diff
@@ -28,20 +28,9 @@
...
-//! * Unary: [`client::UnaryCallBuilder`]
-//! * Client Streaming: [`client::ClientStreamingCallBuilder`]
-//! * Server Streaming: [`client::ServerStreamingCallBuilder`]
-//! * Bidirectional Streaming: [`client::BidiCallBuilder`]
-//!
-//! Each call builder implements [`client::CallBuilder`] which can be used to configure
-//! the call.  Each one also provides an [`IntoFuture`] implementation to
-//! actually begin the call.  See the documentation for each type for additional
-//! usage information.
+//! * [`client`] - Types and call builders for client-side RPCs
```
dfawley: Same nit here: `Client-side types and call builders for RPCs.`

dfawley (approved):
Just a couple tiny nits otherwise LGTM

grpc-protobuf/src/client/mod.rs
arjan-bal: Updated.

grpc-protobuf/src/client/mod.rs
arjan-bal: Done.

grpc-protobuf/src/lib.rs
arjan-bal: Updated.

PR #2818, opened 2026-08-19
grpc-protobuf: add server codegen utils

arjan-bal:
This change introduces the types required for server codegen (see [this branch](https://github.com/grpc/grpc-rust/compare/master...arjan-bal:server-proto-api) for the full set of changes).

Additionally, this PR mvoes the `ServerStatus` from the `grpc` crate to the `grpc-protobuf` crate. `ServerStatus` wraps the standard status type to prevent users from accidentally forwarding client errors from a server handler using the `?` operator.

Finally, this change re-exports the `async_trait` macro from the `grpc` crate (similar to `tonic`). Because `async trait` implementations require` #[async_trait]`, re-exporting it avoids requiring applications to add `async-trait` as a direct dependency in their `Cargo.toml` while using the generated code.


grpc-protobuf/src/status.rs
```diff
@@ -217,6 +223,55 @@ impl StatusError {
...
+}
+
+impl ServerStatusError {
+    /// Creates a new [`ServerStatusError`] with the given code and message.
+    pub fn new(code: StatusCodeError, message: impl Into<String>) -> Self {
+        ServerStatusError(StatusError::new(code, message))
+    }
+
+    /// Creates a new [`ServerStatusError`] from a [`StatusError`].
+    pub fn from_status(status: StatusError) -> Self {
```
dfawley, line 253: We should probably document this and `into` to explain why we didn't `impl From` instead.

grpc-protobuf/src/server/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// Receives the next request message from the stream into `req`.
+    ///
+    /// Returns `Some(Ok(()))` on success, `Some(Err(()))` if the stream
+    /// encountered an error, or `None` if the client has closed the stream.
+    pub async fn recv_into(
+        &mut self,
+        req: &mut impl AsMut<MutProxied = M>,
+    ) -> Option<Result<(), ()>> {
+        let mut res_view = ProtoRecvMessage::from_mut(req);
+        self.rx.dyn_next(&mut res_view).await
```
dfawley: Nit: `req_view` since these are requests instead of responses?  (I assume this was copied from the client `recv_into` which is why it's `res`?)

grpc-protobuf/src/server/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /// Sends a response message on the stream.
+    ///
+    /// Will block if flow control does not allow sending the message. Returns
+    /// an error if the stream has ended or been cancelled.
+    ///
+    /// Note: success does *not* indicate successful receipt of the response by
+    /// the client; it only indicates that the stream has not yet terminated.
+    pub async fn send(&mut self, resp: &impl AsView<Proxied = M>) -> Result<(), ()> {
```
dfawley, line 145: Do we need to do something to silence clippy here?

grpc-protobuf/src/server/bidi.rs
```diff
@@ -0,0 +1,95 @@
...
+    async fn call(
+        &self,
+        requests: GrpcStreamingRequest<Self::Request>,
+        responses: GrpcStreamingResponse<'_, Self::Response>,
+    ) -> ServerStatus;
+}
+
+/// An adapter that wraps a [`BidiStreamingMethod`] to handle incoming
+/// bidirectional-streaming RPCs.
+pub struct BidiStreamingAdapter<M: BidiStreamingMethod> {
```
dfawley: I think we're supposed to not specify the types here unless they're used here.

grpc-protobuf/src/server/unary.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut req = <M::Request as Default>::default();
+
+        if rx
+            .dyn_next(&mut ProtoRecvMessage::from_mut(&mut req))
+            .await
+            .is_none_or(|res| res.is_err())
+        {
+            return trailers_from_status(Err(ServerStatusError::new(
+                StatusCodeError::Internal,
+                "client did not send a request message",
```
dfawley: This is the RPC error we send back to the client?

Maybe "unary call missing request message" or something?

Also should this be using an interceptor to deal with this (and also doing a second recv and expecting to get a `None`)?

https://github.com/grpc/grpc-rust/pull/2602

grpc-protobuf/src/server/unary.rs
```diff
@@ -0,0 +1,127 @@
...
+            )));
+        }
+
+        let mut resp = <M::Response as Default>::default();
```
dfawley: WDYT?  Let's move this up to where `req` is defined and `// TODO: allocate these together on an arena`?

grpc-protobuf/src/server/bidi.rs
arjan-bal: Yes, removed the bounds from all the adapters and their `impl` blocks.

grpc-protobuf/src/server/mod.rs
arjan-bal: Yes, it's a copy/paste error. Fixed.

grpc-protobuf/src/server/mod.rs
arjan-bal: The [clippy docs](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#result_unit_err) suggest replacing the unit struct `()` with a custom Err type that conveys what went wrong, for example:
```rust
#[derive(Debug)]
pub struct EndOfStream;

impl fmt::Display for EndOfStream {
        write!(f, "End of Stream")
    }
}
```

I was thinking of adding two types:
1. `EndOfStream`: Represents either a graceful termination or failure. This will be used in the `send` method.
2. `UnexpectedEndOfStream`: Represents stream failures. This will be used in the`recv` method that returns `Option<Result<M, ()>>` presently, where `None` already represents graceful termination.

We would need to make similar changes in the client APIs for consistency. What do you think?

grpc-protobuf/src/server/unary.rs
arjan-bal: I agree, done.

grpc-protobuf/src/status.rs
arjan-bal: Added to the rustdocs of `from_status` and `into_status` explaining this.

grpc-protobuf/src/server/mod.rs
dfawley: I don't think anything needs changing here, just clippy, probably.  The only thing we need to convey is success vs. failure.  Failure here is always unexpected, as the server method handler is still running, meaning the RPC is still _supposed_ to be working from the server's point of view.  And no matter how it errored, the server should behave the same: return immediately.  In practice, we'll probably drop the future and the handler would never even see this error.

arjan-bal: I’ve suppressed the warning here.

grpc-protobuf/src/server/unary.rs
arjan-bal: Similar to the client-side, I added a `RequestValidator` interceptor in the `grpc` crate that handles request cardinality violations. Now, the interceptor will return an `Err` on the `RecvStream` and forward its own `Trailers` to the client. The `Trailers` returned from here by UnaryAdapter will be ignored.

Made similar changes for all other adapters.

arjan-bal (commented):
I’ve added a request validation interceptor, which expands the scope of this PR a bit. Let me know if you’d prefer me to split that into a separate, smaller PR for easier review.

grpc-protobuf/src/server/bidi.rs
```diff
@@ -78,18 +86,124 @@ where
...
+        &self,
+        headers: RequestHeaders,
+        options: CallOptions,
+        mut tx: &mut dyn DynSendStream,
+        rx: Box<dyn DynRecvStream + 'static>,
+    ) -> Trailers {
+        RequestValidator::new(false)
+            .intercept(headers, options, &mut tx, rx, &self.handle)
+            .make_send()
+            .await
```
dfawley: Can we wrap the handler with the interceptor when registering it instead of doing it this way?  I think that would simplify by removing this type (x4 for each method type)?

dfawley: Otherwise...fun fact, this is more expensive:

```rust
async fn worse() { foo().await }
```

than

```rust
fn better() -> impl Future<Output = ()> { foo() }
```

Even though they're functionally equivalent.  The former incurs extra allocations.

grpc/src/server/stream_util.rs
```diff
@@ -0,0 +1,557 @@
...
+        next: &impl Handle,
+    ) -> Trailers {
+        let (error_tx, mut error_rx) = oneshot::channel();
+        let wrapped_rx = RecvStreamValidator::new(rx, self.is_unary, error_tx);
+
+        let handler_trailers = next.handle(headers, options, tx, wrapped_rx).await;
+        if let Ok(interceptor_trailers) = error_rx.try_recv() {
+            return interceptor_trailers;
+        }
+        handler_trailers
```
dfawley: I don't think this works like we want it to.  On the client the validator works the way it does because the proto code always loops over the whole response stream to get all the message(s) and trailers in a unary response.  On the server, if we are doing unary requests, we should validate that the incoming request has the right cardinality _before_ calling the handler.  Otherwise we're allowing the handler to run when the request may be invalid.

If we want to do this as an interceptor -- and maybe it shouldn't be an interceptor after all(??) -- then we'd want to make this  do two `recv` operations first, then pass a `RecvStream` with the cached message in it to the handler.

And there should be no need to validate anything for non-unary requests.

arjan-bal: Interceptors don't have access to an `&mut dyn RecvMessage` to call the underlying `RecvStream` before the handler calls `next()`. So, I've removed all validation interceptors and reintroduced validation into the unary and server-streaming adapters.

grpc-protobuf/src/server/bidi.rs
arjan-bal: Removed the interceptor now: https://github.com/grpc/grpc-rust/pull/2818#discussion_r3988033230

arjan-bal: I was following the thread on Saurav's PR. I was surprised too.  I've removed the interceptor now: https://github.com/grpc/grpc-rust/pull/2818#discussion_r3988033230

grpc/src/server/stream_util.rs
dfawley: > Interceptors don't have access to an `&mut dyn RecvMessage`

If we wanted to do that, I think they would use some kind of `NopRecvMessage` that would just hold the Bytes from the Buf or something.

grpc-protobuf/src/server/unary.rs
```diff
@@ -0,0 +1,222 @@
...
+                )));
+            }
+            Some(Err(_)) => {
+                return trailers_from_status(Err(ServerStatusError::new(
+                    StatusCodeError::Internal,
+                    "stream failure",
+                )));
+            }
+            Some(Ok(_)) => {}
+        }
```
dfawley, line 114: We need to do a second call and error if a message is received.

grpc-protobuf/src/server/server_streaming.rs
```diff
@@ -0,0 +1,203 @@
...
+                )));
+            }
+            Some(Err(_)) => {
+                return trailers_from_status(Err(ServerStatusError::new(
+                    StatusCodeError::Internal,
+                    "stream failure",
+                )));
+            }
+            Some(Ok(_)) => {}
+        }
```
dfawley, line 108: We need to do a second call and error if a message is received.

arjan-bal: Forgot to add that. Added now.

grpc-protobuf/src/server/unary.rs
arjan-bal: Forgot to add that. Added now.

dfawley (approved):

Issue #2661, opened 2026-05-28
grpc: improve examples


dfawley:
1. The grpc examples share a cargo package with tonic's examples.  Tonic's examples require protoc in your PATH which we'd like to avoid as a strict requirement for grpc.  We should move the grpc examples to `grpc/examples`, or move the tonic examples to `tonic/examples`.

2. The grpc examples currently use a `build.rs` method that requires rebuilding every time, because it outputs the files into `src` and cargo considers them dirty even though they were produced deterministically during the build process.  Instead we should output them into the OUT_DIR, and commit them into another (non-src/...) directory and copy them into OUT_DIR from there when not regenerating from them with protoc (so the examples are trivially runnable).

dfawley:
The second issue is actually a bug in protobuf that was fixed in 35.0:

https://github.com/protocolbuffers/protobuf/issues/26190

ejona86:
(1) is resolved with https://github.com/grpc/grpc-rust/pull/2665 , although you do need to set two separate features to enable. We may still want to split the examples for other reasons.

We did move the output directory (2) is talking about in https://github.com/grpc/grpc-rust/pull/2663, but as dfawley's comment mentions, that is unrelated to the broken dirty detection.

PR #2820, opened 2026-08-19
test: enable missing feature for test build

arjan-bal:
Add missing feature flags to fix `cargo test` when run inside individual crate directories.

CI previously missed this issue because running `cargo nextest` from the workspace root automatically unifies workspace features. 

Additionally, because the `grpc-gcp` example depends on `protoc-gen-rust-grpc` (unlike `tonic` examples), this PR removes `grpc-gcp` from the default features so `cargo test` works out of the box. Note that `examples` crate has no tests.


examples/Cargo.toml
```diff
@@ -294,7 +294,7 @@ grpc-helloworld = ["dep:grpc", "dep:grpc-protobuf", "dep:protobuf"]
 full = ["gcp", "routeguide", "reflection", "autoreload", "health", "grpc-web",
```
dfawley: Is it OK that full doesn't include a feature?  I guess it's also missing "tower" and "h2" already?

But what's the point in having "full" anyway?  Should we delete it and move the things to "default" so it isn't misleading?

examples/README.md
```diff
@@ -100,7 +100,7 @@ Once your credentials are set up, you will need your GCP Project ID, which can
 be found on the main dashboard of the Google Cloud Console. With both of these
 ready, you can run the example like so:
 ```bash
-$ cargo run --bin grpc-gcp-client -- <project-id>
+$ cargo run --bin grpc-gcp-client --features grpc-gcp -- <project-id>
```
dfawley, line 103: This is still a required feature for the binary in the cargo.toml though?  Does that actually mean you have to set it manually or else it won't run?

dfawley (commented):
I still don't understand cargo well enough, it seems....

examples/Cargo.toml
arjan-bal: I'm also not sure what the benefit of the full feature is. It was introduced in #1205 and has been the only `default` feature since optional dependencies were added to the examples crate. Users can already enable all features using Cargo's `--all-features` flag.

I've removed the full feature and set the default directly.

examples/README.md
arjan-bal: Yes, after removing `grpc-gcp` from the `default` set, it must be explicitly enabled. Otherwise the build fails:
```sh
cargo run --bin grpc-gcp-client --features grpc-protobuf-build/build-plugin -- "$PROJECT" 

error: target `grpc-gcp-client` in package `examples` requires the features: `grpc-gcp`
Consider enabling them by passing, e.g., `--features="grpc-gcp"`
```

dfawley: That's kind of crazy.  Maybe we should have a separate crate for each example?  I think that could simplify all this dependency management stuff a lot.

But we should also still think about doing (1) from #2661 and splitting grpc examples from tonic.

Maybe we move the tonic examples either under `tonic/` or rename to `tonic-examples`?

dfawley (approved):

PR #2822, opened 2026-08-19
feat(tonic-xds): expose cluster TLS for custom connectors

LYZJU2019:
## Motivation

Customized transports (e.g. a custom HTTP connector via `build_transport_channel`) need the same gRFC-A29 mTLS as the built-in gRPC connector, but the per-cluster TLS config, cert-provider registry, and SAN matchers were crate-internal.

## Solution

Expose a per-cluster TLS view (feature-gated behind `_tls-any`).

The cert-provider registry is threaded ambiently through the discovery layer, so `build_transport_channel` keeps its public signature and the registry/SAN-matcher types stay private. The verifier is built once per CDS update and `Arc`-cloned per connection — the per-RPC hot path is untouched.

## Testing Done

`cargo test` (default + `tls-ring,testutil`), clippy, fmt, rustdoc `-D warnings`, and `check-external-types --all-features` all green.


tonic-xds/Cargo.toml
```diff
@@ -114,6 +114,7 @@ allowed_external_types = [
     # not major released
     "prost::*",
     "opentelemetry::*",
+    "rustls::verify::ServerCertVerifier",
```
YutaoMa, line 117: Let's also re-export the type from `tonic-xds` so user code can use that directly.

tonic-xds/src/client/endpoint.rs
YutaoMa: Now that `ClusterConfig` also include tls related info, output some info from the `Debug` impl here for better debuggability.

tonic-xds/Cargo.toml
LYZJU2019: Addressed in 51f67df0

tonic-xds/src/client/endpoint.rs
LYZJU2019: Addressed in 2c95ac89

YutaoMa (approved):

PR #2825, opened 2026-08-20
feat(tonic-xds): add public bootstrap config builder & release alpha.3

YutaoMa:
## Motivation

Releasing alpha.3 to include the recent custom transport API additions, in order to test in other crates.

## Solution

One gap identified while assessing the API production readiness: currently the only way to construct a bootstrap config to instantiate an xDS channel from is to go through a A27-formatted JSON string. This is more error prone than a typed builder, which is now introduced in this PR. We acknowledge that other gRPC implementations such as grpc-go explicitly marked its typed config builder as test only / experimental in anticipation of needing to change the config shapes, so here we marked relevant enums and fields as `#[non_exhaustive]` and guarded them with checks to ensure forward compatibility.

### DIFF
--- a/tonic-xds/src/xds/bootstrap.rs
+++ b/tonic-xds/src/xds/bootstrap.rs
@@ -57,18 +57,52 @@ const ENV_BOOTSTRAP_CONFIG: &str = "GRPC_XDS_BOOTSTRAP_CONFIG";
 /// let config = BootstrapConfig::from_json(json).unwrap();
 /// ```
 ///
+/// # Inspecting
+///
+/// The fields are private so new bootstrap keys can be added without breaking
+/// changes. The accessors below report what a loaded config will act on.
+///
+/// ```rust
+/// use tonic_xds::BootstrapConfig;
+///
+/// let json = r#"{
+///   "xds_servers": [{"server_uri": "xds.example.com:443"}],
+///   "node": {"id": "node-1"}
+/// }"#;
+/// let config = BootstrapConfig::from_json(json).unwrap();
+/// assert_eq!(config.server_uri(), "xds.example.com:443");
+/// assert_eq!(config.node_id(), "node-1");
+/// ```
+///
+/// # Building programmatically
+///
+/// [`BootstrapConfig::builder`] constructs a config from typed values,
+/// without needing a JSON string. It covers the
+/// same bootstrap keys [`from_json`] acts on, per gRFC A27.
+///
+/// ```rust
+/// use tonic_xds::{BootstrapConfig, ChannelCredentialType};
+///
+/// let config = BootstrapConfig::builder("xds.example.com:443")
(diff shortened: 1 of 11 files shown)


tonic-xds/src/xds/bootstrap.rs
```diff
@@ -301,13 +400,251 @@ impl BootstrapConfig {
...
+    }
+
+    /// Sets the credentials offered to the xDS server, in preference order.
+    ///
+    /// Replaces any previously set credentials. Unset, the client connects to
+    /// the management server in plaintext, matching a bootstrap with no
+    /// `channel_creds`; pass [`ChannelCredentialType::Tls`] for a secured
+    /// server.
+    #[must_use]
+    pub fn channel_creds(mut self, creds: impl IntoIterator<Item = ChannelCredentialType>) -> Self {
```
ankurmittal, line 603: Picking `Tls` or `GoogleDefault` doesn't guarantee TLS when `server_uri` is `http://`.

`ensure_secure_server_uri` (`xds-client/src/transport/tonic.rs:296`) only prepends
`https://` when the URI has no scheme, so an explicit `http://` stays plaintext. But
`secure` is `self.tls_config.is_some()` (`:313`), so `requires_secure_transport()` sees
`secure == true` and lets caller-supplied call credentials go out over that plaintext
connection.

tonic-xds/src/xds/bootstrap.rs
```diff
@@ -301,13 +400,251 @@ impl BootstrapConfig {
             })
     }
 
-    /// Returns `true` if the first server's selected credential is TLS.
-    pub(crate) fn use_tls(&self) -> bool {
+    /// Returns `true` if the connection to the xDS server uses TLS.
+    pub fn use_tls(&self) -> bool {
```
ankurmittal, line 509: [A27](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md?plain=1#L125-L127) requires `channel_creds`, and requires at least one type the client supports:

> This field is required and must contain at least one channel creds type that the
> client supports.

The field is `#[serde(default)]`, and `selected_credential` skips unknown types, so
missing, empty, and unknown-only lists all reach `use_tls() == false`. With an `http://`
or scheme-less URI that's a plaintext control-plane connection rather than a validation
error.

YutaoMa: true, I've updated `validate` to error on empty channel creds.

tonic-xds/src/xds/bootstrap.rs
YutaoMa: good point, turns out tonic decides TLS on scheme regardless of TLS config. In the new commit I chose to surface the mismatch as error if `http` + tls cred type or vice versa.

ankurmittal (approved):

PR #2827, opened 2026-08-20
Add dead-code ignore annotation to codegen.

nathanielford:
## Motivation

When building a client with `grpc-rust` and a server with `tonic`, the un-ignored client code generates a warning:  `unused import: grpc::client::*`.

This is because Tonic's build process deals with these warnings later, but `grpc-rust` build process has not put anything in place for that yet. 

## Solution

Add the same annotations to the generated client code that exists on the generated server code (see lines 713-720 in the same file), during code generation.  We may want to revisit this later, but this is the most straightforward solution for the moment.

### DIFF
--- a/protoc-gen-rust-grpc/src/cpp_source/src/grpc_rust_generator.cc
+++ b/protoc-gen-rust-grpc/src/cpp_source/src/grpc_rust_generator.cc
@@ -392,11 +392,9 @@ static void GenerateClient(const Service &service, Printer &printer,
       /// Generated client implementations.
       pub mod $client_mod$ {
           #![allow(
-              unused_variables,
               dead_code,
               missing_docs,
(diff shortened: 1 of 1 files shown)


arjan-bal (approved):
LGTM, with a couple of minor comments.

protoc-gen-rust-grpc/src/cpp_source/src/grpc_rust_generator.cc
```diff
@@ -391,6 +391,13 @@ static void GenerateClient(const Service &service, Printer &printer,
       R"rs(
       /// Generated client implementations.
       pub mod $client_mod$ {
+          #![allow(
+              unused_variables,
```
dfawley: Do we need all these? I wouldn't have expected the unused variables or imports.

PR #2836, opened 2026-08-25
xds-client: separate resource worker from ADS server task

W4lspirit:
## Motivation

The current worker owns one ADS stream directly. Federation and fallback need physical streams with independent reconnect, generation, nonce/version, backoff, and request state. Separating that architecture from routing keeps the change reviewable and lets us verify compatibility before enabling multiple servers.

Depends on #2835. Because the dependency branch exists only on the fork, this draft temporarily targets master; after #2835 lands, it will be rebased so the diff contains only this refactor.

## Description

This PR separates resource/watch coordination from a dedicated task that owns the physical ADS stream lifecycle. The server task owns transport construction, reconnect backoff, request snapshots, stream generations, and wire I/O. The resource worker retains subscription aggregation, cache transitions, watcher delivery, and ACK/NACK construction.

The implementation intentionally activates only the first configured server. It does not add fallback, bootstrap federation parsing, authority routing, or xdstp behavior.

## Compatibility

The same watches produce the same initial and incremental requests, ACK/NACK ordering, reconnect version and nonce behavior, watcher events, ProcessingDone flow control, and A78 connection metrics. Stream write failures enter the same closed-stream and reconnect path as receive failures.

Bounded task-channel updates are coalesced by resource type. Dirty updates and removals remain pending until they are successfully queued, and generation-tagged ACK/resume commands cannot affect a replacement stream.

## Validation

- Initial request, update, ACK/NACK, and ProcessingDone tests remain unchanged.
- Added parity coverage for reconnect version/nonce state and continued watcher delivery.
- Added coverage for A78 connected/server-failure metric transitions.
- Added coverage for full command-channel dirty resend retention.
- Added coverage for stream write failure and resubscription.
- Added coverage proving only the first configured server is active.
- cargo test -p xds-client --all-features
- cargo clippy -p xds-client --all-targets --all-features -- -D warnings

### DIFF
--- a/xds-client/src/client/config.rs
+++ b/xds-client/src/client/config.rs
@@ -24,31 +24,152 @@
 
 //! Configuration for the xDS client.
 
+use std::any::Any;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::sync::Arc;
 use std::time::Duration;
 
 use crate::client::retry::RetryPolicy;
 use crate::message::Node;
 
 /// Configuration for an xDS management server.
-#[derive(Debug, Clone)]
+///
+/// Equality and hashing currently cover the URI and transport configuration.
+/// This is not yet the complete gRFC A47 server definition: known server
+/// features must also participate once they are modeled.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
 #[non_exhaustive]
 pub struct ServerConfig {
     uri: String,
-    // Future extensions per gRFC:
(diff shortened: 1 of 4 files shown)


ejona86:
I've updated my branch https://github.com/grpc/grpc-rust/compare/master...ejona86:grpc-rust:xds-client-reorg to bound the events to watchers, and sent out #2881 and #2882. The third commit (not yet sent out, since it is dependent on the others) is the real interesting one though.

PR #2837, opened 2026-08-25
xds-client: reset backoff only after first response

ejona86:
Per gRFC A78, establishing an ADS stream is not sufficient to declare a connection working.

### DIFF
--- a/xds-client/src/client/worker.rs
+++ b/xds-client/src/client/worker.rs
@@ -772,7 +772,7 @@ where
     async fn run_connected<S: TransportStream>(
         &mut self,
         mut stream: S,
-        &mut healthy: bool,
+        healthy: &mut bool,
     ) -> ConnectedOutcome {
         // Whether at least one response was received on this stream. Per gRFC
         // A78 a stream that fails *after* receiving a response is not counted as
@@ -792,9 +792,9 @@ where
                     match result {
                         Ok(Some(bytes)) => {
                             if !saw_response {
-                                if !healthy {
-                                    worker.recorder.record_connected(true);
-                                    healthy = true;
+                                if !*healthy {
+                                    self.recorder.record_connected(true);
+                                    *healthy = true;
                                 }
                                 saw_response = true;
                             }
--- a/xds-client/src/client/worker.rs
+++ b/xds-client/src/client/worker.rs
@@ -697,7 +697,12 @@ where
                 }
             };
 
-            match self.run_connected(stream, &mut healthy).await {
+            if !healthy {
+                self.recorder.record_connected(true);
+                healthy = true;
+            }
+
+            match self.run_connected(stream).await {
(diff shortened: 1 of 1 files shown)


xds-client/src/client/worker.rs
```diff
@@ -700,19 +697,16 @@ where
                 }
             };
 
-            if !healthy {
```
W4lspirit, line 703: >Whether or not the xDS client currently has a working ADS stream to the xDS server. For a given server, this will be set to 1 when the stream is initially created. It will be set to 0 when we have a connectivity failure or when the ADS stream fails without seeing a response message, as per [A57](https://github.com/grpc/proposal/blob/master/A57-xds-client-failure-mode-behavior.md). Once set to 0, it will be reset to 1 when we receive the first response on an ADS stream.

On the first stream creation, we should mark it as healthy even if we did not see any response

ejona86: Hmm... it was broken before, because it should have recorded even before creating a connection. The delay between getting the response headers and getting the first response should be pretty small. I agree it'd be nice to fix that, but it looks more invasive, because it looks like all `return` statements don't properly update the metrics either.

I don't think PR really makes the metrics any worse and all the lines changed here will remain when the metrics are fixed, so I'm going to keep as-is.

YutaoMa: I think it would be worse in the failure mode that if the connection is established but never responded (RST_STREAM, timeouts, etc.) there won't be unhealthy metrics visible. is it possible to keep the "when the stream is initially created" special handling in this PR? I think something similar to grpc-go's `firstStreamCreate` atomic should be quick to implement, and we can fix the rest of the issues like `return` in follow-up PRs.

ejona86: I didn't change that behavior... The code already previously didn't report getting stuck on the first RPC. I only delayed when we mark it as _healthy_. Marking unhealthy was unchanged.

YutaoMa: true that if the stream was never established there is no report, but if the stream is established (headers received) but dropped before response arrives, previously there was metrics but now none, because 

```rust
fn record_unhealthy(&self, healthy: &mut bool) {
    if *healthy {
        self.recorder.record_server_failure();
        self.recorder.record_connected(false);
        *healthy = false;
    }
}
```

has a `if *healthy` guard there. so something like this

```rust
let server = /* servers.recv() */;
server.responses.send(Err(Error::Connection("RST_STREAM".into()))).unwrap();
```

won't be observable if we delay the initial healthy marking.

ejona86: I've reverted the health changes, because it is too broken for me to address right now. Now only the backoff is fixed.

YutaoMa (approved):
thanks, I can work on fixing the metrics

PR #2840, opened 2026-08-27
grpc/service_config: make LB config parsing produce non-optional value

arjan-bal:
Contributes to: https://github.com/grpc/grpc-rust/issues/2761

This change updates `LbConfigSerde` parsing to return a non-optional value, simplifying its use for load balancers that require a child configuration. Consequently, the `load_balancing_config` field in `ServiceConfigSerde` is now optional.

Since `load_balancing_config` is a `repeated` field in the Protobuf definition, and Protobuf cannot differentiate between an unset repeated field and an empty list, a helper function `deserialize_load_balancing_config` is used to parse this optional field and treat an empty list as `None`. This avoids parsing errors and enables falling back to the deprecated `load_balancing_policy` field if it is set.

### DIFF
--- a/grpc/src/client/service_config/mod.rs
+++ b/grpc/src/client/service_config/mod.rs
@@ -57,7 +57,7 @@ impl ServiceConfig {
     // Chooses load balancing configuration per gRPC specification rules.
     pub(crate) fn lb_config(&self) -> (Arc<DynLbPolicyBuilder>, Option<DynLbConfig>) {
         // Choose LbConfig if present.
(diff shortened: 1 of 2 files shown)


nathanielford (approved):

nathanielford (changes requested):
After thinking about this today I think we might want to go about this a bit differently, to avoid exposing serde and to keep the internals where they should be. I have a PR almost completed to address this, but I want to hold off making changes until we can discuss it again.

dfawley (approved):

PR #2841, opened 2026-08-28
Add `Endpoint::eager_connect_errors` to surface lazy-channel connection failures immediately

### LINKED ISSUE #2839: tonic: `Reconnect::poll_ready` returns `Ok(())`  on connection error, breaking tower's p2c  `Balance`  failover
[PR #458 — "fix(transport): reconnect lazy connections after first failure"](https://github.com/grpc/grpc-rust/pull/458) changed `tonic`'s internal `Reconnect` service so, if lazy, when it finds a connection error it returns `Poll::Ready(Ok(()))` in `poll_ready`, deferring returning the error to `call`. Tower's p2c `Balance`, which sits directly on top of these services in `Channel::balance`, relies on `poll_ready` returning `Err` to detect a broken endpoint and route around it. Because `Reconnect::poll_ready` never does this for lazy/reconnecting endpoints, `Balance` can select a broken endpoint as "ready" and dispatch a request to it via `Service::call`, which then fails and forces the caller to retry the RPC instead of tower transparently failing over to a healthy endpoint.

This is affecting services that have a retry budget, since such budget is consumed by connection errors form request errors and are forced to consume such budget in connection errors.

###  Tonic's `src/transport/channel/service/reconnect.rs`

`Reconnect` tracks a one-shot error field and lazy/has-been-connected flags:

```rust
// lines 36-47
pub(crate) struct Reconnect<M, Target>
where
    M: Service<Target>,
    M::Error: Into<crate::BoxError>,
{
    mk_service: M,
    state: State<M::Future, M::Response>,
    target: Target,
    error: Option<crate::BoxError>,
    has_been_connected: bool,
    is_lazy: bool,
}
```

`poll_ready` has two places where it returns `Ok(())` despite a connection error:

```rust
// lines 86-91
fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
    let mut state;

    if self.error.is_some() {
        return Poll::Ready(Ok(()));
    }
```

If a previous poll already recorded an error, any *subsequent* `poll_ready` call short-circuits to `Ok(())` — the caller (Buffer worker / balancer) is told the service is ready, even though it is permanently broken until the next `call` happens to consume the stored error.

```rust
// lines 109-133
State::Connecting(ref mut f) => {
    trace!("poll_ready; connecting");
    match Pin::new(f).poll(cx) {
        Poll::Ready(Ok(service)) => {
            state = State::Connected(service);
        }
        Poll::Pending => {
            trace!("poll_ready; not ready");
            return Poll::Pending;
        }
        Poll::Ready(Err(e)) => {
            trace!("poll_ready; error");

            state = State::Idle;

            if !(self.has_been_connected || self.is_lazy) {
                return Poll::Ready(Err(e.into()));
            } else {
                let error = e.into();
                tracing::debug!("reconnect::poll_ready: {:?}", error);
                self.error = Some(error);
                break;
            }
        }
    }
}
```

```rust
// lines 156-161
    self.state = state;
    }

    self.state = state;
    Poll::Ready(Ok(()))
}
```

When a connect attempt fails and `has_been_connected || is_lazy` is true, the error is stashed in `self.error`, the loop `break`s, and execution falls through to line 160-161, which still returns `Poll::Ready(Ok(()))`. Only the non-lazy and never-connected case (line 124-125) returns `Err` immediately from `poll_ready`.

### Towers p2c balancer

Tonic uses [tower's p2c balancer](https://github.com/tower-rs/tower/blob/master/tower/src/balance/p2c/service.rs) which considers a service to be ready by calling `poll_ready` on it (see [here](https://github.com/tower-rs/tower/blob/df06d70dbea345facbffb5881fe8647f53bf424d/tower/src/balance/p2c/service.rs#L215)). An endpoint being ready if its service returns `Poll::Ready(Ok())` (see [here](https://github.com/tower-rs/tower/blob/df06d70dbea345facbffb5881fe8647f53bf424d/tower/src/ready_cache/cache.rs#L338)). 

In tonic, that inner service is `Connection` which as said before, if lazy, on connection error will return `Poll::Ready(Ok())` and tower's balancer won't try to pick a different one.

hermeGarcia:
Fixes #2839.

Adds `Endpoint::eager_connect_errors(bool)`. When enabled, a lazily-connecting channel (`Endpoint::connect_lazy` / `connect_with_connector_lazy`) reports a failed connection attempt to the caller right away, instead of waiting till a call is made. This allows load balancers like tower's p2c to identify healthy endpoints before making an actual call.  

A new `fail_early` field/parameter on the internal `Reconnect` service, set from `endpoint.eager_connect_errors` at construction time in `Connection::new`.

### DIFF
--- a/tonic/src/transport/channel/service/reconnect.rs
+++ b/tonic/src/transport/channel/service/reconnect.rs
@@ -44,6 +44,7 @@ where
     error: Option<crate::BoxError>,
     has_been_connected: bool,
     is_lazy: bool,
+    fail_early: bool,
 }
 
 #[derive(Debug)]
@@ -58,7 +59,7 @@ where
     M: Service<Target>,
     M::Error: Into<crate::BoxError>,
 {
-    pub(crate) fn new(mk_service: M, target: Target, is_lazy: bool) -> Self {
+    pub(crate) fn new(mk_service: M, target: Target, is_lazy: bool, fail_early: bool) -> Self {
         Reconnect {
             mk_service,
             state: State::Idle,
@@ -66,6 +67,7 @@ where
             error: None,
             has_been_connected: false,
             is_lazy,
+            fail_early,
         }
     }
 }
@@ -121,7 +123,9 @@ where
 
                             state = State::Idle;
 
-                            if !(self.has_been_connected || self.is_lazy) {
+                            if self.fail_early {
+                                return Poll::Ready(Err(e.into()));
+                            } else if !(self.has_been_connected || self.is_lazy) {
                                 return Poll::Ready(Err(e.into()));
                             } else {
                                 let error = e.into();
--- a/tonic/src/transport/channel/service/reconnect.rs
+++ b/tonic/src/transport/channel/service/reconnect.rs
@@ -70,6 +70,10 @@ where
             fail_early,
         }
     }
+
+    fn return_connection_errors_on_poll(&self) -> bool {
+        self.fail_early || !(self.has_been_connected || self.is_lazy)
+    }
 }
 
 impl<M, Target, S, Request> Service<Request> for Reconnect<M, Target>
@@ -123,9 +127,7 @@ where
 
                             state = State::Idle;
 
-                            if self.fail_early {
-                                return Poll::Ready(Err(e.into()));
-                            } else if !(self.has_been_connected || self.is_lazy) {
+                            if self.return_connection_errors_on_poll() {
                                 return Poll::Ready(Err(e.into()));
                             } else {
                                 let error = e.into();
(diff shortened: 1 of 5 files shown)


tonic/src/transport/channel/service/reconnect.rs
```diff
@@ -121,7 +127,7 @@ where
 
                             state = State::Idle;
 
-                            if !(self.has_been_connected || self.is_lazy) {
+                            if self.return_connection_errors_on_poll() {
                                 return Poll::Ready(Err(e.into()));
```
LucioFranco, line 131: I think this might cause problems, such as iirc tower will remove this connection from the readycache and make it never available again. I don't think that is the behavior you want right?

hermeGarcia: Yes, we definitely don't want that.  
Thanks for pointing it out!

PR #2842, opened 2026-08-31
grpc: allow external TLS implementations with __unstable

geertj:
## Motivation

Grpc-rust currently supports rustls for transport security. Some environments want to use grpc-rust but standardize on other TLS libraries (such as BoringSSL).

## Solution

To provide a TLS implementation, an external crate must implement the `ChannelCredentials` (TLS client) and `ServerCredentials` (TLS server) traits. These traits depend on a number of dependent traits, as well as the `Internal` sealing token. This change exposes the required items in the `grpc::__unstable` module, subject to the `__unstable` feature flag.

### DIFF
--- a/grpc/src/credentials/mod.rs
+++ b/grpc/src/credentials/mod.rs
@@ -248,11 +248,18 @@ pub(crate) mod common {
 
 /// Contains information about a [`ChannelCredentials`] or
 /// [`ServerCredentials`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct ProtocolInfo {
     security_protocol: &'static str,
 }
 
 impl ProtocolInfo {
+    #[cfg(feature = "__unstable")]
+    pub const fn new(security_protocol: &'static str) -> Self {
+        Self { security_protocol }
+    }
+
+    #[cfg(not(feature = "__unstable"))]
     pub(crate) const fn new(security_protocol: &'static str) -> Self {
         Self { security_protocol }
     }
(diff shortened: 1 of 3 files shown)


grpc/src/credentials/mod.rs
```diff
@@ -248,11 +248,18 @@ pub(crate) mod common {
 
 /// Contains information about a [`ChannelCredentials`] or
 /// [`ServerCredentials`].
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
dfawley, line 251: I'm not sure whether we can commit to all of these derives at this point.  I'd need to look at what else we need to put in this.  Can you say why you added each of these?

grpc/src/credentials/mod.rs
```diff
@@ -248,11 +248,18 @@ pub(crate) mod common {
...
 pub struct ProtocolInfo {
     security_protocol: &'static str,
 }
 
 impl ProtocolInfo {
+    #[cfg(feature = "__unstable")]
+    pub const fn new(security_protocol: &'static str) -> Self {
```
dfawley, line 258: This would be violating the convention we're trying to follow that everything that is behind this feature flag should be named accordingly.  I think we could name this flavor `unstable_new` or something instead?  (And then always `pub(crate) new`)

grpc/src/credentials/mod.rs
geertj: The derives were added out of habit rather than a concrete requirement. I have removed all except Debug, which I kept as it matches the other public info/metadata types in the crate.

grpc/src/credentials/mod.rs
geertj: Updated. `new` is now `pub(crate)` unconditionally, and the public unstable constructor is named `unstable_new`.

PR #2843, opened 2026-08-31
feat(tonic-xds): support custom control-plane connectors

youyuanwu:
## Motivation

`tonic-xds` always used tonic's built-in connector for the ADS control-plane connection. This prevented applications from supplying a custom transport or managing control-plane TLS themselves.

## Solution

- Add `TonicTransportBuilder::with_connector`.
- Add `XdsChannelBuilder::with_control_plane_connector`.
- Preserve tonic channel setup, keepalives, timeouts, reconnects, and ADS call credentials.
- Leave transport security and URI handling to the custom connector.
- Keep discovered backend connections unchanged.

The connector is cloned for each ADS connection and implements the same `Service<Uri>` interface accepted by tonic's `Endpoint::connect_with_connector`.

## Testing

- Default `tonic-xds` and `xds-client` test suites
- `tonic-xds` with `tls-ring` and `tls-aws-lc`
- Clippy and rustdoc with warnings denied
- Workspace examples and no-default-features build
- `cargo-check-external-types` using the CI nightly toolchain

### DIFF
--- a/tonic-xds/Cargo.toml
+++ b/tonic-xds/Cargo.toml
@@ -69,6 +69,7 @@ tonic = { version = "0.14", features = [ "server", "channel", "tls-ring" ] }
 tonic-prost = "0.14"
 tonic-prost-build = "0.14"
 async-stream = "0.3"
+hyper-util = { version = "0.1", features = ["tokio"] }
 tempfile = "3.27.0"
 rcgen = "0.14"
 google-cloud-auth = { version = "1.9", default-features = false }
--- a/tonic-xds/src/client/channel.rs
+++ b/tonic-xds/src/client/channel.rs
@@ -213,6 +213,7 @@ const _: fn() = || {
     fn assert_send_sync<T: Send + Sync>() {}
     assert_grpc_service::<XdsChannelGrpc>();
     assert_send_sync::<XdsChannelGrpc>();
+    assert_send_sync::<XdsChannelBuilder>();
 };
 
 /// Builder for creating an [`XdsChannel`] or [`XdsChannelGrpc`].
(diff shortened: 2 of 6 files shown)


YutaoMa:
@youyuanwu, thanks for the PR! For the TLS use case I'm planning to add gRFC A65 support properly #2869. My question here would be do you anticipate any use case not covered by A65? Depending on the answer the API surface in this PR may need to be changed, to me personally it feels a bit risky to leave the transport security to the custom connector.

youyuanwu:
@YutaoMa We would like to onboard on tonic-xds, but for tls use case we have hard requirement to use openssl. #2869 does not seem to cover this particular use case.
We also want to use non-tcp transport including the control-plane in the future.

I think the better implementation is to support custom transport for all clients (control plane and data plane), and tls would be a default implementation of the custom transport. (I have not looked at the server apis yet, we need server-side custom transport as well).

YutaoMa:
I see, TLS through openssl is a use case I'm good with supporting, in that case I'm good with a custom connector API. Overall this PR's direction makes sense just have these issues imo:

1. currently the connector interface is unaware of the call credentials mechanism added in #2705 , need some hook to wire them together.
2. `connect_with_connector` skips tonic's `uds_connector` so UDS support is broken under thsi path (tonic hardcodes a  placeholder `http://tonic` for Uds Uri)
3. also `secure && !tls_configured` checks are skipped so if no TLS feature flag is enabled but target URI is https, it will be silently downgraded to plaintext which is a security risk.

Not adding these comments as individual line comments because I think overall they can be solved altogether at a higher level: if you extract out the `connect_with_connector` as the common path, just feed it the tonic default connector if no custom one is provided, then you'll be able to reuse all the existing logic of call creds checks, UDS routing, etc.

PR #2848, opened 2026-09-02
grpc: resolve simple clippy warnings

nathanielford:
## Motivation

Clean up simple, non-invasive clippy warnings in the `grpc` crate so development and pre-commit checks don't trip on unrelated warnings.

## Solution

Resolve simple warnings across 4 categories in `grpc/src`:
- `clippy::semicolon_if_nothing_returned`: Add trailing semicolons on statements returning unit (`()`).
- `clippy::unnecessary_semicolon`: Remove redundant semicolons.
- `clippy::elidable_lifetime_names`: Elide redundant lifetime annotations on method implementations.

More complex lints (e.g. `clippy::large_enum_variant`) are deferred. 

These lints have been added to the Clippy configuration so they run in CICD now.

### DIFF
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -14,6 +14,11 @@ categories = ["web-programming", "network-programming", "asynchronous"]
 rust-version = { workspace = true }
 edition = "2024"
 
+[lints.clippy]
+elidable_lifetime_names = "warn"
+semicolon_if_nothing_returned = "warn"
+unnecessary_semicolon = "warn"
+
 [package.metadata.cargo_check_external_types]
 allowed_external_types = [
     "bytes::*",
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -15,9 +15,9 @@ rust-version = { workspace = true }
 edition = "2024"
 
 [lints.clippy]
-elidable_lifetime_names = "warn"
-semicolon_if_nothing_returned = "warn"
-unnecessary_semicolon = "warn"
+elidable_lifetime_names = "deny"
+semicolon_if_nothing_returned = "deny"
+unnecessary_semicolon = "deny"
 
 [package.metadata.cargo_check_external_types]
 allowed_external_types = [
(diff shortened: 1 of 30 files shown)


grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -525,17 +525,16 @@ mod test {
...
     // 1. Creates a work scheduler.
     // 2. Creates a fake channel that acts as a channel controller.
-    // 3. Creates an StubPolicyBuilder with StubFuncs that each test will define
-    //    and name of the test.
-    // 4. Creates an EndpointSharder with StubPolicyBuilder passed in as the
-    //    child policy.
+    // 3. Creates an StubPolicyBuilder with StubFuncs that each test will define and
+    //    name of the test.
+    // 4. Creates an EndpointSharder with StubPolicyBuilder passed in as the child
+    //    policy.
```
dfawley: I think you're wrapping text at >80 columns.  This should be reverted (stick to 80).

grpc/src/client/load_balancing/child_manager.rs
```diff
@@ -525,17 +525,16 @@ mod test {
...
     // 5. Creates a ChildManager with the EndpointSharder.
     //
     // Returns the following:
     // 1. A receiver for events initiated by the LB policy (like creating a new
     //    subchannel, sending a new picker etc).
-    // 2. The ChildManager to send resolver and subchannel updates from the
-    //    test.
+    // 2. The ChildManager to send resolver and subchannel updates from the test.
```
dfawley: same

dfawley (commented):
IMO we shouldn't do this unless we're also going to enable the clippy things you're fixing at the same time in our CI.  Otherwise we'll just regress again.

If we don't agree these clippies are worth having then we probably should just leave these things alone.  These three do seem pretty reasonable to me, though.

grpc/Cargo.toml
```diff
@@ -14,6 +14,11 @@ categories = ["web-programming", "network-programming", "asynchronous"]
 rust-version = { workspace = true }
 edition = "2024"
 
+[lints.clippy]
+elidable_lifetime_names = "warn"
```
dfawley: Will "warn" here make our CI fail if something is violated? Or just output a warning to logs we'll never see?

nathanielford: Nice catch - 'warn' will only output warning logs in this case. I've moved it up to 'deny'.

dfawley (approved):

PR #2849, opened 2026-09-02
feat(tonic-xds): make outlier detection transport-agnostic

LYZJU2019:
## Motivation

Outlier detection records a per-call success/failure via `svc.record_outcome(result.is_ok())` — a transport-level signal that counts an `Ok(503)` or a gRPC trailers-only error (`200` + `grpc-status != 0`) as a success. That is wrong for a non-gRPC transport, so a plain-HTTP consumer (e.g. an xDS HTTP data plane) cannot use OD correctly. 

## Solution

Introduce a pluggable `OutcomeClassifier` seam so the transport decides what a call outcome means for OD.

Scope: this is the classification seam only. Threading the classifier through `XdsChannelBuilder` and the OD `Discover` is deferred to a follow-up, since the OD load balancer is not yet on the production channel path.

## Testing Done

- 6 `GrpcOutcomeClassifier` unit tests (transport error; 2xx without status; trailers-only `0` / non-zero / unparseable; non-2xx).
- 2 load-balancer wiring tests: a `Failure` verdict ejects an endpoint whose transport calls all succeed; an `Ignore` verdict never ejects endpoints whose transport calls all fail.
- Full `tonic-xds` lib suite green (423 tests); clippy clean (default and `--all-features`); rustdoc intra-doc links resolve.

### DIFF
--- a/tonic-xds/src/client/loadbalance/loadbalancer.rs
+++ b/tonic-xds/src/client/loadbalance/loadbalancer.rs
@@ -1215,14 +1215,15 @@ mod tests {
         for _ in 0..100 {
             let _ = lb.call("hello").await;
         }
-        registry.run_housekeeping();
-        let _ = poll_ready_now(&mut lb);
 
-        assert_eq!(
-            lb.ejected.len(),
-            0,
-            "Ignore must record nothing, so no ejection can occur",
-        );
+        // Ignore records neither success nor failure, so the counters stay at
+        // zero. A "not ejected" check can't prove this: a classifier that
+        // recorded a success on each failing call would also never eject.
+        // `add_channel` is get-or-create, so this reads the state the load
+        // balancer has been recording into.
+        for port in 8080..=8084 {
+            assert_eq!(registry.add_channel(addr(port)).counters(), (0, 0));
+        }
     }
 
     #[tokio::test]
--- a/tonic-xds/src/client/loadbalance/loadbalancer.rs
+++ b/tonic-xds/src/client/loadbalance/loadbalancer.rs
@@ -68,6 +68,9 @@ use crate::client::loadbalance::channel_state::{
 };
 use crate::client::loadbalance::errors::LbError;
 use crate::client::loadbalance::keyed_futures::KeyedFutures;
+use crate::client::loadbalance::outcome::{
+    CallOutcome, HealthOutcome, OutcomeClassifier, OutcomeSource,
+};
 use crate::client::loadbalance::outlier_detection::{OutlierDetector, OutlierStatsRegistry};
 use crate::client::loadbalance::pickers::ChannelPicker;
 use crate::xds::resource::outlier_detection::OutlierDetectionConfig;
@@ -127,6 +130,11 @@ pub(crate) struct LoadBalancer<D, C: Connector, Req> {
     /// and nothing reads from `eject_rx`.
     outlier: OutlierDetector,
     picker: Arc<dyn ChannelPicker<ReadyChannel<C::Service>, Req> + Send + Sync>,
+    /// Transport-specific mapping of each call's result to an outlier-detection
+    /// health verdict. gRPC uses `GrpcOutcomeClassifier`; HTTP and other
+    /// transports inject their own so their status codes are interpreted
+    /// correctly.
+    outcome_classifier: Arc<dyn OutcomeClassifier>,
 }
 
 impl<D, C, Req> LoadBalancer<D, C, Req>
@@ -148,6 +156,7 @@ where
         connector: Arc<C>,
         picker: Arc<dyn ChannelPicker<ReadyChannel<C::Service>, Req> + Send + Sync>,
         config: Arc<ArcSwap<OutlierDetectionConfig>>,
+        outcome_classifier: Arc<dyn OutcomeClassifier>,
     ) -> Self {
         let (registry, eject_rx) = OutlierStatsRegistry::new(config);
         let outlier = OutlierDetector::new(registry, eject_rx);
@@ -160,6 +169,7 @@ where
             ejected: KeyedFutures::new(),
             outlier,
             picker,
+            outcome_classifier,
         }
     }
 
@@ -340,7 +350,7 @@ where
     D::Error: Into<tower::BoxError>,
     C: Connector + Send + Sync + 'static,
     C::Service: Service<Req> + Clone + Send + 'static,
-    <C::Service as Service<Req>>::Response: Send + 'static,
+    <C::Service as Service<Req>>::Response: Send + 'static + OutcomeSource,
     <C::Service as Service<Req>>::Error: Into<tower::BoxError>,
     <C::Service as Service<Req>>::Future: Send + 'static,
     Req: Send + 'static,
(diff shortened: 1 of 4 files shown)


tonic-xds/src/client/loadbalance/outcome.rs
```diff
@@ -0,0 +1,174 @@
+/*
+ *
+ * Copyright 2025 gRPC authors.
```
ankurmittal: 2026?

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -1080,6 +1152,79 @@ mod tests {
...
+        let _ = poll_ready_now(&mut lb);
+
+        assert!(
+            lb.ejected.len() >= 1,
+            "a success classified as Failure should drive ejection",
+        );
+    }
+
+    #[tokio::test]
+    async fn outcome_classifier_ignore_records_nothing() {
```
ankurmittal, line 1189: Every transport call fails and the test asserts no endpoint was ejected. If
`Ignore` recorded a success instead of nothing, there would still be no
ejection.

tonic-xds/src/lib.rs
```diff
@@ -206,6 +206,9 @@ pub use client::channel::{
 pub use client::endpoint::{
     ClusterConfig, Connector, EndpointAddress, EndpointChannel, MakeConnector,
 };
+pub use client::loadbalance::outcome::{
```
ankurmittal: `LoadBalancer::new` is `pub(crate)`, so nothing outside the crate can inject a
classifier, but all four types are re-exported from the crate root.

Two of them are frozen as shaped: `#[non_exhaustive]` sits on the `CallOutcome`
enum rather than on the `Response` variant, so the deferred follow-up can't add
a `trailers` field, and `HealthOutcome` has no `#[non_exhaustive]` at all.

tonic-xds/src/client/loadbalance/outcome.rs
```diff
@@ -0,0 +1,174 @@
...
+    fn classify(&self, outcome: CallOutcome<'_>) -> HealthOutcome {
+        match outcome {
+            CallOutcome::Error => HealthOutcome::Failure,
+            CallOutcome::Response { status, headers } => {
+                if !status.is_success() {
+                    return HealthOutcome::Failure;
+                }
+                let ok = headers
+                    .get("grpc-status")
+                    .is_none_or(|v| v.to_str().ok().and_then(|s| s.parse::<u32>().ok()) == Some(0));
```
ankurmittal: `parse::<u32>()` accepts leading zeros. The protocol says the value is decimal
encoded "without any leading zeros", and tonic's `Code::from_bytes` has no
`(b'0', b'0')` arm, so it maps `00` to `Unknown`. The RPC fails and outlier
detection counts a success.

tonic-xds/src/client/loadbalance/outcome.rs
```diff
@@ -0,0 +1,174 @@
...
+pub(crate) trait OutcomeSource {
+    /// Borrow this response as a [`CallOutcome`].
+    fn call_outcome(&self) -> CallOutcome<'_>;
+}
+
+impl<B> OutcomeSource for http::Response<B> {
+    fn call_outcome(&self) -> CallOutcome<'_> {
+        CallOutcome::Response {
+            status: self.status(),
+            headers: self.headers(),
```
ankurmittal, line 121: `call_outcome()` reads `self.headers()`, and it runs on the value returned by
`svc.call(req).await` — before the body is consumed, so trailers don't exist
yet. The protocol puts the status there for everything that isn't
Trailers-Only:

> Response → (Response-Headers \*Length-Prefixed-Message Trailers) / Trailers-Only

`is_none_or` at line 97 maps an absent `grpc-status` to success, so a streaming
call that fails after the first message, or a unary call where the server sent
headers before erroring, is recorded as healthy.

ankurmittal (approved):

ankurmittal (dismissed):
did you forget to push the commit

tonic-xds/src/client/loadbalance/outcome.rs
LYZJU2019: addressed in 7363d8cb

tonic-xds/src/client/loadbalance/outcome.rs
LYZJU2019: addressed in f6f59a7e

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: addressed in ce8469b2

tonic-xds/src/lib.rs
LYZJU2019: addressed in f0c27505

tonic-xds/src/client/loadbalance/outcome.rs
LYZJU2019: Not addressed here: reading trailers requires draining the response body after `call` returns, which is out of scope for this classification seam and is the same limitation as the `result.is_ok()` path it replaces (a trailer-delivered failure was already counted healthy). It lands with the trailer-aware follow-up — `CallOutcome::Response` is kept crate-internal (see f0c27505) so a `trailers` field can be added then without a breaking change.

tonic-xds/src/client/loadbalance/outcome.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// Maps a transport [`CallOutcome`] to a [`HealthOutcome`] for outlier
+/// detection.
+///
+/// This is the transport seam for OD accounting: the built-in
+/// [`GrpcOutcomeClassifier`] interprets gRPC status, while a non-gRPC transport
+/// (e.g. plain HTTP) supplies its own to interpret HTTP status codes (`5xx` =
+/// failure, `4xx` = ignore, ...) without touching the outlier-detection engine.
+pub(crate) trait OutcomeClassifier: Send + Sync + 'static {
```
YutaoMa: Let's bound it by `Debug` as well to mirror `RetryClassifier`

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -411,6 +432,35 @@ mod tests {
...
+    use crate::client::loadbalance::outcome::GrpcOutcomeClassifier;
+    use std::sync::LazyLock;
+
+    /// The `&'static str` mock responses carry no HTTP metadata, so they satisfy
+    /// the load balancer's `OutcomeSource` bound as a 200 with no headers — a
+    /// success under `GrpcOutcomeClassifier`, matching the pre-classifier
+    /// `record_outcome(result.is_ok())` behavior.
+    static EMPTY_HEADERS: LazyLock<http::HeaderMap> = LazyLock::new(http::HeaderMap::new);
+
+    impl OutcomeSource for &'static str {
```
YutaoMa: create a new type `MockResponse` for this. this `impl for` foreign primitive can cause conflict later.

tonic-xds/src/client/loadbalance/loadbalancer.rs
```diff
@@ -1080,6 +1152,80 @@ mod tests {
...
+        drive_to_ready(&mut lb, &connector).await;
+
+        for _ in 0..100 {
+            lb.call("hello").await.unwrap();
+        }
+        registry.run_housekeeping();
+        let _ = poll_ready_now(&mut lb);
+
+        assert!(
+            lb.ejected.len() >= 1,
```
YutaoMa: nit: `lb.ejected.is_empty()`

tonic-xds/src/client/loadbalance/outcome.rs
LYZJU2019: Addressed in 0575e6e5364fa860087ee13b6c271b76d96a0999

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Addressed in 4c8a090801a169f48531b9f5dc06a8c422c75342

tonic-xds/src/client/loadbalance/loadbalancer.rs
LYZJU2019: Addressed in 4c8a090801a169f48531b9f5dc06a8c422c75342

YutaoMa (approved):

PR #2850, opened 2026-09-03
grpc/client/load_balancing: add endpoint filtering utils

arjan-bal:
Contributes to: #2761

This change add utilities to group endpoints using hierarchical paths stored in `Endpoint` `Attributes`. This will get used in the [upcoming priority balancer](https://github.com/grpc/grpc-rust/compare/master...arjan-bal:hierarchy?expand=1).

### DIFF
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -40,6 +40,7 @@ use crate::core::Address;
 use crate::metadata::MetadataMap;
 use crate::rt::GrpcRuntime;
 
+pub(crate) mod hierarchy;
 pub(crate) mod subchannel_sharing;
 
 pub mod child_manager;
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -43,8 +43,8 @@ use crate::rt::GrpcRuntime;
 pub(crate) mod subchannel_sharing;
 
 pub mod child_manager;
+pub mod endpoint_filtering;
 pub mod graceful_switch;
-pub mod hierarchy;
 pub mod lazy;
 pub mod pick_first;
 pub mod registry;
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -40,11 +40,11 @@ use crate::core::Address;
 use crate::metadata::MetadataMap;
 use crate::rt::GrpcRuntime;
 
-pub(crate) mod hierarchy;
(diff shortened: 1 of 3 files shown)


grpc/src/client/load_balancing/mod.rs
```diff
@@ -40,6 +40,7 @@ use crate::core::Address;
 use crate::metadata::MetadataMap;
 use crate::rt::GrpcRuntime;
 
+pub(crate) mod hierarchy;
```
dfawley: This will need to be exposed via `__unstable`, will it not?  Or will we be moving it to the `grpc-xds` crate completely?

arjan-bal: I've made the module `pub` now. According to [gRFC A56](https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#proposal):

> Note that this policy is not xDS-specific. It does not interact with the XdsClient and contains no xDS-specific code. It is intended to be a generic policy that could be used in any non-xDS context.

Based on this, I'm planning to keep both the `priority` LB and the `hierarchy` utilities in the core `grpc` crate. For reference, Java and Go place the LB under their xDS packages/subdirectories, whereas C++ defines it outside xDS. Let me know if you'd prefer moving both modules into the `grpc-xds` crate instead.

grpc/src/client/load_balancing/mod.rs
```diff
@@ -44,6 +44,7 @@ pub(crate) mod subchannel_sharing;
 
 pub mod child_manager;
 pub mod graceful_switch;
+pub mod hierarchy;
```
dfawley: This name is probably too generic.  In C++ apparently it's in `address_filtering` and in Java it's `AddressFilter` -- but I think those were named that way before we did the happy eyeballs stuff that made everything deal in "endpoints" instead of "addresses".

https://github.com/grpc/grpc/blob/a65db1b513baec11e0d9a726cd4f482f780c88e9/src/core/load_balancing/address_filtering.h#L107

https://github.com/grpc/grpc-java/blob/4fe076d0fd887f6d2ac9aa66a5676222ae2461bb/xds/src/main/java/io/grpc/xds/AddressFilter.java#L65

So maybe `endpoint_filtering` to "match" (but we'd use the more precise/updated nomenclature) would be best.

arjan-bal: Renamed the module to `endpoint_filtering`.

dfawley (approved):
LGTM. Please update the PR title / commit message appropriately.

PR #2851, opened 2026-09-07
grpc/child_manager: use generic child builder

arjan-bal:
Contributes to: https://github.com/grpc/grpc-rust/issues/2761

Make `ChildManager` generic over the `LbPolicyBuilder` type it manages.

Previously, type checking for LB policies and their configuration types relied on runtime invariants enforced by `DynAdapter`. Parameterizing `ChildManager` provides compile-time guarantees for these types. 

As a result:
- `RoundRobin` now manages concrete, non-type-erased `PickFirst` children.
- [`Priority`](https://github.com/grpc/grpc-rust/compare/master...arjan-bal:hierarchy) will manage `GracefulSwitch` children.

### Additional Changes

- Updated `RoundRobin` tests to use the real `PickFirst` balancer rather than test stubs.

### DIFF
--- a/grpc/Cargo.toml
+++ b/grpc/Cargo.toml
@@ -51,6 +51,7 @@ __unstable = []
 gzip = ["dep:flate2"]
 
 [dependencies]
+auto_impl = { version = "1", default-features = false }
 base64 = "0.22"
 bytes = "1.10.1"
 flate2 = { version = "1.0", optional = true, default-features = false, features = [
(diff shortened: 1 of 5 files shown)


dfawley (commented):
After chatting with @LucioFranco, it's probably best to remove the `auto_impl` usage for now.  It seems like people take the proc macro costs pretty seriously, and we also might be able to tweak the way we use serde to avoid it, too...  But if we start adopting this crate, then we'll have more work to do if we do want to do that.

dfawley (approved):
Everything else LGTM, so approving for when that's removed.

PR #2854, opened 2026-09-09
grpc-google: fix cargo toml for publishing

arjan-bal:
This change also unexports the `TokenProvider` trait, which was only used to mock credential-fetching logic in tests.

### DIFF
--- a/grpc-google/Cargo.toml
+++ b/grpc-google/Cargo.toml
@@ -1,11 +1,10 @@
 [package]
 name = "grpc-google"
-version = "0.0.0-alpha"
+version = "0.10.0"
 edition = "2024"
 authors = ["gRPC authors"]
 license = "MIT"
 rust-version = { workspace = true }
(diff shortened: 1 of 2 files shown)


dfawley (approved):

ejona86:
We want to publish a `grpc-google` crate that is "Google Cloud Platform (GCP) Credentials implementation for gRPC"? That is not obvious by its name, and I expect there could be other things in the future that are grpc+google that would wouldn't want to be in this crate. Should we rename this before we publish it?

arjan-bal:
> We want to publish a `grpc-google` crate that is "Google Cloud Platform (GCP) Credentials implementation for gRPC"? That is not obvious by its name, and I expect there could be other things in the future that are grpc+google that would wouldn't want to be in this crate. Should we rename this before we publish it?

I'll bring this up in this weeks maintainers meeting.

PR #2860, opened 2026-09-10
grpc: unlink private symbol in public rustdoc

arjan-bal:
## Why
`RouterBuilder` is private, so it can't be linked in a public rustdoc.

https://github.com/grpc/grpc-rust/blob/900ff055dcc904468747d1ed2a22ffd41486221d/grpc/src/server/router.rs#L48-L49

### DIFF
--- a/grpc/src/server/interceptor.rs
+++ b/grpc/src/server/interceptor.rs
@@ -90,8 +90,8 @@ impl<T: Handle + Sized> HandleExt for T {}
 
 /// A no-op interceptor that simply delegates to the next handler.
 ///
-/// This is the default interceptor used by [`RouterBuilder`](crate::server::RouterBuilder)
-/// when no interceptor has been added.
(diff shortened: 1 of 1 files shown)


nathanielford (approved):
Looks good!

grpc/src/server/interceptor.rs
```diff
@@ -90,8 +90,8 @@ impl<T: Handle + Sized> HandleExt for T {}
 
 /// A no-op interceptor that simply delegates to the next handler.
 ///
-/// This is the default interceptor used by [`RouterBuilder`](crate::server::RouterBuilder)
-/// when no interceptor has been added.
+/// This is the default interceptor used by `RouterBuilder` when no interceptor
+/// has been added.
```
dfawley, line 94: I think this failed to address the concern; it just fixes the build.  Why are we mentioning an internal-only symbol in our public documentation?

Should this be `[ServerBuilder]`?  That's where we expose it externally.

PR #2861, opened 2026-09-10
fix(grpc): remove reference to private RouterBuilder in Identity doc

sauravzg:
Remove reference to private RouteBuilder altogether, there's no need to couple Identity interceptor with it's usage as a default generic in ServerBuilder

### DIFF
--- a/grpc/src/server/mod.rs
+++ b/grpc/src/server/mod.rs
@@ -49,6 +49,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
+use tokio::sync::watch;
 use tonic::async_trait;
 
 use crate::core::ConnectionInfo;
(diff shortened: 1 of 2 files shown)


grpc/src/server/mod.rs
```diff
@@ -192,7 +192,8 @@ impl GracefulCoordinator {
 }
 
 impl Server {
-    /// Creates a [`ServerBuilder`](builder::ServerBuilder) with no interceptors.
+    /// Creates a [`ServerBuilder`](builder::ServerBuilder) with an
+    /// [`Identity`](interceptor::Identity) interceptor which behaves like a no-op.
     pub fn builder() -> builder::ServerBuilder<interceptor::Identity> {
```
dfawley: (A pet peeve I have with Gemini is that it fails to `use` things quite often for some reason, when humans normally would.)

Please `use builder::ServerBuilder` and `interceptor::Identity` and simplify these things.  E.g. the doc no longer needs `[]()` - it can just do `[]` and work automagically.  Also there's other stuff in this file, like we should probably `use tokio::sync::watch` and make the code say `watch::Sender` and `watch::channel()`.

sauravzg: Done.

dfawley (approved):
Thanks!

PR #2862, opened 2026-09-10
grpc/service_config: memoize default_lb_policy via LazyLock

nathanielford:
## Motivation

`ServiceConfig::default_lb_policy()` is invoked whenever a service config omits load balancing configuration, specifies an empty candidate list (`loadBalancingConfig: []`), or when name resolution completes without providing a service config. Re-parsing empty JSON and re-allocating identical configuration objects on every fallback resolution is redundant.

## Solution

Memoize the resolved `(Arc<DynLbPolicyBuilder>, Option<DynLbConfig>)` using `std::sync::LazyLock`:
- The `pick_first` builder lookup and configuration parsing executes once on first access.
- Subsequent calls to `ServiceConfig::default_lb_policy()` return a clone of the memoized tuple with zero mutex locks, zero JSON AST allocations, and zero re-parsing passes.


grpc/src/client/service_config/mod.rs
```diff
@@ -26,13 +26,24 @@ pub(crate) mod duration;
...
 use crate::client::load_balancing::ParsedJsonLbConfig;
 use crate::client::load_balancing::pick_first;
 
+static DEFAULT_PICK_FIRST: LazyLock<(Arc<DynLbPolicyBuilder>, Option<DynLbConfig>)> =
+    LazyLock::new(|| {
+        let builder = GLOBAL_LB_REGISTRY
+            .get_policy(pick_first::POLICY_NAME)
+            .expect("pick_first policy must be registered");
+        let default_json = ParsedJsonLbConfig::from_value(serde_json::json!({}));
+        let parsed_config = builder.parse_config(&default_json).ok().flatten();
```
arjan-bal: We should probably panic instead of ignoring an unexpected parsing error.

arjan-bal (approved):
LGTM with a minor comment.

grpc/src/client/service_config/mod.rs
nathanielford: Done!

PR #2863, opened 2026-09-10
grpc/load_balancing: move LB config selection into registry

nathanielford:
## Motivation

Child selection and configuration parsing for load balancing policies were previously embedded within private `ServiceConfigSerde` deserialization logic. This created several architectural limitations:

- **Inaccessible LB Config Selection**: No reuse of LB configuration resolution for child balancers of composite policies (e.g., xDS locality balancers, priority balancers).
- **Lack of Child Policy Filtering**: Parent policies could not enforce child eligibility constraints per gRFC A24.
- **Public API Coupling to Serde**: `LbPolicyBuilder::parse_config` accepted `&ParsedJsonLbConfig`, leaking `serde::de::DeserializeOwned`.
- **Allocation Inefficiencies**: Child list scanning eagerly deserialized `Vec<HashMap<String, Value>>` and deep-cloned JSON ASTs during configuration parsing.

### Architectural Shift

- **LB Configuration Selection Moved to Registry**: Child policy selection is centralized in `GLOBAL_LB_REGISTRY`, making resolution reusable for both top-level channel configurations and nested child balancers.
- **Filtering and Halting Invariants (gRFC A24)**: Selection supports policy filtering predicates and immediately halts on the first registered, permitted policy, failing if that policy's configuration parsing fails.
- **Decoupled Public API**: Public load balancing traits are decoupled from external serialization crates by passing borrowed string slices (`LbConfigJson<'a>`) rather than Serde-bound structures.
- **Zero-AST `oneOf` Enforcement**: Child policy list parsing enforces single-property (`oneOf`) object entries without intermediate AST allocations.
- **Mandatory Child Resolution Helper**: Added `select_required_child` returning `Result<ParsedLbConfig, String>` for composite balancers where child policies are mandatory.

### Usage Example

**Selecting a required child policy in a composite LB policy:**
```rust
let lb = GLOBAL_LB_REGISTRY.select_required_child(raw_child_policies_json)?;
let child_policy = lb.builder.build(options);
// Apply parsed configuration: lb.config
```

**Hierarchical child selection with a policy filter:**
```rust
let parsed_lb = GLOBAL_LB_REGISTRY.select_child_with_filter(
    raw_child_policies_json,
    Some(|name: &str| name == "round_robin" || name == "pick_first"),
)?;

if let Some(lb) = parsed_lb {
    let child_policy = lb.builder.build(options);
    // Apply parsed configuration: lb.config
}
```

**Implementing an `LbPolicyBuilder`:**
```rust
impl LbPolicyBuilder for PickFirstBuilder {
    type LbPolicy = PickFirstLbPolicy;

    fn parse_config(
        &self,
        config: &LbConfigJson<'_>,
    ) -> Result<Option<PickFirstConfig>, String> {
        serde_json::from_str(config.as_str())
            .map(Some)
            .map_err(|e| format!("Invalid pick_first config: {e}"))
    }
}
```

### DIFF
--- a/grpc/src/client/load_balancing/mod.rs
+++ b/grpc/src/client/load_balancing/mod.rs
@@ -24,7 +24,6 @@
 
 use core::panic;
 use std::any::Any;
-use std::error::Error;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::sync::Arc;
@@ -77,7 +76,7 @@ pub trait LbPolicyBuilder: Send + Sync + Debug + 'static {
     /// default implementation returns Ok(None).
     fn parse_config(
         &self,
-        _config: &ParsedJsonLbConfig,
+        _config: &LbConfigPayload<'_>,
     ) -> Result<Option<<Self::LbPolicy as LbPolicy>::LbConfig>, String> {
         Ok(None)
     }
@@ -178,43 +177,37 @@ pub trait WorkScheduler: Send + Sync + Debug {
     fn schedule_work(&self, data: Option<WorkData>);
 }
 
-/// Abstract representation of the configuration for any LB policy, stored as
-/// JSON.  Hides internal storage details and includes a method to deserialize
-/// the JSON into a concrete policy struct.
-#[derive(Debug)]
-pub struct ParsedJsonLbConfig {
-    value: serde_json::Value,
+/// Borrowed JSON payload containing the configuration for an LB policy.
+///
+/// Hides internal serialization libraries and guarantees zero-allocation access.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct LbConfigPayload<'a>(&'a str);
+
+impl<'a> LbConfigPayload<'a> {
+    pub const EMPTY: Self = Self("{}");
+
+    #[inline]
+    pub const fn new(raw_json: &'a str) -> Self {
+        Self(raw_json)
+    }
+
+    #[inline]
+    pub const fn empty() -> Self {
+        Self::EMPTY
+    }
+
+    #[inline]
+    pub const fn as_str(&self) -> &'a str {
+        self.0
+    }
 }
 
-impl ParsedJsonLbConfig {
-    /// Creates a new ParsedJsonLbConfig from the provided JSON string.
-    pub fn new(json: &str) -> Result<Self, String> {
-        match serde_json::from_str(json) {
-            Ok(value) => Ok(ParsedJsonLbConfig { value }),
-            Err(e) => Err(format!("failed to parse LB config JSON: {e}")),
-        }
-    }
-
-    pub fn from_value(value: serde_json::Value) -> Self {
-        Self { value }
-    }
-
-    /// Converts the JSON configuration into a concrete type that represents the
-    /// configuration of an LB policy.
-    ///
-    /// This will typically be used by the LB policy builder to parse the
-    /// configuration into a type that can be used by the LB policy.
-    pub fn convert_to<T: serde::de::DeserializeOwned>(
-        &self,
-    ) -> Result<T, Box<dyn Error + Send + Sync>> {
-        let res: T = match serde_json::from_value(self.value.clone()) {
-            Ok(v) => v,
-            Err(e) => {
-                return Err(format!("{e}").into());
-            }
-        };
-        Ok(res)
-    }
+/// A successfully resolved load balancing policy builder and its parsed configuration.
+#[derive(Clone, Debug)]
+pub struct ParsedLbConfig {
+    pub builder: Arc<DynLbPolicyBuilder>,
+    pub config: Option<DynLbConfig>,
 }
 
 /// Controls channel behaviors.
@@ -474,7 +467,7 @@ impl<B: LbPolicyBuilder + ?Sized> LbPolicyBuilder for Arc<B> {
 
     fn parse_config(
         &self,
-        config: &ParsedJsonLbConfig,
+        config: &LbConfigPayload<'_>,
     ) -> Result<Option<<B::LbPolicy as LbPolicy>::LbConfig>, String> {
         (**self).parse_config(config)
     }
--- a/grpc/src/client/load_balancing/registry.rs
+++ b/grpc/src/client/load_balancing/registry.rs
@@ -32,10 +32,11 @@ use crate::client::load_balancing::ChannelController;
 use crate::client::load_balancing::DynLbConfig;
 use crate::client::load_balancing::DynLbPolicy;
 use crate::client::load_balancing::DynLbPolicyBuilder;
+use crate::client::load_balancing::LbConfigPayload;
 use crate::client::load_balancing::LbPolicy;
 use crate::client::load_balancing::LbPolicyBuilder;
 use crate::client::load_balancing::LbPolicyOptions;
-use crate::client::load_balancing::ParsedJsonLbConfig;
+use crate::client::load_balancing::ParsedLbConfig;
 use crate::client::load_balancing::WorkData;
 use crate::client::load_balancing::subchannel::Subchannel;
 use crate::client::load_balancing::subchannel::SubchannelState;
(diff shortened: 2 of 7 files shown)


grpc/src/client/load_balancing/registry.rs
```diff
@@ -73,6 +74,157 @@ impl LbPolicyRegistry {
...
+    ///
+    /// ```ignore
+    /// use grpc::client::load_balancing::GLOBAL_LB_REGISTRY;
+    ///
+    /// let candidates_json = r#"[
+    ///     {"xds": {}},
+    ///     {"round_robin": {}}
+    /// ]"#;
+    ///
+    /// // A parent policy (e.g. grpclb) can filter candidates to only allow specific children.
```
arjan-bal, line 93: `grpclb` is deprecated in other languages, and we may not implement in Rust. We should probably not mention in the rustdoc.

grpc/src/client/load_balancing/registry.rs
```diff
@@ -73,6 +74,157 @@ impl LbPolicyRegistry {
...
+    ///     )
+    ///     .unwrap();
+    ///
+    /// assert!(selected.is_some());
+    /// assert_eq!(selected.unwrap().builder.name(), "round_robin");
+    /// ```
+    pub fn select_candidate_with_filter<F>(
+        &self,
+        candidate_list_json: &str,
+        filter: Option<F>,
```
arjan-bal, line 107: I'm not sure if we need a filter right now. Round robin doesn't have an LB config and `grpclb` is not going to be implemented in Rust. Maybe we should add this method later, when we have a consumer for it?

grpc/src/client/load_balancing/registry.rs
```diff
@@ -73,6 +74,157 @@ impl LbPolicyRegistry {
...
     }
+
+    /// Evaluates an ordered candidate list against the registry.
+    ///
+    /// Halts candidate evaluation immediately on the first supported policy
+    /// if configuration parsing fails (gRFC A24).
+    ///
+    /// # Example
+    ///
+    /// ```ignore
```
arjan-bal, line 85: Can we make this a runnable example or use `no_run` to ensure the code compiles? This would help catching regressions.

grpc/src/client/load_balancing/registry.rs
```diff
@@ -73,6 +74,157 @@ impl LbPolicyRegistry {
...
+                filtered_out.join(", ")
+            ),
+            (true, true) => "No supported load balancing policy found in config.".to_string(),
+        };
+
+        Err(err_msg)
+    }
+
+    /// Evaluates an ordered candidate list against the registry without a filter.
+    pub fn select_candidate(
```
arjan-bal, line 176: We should document the return type in more detail. Specifically, the difference between `Ok(None)` and `Err(_)`.

arjan-bal: We should probably use the term "child" or "load_balancer" instead of "candidate" to be consistent with A24.

grpc/src/client/load_balancing/mod.rs
```diff
@@ -178,45 +177,39 @@ pub trait WorkScheduler: Send + Sync + Debug {
...
-#[derive(Debug)]
-pub struct ParsedJsonLbConfig {
-    value: serde_json::Value,
-}
+/// Borrowed JSON payload containing the configuration for an LB policy.
+///
+/// Hides internal serialization libraries and guarantees zero-allocation access.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct LbConfigPayload<'a>(&'a str);
```
arjan-bal, line 185: nit: I don't think `Payload` adds much meaning to the name. Using `JSON` instead (e.g., `LbConfigJson`) would make its purpose much clearer, as it explicitly signals the data format.

grpc/src/client/service_config/serde_bindings.rs
```diff
@@ -116,37 +114,34 @@ impl<'de> Deserialize<'de> for LbConfigSerde {
...
+            Some(RawOrValue::Value(val)) => {
+                if val.is_null() {
+                    return Ok(LbConfigSerde(None));
+                }
+                std::borrow::Cow::Owned(val.to_string())
             }
+        };
+
+        match GLOBAL_LB_REGISTRY.select_candidate(raw_str.as_ref()) {
+            Ok(Some(parsed)) => Ok(LbConfigSerde(Some(LbInnerConfig {
```
arjan-bal, line 138: Can we replace the `LbInnerConfig` struct with `ParsedLbConfig`?

grpc/src/client/load_balancing/registry.rs
arjan-bal, line 176: Can we have another helper method that returns `Result<ParsedLbConfig, String>` after converting the `Ok(None)` to an error? Most consumers would need to do this anyways.

grpc/src/client/load_balancing/mod.rs
```diff
@@ -178,45 +177,39 @@ pub trait WorkScheduler: Send + Sync + Debug {
...
-/// the JSON into a concrete policy struct.
-#[derive(Debug)]
-pub struct ParsedJsonLbConfig {
-    value: serde_json::Value,
-}
+/// Borrowed JSON payload containing the configuration for an LB policy.
+///
+/// Hides internal serialization libraries and guarantees zero-allocation access.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(transparent)]
```
arjan-bal: The default `#[repr(Rust)]` is already pretty good at optimizing layout. Unless we want to do unsafe pointer transmutes, we should probably avoid using `#[repr(transparent)]`.

grpc/src/client/load_balancing/registry.rs
nathanielford: Good context! I've pulled references to it - mostly just using it as an obvious example, so this isn't an issue.

grpc/src/client/load_balancing/registry.rs
nathanielford: Is there a way to do this given that `load_balancing` mod is pub(crate)?  I added a note here as to why it's not runnable rn but if there is a better way to do this I'm happy to change it!

grpc/src/client/load_balancing/registry.rs
nathanielford: Done.

grpc/src/client/load_balancing/mod.rs
nathanielford: I've renamed everything - wasn't keen on anything I came up with anyway, so happy to have input here!

grpc/src/client/service_config/serde_bindings.rs
nathanielford: Done!

grpc/src/client/load_balancing/mod.rs
nathanielford: Removed!

grpc/src/client/load_balancing/registry.rs
nathanielford: I've updated it to `child`, save where it appears in errors, where I use 'load balancer'.

nathanielford: Done!

grpc/src/client/load_balancing/registry.rs
arjan-bal: Maybe you could use the `__unstable` mod?
https://github.com/grpc/grpc-rust/blob/0e49d3014acd70acc8421d43ed117b2145966bbd/grpc/src/lib.rs#L80-L83

grpc/src/client/load_balancing/registry.rs
```diff
@@ -75,72 +75,90 @@ impl LbPolicyRegistry {
...
+    ///
     /// # Example
     ///
-    /// ```ignore
+    /// ```ignore (cannot be compiled as doctest because `load_balancing` is pub(crate))
     /// use grpc::client::load_balancing::GLOBAL_LB_REGISTRY;
     ///
-    /// let candidates_json = r#"[
+    /// let children_json = r#"[
     ///     {"xds": {}},
```
arjan-bal, line 107: nit: Can we use the name of a real LB here? Seeing `xds` here may confuse readers, since there's no LB named `xds`, but there is a resolver with `xds` as the scheme.

arjan-bal (approved):
LGTM! I'd like to get a review from @dfawley before we merge this.

grpc/src/client/service_config/serde_bindings.rs
```diff
@@ -88,17 +83,11 @@ fn default_max_connections_per_subchannel() -> SerdeU32 {
...
 
-#[derive(Debug, Clone)]
-pub(crate) struct LbInnerConfig {
-    pub(crate) builder: Arc<DynLbPolicyBuilder>,
-    pub(crate) config: Option<DynLbConfig>,
-}
-
 #[derive(Debug, Clone, Default)]
-pub(crate) struct LbConfigSerde(Option<LbInnerConfig>);
+pub(crate) struct LbConfigSerde(Option<ParsedLbConfig>);
```
arjan-bal, line 87: If the types in the `serde_bindings` module are only intended for use by the parent `service_config` module, we should probably change the visibility of `serde_bindings` from `pub(crate)` to private within `service_config`.

Update: I realised that there are no non-serde versions of method config, health config and retry policy. Are consumers expected to use `MethodConfigSerde` etc. directly?

nathanielford: I expect that these will be exposed by the top-level ServiceConfig similar to `lb_config()`, and will have their own types (i.e. `MethodConfig`). I think unlike the lb policy configurations these don't need to support custom deserialization code app developers would write? So we can keep the deserialization hidden in `serde_bindings.rs`, but if we need to we can do something similar when we get there.

grpc/src/client/load_balancing/registry.rs
nathanielford: Done! Used `weighted_round_robin_experimental`, but we can update this later if there is one that makes more sense.

grpc/src/client/load_balancing/registry.rs
nathanielford: Done!

grpc/src/client/load_balancing/registry.rs
nathanielford: I could go either way here. I'm a little loathe to pull it so that we don't allow or accidentally do something later that is more awkward because the support doesn't exist. But lmk what you think.

dfawley (changes requested):
As discussed offline, I don't think this is the right approach.  I think parsing the config is a function of the config, not the registry.

Also this PR does other pretty significant things besides what's on the tin and should have been broken up.  Now is also not the time to be discussing some of the changes it's proposing, like removing serde from the LB API.  Once we have a feature complete gRPC implementation and we're ready to talk more seriously about stabilizing that API (which is still unstable in all other languages), we can deal with things like that.

PR #2865, opened 2026-09-11
grpc: fix broken rustdoc links in the server module

sauravzg:
Fix reference to private interceptor and remove now deleted call. 

We may need to revisit making `Identity` and `Interceptor` public at some point, since we expose a private symbol `Identity` via `ServerBuilder` public. Ideally, this may require making interceptors pub instead of pub crate


arjan-bal (approved):
LGTM

Issue #1237, opened 2023-01-14
Add compression level support


zamazan4ik:
## Feature Request
Add compression-level support to the supported compression algorithms.

### Motivation

Right now there is no option to specify compression-level support if compression is enabled. Some default compression level is used. As a user want to be able to configure compression level so I can choose in tradeoff between compression time and resulting size.

### Proposal

Add compression level support to `compress` method.

PR #2866, opened 2026-09-12
tonic: support configurable gzip compression levels

sakateka:
Updates #1237

## Motivation

Allow applications to choose the gzip compression level for requests and
responses. Configurable deflate and zstd levels remain follow-up work.

## Solution

Add `CompressionConfig` and `send_compressed_with_config` to core and generated
clients and servers:

```rust
use tonic::codec::GzipLevel;

let client = client.send_compressed_with_config(GzipLevel::FAST.into());
let server = server.send_compressed_with_config(GzipLevel::BEST.into());
```

`GzipLevel::try_from(u32)` accepts levels in `0..=9` and returns
`InvalidGzipLevel` for invalid values. `NONE`, `FAST`, and `BEST` provide named
constants for common choices.

Store settings separately to preserve the existing `CompressionEncoding` enum,
including numeric casts, and the existing `send_compressed` signatures. Default
compression behavior is unchanged.

Negotiation retains the configured level for outgoing messages. Re-enabling an
encoding replaces its settings without changing its position. Headers still use
`gzip`, and decoding is independent of the sender's compression level.

Regenerate the checked-in clients and servers. A separate CI commit initializes
the MSVC environment for the Windows protoc plugin build.

## Validation

- 84 tonic unit tests and 70 compression tests passed.
- Tonic and tonic-build doctests passed.
- `cargo-semver-checks` passed for tonic and tonic-build against 0.14.5.
- Clippy passed with warnings denied, including all combinations of compression
  features and builds without compression.
- Formatting, rustdoc with default features and compression enabled, and
  code-generation reproducibility checks passed.
- `cargo check --workspace --features full` passed with protoc 32.0.


dfawley:
I don't believe any other grpc implementation has configurable gzip compression levels, so something like this would probably need to be agreed upon as a cross-language addition via a [gRFC](https://github.com/grpc/proposal#grpc-rfcs).  Also, we are not generally adding any significant new features to tonic itself, to focus on development of the grpc crate, so I will close this.

sakateka:
@dfawley, would you consider changing tonic’s default gzip compression level from 6 to 1 as a smaller alternative? I couldn’t find a recommended numeric gzip level in the existing gRFCs, and the [compression specification](https://github.com/grpc/grpc/blob/master/doc/compression.md#compression-levels-and-algorithms) does not prescribe one.

We tested gzip levels 1–6 on a captured 16.7 MB protobuf response, reproducing tonic 0.14.6’s compression path with flate2/miniz_oxide. In a release build, with 80 measured samples per level after warmup:

| Level | Median compression time | Compressed size |
|---|---:|---:|
| gzip-6 | 162.927 ms | 747,572 bytes |
| gzip-1 | 30.984 ms | 1,077,700 bytes |

That is **5.26× faster compression**, at the cost of a **44.2% larger compressed payload**. These measurements cover compression alone. For our CPU-bound server workload, level 6 is a very expensive default and compression becomes a bottleneck; level 1 offers a much better trade-off.

Also, grpc-go already supports configurable gzip levels through [`encoding/gzip.SetLevel`](https://pkg.go.dev/google.golang.org/grpc/encoding/gzip#SetLevel), available since v1.11.0. It allows selecting level 1 during initialization, which might be a useful reference for tonic as well.

PR #2868, opened 2026-09-14
grpc/client/load_balancing: add priority LB

### LINKED ISSUE #2761: gRPC-xDS: Priority LB Policy (gRFC A56)

arjan-bal:
Fixes: https://github.com/grpc/grpc-rust/issues/2761

This change adds the priority load balancing policy as described in [gRFC A56]( https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md) and [gRFC A115](https://github.com/grpc/proposal/blob/master/A115-remove-priority-lb-child-policy-cache.md).


grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+use crate::client::service_config::serde_bindings::LbConfigSerde;
+use crate::client::service_config::serde_bindings::LbInnerConfig;
+use crate::core::Address;
+
+/// Configuration for an individual child under the `priority_experimental`
+/// LB policy.
+///
+/// Corresponds to the protobuf message
+/// `PriorityLoadBalancingPolicyConfig.Child` defined in
+/// [gRFC A56 §LB Policy Configuration]:
```
dfawley, line 71: Not sure how I feel about the section marker symbol here (and below/elsewhere).  IMO we should stick to lower ASCII characters unless necessary.  WDYT?  There are 20+ of these in this PR.

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+    ///   https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md
+    /// [gRFC A56]:
+    ///   https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md
+    #[serde(default)]
+    pub(super) ignore_reresolution_requests: bool,
+
+    /// The child load balancing policy configuration, specifying the policy
+    /// to instantiate (e.g., `round_robin`, `pick_first`, `weighted_target`)
+    /// and its policy-specific configuration.
+    pub(super) config: ChildLbConfig,
```
dfawley, line 102: Super nit: keep the order the same as it's described above. It doesn't matter here because there are only 2 fields, but if there were 10 fields it would be pretty helpful.

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+    /// This prevents lower-priority or failover children from triggering
+    /// unnecessary name resolution when they enter `TRANSIENT_FAILURE`.
+    /// See [gRFC A37] and [gRFC A56] for details.
+    ///
+    /// [gRFC A37]:
+    ///   https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md
+    /// [gRFC A56]:
+    ///   https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md
+    #[serde(default)]
+    pub(super) ignore_reresolution_requests: bool,
```
dfawley, line 97: Why are these fields not private?  Looks like it's for parsing tests in the priority policy?  Should we try to change that to remove the possibility of inter-mixing responsibilities?

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+
+/// Parsed load balancing policy configuration for a child balancer.
+///
+/// Deserializes using [`LbConfigSerde`] and ensures that a supported child
+/// load balancing policy is present (i.e. `as_ref()` is `Some`), failing
+/// deserialization otherwise.
+#[derive(Debug, Clone)]
+pub(super) struct ChildLbConfig {
+    pub(super) builder: Arc<DynLbPolicyBuilder>,
+    pub(super) config: Option<DynLbConfig>,
```
dfawley, line 113: Similar: why pub(super)?  This module should ideally conceal all of this.

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+            return Err(
+                "priority child balancer received update with missing LB config".to_owned(),
+            );
+        };
+        self.ignore_reresolution_requests = config.ignore_reresolution_requests;
+        let mut wrapped_controller =
+            WrappedController::new(channel_controller, self.ignore_reresolution_requests);
+        let gs_cfg = GracefulSwitchLbConfig::new(
+            config.config.builder.clone(),
+            config.config.config.clone(),
```
dfawley, line 204: 👀

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+
+/// Builder for [`ChildPolicy`].
+///
+/// Produces wrapped LB policy instances that wrap a [`GracefulSwitchPolicy`]
+/// and filter re-resolution requests if
+/// [`ChildConfig::ignore_reresolution_requests`] is set to `true`.
+///
+/// This builder is used internally by the priority policy to instantiate
+/// children managed by [`ChildManager`], and is not registered in
+/// [`GLOBAL_LB_REGISTRY`].
```
dfawley, line 138: Is there any reason to have a LbPolicyBuilder then?  Maybe just `ChildPolicy::new()` instead?

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        "priority_child_lb"
+    }
+
+    /// Config parsing is a no-op here because child configurations are
+    /// deserialized and validated as part of `PriorityConfig` in the parent
+    /// priority policy.
```
dfawley, line 162: Another sign this probably doesn't want to be a policy builder - it actually has a config, but we don't parse it the normal way.

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+        update: ResolverUpdate,
+        config: Option<&Self::LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), String> {
+        let Some(config) = config else {
+            return Err(
+                "priority child balancer received update with missing LB config".to_owned(),
+            );
+        };
+        self.ignore_reresolution_requests = config.ignore_reresolution_requests;
```
dfawley, line 199: I'm not sure what it means if this value changes.

What if the value changes and the child policy type changes at the same time?  Do we want to effect the change of this flag immediately before we have switched to the new child policy?  Or should the setting only apply to the new child?  Does the spec cover this at all?

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+        &mut self,
+        subchannel: Arc<dyn Subchannel>,
+        state: &SubchannelState,
+        channel_controller: &mut dyn ChannelController,
+    ) {
+        let mut wrapped_controller =
+            WrappedController::new(channel_controller, self.ignore_reresolution_requests);
+        self.graceful_switch
+            .subchannel_update(subchannel, state, &mut wrapped_controller);
+    }
```
dfawley, line 220: This can be deleted after #2873 :)

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+            WrappedController::new(channel_controller, self.ignore_reresolution_requests);
+        self.graceful_switch.exit_idle(&mut wrapped_controller);
+    }
+}
+
+/// A [`ChannelController`] proxy that conditionally filters re-resolution
+/// requests.
+struct WrappedController<'a> {
+    channel_controller: &'a mut dyn ChannelController,
+    ignore_reresolution_requests: bool,
```
dfawley, line 239: Observation: another choice would be to only use a wrapper when we are ignoring requests.  I _think_ we could have a helper to make that a 1-liner, still (and a shorter one at that):

```rust
impl ChildPolicy {
  fn child_controller(&self, channel_controller: &mut dyn ChannelController) -> &mut dyn ChannelController {
    if self.ignore_reresolution_requests {
      IgnoreReresolutionController::new(channel_controller)
    } else {
      channel_controller
    }
  }
}
```

This would mean that every call into the child has to have this branch, so maybe it's better as-is?  But then we remove this bool and the need for a check internally.

grpc/src/client/load_balancing/priority/child.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn test_child_config_empty_policy_fails() {
+        let json = r#"{
+            "config": [],
+            "ignoreReresolutionRequests": false
+        }"#;
+        let res: Result<ChildConfig, _> = serde_json::from_str(json);
+        assert!(res.is_err());
+    }
+
+    /// Runs an end-to-end test using a [`ChildPolicy`] wrapping a `pick_first`
```
dfawley, line 307: I don't think we should call tests end-to-end unless they include a channel.  I think this would be a module-level test or a component test or something.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! The priority LB policy manages an ordered list of child policies. It routes
+//! RPCs to the highest-priority child that is reachable (in `READY` or `IDLE`
+//! state). If higher-priority children are unavailable, fail, or take too long
+//! to connect, the policy fails over to lower-priority children.
+//!
+//! While originally developed to support xDS priority failover across
+//! localities or aggregate clusters (see [gRFC A27] and [gRFC A37]), this
+//! policy contains no xDS-specific logic and can be used generically in any
+//! hierarchical load balancing context.
+//!
```
dfawley, line 42: This feels unnecessary to call out.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! # Hierarchical Addresses
+//!
+//! Endpoints received in a [`ResolverUpdate`] are delivered as a flat list.
+//! When nested load balancers are arranged hierarchically, each endpoint
+//! address is annotated with a hierarchical path attribute (see [gRFC A56:
```
dfawley, line 47: Correction: each endpoint ~address~ is annotated

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! Hierarchical Addresses]).
+//!
+//! [`PriorityPolicy`] groups endpoint addresses by matching the first element
```
dfawley, line 50: ...groups _endpoints_ ~addresses~ by...

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! in each address's path against the child names, strips that element, and
```
dfawley, line 51: endpoint

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! routes the sub-list of addresses down to the corresponding child policy.
```
dfawley, line 52: So do all endpoints have to have a hierarchy element to them?  Maybe it should be stated explicitly in this section.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//!
+//! # Child Lifetime Management & Lazy Creation
+//!
+//! To avoid unnecessary resource consumption and connection churn:
+//! - Lazy creation: Child policies are not instantiated upfront for every
+//!   configured priority. Instead, each child is created on-demand only when
+//!   the priority selection algorithm needs to attempt using that priority tier
+//!   (see [gRFC A56 §Child Lifetime Management]).
```
dfawley, line 60: Remind me what you were saying about using the `Lazy` policy here?  Why don't we?

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//! - Failback & Deactivation: When a higher-priority child becomes `READY` or
```
dfawley, line 61: We don't use "failback" in any of our gRFCs AFAIK.  You should ideally use the same terminology we use in our gRFCs to make it easier to compare these comments with the gRFC.  In this case it seems to be "switching to a higher priority child".  Or we can adopt "failback" in the gRFC -- though I find it to be a really weird word, but I did still know what it meant without context for some strange reason, so.....


But thinking about it further a couple minutes later...it's probably more accurate the way it is in the gRFC, because we can switch to a higher-priority child not only to recover from fail-over, but after a config update that reorders the priorities.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+//!   `IDLE`, previously active lower-priority children are not immediately
+//!   destroyed. Doing so would cause costly connection re-establishment if
+//!   priorities flap. Instead, lower-priority children enter
+//!   [`ChildState::Deactivated`] and run a 15-minute timer. If the
+//!   higher-priority child fails within 15 minutes, the lower-priority child is
+//!   reactivated instantly. If the timer expires without reactivation, the
+//!   child is cleaned up.
+//! - Unconfigured Child Removal: Per [gRFC A115], children omitted from a
+//!   configuration update are immediately destroyed and removed from the child
+//!   manager, completely bypassing the retention cache.
```
dfawley, line 71: Maybe this wants another sentence about how children already in the retention cache are also removed immediately if they are removed from the config?  (Assuming we got that right in the code....I haven't gotten there yet)

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+ *
+ */
+
+//! Priority Load Balancing Policy (`priority_experimental`).
+//!
+//! This module implements the `priority_experimental` load balancing policy as
+//! specified in [gRFC A56: `priority_experimental` LB policy] and updated by
+//! [gRFC A115: disable Priority LB policy child policy retention cache].
+//!
+//! # Overview
```
dfawley, line 31: In addition to / instead of all the comments I made below here (in code, but earlier chronologically), I also think we could just point to the gRFC and say "go read how it works there" instead of putting all these details here.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+
+/// Failover timeout for a child attempting to connect (10 seconds).
+///
+/// Per [gRFC A56 §Child Connectivity State Tracking], each child has a
+/// 10-second failover timer that starts when it begins attempting to connect.
+/// While this timer is active, the priority selection algorithm waits for
+/// this child to connect before failing over to lower priorities.
+///
+/// [gRFC A56 §Child Connectivity State Tracking]:
+///   https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#child-connectivity-state-tracking
```
dfawley, line 149: This is probably unnecessary.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+/// priority flaps.
+///
+/// Note: Under [gRFC A115], this timeout applies only to failover
+/// deactivations (Case 1); children removed from the configuration (Case 2)
+/// are dropped immediately.
+///
+/// [gRFC A56 §Child Lifetime Management]:
+///   https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#child-lifetime-management
+/// [gRFC A115]:
+///   https://github.com/grpc/proposal/blob/master/A115-remove-priority-lb-child-policy-cache.md
```
dfawley, line 166: Similar

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+        Ok(())
+    }
+}
+
+/// Internal tracking data and state for a configured priority child.
+#[derive(Debug)]
+struct ChildData {
+    /// The current lifecycle and connectivity state of the child policy.
+    state: ChildState,
+    /// The active dynamic LB configuration for this child.
```
dfawley, line 231: Nit: "current"?  Active implies it's running?  And remove the word "dynamic"?

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+    child_config: ChildConfig,
+    /// The latest name resolver update received for this child.
+    ///
+    /// Preserved so that if this child is deactivated and later reactivated,
+    /// or if it was uninitialized, it can be instantiated with the most
+    /// recent endpoints and service configuration.
```
dfawley, line 237: Presumably this is also here for lower-priority things that we fallback to when needed?  Maybe just delete this blob?

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+}
+
+impl Timer {
+    /// Spawns a new timer for the given duration that schedules work on
+    /// completion.
+    fn new(duration: Duration, work_scheduler: Arc<dyn WorkScheduler>, rt: GrpcRuntime) -> Timer {
+        let rt_clone = rt.clone();
+        let task_handle = rt.spawn(Box::pin(async move {
+            rt_clone.sleep(duration).await;
+            work_scheduler.schedule_work(Some(Box::new(PriorityTimerWork)));
```
dfawley, line 282: Does this not need a payload with it to deal with races?

What if we start a timer and it fires approximately the same time a new update comes in.  The update triggers a new P0 policy to go into connecting.  We drop the timer.  But the timer already fired.  (Heck, maybe it fired 300ms ago and we have a long work queue.)  Then we start the timer for the new P0 policy in connecting.  But then almost immediately we get our work() called because the old timer fired.  Way before the 10 seconds.

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+/// The `priority_experimental` load balancing policy instance.
+///
+/// Manages a prioritized collection of child policies.
+#[derive(Debug)]
+struct PriorityPolicy {
+    child_mgr: ChildManager<String, ChildBuilder>,
+    child_data: HashMap<String, ChildData>,
+    /// Current priority hierarchy: list of child names sorted from highest
+    /// priority (0) to lowest.
+    priorities: Vec<String>,
```
dfawley, line 406: What's the need for a map internally?  Could we have `Vec<(String, ChildData)>` instead?  I guess that means an O(n^2) resolver_update?

grpc/src/client/load_balancing/priority/mod.rs
```diff
@@ -0,0 +1,864 @@
...
+
+    fn resolver_update(
+        &mut self,
+        update: ResolverUpdate,
+        config: Option<&Self::LbConfig>,
+        channel_controller: &mut dyn ChannelController,
+    ) -> Result<(), String> {
+        let Some(config) = config else {
+            return Err("priority balancer received update with missing LB config".to_owned());
+        };
```
dfawley, line 427: For discussion, not related to this PR directly.

This pattern crops up a lot.

I wonder if we can move it all to one place by making the config parameter `&Self::LbConfig` and making the `Option` just be part of the `DynLbPolicy`?

dfawley (commented):
I'm not fully done with the review, but wanted to send this out now before heading out for the weekend...

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: I asked Gemini to generate rustdocs for this module by providing the two gRFCs as context. It paired each function and struct with its corresponding gRFC section, which seems nice, though it's slightly verbose.

I've since replaced the section markers with parenthesized section names, e.g., `(Section ABC)`.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Done.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Yes, they are exported only for tests in the parent module. I've removed the validation of these private fields from `mod/test.rs` and added new tests that validate them in `child.rs`. This allows us to keep the fields private.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Refactored tests and un-exported the fields.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: The `ChildManager` requires an `LbPolicyBuilder` to instantiate new children while processing resolver updates.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: I've renamed structs and variables to differentiate b/w priority child config, child config and LB config types. This should makes it clearer

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Merged master.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Removed the term "end-to-end". I had asked Gemini to use the real pickfirst balancer instead of a mock LB, which is probably why the term came up in the doc.

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: Such a helper wouldn't compile because `IgnoreReresolutionController` is created on the stack, inside `child_controller`. It doesn't live long enough for the return reference to be valid. We have two options if we want to move the conditional logic out of the wrapper:
1. Check `self.ignore_reresolution_request` at each callsite before wrapping, or
2. Wrap the child calls in a closure which is executed in the helper's stack, keeping the reference to alive.
```rust
impl ChildPolicy {
    /// Invokes `op` on the wrapped child policy.
    ///
    /// If this child is configured to ignore re-resolution requests, the child
    /// sees an [`IgnoreResolutionRequestsController`] instead of the channel's
    /// controller; otherwise the controller is passed through untouched.
    fn with_child<R>(
        &mut self,
        channel_controller: &mut dyn ChannelController,
        op: impl FnOnce(&mut GracefulSwitchPolicy, &mut dyn ChannelController) -> R,
    ) -> R {
        if !self.ignore_reresolution_requests {
            return op(&mut self.graceful_switch, channel_controller);
        }
        let mut controller = IgnoreResolutionRequestsController { channel_controller };
        op(&mut self.graceful_switch, &mut controller)
    }
}
```
Callsite example:
```rust
self.with_child(channel_controller, |gs, cc| {
    gs.resolver_update(update, Some(&gs_cfg), cc)
})
```

I feel the existing approach seems easier to understand than both these options, but option 1 also seems fine.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Removed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Removed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Fixed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Yes, removed the second paragraph.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: That's a good idea. I've replaced the map with a single `Vec<ChildData>`. During resolver updates, the old children are temporarily moved into a `HashMap` to avoid an $O(n^2)$ lookup cost. This eliminates several `panic!` calls that previously handled cases where the two data structures became desynchronized.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Agreed, removed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Removed all the sections except the overview.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Fixed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Fixed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Fixed.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Yes, otherwise they will be dropped when the `endpoint_filtering` module is used to group them by path. Added a line to mention this.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: I've removed this section now.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Yes, cached children are also dropped immediately. I've removed this section now.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: The child LBs store the task cancellation handle and  the deadline. There are two methods that handle timer expiration: `handle_connectivity_timer` and `handle_deactivation_timer`. Both are invoked each time the priority LB's `work()` method is called. These methods iterate through each child LB with a pending timer and check whether its deadline has passed. If so, the child's state is transitioned.

If a new timer is started before the priority LB processes work, the child's deadline is updated. When processing occurs, the two methods will skip the child LB because its deadline is in the future.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Yes, I think this could work. We would probably need to remove the default `parse_config` impl that returns `None`.
https://github.com/grpc/grpc-rust/blob/d339c4791c502e4f204c497ce343dd3cc8753eb7/grpc/src/client/load_balancing/mod.rs#L79-L84

grpc/src/client/load_balancing/priority/child.rs
arjan-bal: > Does the spec cover this at all?

[The spec](https://github.com/grpc/proposal/blob/master/A37-xds-aggregate-and-logical-dns-clusters.md#priority-policy-reresolution-handling) doesn't talk about the value changing. It mentions that the value is set for all priorities coming from EDS clusters.

>  Do we want to effect the change of this flag immediately before we have switched to the new child policy? 

All future resolver updates will be forwarded to the new child. So, ideally:
1. The value set by a resolver update should be used immediately.
2. Re-resolution requests from old children should be ignored.

Go seems to be immediately updating the value here: [1](https://github.com/grpc/grpc-go/blob/059a3022db217da742d0689ebfcbaf9c6f4ed74d/internal/xds/balancer/priority/balancer_child.go#L87-L96), [2](https://github.com/grpc/grpc-go/blob/059a3022db217da742d0689ebfcbaf9c6f4ed74d/internal/xds/balancer/priority/balancer_child.go#L112-L118).
And Go's gracefulswitch [ignores re-resolution requests](https://github.com/grpc/grpc-go/blob/059a3022db217da742d0689ebfcbaf9c6f4ed74d/internal/balancer/gracefulswitch/gracefulswitch.go#L399-L403) from the inactive child.  The rust gracefulswith doesn't seem to be doing this.

grpc/src/client/load_balancing/priority/mod.rs
arjan-bal: Here are the problems I ran into:
* Child manager's `exit_idle` would cause all the lazy children to start connecting. We would need a child manager API to filter children before calling `exit_idle`. This API is also needed to select a particular child to start connecting during the priority picking algorithm.
* Deactivation would require removing the child from the lazy balancer and re-inserting it. This in turn requires priority to continue storing the latest resolver update for each child.
* In the [priority picking algorithm](https://github.com/grpc/proposal/blob/master/A56-priority-lb-policy.md#algorithm-for-choosing-a-priority), IDLE children are treated differently from uninitialized ones. An IDLE child causes lower priorities to be de-activated, while an unitialized child doesn't. We need a way to differentiate the two cases.

PR #2869, opened 2026-09-15
feat(tonic-xds): support A65 ADS TLS credentials

YutaoMa:
## Motivation

Ref: #2444 

Previously data plane TLS (A29) has been supported in `tonic-xds`, but control plane TLS (A65) was missing. The TLS config in bootstrap is parsed and validated but only translated to a "system root only" TLS config, no user configurable TLS config support.

## Solution

This PR re-uses the existing FileWatcher module from A29 (as per the spec of A65) for control plane connection if TLS is specified in the bootstrap config.

Some highlights regarding the code change design:
1. Because we still doesn't support `GoogleDefaultCredentials`, previously this config is parsed and validated as "supported" in the bootstrap config, we now change it to an unsupported variant, making it fall through to use "tls" or "insecure" if specified, or NACK the config otherwise.
2. Bootstrap config for `tls` variant now include the cert file paths fields.
3. extracted the shared abstraction between A29 and A65 file watcher to a separate cert_provider_config module, so the A65 support reuses the logic but isn't coupled to the A29 CertProvider interface exactly. This matches the other gRPC implementations.


tonic-xds/src/client/channel.rs
```diff
@@ -1104,6 +1158,50 @@ mod tests {
         assert!(config.call_creds.is_some());
     }
 
+    #[cfg(feature = "_tls-any")]
+    #[test]
+    fn a65_missing_certificate_file_fails_channel_build() {
```
ankurmittal, line 1163: There's no test that connects to a TLS ADS server with `ca_certificate_file` and
an identity set and checks they're the credentials used.

tonic-xds/src/xds/cert_provider/file_watcher.rs
```diff
@@ -146,7 +104,7 @@ fn refresh_once(config: &FileWatcherConfig, cached: &ArcSwap<CertificateData>) {
         Ok(data) => cached.store(Arc::new(data)),
```
ankurmittal, line 104: `refresh_once` caches the bytes without parsing them, and the cert and key are
two separate reads:

```rust
Some(Identity::new(read_file(cert_path)?, read_file(key_path)?))
```

A rotation between the two pairs a new cert with an old key, and that replaces
the good snapshot. Reconnects fail until the next refresh, 600s by default.

original master branch only used this for the data plane. Now the xDS connection is behind it.

tonic-xds/src/client/channel.rs
YutaoMa: right, I've updated `xds-test-util` to have a TLS xDS server and added an e2e test case for A65.

tonic-xds/src/xds/cert_provider/file_watcher.rs
YutaoMa: Good point, I've added cert validation after each read and make it keep using existing certs if validation failed (same logic as from grpc-go). One specific highlight is that I didn't yet add a retry immediately after cert read failure, because other grpc implementations don't have it (grpc-go have a TODO for that, no gRFC yet) so I decided to keep it simple for now.

ankurmittal (approved):

PR #2871, opened 2026-09-17
grpc-xds: add virtual-host domain matching

YutaoMa:
## Motivation

Ref: #2754 

For gRFC A74 XdsDependencyManager, to resolve the channel's virtual host, we'll need to run string matching per Envoy `StringMatcher` spec.

## Solution

To reduce the size of the upcoming A74 PR, I'm sending this `StringMatcher` PR first. The implementation mostly follows the design we used in `tonic-xds`, with these highlights:

1. Since the same string matching logic is shared between the virtual host selection and A29 SAN matching for mTLS, we extracted a common `StringMatcher` validated resource type and a pure `matcher` module.
2. Used the Rust idiomatic "new type" pattern for `DomainMatchScore` and `SafeRegex`: the former encapsulate the match ordering semantics, and the latter to reject empty regex pattern and conform to the "only match the entire input" Envoy spec.


YutaoMa:
Hi @dfawley @arjan-bal this is the first PR of the `XdsDependencyManager` work. Let me know if you have any thoughts and comments on it

grpc-xds/src/lib.rs
```diff
@@ -42,12 +42,19 @@ pub(crate) mod generated {
     include!(concat!(env!("OUT_DIR"), "/generated/mod.rs"));
 }
 
+/// Shared Envoy string and regular-expression matching primitives.
+/// Used in both routing and TLS SAN matching.
+pub(crate) mod matcher;
```
arjan-bal, line 47: nit: We can omit `pub(crate)` in the root module as it's a no-op.

grpc-xds/src/matcher/mod.rs
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+//! Runtime matching on validated xDS values for routing and TLS SAN matching.
+
+// TODO: remove once consumed by routing and TLS SAN matching
+#![allow(dead_code)]
+
+mod string;
+
+pub(crate) use string::{ends_with_ignore_ascii_case, starts_with_ignore_ascii_case};
```
arjan-bal, line 32: Is there a reason to re-export them instead of exporting the string module directly? My understanding is that re-exporting allows us to provide better organization for external consumers of our API. Since these modules are only consumed within this crate, re-exporting shouldn't be necessary.

grpc-xds/src/resource/safe_regex.rs
```diff
@@ -0,0 +1,136 @@
...
+
+use regex::Regex;
+use xds_client::Error;
+
+const ANCHOR_PREFIX: &str = r"\A(?:";
+const ANCHOR_SUFFIX: &str = r")\z";
+
+/// A regex that can only match an entire input, as required by Envoy.
+#[derive(Clone)]
+pub(crate) struct SafeRegex(Regex);
```
arjan-bal, line 37: nit: Can we use a regular struct instead of a tuple? The code is more readable when the field has a name.

grpc-xds/src/resource/string_matcher.rs
```diff
@@ -0,0 +1,137 @@
...
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ *
+ */
+
+//! Validated Envoy string matcher values shared by header and SAN configuration.
+
+use xds_client::Error;
+
+use super::safe_regex::{SafeRegex, compile_regex};
```
arjan-bal, line 29: Could you set the import granularity to `item` for both Rust Analyzer and rustfmt in your IDE? We use this setting across all of our gRPC crates to keep imports consistent.

grpc-xds/src/resource/string_matcher.rs
```diff
@@ -0,0 +1,137 @@
...
+    /// Parses and validates an `envoy.type.matcher.v3.StringMatcher`.
+    ///
+    /// Returns an error if the `match_pattern` oneof is unset or carries an
+    /// unsupported variant, a prefix/suffix/contains value is empty, or a
+    /// `safe_regex` fails to compile.
+    pub(crate) fn from_proto(proto: StringMatcherView<'_>) -> xds_client::Result<Self> {
+        let ignore_case = proto.ignore_case();
+        match proto.match_pattern() {
+            MatchPatternOneof::Exact(value) => Ok(Self::Exact {
+                value: value.to_str().unwrap_or_default().to_string(),
```
arjan-bal, line 53: Since proto3 strings must be valid UTF-8, the protobuf library rejects non-UTF-8 strings during parsing. However, it is possible to construct in-memory structs containing invalid strings by using the safe and infallible [ProtoString::from_utf8_unchecked(&[u8])](https://docs.rs/protobuf/4.36.2-release/protobuf/struct.ProtoStr.html#method.from_utf8_unchecked) method.

Silently resetting these invalid values to an empty string can lead to unexpected behavior. Could we instead return a validation error in these cases?

grpc-xds/src/resource/safe_regex.rs
```diff
@@ -0,0 +1,136 @@
...
+impl fmt::Debug for SafeRegex {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let anchored = self.0.as_str();
+        let pattern = &anchored[ANCHOR_PREFIX.len()..anchored.len() - ANCHOR_SUFFIX.len()];
+        f.debug_tuple("SafeRegex").field(&pattern).finish()
+    }
+}
+
+/// Compiles a `RegexMatcher` pattern, rejecting the empty pattern.
+pub(crate) fn compile_regex(pattern: &str, kind: &str) -> xds_client::Result<SafeRegex> {
```
arjan-bal, line 62: Can this be a function in `SafeRegex`'s  `impl` block?

grpc-xds/src/matcher/string.rs
```diff
@@ -0,0 +1,170 @@
...
+                } else {
+                    input.contains(value)
+                }
+            }
+            Self::SafeRegex(regex) => regex.is_match(input),
+        }
+    }
+}
+
+pub(crate) fn starts_with_ignore_ascii_case(input: &str, prefix: &str) -> bool {
```
arjan-bal, line 67: Instead of having these match functions in a separate module, can we instead use the `StringMatcher` in `virtual_host.rs`. We can export infallible constructors that create specific variants.
```rust
 impl StringMatcher {
  /// Matches inputs equal to `value`.
  ///
  /// `ignore_case` uses ASCII case folding. Unlike [`Self::from_proto`],
  /// these constructors perform no validation: callers are responsible for
  /// supplying a non-empty value where the xDS API requires one.
  #[inline]
  pub(crate) fn exact(value: impl Into<String>, ignore_case: bool) -> Self {
      Self::Exact { value: value.into(), ignore_case }
  }

  /// Matches inputs starting with `value`.
  #[inline]
  pub(crate) fn prefix(value: impl Into<String>, ignore_case: bool) -> Self {
      Self::Prefix { value: value.into(), ignore_case }
  }

  /// Matches inputs ending with `value`.
  #[inline]
  pub(crate) fn suffix(value: impl Into<String>, ignore_case: bool) -> Self {
      Self::Suffix { value: value.into(), ignore_case }
  }

  /// Matches inputs containing `value`.
  #[inline]
  pub(crate) fn contains(value: impl Into<String>, ignore_case: bool) -> Self {
      Self::Contains { value: value.into(), ignore_case }
  }
}
```
In virutal_host.rs, you can do something like this:
```rust
StringMatcher::suffix(&pattern[1..], true).is_match(authority)
```

This will cause a heap allocation, but it's once per config update and not per RPC.

YutaoMa:
Accidentally closed when I'm cleaning up my fork's branches, reopening

grpc-xds/src/lib.rs
YutaoMa: valid, removed them in the new commit

grpc-xds/src/matcher/mod.rs
YutaoMa: It's mostly a dev hygiene thing that I've being doing in Rust code, so that it's easier to see all the public interface from a module in a single file, and offers a minor degree of encapsulation that hides the submodule paths. I don't mind either layout tbh, if grpc crates has an established pattern I can switch to that.

grpc-xds/src/resource/safe_regex.rs
YutaoMa: this is idiomatic style for Rust new type style though, I see the same pattern used in grpc crates as well. imo named field is useful when having multiple fields or unclear field semantics, and in case like this the field type `Regex` is self-explanatory.

grpc-xds/src/resource/safe_regex.rs
YutaoMa: Good point, in fact now I think of it this can be a `TryFrom` or `FromStr` impl for idiomatic Rust style. Let me change to that instead.

grpc-xds/src/matcher/string.rs
YutaoMa: was keeping the matching logic separated from the proto validation logic for encapsulation, but I'm good with re-unifying them for simplicity as well. Agree with the heap allocation being not a blocker.

grpc-xds/src/resource/string_matcher.rs
YutaoMa: good point about that function creating a problem (although to me that's an API design issue in protobuf lib, Rust `unsafe` function can and should be used to guard type invariant as well), I've changed the behavior to return validation error instead of defaulting to empty str.

grpc-xds/src/matcher/mod.rs
YutaoMa: (given we are moving this to be along with StringMatcher in resource anyway not an issue anymore)

grpc-xds/src/resource/string_matcher.rs
YutaoMa: Sure, also added rustfmt config file along with the change, and kept the fmt run in a separate commit for reviewability. If you'd rather the fmt run itself to be in another PR I can extract it out as well.

PR #2873, opened 2026-09-17
grpc/lb: add SubchannelUpdate and remove LbPolicy::subchannel_update

dfawley:
This change leverages the existing WorkScheduler to deliver subchannel updates instead of requiring LB policies to know which child created which subchannel.

This PR is split across 6 commits that touch different pieces in isolation: the LB module definition, channel impl, the different LB policies, and finally the existing test fakes to keep them working.

I have a follow-on PR that shares a **ton** of duplicated testing code which I noticed is common between all the LB policies.  I'll send that after this is done, but don't look _too too_ hard at `test_utils.rs` or the tests and suggest nits/minor improvements (until that PR).


grpc/src/client/load_balancing/mod.rs
```diff
@@ -221,7 +217,21 @@ impl ParsedJsonLbConfig {
 /// Controls channel behaviors.
 pub trait ChannelController: Send + Sync {
     /// Creates a new subchannel and returns its current state.
-    fn new_subchannel(&mut self, address: &Address) -> (Arc<dyn Subchannel>, SubchannelState);
+    ///
+    /// Whenever the subchannel changes state, the channel calls
+    /// [`schedule_work`](WorkScheduler::schedule_work) on `work_scheduler`
+    /// with a [`SubchannelUpdate`](subchannel::SubchannelUpdate) describing the
+    /// new state.  Policies should generally pass the WorkScheduler they were
+    /// given in [`LbPolicyOptions`] so the update is routed back to them.
```
arjan-bal, line 225: Why does the load balancer need to provide a WorkScheduler back to the channel? Since the channel already supplied one when creating the LB, can’t the channel assume it's the same one?

grpc/src/client/load_balancing/mod.rs
```diff
@@ -173,9 +170,8 @@ pub type WorkData = Box<dyn WorkDataTrait>;
 /// the LbPolicy needs to provide an update without waiting for an update
 /// from the channel first.
 pub trait WorkScheduler: Send + Sync + Debug {
-    // Schedules a call into the LbPolicy's work method.  If there is already a
-    // pending work call that has not yet started, this may not schedule another
-    // call.
+    /// Schedules a call into the LbPolicy's work method.  Multiple work calls
+    /// carrying a `data` payload of `None` may be coalesced with one another.
     fn schedule_work(&self, data: Option<WorkData>);
```
arjan-bal, line 175: What do you think about turning data into an enum to provide stronger type safety for common cases?
```rust
#[non_exhaustive]
enum WorkItem {
  SubChannelUpdate(SubchannelState)
  Work(WorkData)
  None
}
```

grpc/src/client/load_balancing/mod.rs
dfawley: This allows direct forwarding to a child policy that creates a subchannel.  Otherwise how does a parent know how to route the subchannel update?  It needs to intercept new_subchannel and then...remember which subchannel was created by which child (and so there's no benefit here).

This way the routing information can be built into the work scheduler itself.

grpc/src/client/load_balancing/mod.rs
dfawley: I don't think this provides any stronger type safety, it just converts a downcast into a match.

I did consider this, but I think it's unnecessary and just adds another type into this system.

EDIT: Also, it requires implementers of `work` that don't ever call `new_subchannel` (which is everyone besides `pick_first` btw) to include `SubchannelUpdate` in their match arms.)

In the future I'll also be adding a subchannel health state struct, and arbitrary payloads for producers.

One thing I _do_ still want to try out is limiting the data that can be sent to the work scheduler provided to `new_subchannel` so that it can _only_ be used for a `SubchannelUpdate`.  But that's a smaller refinement, I think, and will have more bearing when we get to the producer stuff.

grpc/src/client/load_balancing/mod.rs
arjan-bal: Ah, I see. The parent LB needs a way to intercept the `schedule_work` call from the subchannel.

arjan-bal (approved):
LGTM

PR #2875, opened 2026-09-18
xds-client: allow configuring ADS message size limits

ankurmittal:
## Motivation

The ADS stream runs with tonic's default message size limits, so a decoded
response is capped at 4 MiB and there is no way to raise it. A control plane
serving a large resource set exceeds that, failing the stream and forcing a
reconnect that hits the same limit again.

## Solution

Add `with_max_decoding_message_size` and `with_max_encoding_message_size` to
`TonicTransportBuilder`, forwarding to the underlying `Grpc` client. Both are
unset by default, leaving tonic's behaviour unchanged.

## Testing Done

New tests assert that an oversized response fails at the default limit and
succeeds once the limit is raised. `cargo test`, clippy, fmt and rustdoc
`-D warnings` are green for `xds-client`.


YutaoMa (approved):
LGTM

PR #2876, opened 2026-09-18
tonic-xds: expose gRFC A29 cluster security parsing

ankurmittal:
## Motivation

`ClusterTlsConfig` lets a custom connector reuse the crate's gRFC A29 handling instead of re-implementing SAN matching and chain validation. The only way to get one is from `MakeConnector::make_connector`, since it borrows the channel's cert-provider registry and has no public constructor.

Code that runs its own ADS stream already holds the `Cluster` and the certificate providers, but it never calls `make_connector`, so it cannot get a `ClusterTlsConfig`. That leaves copying the A29 parsing, which is about 1300 lines across `security.rs`, `san_matcher.rs` and `string_matcher.rs`, all private to the crate.

## Solution

Make `ClusterSecurityConfig` public, with private fields and accessors.

`from_cluster_bytes` parses the `transport_socket` of a serialized `Cluster`. It returns `Ok(None)` when the cluster has no transport socket, and otherwise fails with the same NACK errors as before. It takes the encoded resource rather than a decoded `Cluster` so that the envoy protobuf types stay out of this crate's public API, which `check-external-types` enforces. A caller driving its own ADS stream has those bytes already, since they are what the management server sent.

`build_verifier` takes the CA provider as an argument. The registry exists so that `ClusterTlsConfig` can look that provider up by name, so passing it directly means no registry is needed.

`SanMatcher` and the string matchers stay private. A caller never handles them: `build_verifier` returns a verifier that already enforces the cluster's matchers.

`ClusterTlsConfig` now calls these methods rather than reading the fields, so the registry lookup is all it still does itself.

## Testing Done

New tests cover plaintext, server-authentication-only, mTLS, a missing validation context, and bytes that are not a `Cluster`. `cargo test`, clippy, fmt and rustdoc `-D warnings` are green for `tonic-xds` with `tls-ring`, and `cargo check` and `cargo clippy` are clean with `--no-default-features`.

## Notes for reviewers

The signature taking `&[u8]` is the part most likely to look wrong. `&Cluster` reads better, but `allowed_external_types` in `tonic-xds/Cargo.toml` lists no `envoy_types` entry, so a decoded `Cluster` in a public signature fails `check-external-types`. Adding an entry would make this the first public API in the crate to expose an envoy type, and would tie the crate's semver to a 0.x dependency, so the bytes avoid that. Happy to switch if you would rather allowlist the type.

This does mean the caller decodes the `Cluster` and `from_cluster_bytes` decodes it again. That is once per CDS update rather than per request, and a connector is rebuilt on that same edge, so the cost lands in the same place.

`ClusterSecurityConfig` keeps the fields private, so the four existing call sites use the accessors. Only one needed more: a test helper in `cluster_discovery.rs` built the struct literally, and now uses a `#[cfg(test)]` constructor.


tonic-xds/src/xds/resource/security.rs
```diff
@@ -291,7 +354,45 @@ mod tests {
...
+        let cluster = cluster_bytes(Some(wrap_upstream(CommonTlsContext::default())));
+        assert!(ClusterSecurityConfig::from_cluster_bytes(&cluster).is_err());
+    }
+
+    #[test]
+    fn from_cluster_bytes_rejects_bytes_that_are_not_a_cluster() {
+        assert!(ClusterSecurityConfig::from_cluster_bytes(&[0xff, 0xff]).is_err());
+    }
+
     fn absent_transport_socket_yields_plaintext() {
```
YutaoMa, line 397: this test's `#[test]` is taken away

YutaoMa: reminder for myself to change the module level `#![allow(dead_code)]` to field level instead.

tonic-xds/src/xds/resource/security.rs
```diff
@@ -45,19 +45,81 @@ const TLS_TRANSPORT_SOCKET_NAME: &str = "envoy.transport_sockets.tls";
...
+    /// driving its own ADS stream has these bytes already: they are what the
+    /// management server sent.
+    ///
+    /// Returns `Ok(None)` when the cluster declares no transport socket and so
+    /// connects in plaintext, and an error for any A29 NACK condition.
+    ///
+    /// [`XdsChannel`](crate::XdsChannel) parses clusters itself and hands the
+    /// result to a connector, so this is for a caller outside that flow, which
+    /// holds the resource and its own providers.
+    pub fn from_cluster_bytes(cluster: &[u8]) -> Result<Option<Self>, Error> {
```
YutaoMa, line 75: This `Error` here is `xds_client::Error`. Now it's in the public signature we should consider re-export it in `lib.rs` for usability.

YutaoMa (approved):

PR #2877, opened 2026-09-18
fix(transport): drain connections before the grace timeout

AnasTaimah:
## Problem

When both connection age and grace limits are set, connections keep accepting new RPCs until the combined duration elapses and are then force-closed. No GOAWAY/drain phase runs at the age limit, so calls arriving near the cutoff can fail instead of moving to a fresh connection.

## Solution

Begin graceful draining at the age limit. Allow existing calls to finish during grace, closing early when they complete and force-closing any remaining streams at the deadline. An unset age leaves connections unrestricted; an unset grace allows draining without a forced deadline.

## How to test

Hold a request open on one HTTP/2 connection with a 10-second age and 5-second grace. After the age limit, new streams should be rejected with GOAWAY while the held response can still complete successfully. Repeat with the response blocked past 15 seconds and verify that the connection closes.


ejona86 (commented):
I haven't reviewed the tests, but connection_timeout_future() was clearly broken and the new code seems good.

